PR #13, opened 2017-04-06
cmd: use AWS not Aws

philips:
AWS is an acronym and should be all capitals or all lower case due to
the Go style guide.


bigkraig:
👍

PR #23, opened 2017-04-07
WIP: Refactor code into smaller packages

joshrosso:
- First pass at breaking code into smaller packages.
- Tests need to be re-thought and re-written with the new package model

The pkg setup is now
```
/
- awsutil/
- controller/
  - alb/
  - config/
  - util/
- log/
- main.go
```

Let me know your thoughts @bigkraig and if you have some bandwidth to check it out @philips.

A few observations / todo:
- Not to crazy about `controller/util` could possible be rethought where it should belong. Feel like this will become a massive catch-all.
- Could wrap the `github.com/aws/aws-sdk-go/aws/awsutil` (we only use 2 functions) into our `awsutil` to avoid confusion.
- Need to go through code and focus on making it more idiomatic in general.


joshrosso:
@bigkraig take a look at [639b56](https://github.com/coreos/alb-ingress-controller/pull/23/commits/639b56519bb1c91af5442d0b2104968f7a148eff) when you have a moment and let me know if it LGTyou.

This is an example of what we discussed this morning around our AWS client. I'm moving other service interactions to this model.

bigkraig:
Looking good!

PR #53, opened 2017-04-14
added acm policy

code0x9:
need AWS CertificateManager read permission when using "alb.ingress.kubernetes.io/certificate-arn" annotation.

### DIFF
--- a/examples/iam-policy.json
+++ b/examples/iam-policy.json
@@ -62,10 +62,7 @@
         {
             "Effect": "Allow",
             "Action": [
-                "acm:DescribeCertificate",
-                "acm:ListCertificates",
-                "acm:GetCertificate",
-                "acm:ListTagsForCertificate"
+                "acm:DescribeCertificate"
             ],
             "Resource": "*"
         }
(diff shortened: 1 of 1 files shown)


examples/iam-policy.json
```diff
@@ -1,63 +1,73 @@
...
+                "route53:ListHostedZones",
+                "route53:ListHostedZonesByName",
+                "route53:ListResourceRecordSets"
+            ],
+            "Resource": "*"
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "acm:DescribeCertificate",
```
joshrosso: Do we need actions other than this regarding ACM? I'm not sure the code uses the ones below.

code0x9: Well, I just copied "AWSCertificateManagerReadOnly" policy from IAM. maybe "acm:GetCertificate" and "acm:ListTagsForCertificate" is not needed.

joshrosso: Yup makes sense. Our only call today [is to DescribeCertificate](https://github.com/coreos/alb-ingress-controller/blob/298a4dca9a3456f35546af41a5e70dfc841c4955/awsutil/acm.go#L40-L46).

Can you remove the others in favor of only retaining minimal rights? Then i'll merge you in.

Thanks again!

code0x9: removed & committed. :)

PR #55, opened 2017-04-14
Fixing a bug when MX/SOA/NS/etc records exist for the hostname we requested

bigkraig:
We may also want to look into a situation where the requested hostname has an existing ALIAS and CNAME record.

### DIFF
--- a/awsutil/route53.go
+++ b/awsutil/route53.go
@@ -170,7 +170,7 @@ func (r *Route53) DescribeResourceRecordSets(zoneID *string, hostname *string) (
 		HostedZoneId:    zoneID,
 		MaxItems:        aws.String("100"),
 		StartRecordName: hostname,
-		StartRecordType: aws.String("A"),
+		StartRecordType: aws.String(route53.RRTypeA),
 	}
 
 	resp, err := r.Svc.ListResourceRecordSets(params)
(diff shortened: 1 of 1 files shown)


awsutil/route53.go
```diff
@@ -168,8 +168,9 @@ func (r *Route53) Delete(in route53.ChangeResourceRecordSetsInput) error {
 func (r *Route53) DescribeResourceRecordSets(zoneID *string, hostname *string) (*route53.ResourceRecordSet, error) {
 	params := &route53.ListResourceRecordSetsInput{
 		HostedZoneId:    zoneID,
-		MaxItems:        aws.String("1"),
+		MaxItems:        aws.String("100"),
 		StartRecordName: hostname,
+		StartRecordType: aws.String("A"),
```
joshrosso: For consistency use constants from https://github.com/aws/aws-sdk-go/blob/master/service/route53/api.go#L13538-L13571

awsutil/route53.go
```diff
@@ -182,7 +183,16 @@ func (r *Route53) DescribeResourceRecordSets(zoneID *string, hostname *string) (
 		return nil, fmt.Errorf("ListResourceRecordSets(%s, %s) returned an empty list", *zoneID, *hostname)
 	}
 
-	return resp.ResourceRecordSets[0], nil
+	for _, record := range resp.ResourceRecordSets {
+		if *record.Type != "CNAME" && *record.Type != "A" {
+			continue
```
joshrosso: For consistency use constants from https://github.com/aws/aws-sdk-go/blob/master/service/route53/api.go#L13538-L13571

PR #74, opened 2017-04-25
Resolve hosted zones regardless of domain name

### LINKED ISSUE #62: Can't create DNS entry with three-part domain name
Hi, thanks for the great product, it's exact what we need and are planning to use it in production.

However, after running it the first time, I got a panic similar to #61. It seems that the issue is in this line: https://github.com/coreos/alb-ingress-controller/blob/master/awsutil/route53.go#L73

getDomain() assumes that the last two components of a hostname is the domain. However, we are playing around in our sandbox AWS account, which owns *.dev.rockset.io and doesn't own *.rockset.io. This causes a failure in `awsutil.Route53svc.GetZoneID` because it's called with rockset.io instead of dev.rockset.io.

joshrosso:
- Goes through and incrementally looks for a hostedZone based on the
domain
- Resolves #62 

@bigkraig This solution works elegantly. I'm working through a small caching issue, which i'm attempting to resolve tonight.

Caching issue will cause minimal code changes, feel free to review when you have a free moment.

### DIFF
--- a/awsutil/route53.go
+++ b/awsutil/route53.go
@@ -58,71 +58,52 @@ func NewRoute53(awsconfig *aws.Config) *Route53 {
 	return &r53
 }
 
-// getDomain looks for the 'domain' of the hostname
-// It assumes an ingress resource defined is only adding a single subdomain
-// on to an AWS hosted zone. This may be too naive for Ticketmaster's use case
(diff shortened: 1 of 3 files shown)


joshrosso:
Ready for validation and merge.

PR #86, opened 2017-05-07
Catch errors when we fail to modify targetgroup targets

bigkraig:


controller/alb/targetgroup.go
```diff
@@ -138,8 +138,8 @@ func (tg *TargetGroup) create(lb *LoadBalancer) error {
 		Port:                       tg.DesiredTargetGroup.Port,
 		Protocol:                   tg.DesiredTargetGroup.Protocol,
 		Name:                       tg.DesiredTargetGroup.TargetGroupName,
-		UnhealthyThresholdCount:    tg.DesiredTargetGroup.UnhealthyThresholdCount,
-		VpcId:                      lb.CurrentLoadBalancer.VpcId,
+		UnhealthyThresholdCount: tg.DesiredTargetGroup.UnhealthyThresholdCount,
+		VpcId: lb.CurrentLoadBalancer.VpcId,
```
joshrosso: Does this need `go fmt`?

controller/ingress.go
```diff
@@ -234,8 +234,8 @@ func assembleIngresses(ac *ALBController) ALBIngressesT {
 		ingressID := namespace + "-" + ingressName
 
 		rs := &alb.ResourceRecordSet{
-			IngressID:                &ingressID,
-			ZoneID:                   zone.Id,
+			IngressID: &ingressID,
+			ZoneID:    zone.Id,
```
joshrosso: Does this need `go fmt`?

bigkraig:
i think my gofmt has lost its mind

joshrosso:
Nope, actually I think my go fmt has lost its mind. I was looking over those things as if they were struct definitions, but that spacing isn't required for composite literal definitions.

Don't really know how the spacing got that way in the first place.

PR #94, opened 2017-05-10
#92 Ignore master nodes

### LINKED ISSUE #92: Stop using master nodes
Stop master nodes being added to the target group. Master nodes should not handle this traffic.

owenhaynes:
Removes master nodes from registered instances.

Resolves #92

### DIFF
--- a/controller/ingress.go
+++ b/controller/ingress.go
@@ -234,8 +234,8 @@ func assembleIngresses(ac *ALBController) ALBIngressesT {
 		ingressID := namespace + "-" + ingressName
 
 		rs := &alb.ResourceRecordSet{
-			IngressID:                &ingressID,
-			ZoneID:                   zone.Id,
+			IngressID: &ingressID,
+			ZoneID:    zone.Id,
 			CurrentResourceRecordSet: resourceRecordSet,
 		}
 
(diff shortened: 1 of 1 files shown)


treyhyde:
The logic for Service: LoadBalancer in kube aws is to exclude nodes that are marked with SchedulingDisabled which usually includes the masters.  I think that covers a broader range of cases.  This PR will still probably include cordoned nodes in the pool for instance.

owenhaynes:
Don't mind changing it to scheduling disabled. Will give it a test

owenhaynes:
so check for the following taints on the nodes:
node-role.kubernetes.io/master:NoSchedule
node-role.kubernetes.io/node:NoSchedule
?

treyhyde:
I understand there are some mechanical differences between NodeNotSchedulable not 1.5 and and the more mature taints in 1.6.  I'm looking at my 1.5 cluster right now, and marking a node as not schedulable doesn't' seem to add a taint or an annotation.


re kops, there was some healthy discussion about taints and marking masters as schedulable or not... I can't find it right now.  Maybe both approaches are warranted.

joshrosso:
Hi @treyhyde @owenhaynes 

The exclusion of master nodes makes sense. Can we reach consensus on what annotations are required for 1.5+ as discussed above and add them into this PR?

Thanks for the work, looking forward to merging this!

joshrosso:
@owenhaynes @treyhyde 

Ping on this. Does it make sense to simply detect both annotations?

Or is there only 1 you feel we need. Most of the deployments i'm testing against are self-hosted k8s so the same principals don't apply. Would love to have this if it helps compatibility with kops and similar.

darkofabijan:
Would love to see solution for this :) Exposing masters in this way seems like a pretty bad idea.

miry:
is it possible to have node selector? ideally I would like to have registered nodes where my containers are running. instead of include all nodes.

joshrosso:
@miry for the near term, we're planning to go with what's described here: https://github.com/coreos/alb-ingress-controller/pull/197/commits/32655dfc057bfad61f45bee670ad1e97bb997fab

PR #96, opened 2017-05-11
Update vendored dependencies to work towards kubernetes 1.6 compatibility

cmaloney:
Currently trying to run this with a kops 1.6.0-beta.1 deployed kubernetes 1.6.0 cluster and getting quite a bit of odd behavior.  This is my work towards fixing that

Some of the things seen
 - Ingresses being updated but the ALB ingress controller not having it's OnUpdate event fire.
 - Weirdness in syncing the rules (Some rules appear, some don't. Some point to the wrong service. All targets exist)

Hope is to update the vendored library to hopefully fix the update issues (The kubernetes/ingress repo changed the kubernetes go client version), and then work on the issues around syncing rules.

PR #97, opened 2017-05-11
Switch to smaller base docker container

cmaloney:
Follows changes in
https://github.com/kubernetes/ingress/blob/master/controllers/gce/Dockerfile
which is what this seems to be based on.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,11 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


cmaloney:
I'd rather keep them separate / will pull this out of the update-deps branch. The update-deps I'm still validating a bit (With the change the extensions import + type casting needs to change everywhere it exists). Pulled it out of #96.

joshrosso:
Cool, this lgtm.

@bigkraig Any reservations on slimming down to alpine? If not, let's merge.

PR #99, opened 2017-05-13
Share AWS session between clients, mainly for performance

joshuaswaney:
This simplifies creation of AWS clients by re-using the session, and it should speed up initialization a bit.

My first time working in go, so feel free to comment on anything


joshrosso:
Thanks for the PR, @joshuaswaney.

Works well on my end.

PR #108, opened 2017-05-24
Do not attempt to cast nil errors.

joshrosso:


joshrosso:
@bigkraig, this checks out in my cluster. Please review when you can.

PR #110, opened 2017-05-25
WIP: Refine rule (path) modifications to be stable

joshrosso:
Goal is to allow for path modification without the need to completely delete / recreate the ingress.

### DIFF
--- a/controller/alb/rules.go
+++ b/controller/alb/rules.go
@@ -1,6 +1,7 @@
 package alb
 
 import (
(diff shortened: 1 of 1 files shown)


joshrosso:
https://github.com/coreos/alb-ingress-controller/pull/110/commits/30a6a1a4959f20b84400a4accb96577099f073af address incorrect index mentioned in https://github.com/coreos/alb-ingress-controller/issues/75.

PR #117, opened 2017-06-12
fix AWS env variable name in the helm chart

sdomula:
In the helm chart the env variable for the  secret access key is set as `AWS_ACCESS_SECRET_KEY` but should be  `AWS_SECRET_ACCESS_KEY`.


bigkraig:
thanks!

PR #120, opened 2017-06-19
Fix for nil annotations

arbarlow:
This is a small, quick fix for the ALB ingress to error gracefully when an ingress has no annotations, providing some feedback to help the user with adding the minimum required annotations.

Happy to fix higher up the chain if need be, perhaps ingress validation is required?


bigkraig:
Thanks!

PR #126, opened 2017-06-23
reconcile: don't stop due to single failures

joshrosso:
- This commit ensures that ingress resources that have failed during
reconcile are captured in their own array for logging. Additionally,
rather than bubbling the error up and brining a Reconcile event to a
complete halt, just that ingress will be stopped, logged, and the rest
are able to continue reconciling.


joshrosso:
Tested in our clusters -- seems to solve loop fine.

PR #127, opened 2017-06-23
Add `ingresses/status` to fix the error:

andromedarabbit:
> error updating ingress rule: User "system:serviceaccount:kube-system:alb-ingress-controller" cannot update ingresses.extensions/status in the namespace "default". (put ingresses.extensions gophernotes)


joshrosso:
Thanks @andromedarabbit

PR #134, opened 2017-07-11
fail lb if resourcerecordset never populated

justinclayton:
This comes up most often when either the service is missing or NodePort is not set on the service, resulting in endpoints not found.

### DIFF
--- a/controller/alb/loadbalancers.go
+++ b/controller/alb/loadbalancers.go
@@ -1,5 +1,7 @@
 package alb
 
+import "errors"
+
 // LoadBalancers is a slice of LoadBalancer pointers
 type LoadBalancers []*LoadBalancer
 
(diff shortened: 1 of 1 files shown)


controller/alb/loadbalancers.go
```diff
@@ -28,6 +30,12 @@ func (l LoadBalancers) Reconcile() (LoadBalancers, LoadBalancers) {
 			errLBs = append(errLBs, loadbalancer)
 			continue
 		}
+		if loadbalancer.ResourceRecordSet == nil {
```
joshrosso: Perhaps this check is too late.

If `ResourceRecordSet` is nil, can we prevent `Reconcile` from occurring in the first place?

Specifically, could we do a similar check somewhere around here: https://github.com/coreos/alb-ingress-controller/blob/master/controller/ingress.go#L164-L174.

bigkraig:
Closing this PR since we have removed Route53 management in the current release, favoring https://github.com/kubernetes-incubator/external-dns

PR #140, opened 2017-07-13
Make Route53 integration optional

### LINKED ISSUE #9: Make R53 integration optional
As this gets rolled out to the larger masses, the front-to-back integration may a be no go when individuals find that Route 53 is coupled to the ALB creation.

It would be helpful to have a flag (ENV var) that disables state sync against Route 53. Making the alb-ingress-controller only concerned with ELBV2s (ALBs), TargetGroups, Listeners, and Rules.

Helpful for those who:
1. Manage their DNS outside of Route53.
1. Want to plugin their own Route53-handling controller.
1. Prefer to manually alter Route53 records.

/cc @bigkraig

justinclayton:
Fixes #9. This PR allows wholesale disabling of Route53 integration by passing the environment variable `DISABLE_ROUTE53=true` at controller start.

There are a lot of ways this could be solved that are cleaner, so if people aren't happy with this sledgehammer approach I'd love this to be the start of a discussion.

### DIFF
--- a/controller/alb/loadbalancers.go
+++ b/controller/alb/loadbalancers.go
@@ -17,7 +17,7 @@ func (l LoadBalancers) Find(lb *LoadBalancer) int {
 // balancer and its resource record set, target group(s), and listener(s). It returns 2
 // LoadBalancers (slices), the first being the list of all known LoadBalancers and the subset
 // second being of LoadBalancers, from the first list, that failed to reconcile.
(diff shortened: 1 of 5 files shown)


bigkraig:
thanks!

jwhitcraft:
@justinclayton while this is nice, i would have like to see it more of a new annotation so it could be selective as if it's done or not.

But regardless, this is a good first step.

justinclayton:
@jwhitcraft I agree, that would be a good next step.

PR #141, opened 2017-07-14
Update Helm chart push

mgoodness:
* Use latest version of Helm
* Use new Helm plugin installation process to get registry plugin
* Override chart name to get around quay.io naming limitation
* Bump chart version

As before, I can't guarantee Travis will push the chart correctly, but I'm happy to help get it working. Would love to replace our old TM chart with this one.


mgoodness:
ping @joshrosso

PR #145, opened 2017-07-17
Addressing memory leak in listener creation

bigkraig:
Pending the merge of https://github.com/kubernetes/ingress/pull/985


controller/controller.go
```diff
@@ -220,169 +175,120 @@ func (ac *ALBController) ingressToDelete(newList ALBIngressesT) ALBIngressesT {
...
+}
 
-		for _, targetGroup := range targetGroups {
-			tags, err := awsutil.ALBsvc.DescribeTags(targetGroup.TargetGroupArn)
-			if err != nil {
-				glog.Fatal(err)
-			}
+// OverrideFlags configures optional override flags for the ingress controller
+func (ac *ALBController) OverrideFlags(flags *pflag.FlagSet) {
+	flags.Set("update-status-on-shutdown", "false")
```
bigkraig: We want to get this into master, without this the address will get removed on shutdown even though we dont destroy the AWS resources

controller/alb/listener.go
```diff
@@ -172,6 +168,8 @@ func (l *Listener) delete(lb *LoadBalancer) error {
 
 func (l *Listener) needsModification(target *elbv2.Listener) bool {
 	switch {
+	case l.CurrentListener == nil && l.DesiredListener == nil:
+		return false
```
bigkraig: this needs to get into a new PR

bigkraig:
includes commits from #166

PR #150, opened 2017-07-21
update ingress' statuses with alb dns names

matt-deboer:
In combination with `DISABLE_ROUTE53`, this supports external updates of route53 by reflecting the ALBs' dns names in the `status.loadBalancer.ingress` field of the associated ingress object.

(i.e, `kubectl get ingress` will now show a correctly-populated value for the ADDRESS field)

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -13,15 +13,22 @@ import (
 	"github.com/golang/glog"
 	"github.com/spf13/pflag"
 
+	"k8s.io/client-go/kubernetes"
 	api "k8s.io/client-go/pkg/api/v1"
 	extensions "k8s.io/client-go/pkg/apis/extensions/v1beta1"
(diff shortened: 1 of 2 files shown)


justinclayton:
This is going to be needed if anyone wants to leverage [kubernetes-incubator/external-dns](http://github.com/kubernetes-incubator/external-dns).

joshrosso:
Awesome! Spinning up a cluster now to do some validation. Thank you both!

bigkraig:
We've got a similar solution in the works that is dependent on https://github.com/kubernetes/ingress/pull/985. The thought in the ingress library is that the client shouldn't need to be exposed to controllers using it, so occasionally changes up stream need to be made.

matt-deboer:
@bigkraig that sounds like a better fix; I was following `gce` as example, but looks like they're moving that way also.

PR #152, opened 2017-07-27
Support Helm release name = chart name

mgoodness:
Minor update to the Helm chart that eliminates redundancy if the release name is the same as the chart name.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /alb-ingress-controller
 /server
+*.tgz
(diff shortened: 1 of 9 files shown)


mgoodness:
This will (eventually) fix #163

joshrosso:
lgtm @mgoodness, anything else you're planning on adding to this before I merge?

mgoodness:
Nope. I've already pushed it to https://quay.io/repository/mgoodness/alb-ingress-controller-helm and http://charts.opsgoodness.net. Merge away!

PR #158, opened 2017-08-03
empty (default) path should also trigger default rule

### LINKED ISSUE #155: "Default Backends" is not implemented/supported.
**What Happens**: When an ingress is created such as the following: (taken directly from the [kube docs](https://kubernetes.io/docs/concepts/services-networking/ingress/) )
```
kind: Ingress
metadata:
  name: test
spec:
  rules:
  - host: foo.bar.com
    http:
      paths:
      - backend:
          serviceName: s1
          servicePort: 80
```
(notice the lack of a `path` value)
the result is continual errors agains the AWS elastic load balancers API, `ValidationError: A condition value cannot be empty`.

The important part from the cloudwatch event details:
```
"requestParameters": {
   ...
   "conditions": [{"field": "path-pattern", "values": [""]}]}
   ...
}
```

**What should happen**: the alb-ingress-controller should use a default "catch-all" rule for this case.

Additionally, there should probably be some sort of exponential back-off in the event of AWS api failures, as this error can result in many thousands of error requests per hour (and may end up causing other api requests to be throttled)

matt-deboer:
This PR fixes #155 .

### DIFF
--- a/controller/alb/rule.go
+++ b/controller/alb/rule.go
@@ -28,7 +28,7 @@ func NewRule(path extensions.HTTPIngressPath, ingressID *string) *Rule {
 		},
 	}
 
-	if path.Path == "/" {
+	if path.Path == "/" || path.Path == "" {
 		r.IsDefault = aws.Bool(true)
 		r.Priority = aws.String("default")
 	} else {


bigkraig:
Thanks!

PR #166, opened 2017-08-04
Improved logging

bigkraig:
Logging now consistently uses our logging functions and reports proper filename/lines.


joshrosso:
Relates to https://github.com/coreos/alb-ingress-controller/issues/29

bigkraig:
@joshrosso ready for review

awsutil/route53.go
```diff
@@ -173,6 +172,7 @@ func LookupExistingRecord(hostname *string) *route53.ResourceRecordSet {
 	// If zone was resolved, then host exists. Return the respective route53.ResourceRecordSet.
 	rrs, err := Route53svc.DescribeResourceRecordSets(zone.Id, hostname)
 	if err != nil {
+		log.Errorf(err.Error(), "aws")
```
joshrosso: Very close to working as expected. Seems a new error log is causing a false alarm.

Before we go to create a new r53 resource record set we check to see if one already exists.

https://github.com/coreos/alb-ingress-controller/blob/9ac610482e15a7850ad70bb1a2cf0e790cdc2511/controller/alb/resourcerecordset.go#L105

If this returns no existing r53 record (which would be the most common scenario on create) we log an error.

```
E0807 16:47:23.215826       1 route53.go:175] [ALB-INGRESS] [aws] [ERROR]: ListResourceRecordSets(/hostedzone/Z37F6OLDZZO6IG, echo.josh-test-dns.com) did not return any valid records
```

@bigkraig Can we change this to `INFO`? If not i'd settle for a `WARN`.

bigkraig: @joshrosso done

PR #182, opened 2017-08-15
replace any '-' in clustername with '_' before using it

### LINKED ISSUE #181: Using CLUSTER_NAME values containing "-" causes failures to reconcile
When the `cluster-name` config value contains the hyphen character, load-balancer reconciliation will never happen, do to the code below: (from `ELBV2.DescribeLoadBalancers`)

```
if strings.HasPrefix(*loadBalancer.LoadBalancerName, *clusterName+"-") {
  if s := strings.Split(*loadBalancer.LoadBalancerName, "-"); len(s) == 2 {
    if s[0] == *clusterName {
      loadbalancers = append(loadbalancers, loadBalancer)
    }
  }
}
```
An error should be thrown when `-` is used, or else it should be converted to something else, like `_` before being used in the loadbalancer name.

matt-deboer:
Fixes #181 .

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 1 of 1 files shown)


bigkraig:
We picked this one up in the refactor and just throw an error for people who try to use `-`: https://github.com/coreos/alb-ingress-controller/blob/master/pkg/controller/alb-controller.go#L68-L78

PR #183, opened 2017-08-19
WIP: Host based routing

bigkraig:


bigkraig:
I think this is ready to go now, but since this is a large change from any existing deployments, I am concerned about how we communicate this to our users.

The main things are the requirement on using something like external-dns to manage your records and the change of load balancer IDs. Everyone is going to need to build all new load balancers and migrate over.

PR #196, opened 2017-09-07
Fix misclassification of scheme as an optional annotation in docs

BlueMonday:
Scheme is a required annotation. The docs incorrectly classified scheme as an
optional annotation.

See: https://github.com/coreos/alb-ingress-controller/blob/master/pkg/annotations/annotations.go#L256


BlueMonday:
@bigkraig

joshrosso:
Thanks for the contribution @BlueMonday.

PR #210, opened 2017-09-15
add mutex around r/w to ac.ALBIngresses

matt-deboer:


pkg/controller/alb-controller.go
```diff
@@ -107,13 +105,20 @@ func (ac *ALBController) syncALBs() {
...
+		ac.syncALBsWithAWS()
 	}
 }
 
+func (ac *ALBController) syncALBsWithAWS() {
+	albIngresses := albingresses.AssembleIngressesFromAWS(&albingresses.AssembleIngressesFromAWSOptions{
+		Recorder:    ac.recorder,
+		ClusterName: ac.clusterName,
+	})
+	ac.mutex.Lock()
```
joshrosso: @matt-deboer - I'm thinking we should lock before attempting `AssembleIngressesFromAWS` otherwise an in flight reconcile could have changed AWS resources after we gathered them.

Thoughts?

matt-deboer:
makes sense..

Issue #79, opened 2017-05-01
[Feature] Annotation for read timeout


Awk34:
I don't know Go, but I would like to help add an annotation for the load balancer's read idle timeout setting (defaults to 60s). How could I start on trying to get this implemented?

bigkraig:
You'll want to add support for reading the annotation to https://github.com/coreos/alb-ingress-controller/blob/master/controller/config/annotations.go. `SuccessCodes` is probably a good one to replicate.

Adding support for configuring the ALB with this is going to be a little more complicated, since you will need to leverage `LoadBalancerAttribute`'s, which we are not yet using. The corresponding information will need to be saved in the `LoadBalancer` struct, compared in `.needsModification()`, and updated in the `.modify()` func.

joshrosso:
This is almost completed.

PR #218, opened 2017-09-19
WIP: Configurable load balancer attributes (access logging and read timeout)

### LINKED ISSUE #90: Enable access logging for ALBs created with alb-ingress-controller
Hello,

We'd like to enable access logging for ALBs that alb-ingress-controller create. We are currently doing this manually for long-running services, but it would be nice if we could configure that in alb-ingress-controller.

ALBs support logging to S3, see https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#enable-access-logging

antoniobeyah:
Fixes #90 (access logging) and #79 (read timeout)

### DIFF
--- a/pkg/annotations/annotations.go
+++ b/pkg/annotations/annotations.go
@@ -42,6 +42,7 @@ const (
 	clusterTagValue               = "shared"
 	albRoleTagKey                 = "tag:kubernetes.io/role/alb-ingress"
 	albManagedSubnetsCacheKey     = "alb-managed-subnets"
+	attributesKey                 = "alb.ingress.kubernetes.io/attributes"
 )
 
 // Annotations contains all of the annotation configuration for an ingress
@@ -62,6 +63,7 @@ type Annotations struct {
 	SuccessCodes               *string
 	Tags                       []*elbv2.Tag
 	VPCID                      *string
+	Attributes                 []*elbv2.LoadBalancerAttribute
 }
 
 // ParseAnnotations validates and loads all the annotations provided into the Annotations struct.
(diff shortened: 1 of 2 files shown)


pkg/annotations/annotations.go
```diff
@@ -106,6 +109,33 @@ func ParseAnnotations(annotations map[string]string, clusterName string) (*Annot
...
+	var attrs []*elbv2.LoadBalancerAttribute
+	var badAttrs []string
+	rawAttrs := util.NewAWSStringSlice(annotations[attributesKey])
+
+	for _, rawAttr := range rawAttrs {
+		parts := strings.Split(*rawAttr, "=")
+		switch {
+		case *rawAttr == "":
+			continue
+		case len(parts) < 2:
```
joshrosso: You'll always expect `parts` to have length `2` here right?

Possibly best to set this check such that when it's anything other than 2, we return the error. Preventing cases like `attr=value=value2`

antoniobeyah:
@joshrosso how should the controller deal with configuration items there were set outside of the controller?  Looking at the way the reconcile works if someone setup access logging outside the scope of the controller and did not add in the matching annotations access logging would be disabled.  Is this something we should try to handle in code or just add it as a warning in the changelog?

joshrosso:
@antoniobeyah Our approach thus far has been that all ALBs the controller manages always uses the desired state (derived from ingress manifests) as what we'll eventually get to. Meaning, any configuration options made outside of the ALB (ie AWS) should be overwritten by the controller in the next reconcile.

Let me know if this answers your Q.

masterzen:
Hi,

Because we needed to setup the ALB access log, I finished the work started by @antoniobeyah in this PR. Since it is not possible to contribute directly to this PR, I've open a new one: #279 .

I've tested the feature on a live k8s cluster.

Please let me know if there's anything I need to do.
Thanks
Brice

Issue #225, opened 2017-09-27
Security groups not assigned to all instances


olitheolix:
Hi,

I have 3 worker nodes spread over two availability zones. The ingress controller creates the necessary security groups but only assigns it to one worker (or all workers in the first zone - cannot say for sure).

As a result, the ALB deems some pods unhealthy. The problem vanishes once I manually add the controller generated SG to the other instances.

Here is my ingress manifest:
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: something
  namespace: something
  annotations:
    alb.ingress.kubernetes.io/scheme: internet-facing
    alb.ingress.kubernetes.io/subnets: 'my-master-ap-southeast-1a, my-master-ap-southeast-1b'
    alb.ingress.kubernetes.io/healthcheck-path: '/core/healthcheck'
    alb.ingress.kubernetes.io/successCodes: '200'
    alb.ingress.kubernetes.io/backend-protocol: 'HTTP'
    alb.ingress.kubernetes.io/certificate-arn: 'arn:aws:acm:ap-southeast-1:something'
    alb.ingress.kubernetes.io/unhealthy-threshold-count: '2'
    alb.ingress.kubernetes.io/healthy-threshold-count: '2'
    alb.ingress.kubernetes.io/healthcheck-timeout-seconds: '2'
    alb.ingress.kubernetes.io/healthcheck-interval-seconds: '15'
  labels:
    app: something-ingress
spec:
  rules:
  - host: "something.net"
    http:
      paths:
      - path: /
        backend:
          serviceName: "api"
          servicePort: 80

```

This issue seems related to #6 and #100.

I use `1.0-alpha.2`

joshrosso:
Thx for your work on this!

olitheolix:
You are welcome.

PR #229, opened 2017-10-02
Bugfix for #225: iterate over all instances in all reservations.

olitheolix:
The original code assumed all workers were in the first reservation set. As a result, some instances were not assigned the necessary SGs to communicate with the ALB.

This PR now loops over all instances in all reservations and solves #225


joshrosso:
@olitheolix makes sense. Will test this out today.

joshrosso:
Thanks for the work @olitheolix. Verified this worked as expected in a multi-worker cluster.

PR #248, opened 2017-10-18
Stabilize Listener create, update, delete

joshrosso:


joshrosso:
Focusing this on listener create, update, delete as target group is a much larger effort.

PR #252, opened 2017-10-19
Fix TargetGroup and Rule reconciliation

joshrosso:
- TargetGroups and rules now reconcile correctly adding the following
correct behaviors.

    - Unused TGs are deleted

    - When a new service is pointed to, the existing rules are updates with
the new service's TG

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@
 
 all: container
 
-TAG?=1.0-alpha.3
+TAG?=1.0-alpha.5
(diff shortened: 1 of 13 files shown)


joshrosso:
This is pretty close and in my testing seems stable.

There is however a no-op rule modification firing off unnecessarily when you're using multiple rules.

I've narrowed it down to the deepequals throwing false due to order of rule condition attributes (sorting issue).

PR #258, opened 2017-11-02
Ensure that load balancer name consists of valid characters

abuchanan920:
Ran into an issue where clusters with "."'s in their name could run afoul of the valid naming scheme for load balancers.

This fix ensures that any invalid (non-alphanumeric) characters in the namespace or ingress name are removed and that any invalid (non-alphanumeric) characters in the cluster name are replaced with "-"'s.

### DIFF
--- a/pkg/alb/loadbalancer/loadbalancer.go
+++ b/pkg/alb/loadbalancer/loadbalancer.go
@@ -5,8 +5,8 @@ import (
 	"encoding/hex"
 	"fmt"
 	"reflect"
+	"regexp"
 	"sort"
-	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


pkg/alb/loadbalancer/loadbalancer.go
```diff
@@ -597,10 +596,12 @@ func createLBName(namespace string, ingressName string, clustername string) stri
 	hasher := md5.New()
 	hasher.Write([]byte(namespace + ingressName))
 	hash := hex.EncodeToString(hasher.Sum(nil))[:4]
+
+	r, _ := regexp.Compile("[[:^alnum:]]")
```
bigkraig: I think we will need to catch the error here

abuchanan920: Apologies for the late reply. I must have missed the notification.

Just checking - what would you like me to do with the error? Since we are compiling a static regex that we know compiles, there didn't seem to be much point in checking for an error, but I'm admittedly not very familiar with Go. Happy to adjust to whatever you like here.

PR #260, opened 2017-11-11
Replace 'their' with 'there'

johananl:
I stumbled upon this typo in a logging message while experimenting with the controller (nice job, by the way!). Thought I should let you guys know :-)


joshrosso:
Thanks!

PR #267, opened 2017-11-17
Configurable Attributes

r0fls:
Picking up where https://github.com/coreos/alb-ingress-controller/pull/218 was left off.

### DIFF
--- a/pkg/alb/loadbalancer/loadbalancer.go
+++ b/pkg/alb/loadbalancer/loadbalancer.go
@@ -398,6 +398,7 @@ func (lb *LoadBalancer) modify(rOpts *ReconcileOptions) error {
 			if err := albelbv2.ELBV2svc.UpdateAttributes(lb.Current.LoadBalancerArn, lb.DesiredAttributes); err != nil {
 				rOpts.Eventf(api.EventTypeWarning, "ERROR", "%s tag modification failed: %s", *lb.Current.LoadBalancerName, err.Error())
 				lb.logger.Errorf("Failed ELBV2 (ALB) tag modification: %s", err.Error())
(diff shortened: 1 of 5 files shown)


pkg/alb/loadbalancer/loadbalancer.go
```diff
@@ -373,6 +392,18 @@ func (lb *LoadBalancer) modify(rOpts *ReconcileOptions) error {
 				log.Prettify(lb.CurrentTags))
 		}
 
+		// Modify Attributes
+		if needsMod&attributesModified != 0 {
+			lb.logger.Infof("Start ELBV2 tag modification.")
+			if err := albelbv2.ELBV2svc.UpdateAttributes(lb.Current.LoadBalancerArn, lb.DesiredAttributes); err != nil {
+				rOpts.Eventf(api.EventTypeWarning, "ERROR", "%s tag modification failed: %s", *lb.Current.LoadBalancerName, err.Error())
+				lb.logger.Errorf("Failed ELBV2 (ALB) tag modification: %s", err.Error())
```
bigkraig: It needs to return the err here

pkg/aws/elbv2/elbv2.go
```diff
@@ -220,3 +249,14 @@ func (e *ELBV2) UpdateTags(arn *string, old util.Tags, new util.Tags) error {
 
 	return nil
 }
+
+// UpdateAttributes compares the new (desired) attributes against the old (current) tags. It then adds and
```
bigkraig: /tags/attributes/

pkg/alb/loadbalancer/loadbalancer.go
r0fls: ✅

PR #269, opened 2017-11-17
Support a subset of annotations on the service resources

### LINKED ISSUE #93: Health Checks do not work if using multiple pods on routes
e.g

  - host: test.com
    http:
      paths:
      - path: "/api/*"
        backend:
          serviceName: api
          servicePort: 80
      - path: "/*"
        backend:
          serviceName: frontend
          servicePort: 80

You can only set one health-check on the ingress so if service api does not have the same route for a health check as frontend you get a unable system

bigkraig:
This should allow users to specify custom healthchecks per service and resolve #93.

### DIFF
--- a/docs/resources.md
+++ b/docs/resources.md
@@ -1,4 +1,4 @@
-# Ingress Resources
+# Resources
 
 This document covers how ingress resources work in relation to The ALB Ingress Controller.
 
@@ -35,9 +35,9 @@ The host field specifies the eventual Route 53-managed domain that will route to
 
 ## Annotations
 
-The ALB Ingress Controller is configured by Annotations on the `Ingress` resource object. Some are required and some are optional. All annotations use the namespace `alb.ingress.kubernetes.io/`.
+The ALB Ingress Controller is configured by Annotations on the `Ingress` and `Service` resource objects. Some are required and some are optional. All annotations use the namespace `alb.ingress.kubernetes.io/`.
 
-### Required Annotations
+### Required Ingress Annotations
 
 ```
 alb.ingress.kubernetes.io/scheme
@@ -47,7 +47,7 @@ Required annotations are:
 
 - **scheme**: Defines whether an ALB should be `internal` or `internet-facing`. See [Load balancer scheme](http://docs.aws.amazon.com/elasticloadbalancing/latest/userguide/how-elastic-load-balancing-works.html#load-balancer-scheme) in the AWS documentation for more details.
 
-### Optional Annotations
+### Optional Ingress Annotations
 
 ```
 alb.ingress.kubernetes.io/backend-protocol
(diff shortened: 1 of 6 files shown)


docs/resources.md
```diff
@@ -1,4 +1,4 @@
-# Ingress Resources
+# Resources
 
 This document covers how ingress resources work in relation to The ALB Ingress Controller.
```
joshrosso: ingress -> ingress and service

pkg/annotations/annotations.go
```diff
@@ -114,6 +115,61 @@ func ParseAnnotations(annotations map[string]string, clusterName string) (*Annot
 	return a, nil
 }
 
+// ParseServiceAnnotations validates and loads all the annotations provided into the Annotations struct.
```
joshrosso: This comment should mention it parses all annotations relevant to the service resource.

Also the `annotations.ParseAnnotations` should likely be refactored to `annotations.ParseIngressAnnotations` for consistency.

[or] you could combine these 2 functions as they're quite similar. Just needs to be smarter on which annotations to parse based on what's passed. I'm fine either way, as long as the intent of `annotations.ParseServiceAnnotations` vs `annotations.ParseAnnotations` is clear.

joshrosso:
Minor changes requested. Spinning up a test cluster now to do some verification.

joshrosso:
The only issue i've hit when testing the controller **doesn't** fallback on the ingress resource as they should.

For example, I've set my `healthcheck-port` to `8880`. In the ingress resource.
```
$ kubectl get ing -n echoserver echoserver -o yaml
kind: Ingress
metadata:
  annotations:
    alb.ingress.kubernetes.io/healthcheck-interval-seconds: "30"
    alb.ingress.kubernetes.io/healthcheck-port: "8880"
```

But in AWS, it's still using the traffic port as seen below.

image

If I add this setting to the service, it will be respected.

```
apiVersion: v1
kind: Service
metadata:
  annotations:
    alb.ingress.kubernetes.io/healthcheck-port: '8880'
```

image

But after removal from the service resource, even though that setting is still in the ingress resource, it goes back to the default (traffic port).

```
$ kubectl get -o yaml -n echoserver svc echoserver
apiVersion: v1
kind: Service
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"Service","metadata":{"annotations":{},"name":"echoserver","namespace":"echoserver"},"spec":{"ports":[{"port":80,"protocol":"TCP","targetPort":8080}],"selector":{"app":"echoserver"},"type":"NodePort"}}
  creationTimestamp: 2017-11-18T00:33:17Z
  name: echoserver
  namespace: echoserver
  resourceVersion: "6811"
  selfLink: /api/v1/namespaces/echoserver/services/echoserver
  uid: 12541ce2-cbf8-11e7-8ebd-0aeebdf18292
spec:
```

image

Perhaps this is a small issue inside of the new `annotations.Join`?

bigkraig:
Gotta be. That’s what I get for not writing a test function.

rahulsen:
Why are we not simply specifying the custom healthcheck per targetGroup inside the ingress file itself ? Why the need to use an annotation from the service definition? Wont this be so much cleaner and intuitive like this:
```
#ingress yaml
kind: Ingress
metadata:
  name: myIngress
  namespace: default
  annotations:
spec:
  rules:
  - host: myhost.com
    http:
      paths:
      - path: /path1
        backend:
          serviceName: service1-stg
          servicePort: 8080
          healthCheckPath: /path1/health
      - path: /path2
        backend:
          serviceName: service2-stg
          servicePort: 9000
          healthCheckPath: /path2/health
```

bigkraig:
@rahulsen agreed, but its not supported in the ingress spec, i know they are looking at extending it

Issue #136, opened 2017-07-12
Set stickiness attribute on target group


tuvistavie:
Hi, I need ALB sticky sessions in my application but cannot find any way to set the relevant attribute on the target group.
Is there a particular reason why it is not possible, or is it simply not implemented yet?

Thank you.

bigkraig:
Most of the features have been added on an as needed basis. Thanks for the request!

joshrosso:
Considering the ingress traffic will hit a nodeport...I'm not sure how this would be possible as we can't guarantee the endpoint (pod) it'll land on.

Going to backlog this for now, but let me know if my assessment is invalid.

tomfriedhof:
Bumping this feature request. It would be nice to set Sticky with the ingress definition. 

I'm using socket.io, which appears to only work when I manually edit the Target Group and enable sticky for the target group using socket.io.

bigkraig:
Please test with `1.0-alpha.9` and let me know how it goes

odzhu:
@bigkraig, guys has this been released? How to enable it?

syst0m:
I am also interested in how to enable this, @bigkraig

bigkraig:
@odzhu & @syst0m you want to use the target group attribute annotation: `alb.ingress.kubernetes.io/target-group-attributes`. See https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/ingress/annotation/#target-group-attributes on how to configure it and https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#target-group-attributes for the attributes.

PR #275, opened 2017-11-26
configurable target group attributes

r0fls:
This will allow setting the deregistration delay and other attributes for the target groups. Here is the complete list of supported attributes: http://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#target-group-attributes


ehbello:
Hello, Is there any intention to fix and include this pull request soon?

ehbello:
Fixes #136

PR #276, opened 2017-11-28
explicit target deregistration

khrisrichardson:
Borrowing some of the logic from the recent [NLB support](https://github.com/kubernetes/kubernetes/commit/63d4b85bf4f2894e65610a5446050a8844ca78cd) added to kubernetes, namely:

* exclusion of targets already marked for deregistration
* explicit deregistration

### DIFF
--- a/pkg/util/types/types.go
+++ b/pkg/util/types/types.go
@@ -131,3 +131,16 @@ func (s Subnets) String() string {
 	}
 	return out
 }
+
+func Difference(a, b AWSStringSlice) (ab AWSStringSlice) {
+	mb := map[string]bool{}
+	for _, x := range b {
+		mb[*x] = true
+	}
+	for _, x := range a {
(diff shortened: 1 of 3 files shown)


pkg/util/types/types.go
```diff
@@ -131,3 +131,16 @@ func (s Subnets) String() string {
 	}
 	return out
 }
+
+func Difference(a, b AWSStringSlice) (ab AWSStringSlice) {
+	mb := map[string]bool{}
+	for _, x := range b {
+		mb[*x] = true
+	}
+	for _, x := range a {
```
r0fls: shouldn't this just be `=` instead of `:=`?

joshrosso: `x` is local to the scope of the `for` so `:=` is appropriate.

joshrosso:
Changes LGTM.

Doing some validation in a test cluster now.

PR #278, opened 2017-11-30
Associate WAF ACL

masterzen:
Hi,

This PR is a work on top of @athal7 #214 PR, which aim to fix the panic and finish the work. I couldn't contribute to @athal7's original PR, so I had to open a new one.

It is rebased on top of latest master and has been proof-tested on a live aws k8s cluster.

It would awesome if any of you could take the time to review it and possibly merge it.
Let me know if there's anything I need to do.

Thanks,
Brice

PR #279, opened 2017-11-30
Manage ALB attributes

masterzen:
Hi,

As for the WAF ACL, we needed the feature to control the ALB access log settings, so I finished the work started by @antoniobeyah in #218 .

This PR is actually based on top of #278 because that's what we use in production (so only the last 3 commits are part of this feature). I will/can rebase it to master if needed.

This allows to set the following annotations:
~~~
alb.ingress.kubernetes.io/attributes: access_logs.s3.enabled=true,access_logs.s3.bucket=our-access-logs,access_logs.s3.prefix=blahblah
~~~

This work has been tested (slightly) in a live environment.
Please feel free to review this contribution, and do not hesitate to comment.

Thanks!
Brice


masterzen:
Oops, I just noticed that @r0fls did exactly what I just did in this PR.

My bad, feel free to close this PR in favor of #267.

The only difference between #267 and this one (beside some cosmetic differences) is that in this one we get the attributes in albingress instead of getting them in the loadbalancer package.

flowjo-sethd:
What are the permissions of the bucket for making this work?

Issue #257, opened 2017-10-25
Ingress do some unnecessary update for the idle timeout


piclemx:
```
Name:            name
Namespace:        ns
Address:        alb_address
Default backend:    default-http-backend:80 (<none>)
Rules:
  Host                        Path    Backends
  ----                        ----    --------
  host
                            /     service:80 (<none>)
Annotations:
Events:
  FirstSeen    LastSeen    Count    From            SubObjectPath    Type        Reason    Message
  ---------    --------    -----    ----            -------------    --------    ------    -------
  1d        1m        762    ingress-controller            Normal        MODIFY    Connection idle timeout updated to 600
```

This shouldn't have been updated 762 times. 

This not hurt anything but making unnecessary AWS calls.

bigkraig:
Resolved by #280

PR #280, opened 2017-12-06
exception handling for current idle timeout attribute

khrisrichardson:
With this snippet I was able to demonstrate that since the error isn't handled [here](https://github.com/coreos/alb-ingress-controller/blob/03cf52c2c7c0d21cf9685cc29aebec2bc02fb550/pkg/albingresses/albingresses.go#L134-L138), the value returned for idleTimeout is occasionally the default `0`, thus triggering needless load balancer attribute modifications:

```go
package main

import (
	"fmt"
	"strconv"

	"github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/session"
	"github.com/aws/aws-sdk-go/service/elbv2"
)

const (
	IdleTimeoutKey = "idle_timeout.timeout_seconds"
)

func main() {
	sess := session.Must(session.NewSession())
	svc := elbv2.New(sess)

	arn := aws.String("arn:aws:elasticloadbalancing:us-east-1:234212695392:loadbalancer/app/nonprod1-prd115-controller-17d6/1bca811ec6ea7137")

	idleTimeout := int64(0)
	in := &elbv2.DescribeLoadBalancerAttributesInput{
		LoadBalancerArn: arn,
	}
	attrs, err := svc.DescribeLoadBalancerAttributes(in)
	if err != nil {
		fmt.Printf("%v\n", err.Error())
	}
	for _, attr := range attrs.Attributes {
		if *attr.Key == IdleTimeoutKey {
			idleTimeout, err = strconv.ParseInt(*attr.Value, 10, 64)
			if err != nil {
				fmt.Printf("Failed to retrieve invalid idle timeout from ALB in AWS. Was: %s\n", *attr.Value)
			}
		}
	}
	fmt.Println(idleTimeout)
}
```

```shell
root@kube2iam-test:/go# go run main.go
60
root@kube2iam-test:/go# go run main.go
Throttling: Rate exceeded
	status code: 400, request id: 5aaf3885-d9ff-11e7-a559-2bffe66fa458
0
root@kube2iam-test:/go# go run main.go
60
```

This PR initializes the timeout to nil, and only performs the modification if not nil.

### DIFF
--- a/pkg/albingresses/albingresses.go
+++ b/pkg/albingresses/albingresses.go
@@ -4,6 +4,7 @@ import (
 	"strconv"
 	"sync"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/elbv2"
 
 	extensions "k8s.io/api/extensions/v1beta1"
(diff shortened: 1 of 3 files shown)


pkg/albingresses/albingresses.go
```diff
@@ -131,17 +132,21 @@ func AssembleIngressesFromAWS(o *AssembleIngressesFromAWSOptions) ALBIngresses {
...
+			if err != nil {
+				logger.Fatalf("Failed to retrieve attributes from ALB in AWS. Error: %s", err.Error())
+			}
 			for _, attr := range attrs.Attributes {
 				if *attr.Key == util.IdleTimeoutKey {
-					idleTimeout, err = strconv.ParseInt(*attr.Value, 10, 64)
+					idleTimeoutInt64, err := strconv.ParseInt(*attr.Value, 10, 64)
 					if err != nil {
-						logger.Fatalf("Failed to retrieve invalid idle timeout from ALB in AWS. Was: %s", *attr.Value)
+						logger.Fatalf("Failed to parse idle timeout value from ALB attribute. Was: %s", *attr.Value)
```
r0fls: This line definitely makes sense, but what's the overall purpose for switching idleTimout to a pointer?  Whatever the case, I think the DesiredIdleTimeout and CurrentIdleTimeout should have the same type.

khrisrichardson:
The reason for switching to pointer was that 0 being the default for int64 makes a less than ideal comparator when  checking if the value is set. Thus with a pointer you can compare to nil to see whether it was truly set or not.

piclemx:
This should close #257

joshrosso:
LGTM.

PR #288, opened 2017-12-20
exclude draining and unused targets from current and desired groups

khrisrichardson:
Including draining targets in the list of desired targets for a target group causes deregistered targets to be reregistered, which may adversely affect traffic destined to the load balancer if it is directed to an instance that's about to be terminated.


bigkraig:
@khrisrichardson can you add test cases?

bhvr-fmarceau:
@khrisrichardson any update on the test cases? Can we expect this to be merged soon?

PR #291, opened 2017-12-27
Fix RBAC for controller 

tyrannasaurusbanks:
After installing the chart into our cluster, the controller was complaining of missing permissions on services. I moved the 'services' resources to the more privileged group & the controller starts updating and all issues disappear - so I think this is all it needs.


tyrannasaurusbanks:
Are there any other checks I need to trigger/pass for this change?

PR #297, opened 2018-01-08
Always use local (to cluster) VPC ID for ALBs

joshrosso:
Previously the VPC ID, and thus VPC ALBs would be created within, was
detected by looking up the VPC based on the subnets the controller
intended to use. This approach can create conflicts when the subnets are
looked up by name (or perhaps other tags). This can cause another VPC
with identical subnet names to be resolved.

This new approach resolves the VPC ID based on the current EC2 instance
the alb-ingress-controller is running on.

### DIFF
--- a/pkg/aws/ec2/ec2.go
+++ b/pkg/aws/ec2/ec2.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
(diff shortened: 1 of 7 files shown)


joshrosso:
@jescarri -- testing in progress. Could you be a second set of 👀?

Let me know if the logic makes sense.

pkg/aws/ec2/ec2.go
```diff
@@ -473,38 +491,59 @@ func (e *EC2) CreateNewInstanceSG(sgName *string, sgID *string, vpcID *string) (
...
+	if len(descInstancesOutput.Reservations) < 1 {
+		return nil, fmt.Errorf("When looking up VPC ID could not identify instance. Found %d reservations"+
+			" in AWS call. Should have found atleast 1.", len(descInstancesOutput.Reservations))
+	}
+	if len(descInstancesOutput.Reservations[0].Instances) < 1 {
+		return nil, fmt.Errorf("When looking up VPC ID could not identify instance. Found %d instances"+
+			" in AWS call. Should have found atleast 1.", len(descInstancesOutput.Reservations))
 	}
 
+	vpc = descInstancesOutput.Reservations[0].Instances[0].VpcId
```
jescarri: @joshrosso isn't good check for nil or blank?

Otherwise the logic looks good.

joshrosso: @jescarri I believe i'm already doing this above. 

https://github.com/coreos/alb-ingress-controller/pull/297/files/653c431f3cdb5a316dd69a8ed249695898e4d0ad#diff-15b7ef6a224ea1db965f77e6d392cfadR532

^ can you verify this is an adequate nil check you're thinking of?

jescarri: yep, that one looks for nil or len == 0 on the output, however  no testing of ` descInstancesOutput.Reservations[0].Instances[0].VpcId` != nil or != "", just an extra safeguard, in the event aws returning "" for vpcID.

joshrosso:
@jescarri kk -- one more sanity check. nil and empty check added

joshrosso:
In testing, I found that the controller would become confused in resolving the appropriate security group since the search isn't scoped to the resolved VPC ID.

I've added in logic to ensure the VPC ID is part of the filter when looking up security groups.

joshrosso:
Similarly, I've found the annotation code, when resolving / validating subnets, also queries AWS without filtering by VPC.

I've now added this filter to subnets.

PR #309, opened 2018-01-25
Remove livenessProbe

mgoodness:
The /healthz endpoint currently checks AWS connectivity. That indicates
to me that it should be used as a readinessProbe, not a livenessProbe,
as restarting the Pod may not resolve the underlying issue.

We're also allowing the readinessProbe interval to be adjusted so AWS
throttling can be diminished.


joshrosso:
Thanks!

PR #314, opened 2018-01-30
optionally restrict scheme type by namespace

r0fls:
This allows restricting the scheme to `internal`.  When enabled, the annotation `alb.ingress.kubernetes.io/allow-external` can be added to a namespace to permit `internet-facing` scheme on ingresses in that namespace.


r0fls:
going to modify then reopen this

PR #315, opened 2018-01-30
optionally restrict scheme type

r0fls:
See https://github.com/coreos/alb-ingress-controller/compare/master...r0fls:restrict-scheme-configmap?expand=1#diff-76e731333fb756df3bff5ddb3b731c46 for a description of the changes.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -61,3 +61,20 @@ spec:
 ```
 
 > Currently, you can set only 1 namespace to watch in this flag. See [this Kubernetes issue](https://github.com/kubernetes/contrib/issues/847) for more details.
+
+### Limiting External Namespaces
+
+Setting the `--restrict-scheme` boolean flag to `true` will enable the ALB controller to check the configmap named `alb-ingress-controller-internet-facing-ingresses` for a list of approved ingresses before provisioning ALBs with an internet-facing scheme. Here is an example of that ConfigMap:
+
+```yaml
+apiVersion: v1
+data:
+ mynamespace: my-ingress-name, my-ingress-name-2
+ myothernamespace: my-other-ingress-name
+kind: ConfigMap
+metadata:
+  name: alb-ingress-controller-internet-facing-ingresses
+```
+
+
+That ConfigMap is kept in `default` if unspecified, but can moved to another with the `ALB_CONTROLLER_RESTRICT_SCHEME_CONFIG_NAMESPACE` environment variable. This can also be passed to the command line via the `restrict-scheme-namespace` flag.
--- a/pkg/albingress/albingress.go
+++ b/pkg/albingress/albingress.go
@@ -122,7 +122,7 @@ func NewALBIngressFromIngress(o *NewALBIngressFromIngressOptions) *ALBIngress {
 	}
 
 	// Load up the ingress with our current annotations.
-	newIngress.annotations, err = annotations.ParseAnnotations(o.Ingress.Annotations, o.ClusterName)
+	newIngress.annotations, err = annotations.ParseAnnotations(o.Ingress.Annotations, o.ClusterName, o.Ingress.Namespace, o.Ingress.Name)
 	if err != nil {
 		msg := fmt.Sprintf("Error parsing annotations: %s", err.Error())
 		newIngress.Reconciled = false
--- a/pkg/annotations/annotations.go
+++ b/pkg/annotations/annotations.go
@@ -13,6 +13,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/elbv2"
 	albec2 "github.com/coreos/alb-ingress-controller/pkg/aws/ec2"
+	"github.com/coreos/alb-ingress-controller/pkg/config"
 	albprom "github.com/coreos/alb-ingress-controller/pkg/prometheus"
 	"github.com/coreos/alb-ingress-controller/pkg/util/log"
 	util "github.com/coreos/alb-ingress-controller/pkg/util/types"
@@ -75,7 +76,7 @@ type PortData struct {
 // If there is an issue with an annotation, an error is returned. In the case of an error, the
 // annotations are also cached, meaning there will be no reattempt to parse annotations until the
 // cache expires or the value(s) change.
-func ParseAnnotations(annotations map[string]string, clusterName string) (*Annotations, error) {
+func ParseAnnotations(annotations map[string]string, clusterName string, ingressNamespace string, ingressName string) (*Annotations, error) {
 	if annotations == nil {
 		return nil, fmt.Errorf("Necessary annotations missing. Must include at least %s, %s, %s", subnetsKey, securityGroupsKey, schemeKey)
 	}
@@ -100,7 +101,7 @@ func ParseAnnotations(annotations map[string]string, clusterName string) (*Annot
 		a.setHealthyThresholdCount(annotations),
 		a.setUnhealthyThresholdCount(annotations),
 		a.setPorts(annotations),
-		a.setScheme(annotations),
+		a.setScheme(annotations, ingressNamespace, ingressName),
 		a.setSecurityGroups(annotations),
 		a.setSubnets(annotations, clusterName),
 		a.setSuccessCodes(annotations),
@@ -284,7 +285,7 @@ func (a *Annotations) setPorts(annotations map[string]string) error {
 	return nil
 }
 
-func (a *Annotations) setScheme(annotations map[string]string) error {
+func (a *Annotations) setScheme(annotations map[string]string, ingressNamespace, ingressName string) error {
 	switch {
 	case annotations[schemeKey] == "":
 		return fmt.Errorf(`Necessary annotations missing. Must include %s`, schemeKey)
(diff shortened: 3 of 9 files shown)


docs/configuration.md
```diff
@@ -61,3 +61,20 @@ spec:
 ```
 
 > Currently, you can set only 1 namespace to watch in this flag. See [this Kubernetes issue](https://github.com/kubernetes/contrib/issues/847) for more details.
+
+### Limiting External Namespaces
+
+Setting the `--restrict-scheme` boolean flag to `true` will enable the ALB controller to check configmap named `alb-ingress-controller-internet-facing-ingresses` for a list of approved ingresses before provisioning ALBs with an internet-facing scheme. Here is an example of that ConfigMap:
```
joshrosso: `to check configmap named` -> `to check the configmap named`

docs/configuration.md
```diff
@@ -61,3 +61,20 @@ spec:
...
+data:
+ mynamespace: my-ingress-name
+ myothernamespace: my-other-ingress-name
+kind: ConfigMap
+metadata:
+  name: alb-ingress-controller-internet-facing-ingresses
+```
+
+
+That ConfigMap is kept in `default` if unspecified, but can moved to another with the `ALB_CONTROLLER_RESTRICT_SCHEME_CONFIG_NAMESPACE` environment variable.
```
joshrosso: Perhaps mentions flag here, `--restrict-scheme-namespace` similar to how you've mentioned the flag `--restrict-scheme` above.

docs/configuration.md
```diff
@@ -61,3 +61,20 @@ spec:
...
 > Currently, you can set only 1 namespace to watch in this flag. See [this Kubernetes issue](https://github.com/kubernetes/contrib/issues/847) for more details.
+
+### Limiting External Namespaces
+
+Setting the `--restrict-scheme` boolean flag to `true` will enable the ALB controller to check configmap named `alb-ingress-controller-internet-facing-ingresses` for a list of approved ingresses before provisioning ALBs with an internet-facing scheme. Here is an example of that ConfigMap:
+
+```yaml
+apiVersion: v1
+data:
+ mynamespace: my-ingress-name
```
joshrosso: Based on the code, in order to specify multiple ingresses in the same namespace, we would **not use**:

```
data:
  mynamespace: my-ingress-name, my-ingress-name2
  myothernamespace: my-other-ingress-name
```

Instead, we **would use**:

```
data:
  mynamespace: my-ingress-name, 
  mynamespace: my-ingress-name2
  myothernamespace: my-other-ingress-name
```

If my reading of the code is valid, then perhaps we include an example like the latter above. This way it's obvious how to do multiple ingresses for the same namespaces.

pkg/albingress/albingress.go
```diff
@@ -151,6 +155,18 @@ func NewALBIngressFromIngress(o *NewALBIngressFromIngressOptions) *ALBIngress {
 		Tags:                 newIngress.Tags(),
 	})
 
+	// If internet-facing is restricted to certain namespaces, then validate this is one of them
+	if o.RestrictScheme && *newIngress.annotations.Scheme == "internet-facing" {
```
joshrosso: I think we should consider the following with validating whether `alb.ingress.kubernetes.io/scheme=internet-facing` should be rejected.

- Move this logic to the `annotations` package.
  - Inside the following logic https://github.com/coreos/alb-ingress-controller/blob/master/pkg/annotations/annotations.go#L287-L296
  - See how we've done similar for validating / rejecting cert ARNs (https://github.com/coreos/alb-ingress-controller/blob/master/pkg/annotations/annotations.go#L126-L137)
  - The goal here being to contain all annotations parse / validation in one place. Additionally to catch issues with this annotation early.

pkg/util/types/types.go
```diff
@@ -73,6 +78,36 @@ func (a AWSStringSlice) Hash() *string {
...
+	// creates the clientset
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		logger.Errorf(err.Error())
+	}
+	cm, err := clientset.CoreV1().ConfigMaps(namespace).Get(restrictIngressConfigMap, metav1.GetOptions{})
+	if err != nil {
+		logger.Errorf(err.Error())
+	}
+	return cm.Data
```
joshrosso: I'd like to Maintain a controller-level cache of this configmap returned here.
  - I want to prevent an ingress event, which causes all ingresses to be re-evaluated, from hitting the api server N time(s) to lookup 
  - The result being, most lookups to [IngressAllowedExternal](https://github.com/coreos/alb-ingress-controller/pull/315/files#diff-902bed7fc500e15b13b452a38d1269acR101) would result in a local lookup and **not** a new `go-client` call to the k8s api server.
  - Similar to the previous comment, this logic likely best lives in the `annotations` package, in [validation.go] (https://github.com/coreos/alb-ingress-controller/blob/master/pkg/annotations/validation.go) specifically.

pkg/util/types/types.go
```diff
@@ -73,6 +78,36 @@ func (a AWSStringSlice) Hash() *string {
 	return aws.String(output)
 }
 
+// Get the configmap that holds the whitelisted internet facing
+func GetInternetFacingConfigMap(namespace, ingressName string) map[string]string {
```
joshrosso: If my understanding of the design (see my comment on `configuration.md`) is that we expect individual to name namespace as so:

```
data:
  mynamespace: my-ingress-name, 
  mynamespace: my-ingress-name2
  myothernamespace: my-other-ingress-name
```

I'm uncertain how the returned `map[string]string` here would function.

In that being captured in a map, you would end up with duplicate keys. Specifically, when there are multiple ingresses in the same namespace you wish to allow the `internet-facing` scheme. ( see example above ).

So you'd have a resulting conflict such as https://play.golang.org/p/4LhZeJBhOTX, in our case, i'm guessing the latter key would simply overwrite the former.

Were you able to validate a scenario with multiple ingresses in the same namespace?

pkg/controller/alb-controller.go
```diff
@@ -244,6 +249,14 @@ func (ac *ALBController) Info() *ingress.BackendInfo {
 func (ac *ALBController) ConfigureFlags(pf *pflag.FlagSet) {
 	pf.StringVar(&ac.clusterName, "clusterName", os.Getenv("CLUSTER_NAME"), "Cluster Name (required)")
 
+	rs, _ := strconv.ParseBool(os.Getenv("ALB_CONTROLLER_RESTRICT_SCHEME"))
+	ns := os.Getenv("ALB_CONTROLLER_RESTRICT_SCHEME_CONFIG_NAMESPACE")
+	pf.BoolVar(&ac.restrictScheme, "restrict-scheme", rs, "Restrict the scheme to internal except for whitelisted namespaces (defaults to false)")
+	if ns == "" {
+		ns = "default"
+	}
+	pf.StringVar(&ac.restrictSchemeNamespace, "restrict-scheme-namespace", ns, "The namespace that holds the configmap with the allowed ingresses.")
```
joshrosso: add `Only respected when restrict-scheme is true.`

pkg/util/types/types.go
r0fls: The caching is addressed now here: https://github.com/coreos/alb-ingress-controller/pull/315/files#diff-f7784086caeb4e10b07629efa6b4d14aR295

joshrosso:
@r0fls, we'll need to update some of the annotations tests based on the new function signatures

https://travis-ci.org/coreos/alb-ingress-controller/builds/336374451?utm_source=github_status&utm_medium=notification

pkg/annotations/annotations.go
```diff
@@ -284,15 +285,27 @@ func (a *Annotations) setPorts(annotations map[string]string) error {
...
 	}
 	a.Scheme = aws.String(annotations[schemeKey])
-	return nil
+	cacheKey := fmt.Sprintf("scheme-%s-%s-%s-%s", config.RestrictScheme, config.RestrictSchemeNamespace, ingressNamespace, ingressName)
+	var err error
+	if item := cacheLookup(cacheKey); item != nil {
+		if item.Value() != nil {
+			err = item.Value().(error)
+		} else {
+			err = nil
```
joshrosso: Wouldn't `err` already be `nil` here? Perhaps this `else` block is unneeded.

pkg/annotations/annotations.go
```diff
@@ -284,15 +285,27 @@ func (a *Annotations) setPorts(annotations map[string]string) error {
...
+		}
+	} else {
+		err = a.ValidateScheme(ingressNamespace, ingressName)
+		cache.Set(cacheKey, err, time.Minute*60)
```
joshrosso: If I'm reading this correctly, we only cache when `ValidateScheme` returns an error (`internal-facing` is rejected)?

What do you think about caching a boolean instead? So that every time the ingress is parsed, and its allowed to be `internet-facing` we don't lookup the configmap again. Perhaps this also warrants bringing the cache down to something like `10` minutes?

joshrosso: I'd hope this would also simplify logic, making it more readable.

Thinking it'd become something closer to:

```
if item := cacheLookup(cacheKey); item != nil {
  if item.Value().(bool) {
    return nil  
  }
  return fmt.Errorf("SOME ERROR") 
}

isValid := a.ValidateScheme(ingressNamespace, ingressName)
cache.Set(cacheKey, isValid, time.Minute*10)
if !isValid {
  return fmt.Errorf("SOME ERROR")
}
return nil
``` 

Let me know if that resonates with you @r0fls.

joshrosso: updated that just now ^^

pkg/controller/alb-controller.go
```diff
@@ -244,6 +245,15 @@ func (ac *ALBController) Info() *ingress.BackendInfo {
 func (ac *ALBController) ConfigureFlags(pf *pflag.FlagSet) {
 	pf.StringVar(&ac.clusterName, "clusterName", os.Getenv("CLUSTER_NAME"), "Cluster Name (required)")
 
+	rs, _ := strconv.ParseBool(os.Getenv("ALB_CONTROLLER_RESTRICT_SCHEME"))
```
joshrosso: We should do something with this error?

Perhaps we `panic()` stating something along the lines `ALB_CONTROLLER_RESTRICT_SCHEME` must be true or false. Instead was `X`.

PR #331, opened 2018-02-21
allow disabling the liveness probe

r0fls:
During periods of AWS API throttling, the liveness probe can lead to a crash loop in the controller. This makes the liveness probe configurable, so it can be disabled to prevent the crash loop from happening.


mgoodness:
The livenessProbe was disabled completely in #309

Issue #326, opened 2018-02-07
Error from server (BadRequest): error when creating "alb-ingress-controller.yaml": Deployment in version "v1beta1" cannot be handled as a Deployment: v1beta1.Deployment: Spec: v1beta1.DeploymentSpec: 


holoGDM:
Hello, when im trying to star alb controller according to your documentation im getting:

Error from server (BadRequest): error when creating "alb-ingress-controller.yaml": Deployment in version "v1beta1" cannot be handled as a Deployment: v1beta1.Deployment: Spec: v1beta1.DeploymentSpec: 

What is strange default-backend started with same settings without problem

Changeing it to: apps/v1beta2 getting same error

evilfrog:
@holoGDM 

Changing this:

```
# Maximum number of times to retry the aws calls.
# defaults to 20.
- name: AWS_MAX_RETRIES
  value: 20
```

To this:
```
# Maximum number of times to retry the aws calls.
# defaults to 20.
- name: AWS_MAX_RETRIES
  value: "20"
```
Solved the issue for me.

Note: quotes around numeric value.

holoGDM:
Thank You very much. That solves my problem too :+1:

PR #340, opened 2018-03-05
A few example clarifications

claudenm:
This PR makes a few changes to address issues I came across while running through the tutorial.  I added a note about having more than one subnet, and I added a note about security groups needing http access.  Both pretty obvious from error messages or standard debugging, but I thought it worthwhile to put a note in for future readers.  It also addresses #326 (which was one of the issues I came across).


bigkraig:
Thanks @claudenm!

PR #346, opened 2018-03-15
fix ingress deletion bug

r0fls:


r0fls:
If `i` is the last item in the `ac.ALBIngresses` this will cause a `"slice bounds out of range"` panic, this fixes it

PR #366, opened 2018-04-17
prometheus/managed ingresses by namespace count

jescarri:
Added namespace label to albingress_managed_ingresses metric.


pkg/controller/alb-controller.go
```diff
@@ -165,7 +166,16 @@ func (ac *ALBController) update() {
 	newIngresses = append(newIngresses, ac.ALBIngresses.RemovedIngresses(newIngresses)...)
 
 	// Update the prometheus gauge
-	albprom.ManagedIngresses.Set(float64(len(newIngresses)))
+	ingressesByNamespace := map[string]int{}
+	logger.Infof("Ingress count: %d\n", len(newIngresses))
```
bigkraig: Can you change this to a debug?

jescarri: Fixed.

PR #373, opened 2018-05-01
Fix build info injection [Small]

tyrannasaurusbanks:
I noticed that 1.0-alpha.7 was printing the defaults for build info upon start up, so here's a tiny PR to fix that by switching global build and release vars from private to public.


bigkraig:
Thanks!

PR #374, opened 2018-05-02
Add ignore-host-header annotation

### LINKED ISSUE #371: Add option to ignore Host headers
At the moment if a hostname is specified the listeners get created with a Host header check rule, which forces users of the ALB to send that Host header. The logic is [here](https://github.com/coreos/alb-ingress-controller/blob/01441fa138e7255aa8632b9c9a7147baf47deb03/pkg/alb/rule/rule.go#L56):

```golang
if o.Hostname != "" {
	r.Conditions = append(r.Conditions, &elbv2.RuleCondition{
		Field:  aws.String("host-header"),
		Values: []*string{aws.String(o.Hostname)},
	})
}
```

That impedes using a higher level load balancer on top of the ALB, as the expected host would point to the ALB itself, and the higher load balancer would have a different hostname.

As we're using an NLB on top of the ALB, for supporting VPC Endpoints (aka PrivateLinks), and [external-dns](https://github.com/kubernetes-incubator/external-dns) to create all required hostnames, this blocks our use case.

My proposal is to add an annotation to allow ignoring the Host header check, e.g.:

`alb.ingress.kubernetes.io/ignore-host-header`

riccardofreixo:
When not empty the Host header check won't be enabled as part of the ALB listener routing rules.

This commit closes #371.

### DIFF
--- a/pkg/annotations/annotations.go
+++ b/pkg/annotations/annotations.go
@@ -42,6 +42,7 @@ const (
 	subnetsKey                    = "alb.ingress.kubernetes.io/subnets"
 	successCodesKey               = "alb.ingress.kubernetes.io/successCodes"
 	tagsKey                       = "alb.ingress.kubernetes.io/tags"
+	ignoreHostHeader              = "alb.ingress.kubernetes.io/ignore-host-header"
 	clusterTagKey                 = "tag:kubernetes.io/cluster"
 	clusterTagValue               = "shared"
 	albRoleTagKey                 = "tag:kubernetes.io/role/alb-ingress"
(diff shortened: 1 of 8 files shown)


pkg/annotations/annotations.go
```diff
@@ -612,6 +615,15 @@ func (a *Annotations) setTags(annotations map[string]string) error {
 	return nil
 }
 
+func (a *Annotations) setIgnoreHostHeader(annotations map[string]string) error {
+	if annotations[ignoreHostHeader] == "" {
```
bigkraig: Can we use a ParseBool here rather than defaulting to true if its not an empty string?

https://golang.org/pkg/strconv/#ParseBool

riccardofreixo: Sure, done 8f5f2675f4aae64cb73ade76cba514bdf901efda

riccardofreixo:
Changed it to use ParseBool and added documentation.

PR #380, opened 2018-05-08
Fix so if no connection-idle-timeout is set 

gaganapplatix:
This makes it so we don't try to do SetIdleTimeout since time of 0 is invalid which is what is parsed from annotation.


gaganapplatix:
Fixes https://github.com/coreos/alb-ingress-controller/issues/361

bigkraig:
Thanks!

Issue #383, opened 2018-05-16
Instances are not added/removed from Target Group after it is created


jkemp101:
It looks like something broke between version `1.0-alpha.8` and `67c350d` with target groups getting updated when the available nodes change.  I also tried the latest version `ac36437` and got the same result.  I happened to realize only 1 instance was in the target group even though the cluster had 3 nodes.

With `1.0-alpha.8` I get the following when I increased the instances in the ASG. The target group is modified as expected.
```
I0516 14:26:09.435104       1 loadbalancer.go:194] [ALB-INGRESS] [test/nginx-ingress] [DEBUG]: No modification of ELBV2 (ALB) required.
I0516 14:26:09.435136       1 targetgroup.go:323] [ALB-INGRESS] [test/nginx-ingress] [DEBUG]: Targets need to be changed.
I0516 14:26:09.435143       1 targetgroup.go:159] [ALB-INGRESS] [test/nginx-ingress] [INFO]: Start TargetGroup modification.
I0516 14:26:09.435155       1 targetgroup.go:323] [ALB-INGRESS] [test/nginx-ingress] [DEBUG]: Targets need to be changed.
I0516 14:26:09.649060       1 targetgroup.go:164] [ALB-INGRESS] [test/nginx-ingress] [INFO]: Succeeded TargetGroup modification. ARN: arn:aws:elasticloadbalancing:us-west-2:892789744163:targetgroup/test9k8susw-32596-HTTP-1cd5e8a/3145cca7dfe7b97b | Name: test9k8susw-32596-HTTP-1cd5e8a.
I0516 14:26:09.649093       1 listener.go:112] [ALB-INGRESS] [test/nginx-ingress] [DEBUG]: No listener modification required.
```

But with `67c350d` I get the following when I increase the ASG instance count:
```
I0516 14:22:30.140538       1 loadbalancer.go:703] [ALB-INGRESS] [test/nginx-ingress] [DEBUG]: test9k8susw-test-nginxingr-7831 checking if WAF needs update (old: <nil>, new: <nil>)
I0516 14:22:30.140550       1 loadbalancer.go:217] [ALB-INGRESS] [test/nginx-ingress] [DEBUG]: No modification of ELBV2 (ALB) required.
I0516 14:22:30.140567       1 targetgroup.go:169] [ALB-INGRESS] [test/nginx-ingress] [DEBUG]: No TargetGroup modification required.
[...]

PR #384, opened 2018-05-16
fix DescribeTargetGroupTargetsForArn with empty target list

r0fls:
fixes: #383

### DIFF
--- a/pkg/aws/elbv2/elbv2.go
+++ b/pkg/aws/elbv2/elbv2.go
@@ -35,7 +35,7 @@ type ELBV2API interface {
 	UpdateAttributes(arn *string, new []*elbv2.LoadBalancerAttribute) error
 	RemoveTargetGroup(in elbv2.DeleteTargetGroupInput) error
 	DescribeTagsForArn(arn *string) (util.Tags, error)
-	DescribeTargetGroupTargetsForArn(arn *string, targets []*elbv2.TargetDescription) (util.AWSStringSlice, error)
+	DescribeTargetGroupTargetsForArn(arn *string, targets ...[]*elbv2.TargetDescription) (util.AWSStringSlice, error)
 	RemoveListener(in elbv2.DeleteListenerInput) error
 	DescribeTargetGroupsForLoadBalancer(loadBalancerArn *string) ([]*elbv2.TargetGroup, error)
 	DescribeListenersForLoadBalancer(loadBalancerArn *string) ([]*elbv2.Listener, error)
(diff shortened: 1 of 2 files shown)


bigkraig:
I’m operating off a phone in Costa Rica right now. So my review is limited, but I’d like to get this fixed for you guys. Is there a reason why we only look for len==1 and not join the lists whenever it’s => 1? And if a third party can confirm this fixes it for them I’m good with a phone merge. I can’t hover to see who 👍

r0fls:
@bigkraig 😂 thanks for checking this out while you're on vacation. Jesus and Khris form the 👍 party

jkemp101:
This didn't seem to fix it for me.  Did it work for you guys?  Can we add a test for this situation to catch issues like this in the future?

jkemp101:
The newly added instance is falling into `case elbv2.TargetHealthStateEnumUnused:` and therefore not being added to `result`.

khrisrichardson:
@jkemp101 check out the reasons for state unused [here](https://docs.aws.amazon.com/sdk-for-go/api/service/elbv2/#TargetHealth). I don't believe the target should be added in any of those circumstances, but am willing to hear arguments to the contrary.

```
    // If the target state is unused, the reason code can be one of the following
    // values:
    //
    //    * Target.NotRegistered - The target is not registered with the target
    //    group.
    //
    //    * Target.NotInUse - The target group is not used by any load balancer
    //    or the target is in an Availability Zone that is not enabled for its load
    //    balancer.
    //
    //    * Target.IpUnusable - The target IP address is reserved for use by a load
    //    balancer.
    //
    //    * Target.InvalidState - The target is in the stopped or terminated state.
```

jkemp101:
I assume it is hitting that state because `Target.NotRegistered` which would make sense because it is a new instance that hasn't been added to the target group yet but that is what is needed to be done.  Can you confirm if this is working for you or anyone?  AWS instances fail all of the time.  As far as I can tell if an instance fails and is replaced, the new one will not get added to the target group.  Eventually you would be left with no instances in the target group and your site would go down.

I'm going back to CLBs via regular k8s LoadBalancer services for now because of this issue and my other concerns laid out in https://github.com/coreos/alb-ingress-controller/issues/375.  I need to move forward with other parts of my project and will revisit k8s ALB options later.

roffe:
PR fixed my problem with ALB's loosing all the targets when i do a kops rolling upgrade on the cluster and replace all nodes

pkg/aws/elbv2/elbv2.go
```diff
@@ -207,20 +207,13 @@ func (e *ELBV2) DescribeTagsForArn(arn *string) (util.Tags, error) {
...
-		})
-	} else {
-		targetHealth, err = e.DescribeTargetHealth(&elbv2.DescribeTargetHealthInput{
-			TargetGroupArn: arn,
-		})
+	opts := &elbv2.DescribeTargetHealthInput{
+		TargetGroupArn: arn,
+	}
+	for _, targets := range targets {
+		mergedTargets = append(opts.Targets, target...)
```
khrisrichardson: I think you mean `opts.Targets = append....`

r0fls: Whoops! ☕️

r0fls: Bah, I did a `go build cmd` but I guess that wouldn't have caught it anyway...

PR #399, opened 2018-06-20
Add kube2iam documentation

zioalex:
Add kube2iam documentation


zioalex:
@bigkraig why did you close it?

bigkraig:
@zioalex isn't this a duplicate of #400?

zioalex:
Ah yep wrongly it was! My idea was to have 2 different commit...but then they landed in the same PR.

PR #408, opened 2018-06-22
Update version to alpha.9 everywhere

### LINKED ISSUE #407: version update to alpha.9 is not done in all places
version should be updated to latest at below places
build
readme
examples

raoofm:
Update version to alpha.9 everywhere

Fixes #407


bigkraig:
Thanks!

PR #427, opened 2018-06-26
Update broken readme links

bendrucker:
Caught some broken links in the readme and tried to correct them to what seemed like suitable replacement pages


bigkraig:
Thanks!

Issue #429, opened 2018-06-27
Annotation parser does not have correct value for clusterName


drakedevel:
When I run 1.0-beta.1 with autodiscovered subnets, I get this error:

```
I0627 01:24:38.019996       1 event.go:218] Event(v1.ObjectReference{Kind:"Ingress", Namespace:"redacted", Name:"redacted", UID:"redacted", APIVersion:"extensions", ResourceVersion:"14402657", FieldPath:""}): type: 'Warning' reason: 'ERROR' Error parsing annotations: Retrieval of subnets failed to resolve 2 qualified subnets. Subnets must contain the tag:kubernetes.io/cluster/ tag with a value of shared and the tag:kubernetes.io/role/alb-ingress tag signifying it should be used for ALBs Additionally, there must be at least 2 subnets with unique availability zones as required by ALBs. Either tag subnets to meet this requirement or use the subnets annotation on the ingress resource to explicitly call out what subnets to use for ALB creation. The subnets that did resolve were []. (cache hit)
```

(Note the `... must contain the tag:kubernetes.io/cluster/ tag with ...` part)

Given how that log line is formatted, I believe that means that the ValidatingAnnotationFactory has clusterName = "". It looks like NewValidatingAnnotationFactory is called in NewALBController, which is (necessarily) called before ConfigureFlags. Since ConfigureFlags is responsible for populating the clusterName variable, this means the ValidatingAnnotationFactory gets an empty string. If this is correct, then I believe this behavior is a regression introduced with #418 .

I'm not terribly familiar with Go so it's possible I'm missing something with mutable references or something, but then I'm really not sure why I'm getting the above behavior (which is a regression in my deployment).

drakedevel:
I just confirmed the same behavior in 1.0-alpha.9; 1.0-alpha.3 works correctly.

tatobi:
I confirm the same problem as well. The 1.0-alpha.3 works fine, the  1.0-beta.1 does not find the subnets wuith the same (correct) Tags. The new version works only if I explicitely define the subnets in pod deployment manifest yaml file.

bigkraig:
can you guys give me an example of how to reproduce it? we use a common set of security groups and i haven't had to leverage the tagging logic (nor did i write it) so i am not fluent on how it should be working

bigkraig:
[...]

PR #430, opened 2018-06-27
set cluster name in ValidatingANnotationFactory

r0fls:
fixes: #429


bigkraig:
This leaves us setting the `NewValidatingAnnotationFactory` twice, I think the real problem is at https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/pkg/controller/alb-controller.go#L99 which should be `conf.ClusterName` or alternatively (and maybe better) we set the `ac.clusterName` at the start of the func:

```
	ac := &albController{
		awsChecks:   make(map[string]func() error),
+		clusterName: conf.ClusterName,
	}
```

bigkraig:
#431

PR #449, opened 2018-07-10
Support routing directly to pods

### LINKED ISSUE #343: Support Amazon ENI CNI plugin
I'm using `kops` with the new [`--networking amazon-vpc-routed-eni`]( https://github.com/aws/amazon-vpc-cni-k8s)

This only happens when I don't manually add the `alb.ingress.kubernetes.io/security-groups` annotation.

Here is the log:

```
I0313 15:34:44.152237       1 event.go:218] Event(v1.ObjectReference{Kind:"Ingress", Namespace:"echoserver", Name:"echoserver", UID:"0b051e24-26d4-11e8-ab29-0669b4832cd4", APIVersion:"extensions", ResourceVersion:"108737", FieldPath:""}): type: 'Warning' reason: 'ERROR' Error registering targets to target group k8s-32205-HTTP-94c346f: InvalidInstanceID: There are multiple interfaces attached to instance 'i-08a3692459f8f4731'. Please specify an interface ID for the operation instead.
```

Needless to say, the ingress completely fails to work.

I believe the error is coming from here [pkg/alb/targetgroup/targetgroup.go:356](https://github.com/coreos/alb-ingress-controller/blob/master/pkg/alb/targetgroup/targetgroup.go#L356)

bigkraig:
Resolves #343, this annotation will configure the target group to route directly to the service endpoint `pod ip:port`.


liwenwu-amazon:
This PR also solves a [aws-vpc-cni issue](https://github.com/aws/amazon-vpc-cni-k8s/issues/75)

lbernail:
We would definitely use this
I'll try to build a controller from this branch this week to test it and give feedback

Issue #274, opened 2017-11-24
"missing required field, SetSecurityGroupsInput.SecurityGroups."


fredsted:
We initially had 1 k8s master and 1 node, and installed this ingress controller. 

After we added a new k8s node and scaled our deployment, the new AWS instance didn't get the ALB security group added, which results in the ALB timing out in the health check and not routing the traffic to that node.

However, this issue only happens on one of the ingresses. We have 2 other ingresses where the security group was added, and the traffic flows as it should.

I also noticed the ingress controller pod has 200+ restarts (we installed it yesterday.)

In the alb-ingress-controller logs, we get this every minute: (`bots/faulty-ingress` is the one that we're having issues with)

``` 
I1124 13:09:20.631125       1 loadbalancer.go:190] [ALB-INGRESS] [bots/faulty-ingress] [INFO]: Start ELBV2 (ALB) modification.
I1124 13:09:20.631188       1 loadbalancer.go:319] [ALB-INGRESS] [bots/faulty-ingress] [INFO]: Start ELBV2 security groups modification.
E1124 13:09:20.631251       1 loadbalancer.go:325] [ALB-INGRESS] [bots/faulty-ingress] [ERROR]: Failed ELBV2 security groups modification: InvalidParameter: 1 validation error(s) found.
E1124 13:09:20.631287       1 loadbalancer.go:325] [ALB-INGRESS] [bots/faulty-ingress] [ERROR]: - missing required field, SetSecurityGroupsInput.SecurityGroups.
E1124 13:09:20.631298       1 loadbalancer.go:325] [ALB-INGRESS] [bots/faulty-ingress] [ERROR]:
E1124 13:09:20.631358       1 albingress.go:322] [ALB-INGRESS] [bots/faulty-ingress] [ERROR]: Failed to reconcile state on this ingress
E1124 13:09:20.631375       1 albingress.go:324] [ALB-INGRESS] [bots/faulty-ingress] [ERROR]:  - InvalidParameter: 1 validation error(s) found.
E1124 13:09:20.631380       1 albingress.go:324] [ALB-INGRESS] [bots/faulty-ingress] [ERROR]: - missing required field, SetSecurityGroupsInput.SecurityGroups.
E1124 13:09:20.631383       1 albingress.go:324] [ALB-INGRESS] [bots/faulty-ingress] [ERROR]:
I1124 13:09:20.631710       1 controller.go:486] ingress backend successfully reloaded...
``` 

[...]

PR #451, opened 2018-07-11
alb-ingress-controller tag changed

kanolato:
The default `1.0-beta.2` tag makes the pod keep restarting, because the liveness probe fails and Kubernetes tries to heal it. 
The `latest` tag seems to work good. We are not having troubles with the restarts anymore.

### DIFF
--- a/alb-ingress-controller-helm/README.md
+++ b/alb-ingress-controller-helm/README.md
@@ -47,7 +47,7 @@ Parameter | Description | Default
 `awsRegion` | (REQUIRED) AWS region in which this ingress controller will operate | `us-west-1`
 `clusterName` | (REQUIRED) Resources created by the ALB Ingress controller will be prefixed with this string | `k8s`
 `controller.image.repository` | controller container image repository | `quay.io/coreos/alb-ingress-controller`
-`controller.image.tag` | controller container image tag | `0.8`
+`controller.image.tag` | controller container image tag | `latest`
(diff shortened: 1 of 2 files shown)


bigkraig:
Which liveness probe are you using? there was a recent PR to create a new `/alive` endpoint. #406

kanolato:
Im using the default liveness probe, given by the [deployment](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/alb-ingress-controller-helm/templates/controller-deployment.yaml) in the helm template.
```
livenessProbe:
            httpGet:
              path: /alive
              port: 8080
              scheme: HTTP
            initialDelaySeconds: 30
            periodSeconds: 60
```
but its not working on the `1.0-beta.2` it answers with:
```
HTTP/1.1 404 Not Found
Content-Length: 19
Content-Type: text/plain; charset=utf-8
Date: Thu, 12 Jul 2018 02:11:50 GMT
X-Content-Type-Options: nosniff

404 page not found
```
I think it would be nice to have an image tag that handles the liveness probe properly as it is on the template. Thats what im suggesting with the PR. Maybe thats why in #274 he says _I also noticed the ingress controller pod has 200+ restarts (we installed it yesterday.)_

bigkraig:
@kanolato I just pushed a 1.0-beta.3 release that includes the `/alive` endpoint and updates the docs. Thanks for catching this!

PR #453, opened 2018-07-12
Change healthz port

brandonwillard:
We were having the same readiness probe problem as https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/281 and the changes in this PR appear to fix it.

FYI: an alternative fix might be possible by changing the entrypoint in the `alb-ingress-controller` Dockerfile to specify `--healthz-port=8080`.


brandonwillard:
Still seeing some issues with the liveness probe, but the readiness probe works fine.

Issue #253, opened 2017-10-20
Subnet Selection (auto-detection) - Suggestion for Change.


whithajess:
I am definitely a fan of using auto-detection

Wondering if it could be more closely aligned to [Kops](https://github.com/kubernetes/kops) given this is the suggested way to install Kuberenetes on AWS.

The tags given to subnets by default with `kops` are:

```ruby
KubernetesCluster | $CLUSTER_NAME
kubernetes.io/cluster/$CLUSTER_NAME | owned
```

blytheaw:
Kops seems to recommend naming clusters to match the domain name you set up for DNS records in order to ensure uniqueness between clusters. This could commonly be a pretty long cluster name; would this not break the auto-detection feature since the alb-ingress has a character limit in CLUSTER_NAME, which has to match the same tag on the subnets used by Kops?

I'm not sure if this is a separate issue to be discussed.

whithajess:
@blytheaw That was fixed via https://github.com/coreos/alb-ingress-controller/issues/223.

Specifically https://github.com/coreos/alb-ingress-controller/commit/918d034df4f3994ca34f5d70d3de81f74d55e934

If you are still having problems I would suggest raising another issue.

joshrosso:
^ 👍 Let us know!

joshrosso:
Sorry didn't mean to close -- @whithajess, you're suggesting re-using those existing kops annotations, correct?

whithajess:
@joshrosso Yeah basically it would just be nice if it found default tags created by kops, and by default would auto-select them without me having to either manually enter them in either my ingress setup or by tagging AWS resources.

ese:
These tags are a kubernetes convention in aws beyond kops 
https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/aws/tags.go#L30

whithajess:
Oh cool so then follow Kubernetes convention not just Kops :) thanks for the info @ese

cyburgee:
I too would like to see auto detection of subnets work with a cluster tag value of "owned". I can open a PR for the change if it wouldn't cause folks problems

jhohertz:
Being able to use "owned" lifecycle subnets in addition to "shared" would be nice for being able to use w/ kops out of the box. However this would also require being able to change the other tag.

This component expects: kubernetes.io/role/alb-ingress
[...]

PR #455, opened 2018-07-13
Adopt the  kubernetes.io/role/internal-elb and kubernetes.io/role/elb tags

bigkraig:
This PR leverages the common Kubernetes subnet tags to auto-detect internal and internet-facing subnets.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -24,12 +24,12 @@ The controller determines subnets to deploy each ALB to based on an annotation o
 ##### Via tags on the subnets
 When subnet annotations are not present, the controller will attempt to choose the best subnets for deploying the ALBs. It uses the following tag criteria to determine the subnets it should use.
 
-* `kubernetes.io/cluster/$CLUSTER_NAME: = "shared"` `$CLUSTER_NAME` must match the `CLUSTER_NAME` environment variable on the controller.
+- `kubernetes.io/cluster/$CLUSTER_NAME` equal to `shared` or `owned`. `$CLUSTER_NAME` must match the `CLUSTER_NAME` environment variable on the controller.
 
 And one of the following:
 
-* `kubernetes.io/role/internal-elb: ""` For internal load balancers
-* `kubernetes.io/role/elb = ""` For internet-facing load balancers
+- `kubernetes.io/role/internal-elb: ""` For internal load balancers
+- `kubernetes.io/role/elb = ""` For internet-facing load balancers
 
 ### Security Group Selection
 
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -18,19 +18,24 @@ An example policy with the minimum rights can be found at [examples/iam-policy.j
 
 The controller determines subnets to deploy each ALB to based on an annotation or auto-detection.
 
-- annotation: `alb.ingress.kubernetes.io/subnets` may be specified in each ingress resource with the subnet IDs or `Name` tags. This allows for flexibility in where ALBs land. This list of subnets must include 2 or more that exist in unique availability zones. See the [annotations documentation](ingress-resources.md#annotations) for more details.
+##### Via annotation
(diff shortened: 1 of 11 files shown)


docs/setup.md
```diff
@@ -18,19 +18,24 @@ An example policy with the minimum rights can be found at [examples/iam-policy.j
...
+##### Via annotation
+`alb.ingress.kubernetes.io/subnets` may be specified in each ingress resource with the subnet IDs or `Name` tags. This allows for flexibility in where ALBs land. This list of subnets must include 2 or more that exist in unique availability zones. See the [annotations documentation](ingress-resources.md#annotations) for more details.
 
-- auto-detection: When subnet annotations are not present, the controller will attempt to choose the best subnets for deploying the ALBs. It uses the following tag criteria to determine the subnets it should use.
+##### Via tags on the subnets
+When subnet annotations are not present, the controller will attempt to choose the best subnets for deploying the ALBs. It uses the following tag criteria to determine the subnets it should use.
 
-	- `kubernetes.io/cluster/$CLUSTER_NAME`=`shared` where `$CLUSTER_NAME` matches the `CLUSTER_NAME` environment variable from the `alb-ingress-controller.yaml` manifest.
+* `kubernetes.io/cluster/$CLUSTER_NAME: = "shared"` `$CLUSTER_NAME` must match the `CLUSTER_NAME` environment variable on the controller.
```
whithajess: This should be able to be `shared` or `owned` see: https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/aws/tags.go#L43

pkg/aws/albec2/ec2.go
```diff
@@ -16,12 +20,21 @@ import (
...
 
+var cache = ccache.New(ccache.Configure())
+
 const (
 	instSpecifierTag = "instance"
 	ManagedByKey     = "ManagedBy"
 	ManagedByValue   = "alb-ingress"
+
+	tagNameCluster  = "kubernetes.io/cluster"
+	tagValueCluster = "shared"
```
whithajess: Same comment as above ^

pkg/aws/albrgt/rgt.go
```diff
@@ -43,45 +46,77 @@ func (r *RGT) GetResources(clusterName *string) (*Resources, error) {
...
-			req, _ := r.GetResourcesRequest(&params)
-			return req, nil
+		&resourcegroupstaggingapi.GetResourcesInput{
+			ResourceTypeFilters: []*string{
+				aws.String("elasticloadbalancing"),
+			},
+			TagFilters: []*resourcegroupstaggingapi.TagFilter{
+				&resourcegroupstaggingapi.TagFilter{
+					Key:    aws.String("kubernetes.io/cluster/" + *clusterName),
+					Values: []*string{aws.String("owned")},
```
whithajess: At here I think you want `shared` or `owned` right?

bigkraig:
Thanks for the review @whithajess!

argeualcantara:
Considering: 

> A subnet is public if it routes to an internet gateway and it's instances will be reachable if they are attached to an Elastic IP.

Can't we use the routing tables to determine if the subnet is public or private, the public ones have `0.0.0.0/0` set to target `igw-*`. we could use something like this to avoid the need of adding more tags to subnets:

```Go
	params := &ec2.DescribeRouteTablesInput{
		Filters: []*ec2.Filter{
			{
				Name:   aws.String("association.subnet-id"),
				Values: []*string{aws.String(subnetid)},
			},
			{
				Name:   aws.String("route.destination-cidr-block"),
				Values: []*string{aws.String("0.0.0.0/0")},
			},
			{
				Name:   aws.String("route.gateway-id"),
				Values: []*string{aws.String("igw*")},
			},
		},
	}
```

jhohertz:
This would go a long way to allowing this ingress controller to co-exist with k8s internal LBs and other ELB/ALB ingress controllers that use this standard.

whithajess:
fixes #253

PR #470, opened 2018-07-18
fix restrict scheme logic

r0fls:


r0fls:
Here it sets the scheme to `internet-facing` on the annotations even if the validation fails: https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/pkg/annotations/annotations.go#L370

r0fls:
closing this for now

PR #476, opened 2018-07-23
Forked upstream ingress

bigkraig:
The nginx-ingress controller which we leveraged as a library has moved the majority of its functionality into an internal package that we can no longer use. Forked the codebase into this repo and made some adjustments to be more compatible with how its used here.


rifelpet:
I'm seeing some weird behavior while running this branch.


```
...
spec:
  containers:
  - args:
    - /server
    - --ingress-class=alb
...
```
results in:
```
I0724 16:33:31.643784       1 launch.go:112] &{ALB Ingress Controller latest git-9fa0225f git://github.com/kubernetes-sigs/aws-alb-ingress-controller}
I0724 16:33:31.643836       1 launch.go:115] Watching for ingress class: alb
F0724 16:33:31.643843       1 launch.go:119] Please specify --default-backend-service
```

whereas 
```
...
spec:
  containers:
  - args:
    - /server
    - --default-backend-service=my-namespace/alb-ingress-controller-default-backend
    - --ingress-class=alb
...
```
results in:
```
-------------------------------------------------------------------------------
AWS ALB Ingress controller
  Release:    UNKNOWN
  Build:      UNKNOWN
  Repository: UNKNOWN
-------------------------------------------------------------------------------

unknown flag: --default-backend-service
unknown flag: --default-backend-service
Usage of :
      --alb-name-prefix string             Prefix to add to ALB resources (11 alphanumeric characters or less)
      --alsologtostderr                    log to standard error as well as files
      --annotations-prefix string          Prefix of the Ingress annotations specific to the AWS ALB controller. (default "alb.ingress.kubernetes.io")
      --apiserver-host string              Address of the Kubernetes API server.
                                           Takes the form "protocol://address:port". If not specified, it is assumed the
                                           program runs inside a Kubernetes cluster and local discovery is attempted.
      --aws-api-debug                      Enable debug logging of AWS API
      --aws-max-retries int                Maximum number of times to retry the AWS API. (default 10)
      --aws-sync-period duration           Period at which the controller refreshes the state from AWS. (default 1h0m0s)
      --cluster-name string                Kubernetes cluster name (required)
      --election-id string                 Election id to use for Ingress status updates. (default "ingress-controller-leader")
      --health-check-path string           URL path of the health check endpoint.
                                           Configured inside the NGINX status server. All requests received on the port
                                           defined by the healthz-port parameter are forwarded internally to this path. (default "/healthz")
      --health-check-period duration       Period at which the controller executes AWS health checks for its healthz endpoint. (default 1m0s)
      --healthz-port int                   Port to use for the healthz endpoint. (default 10254)
      --ingress-class string               Name of the ingress class this controller satisfies.
                                           The class of an Ingress object is set using the annotation "kubernetes.io/ingress.class".
                                           All ingress classes are satisfied if this parameter is left empty.
      --kubeconfig string                  Path to a kubeconfig file containing authorization and API server information.
      --log_backtrace_at traceLocation     when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                     If non-empty, write log files in this directory
      --logtostderr                        log to standard error instead of files (default true)
      --profiling                          Enable profiling via web interface host:port/debug/pprof/ (default true)
      --restrict-scheme                    Restrict the scheme to internal except for whitelisted namespaces
      --restrict-scheme-namespace string   The namespace with the ConfigMap containing the allowed ingresses. Only respected when restrict-scheme is true. (default "default")
      --stderrthreshold severity           logs at or above this threshold go to stderr (default 2)
      --sync-period duration               Period at which the controller forces the repopulation of its local object stores. (default 30s)
      --sync-rate-limit float32            Define the sync frequency upper limit (default 0.3)
  -v, --v Level                            log level for V logs
      --version                            Show release information about the AWS ALB Ingress controller and exit.
      --vmodule moduleSpec                 comma-separated list of pattern=N settings for file-filtered logging
      --watch-namespace string             Namespace the controller watches for updates to Kubernetes objects.
                                           This includes Ingresses, Services and all configuration resources. All
                                           namespaces are watched if this parameter is left empty.
```


I'm not having much success grepping for these errors so I'm not exactly sure why this is happening, any idea?

bigkraig:
@rifelpet Since I forked the ingress code I was able remove that dependency (it wasn't used), but let me add a parameter back in so that it doesn't throw an error if you provide it.

rifelpet:
@bigkraig that did the trick, thanks!

alb-ingress-controller-helm/values.yaml
```diff
@@ -17,9 +17,10 @@ clusterName: k8s
 controller:
   image:
     repository: quay.io/coreos/alb-ingress-controller
-    tag: "1.0-beta.4"
+    tag: "1.0-beta.5"
```
iggy: This tag doesn't exist in the repo yet. `1.0-beta.4` is the latest.

https://quay.io/repository/coreos/alb-ingress-controller?tab=tags and filter `1.0`

PR #479, opened 2018-07-24
Support KOPS values for private/public subnet tags

argeualcantara:
Kops uses "1", and "true" for older versions, having to change and monitor every subnet to check if the value is empty is not practical at all, with this we check if the tag is the expected for private or public subnets, not it's value. Please consider this change.


bigkraig:
Thanks! I didn't see that we could use wildcards.

m15o:
@bigkraig 
Hi, we are using "kubernetes.io/role/internal-elb" tag with empty value. This PR breaks aws-ingress-controller working in our environment. It seems aws-ingress-controller is unable to find any subnet.

argeualcantara:
@m15o Do you have any logs to share?

m15o:
@argeualcantara 
I re-checkd my logs, my situation is not coming from this PR. Sorry for misunderstanding.

I deleted a EKS cluster and created new one with the same cluster name. I forgot to delete target groups created by the old cluster. Then the new cluster tries to create the target group with the same name and failed.

Issue #471, opened 2018-07-20
Support for default target type


dhawani:
In EKS, target-type = instance doesn't work. We need to add target-type=pod annotation in all ingress controller. Instead, can we add an ability to specify default target-type at controller configuration instead ? 

This is a follow-up request on #343

bigkraig:
Makes sense to me. I am currently working on an upgrade and import of the upstream ingress library into this codebase and I can get on this after that.

bigkraig:
@dhawani LMK how #492 looks to you

PR #492, opened 2018-07-26
Added the option to provide a default target type.

bigkraig:
This also starts to move default values into single locations but there is more work to be done there, as needed.


bigkraig:
Should resolve #471

PR #503, opened 2018-07-30
Kubernetes 1.11 has removed ExternalID

lareeth:
To try and fix https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/502

I'm not sure if we should separate the kubelet version check into a helper function.


bigkraig:
@lareeth My EKS cluster has IDs like `aws:///us-west-2c/i-0f4fd41e227b96c96`, which needs a 3 not a 2. What does yours look like? Is it better to do `len()-1`?

PR #516, opened 2018-07-31
Added missing IAM policy

bigkraig:


rifelpet:
just FYI, not including this policy caused the controller to drop all targets from the target group. I'm wondering if theres a better way we can handle that scenario

bigkraig:
@rifelpet yeah, perhaps the best method would be to throw the error and abort the ingress

Issue #487, opened 2018-07-26
Feature req: HTTP -> HTTPS redirects (via newly added ALB feature)


danopia:
Right now, in order to redirect users to HTTPS, the application's backend had to serve the redirects since until now ALBs didn't support modifying requests. For comparison, the nginx ingress controller supports HTTP->HTTPS redirects from an annotation because nginx itself will happily redirect.

Today a new ALB feature was launched which lets the ALB itself redirect traffic. This could be used for a few different things, but the obvious primary use-case (and the one called out in the blog post) is HTTP->HTTPS redirects. So hopefully there's a potential clean way for alb-ingress-controller to offer this feature.

Feature announcement, supports redirects and static responses without hitting the backend: https://aws.amazon.com/about-aws/whats-new/2018/07/elastic-load-balancing-announces-support-for-redirects-and-fixed-responses-for-application-load-balancer/

Specific documentation for redirects: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-listeners.html#redirect-actions

whithajess:
Please someone do this 🎉

anhpham1509:
@bigkraig Are you planning to have this in v1.0?

bigkraig:
How would you all like to see this implemented? The ingress spec is somewhat limiting.

disha94:
Even I am looking for https redirection with alb-ingress.When are you guys planning to implement this feature??

stevenpall:
@bigkraig I think it would make sense to have this as another ingress annotation?

bigkraig:
The [ingress backend spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#ingressbackend-v1beta1-extensions) has `serviceName` and `servicePort`.

I'm pretty sure service names can't include a `:`, so what if we used that to configure the ALB rule actions?

Something like this

```
paths:
- path: /invalid
  backend:
    serviceName: fixed-response:{"ContentType":"text/plain","StatusCode":503,"MessageBody":"message body"}
- path: /
  backend:
    serviceName: echoserver
    servicePort: 80
```

anhpham1509:
I think it's better to have a boolean annotation and redirect all HTTP requests to HTTPS requests by a listener creation to AWS. What do you think @bigkraig ?

bigkraig:
[...]

PR #542, opened 2018-08-10
Customized ALB actions

bigkraig:
Addresses #487. Provide a method for defining custom ALB rule actions via annotations.


bigkraig:
Still need to implement support on default backends, but everything else should be in order.

docs/ingress-resources.md
```diff
@@ -110,6 +109,8 @@ Optional annotations are:
 
 - **ssl-policy**: Defines the [Security Policy](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html#describe-ssl-policies) that should be assigned to the ALB, allowing you to control the protocol and ciphers.
 
+- **alb.ingress.kubernetes.io/actions.\<ACTION NAME>**: Provides a method for configuring custom actions on a listener, such as for [Redirect Actions](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-listeners.html#redirect-actions). The `<ACTION NAME>` in the annotation must match the `serviceName` in the ingress rules. The value of the annotation is the JSON spec of the action. See the [Action type](https://docs.aws.amazon.com/sdk-for-go/api/service/elbv2/#Action) for documentation on what should be in the JSON. An example for a fixed-response would be: `alb.ingress.kubernetes.io/actions.fixed-response-error: '{"Type": "fixed-response", "FixedResponseConfig": {"ContentType":"text/plain", "StatusCode":"503", "MessageBody":"503 error text"}}'` for a `serviceName` of `fixed-response-error`. _NOTE_ you must set the `servicePort` to `use-annotation`.
```
anhpham1509: I think it's better to have examples and notes of `fixed-respone` and `redirect` clearly in sub-indented bullet points.

internal/alb/rs/rule_test.go
```diff
@@ -22,8 +25,39 @@ func TestNewDesiredRule(t *testing.T) {
...
+			TargetPort: 0,
+			ExpectedRule: Rule{
+				svc: svc{desired: service{name: "fixed-response-action", port: intstr.FromString("use-annotation"), targetPort: 0}},
+				rs: rs{
+					desired: &elbv2.Rule{
+						Priority:  aws.String("default"),
+						IsDefault: aws.Bool(true),
+						Actions: []*elbv2.Action{
+							{
+								Type: aws.String("fixed-response"),
```
anhpham1509: 1 test case for `redirect` also?

internal/ingress/annotations/action/main.go
```diff
@@ -0,0 +1,74 @@
...
+	return &Config{
+		Actions: actions,
+	}, nil
+}
+
+func Dummy() *Config {
+	return &Config{
+		Actions: map[string]*elbv2.Action{
+			"fixed-response-action": &elbv2.Action{
+				Type: aws.String("fixed-response"),
```
anhpham1509: How's about dummy type `redirect`? :-D

docs/ingress-resources.md
gabzim: One of the main reasons we're waiting for this is http -> https redirection. Which I'd say is the biggest use case so far. Maybe that should be a redirect example?

PR #549, opened 2018-08-13
Validate node's providerID in case it's empty

bendrucker:
When the kubelet is not run with `--cloud-provider=aws`, the value of `node.Spec.ProviderID` is `""`. I found this to be the case in our cluster since we didn't need persistent volume support right away and were running without the cloud provider enabled.

In the absence of this validation, the error is fairly hard to identify. The result is a call to `IsNodeHealthy` with an empty string as the instance ID which calls `ec2:DescribeInstanceStatus` with the same input. Amazon rejects that request with a 400. 

I've also added a short section to the setup docs mentioning that the cloud provider is required.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -8,6 +8,10 @@ If you'd prefer an end-to-end walkthrough (example) of setup instead, see [the e
 
 This section details what must be setup in order for the controller to run.
 
+### Kubelet
+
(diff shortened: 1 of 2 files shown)


bigkraig:
nice find! Thanks @bendrucker

Issue #561, opened 2018-08-16
When using VPC with secondary CIDR ranges and target-type: IP, IP fail to register


lbernail:
Hello

We are currently testing the controller in ip-mode to route directly to pods. We get this error in the controller:

```E0816 09:39:05.255601       1 api.go:2803] Failed request: elasticloadbalancing/RegisterTargets, Payload: {  TargetGroupArn: "arn:aws:elasticloadbalancing:us-east-1:12345678:targetgroup/47b3c6ad-xxxxxxxxxxxx/yyyyyyyyyy",  Targets: [{      AvailabilityZone: "all",      Id: "10.x.x.xx",      Port: 5000    },{      AvailabilityZone: "all",      Id: "10.x.x.x",      Port: 5000    }]}, Error: ValidationError: The IP address '10.x.x.x' is within the VPC, and cannot have its Availability Zone overridden to 'all' from 'us-east-1b'```

Our VPC setup uses additional CIDR ranges so it is probably related

lbernail:
Ok after looking at the code it's coming from here:
https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/5d7b84cb06a8a4b89511994aae6138052dec27ef/internal/aws/albelbv2/elbv2.go#L242

This is probably related to our VPC setup. I will investigate some more and update the issue

bigkraig:
Merged the fix. Thanks

PR #562, opened 2018-08-16
[elbv2/PopulateAZ] Support VPC with secondary CIDR ranges

lbernail:
This PR addresses #561

The original code only consider the primary CIDR range and cannot identify pods which are using IPs in secondary CIDR blocks.

This modifies the function to consider all CIDR blocks associated with the VPC.


bigkraig:
Thanks for tracking this one down

PR #564, opened 2018-08-17
walkthrough.md: fix indentation for code block

gyuho:


bigkraig:
@gyuho I had to revert this, it lost the indents on the annotations

Issue #362, opened 2018-04-11
Add official chart to the kubernetes/charts repo


sean-krail:
Add an official chart to the [kubernetes/charts](/kubernetes/charts) repo. I'd rather install the chart the native way using `helm install stable/alb-ingress` (like `helm install `[`stable/nginx-ingress`](/kubernetes/charts/tree/master/stable/nginx-ingress)) than installing a plugin for helm.

sstarcher:
Any chance this could happen before 1.0?

rifelpet:
I can open a kubernetes/charts PR, I'm just not sure if there are any legal considerations that need to be taken with copying the alb-ingress-controller-helm directory over. It'll also require updates to the release process which would have to be addressed by this repo's owners.

Jeeppler:
The AWS ALB Ingress Controller is now for two months in the [Helm Charts Incubator](https://github.com/helm/charts/tree/master/incubator/aws-alb-ingress-controller) repository. It would be nice if the AWS ALB Ingress Controller Chart could be promoted to the Helm Chart Stable repository.

rifelpet:
It looks like its just a matter of promoting the chart using the documented process in charts repo:

https://github.com/helm/charts/blob/master/PROCESSES.md#promoting-a-chart-from-incubator-to-stable

PR #587, opened 2018-09-05
Add packaged helm chart

tooxie:
This will make it possible to install it without need for the `registry` plugin.

---

For people using the [terraform-provider-helm](https://github.com/mcuadros/terraform-provider-helm) to install charts it's not possible to use the [helm app registry plugin](https://github.com/app-registry/appr-helm-plugin) because it's not supported by the provider.


bigkraig:
Hey @tooxie I am not a big fan of adding a binary file to the repo, especially one that is crafted outside of the CI pipeline. Do you think it would be possible to leverage the GitHub releases in some way to achieve a download link?

rifelpet:
For what its worth, #362 suggested moving the chart to the official charts repo and helm/charts#7236 is open yet idle. I would lean towards the official repo option seeing as thats the most common helm chart solution for tools like this.

tooxie:
@rifelpet I didn't know there were ongoing efforts to push this to the official repo already. I'll close this. Thanks!

PR #590, opened 2018-09-06
Filter out redirect loop actions

### LINKED ISSUE #487: Feature req: HTTP -> HTTPS redirects (via newly added ALB feature)

bigkraig:
Resolves #487

### DIFF
--- a/internal/alb/ls/listener.go
+++ b/internal/alb/ls/listener.go
@@ -75,6 +75,8 @@ func NewDesiredListener(o *NewDesiredListenerOptions) (*Listener, error) {
 			Priority:         p,
 			Logger:           o.Logger,
 			ListenerRules:    listener.rules,
+			ListenerProtocol: listener.ls.desired.Protocol,
+			ListenerPort:     o.Port,
 			Rule:             &rule,
 			IgnoreHostHeader: o.IgnoreHostHeader,
 			TargetGroups:     o.TargetGroups,
--- a/internal/alb/rs/rule.go
+++ b/internal/alb/rs/rule.go
@@ -36,20 +36,14 @@ type NewDesiredRuleOptions struct {
 // NewDesiredRule returns an rule.Rule based on the provided parameters.
 func NewDesiredRule(o *NewDesiredRuleOptions) (*Rule, error) {
 	r := &elbv2.Rule{
+		IsDefault: aws.Bool(false),
+		Priority:  aws.String(fmt.Sprintf("%v", o.Priority)),
 		Actions: []*elbv2.Action{
 			{
 				TargetGroupArn: nil, // Populated at creation, since we create rules before we create rules
 				Type:           aws.String(elbv2.ActionTypeEnumForward),
 			},
 		},
-	}
-
-	if o.Priority == 0 {
-		r.IsDefault = aws.Bool(true)
-		r.Priority = aws.String("default")
-	} else {
-		r.IsDefault = aws.Bool(false)
-		r.Priority = aws.String(fmt.Sprintf("%v", o.Priority))
 	}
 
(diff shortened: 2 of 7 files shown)


internal/alb/rs/rules.go
```diff
@@ -79,14 +82,12 @@ func NewDesiredRules(o *NewDesiredRulesOptions) (Rules, int, error) {
 	rs := o.ListenerRules
 	paths := o.Rule.HTTP.Paths
 
-	if len(paths) == 0 {
-		return nil, 0, fmt.Errorf("Ingress doesn't have any paths defined. This is not a very good ingress.")
+	if o.Priority == 0 {
+		o.Priority = 1
```
M00nF1sh: seems hack to me 😄

bigkraig: yeah i am actually not a fan of how the priority is built up. maybe it would be better to build all of the rules and then set the priorities after

internal/alb/rs/rule.go
```diff
@@ -324,3 +318,43 @@ func (r Rule) IsDesiredDefault() bool {
...
+		for _, c := range r.rs.desired.Conditions {
+			if *c.Field == "host-header" {
+				host = c.Values[0]
+			}
+			if *c.Field == "path-pattern" {
+				path = c.Values[0]
+			}
+		}
+
+		if host == nil && *rc.Host != "#{host}" {
```
M00nF1sh: i think the two if for host check can be simplified to be *rc.Host != "#{host}" , since
host can never be "#{host}", according to k8s spec, it need to be a domain name of RFC 3986.

internal/alb/rs/rule.go
```diff
@@ -36,6 +36,8 @@ type NewDesiredRuleOptions struct {
 // NewDesiredRule returns an rule.Rule based on the provided parameters.
 func NewDesiredRule(o *NewDesiredRuleOptions) (*Rule, error) {
 	r := &elbv2.Rule{
+		IsDefault: aws.Bool(false),
```
M00nF1sh: What's is the `IsDefault`'s original purpose? Since it will never be true again..and there is still code referencing it.

internal/alb/ls/listener.go
```diff
@@ -75,6 +75,8 @@ func NewDesiredListener(o *NewDesiredListenerOptions) (*Listener, error) {
 			Priority:         p,
 			Logger:           o.Logger,
 			ListenerRules:    listener.rules,
+			ListenerProtocol: listener.ls.desired.Protocol,
+			ListenerPort:     o.Port,
```
M00nF1sh: minimal comment:
I think it's better be `listener.ls.desired.Port` instead of `o.Port` here, even they are equal.
Which can be understand better 😄, since it's aligned with `listener.ls.desired.Protocol`

PR #605, opened 2018-09-11
Use paginated APIs

cv:
This spawned from https://github.com/kubernetes-sigs/aws-alb-ingress-controller/pull/600#issuecomment-420448144, where I found out there was an outdated paginated API access pattern in use. None of these changes should affect behavior.

### DIFF
--- a/internal/aws/albelbv2/elbv2.go
+++ b/internal/aws/albelbv2/elbv2.go
@@ -15,7 +15,6 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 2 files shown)


bigkraig:
@spiffxp should tide be taking it from here?

spiffxp:
>should tide be taking it from here?

@bigkraig I think you've got it figured out now, for future reference: https://github.com/kubernetes/test-infra/blob/master/prow/cmd/tide/pr-authors.md

We're working to standardize on roughly: (green contexts) + lgtm + approved - do-not-merge/*

bigkraig:
Thanks @spiffxp, I got held up because there was some final auth that tide needed and it didnt resync after I had added it, everything has been smooth sailing since that! 👍

PR #607, opened 2018-09-12
Subnet tag doc update

### LINKED ISSUE #606: [1.0-beta.6] error parsing annotations: Retrieval of subnets failed to resolve 2 qualified subnets.
In `1.0-beta.6`, I was receiving the following error.

```
I0912 04:45:56.361403       1 event.go:221] Event(v1.ObjectReference{Kind:"Ingress", Namespace:"echoserver", Name:"echoserver", UID:"92ecc697-b63a-11e8-b7c1-129bc0ce9c1e", APIVersion:"extensions/v1beta1", ResourceVersion:"902067", FieldPath:""}): type: 'Warning' reason: 'ERROR' error parsing annotations: Retrieval of subnets failed to resolve 2 qualified subnets. Subnets must contain the kubernetes.io/cluster/<cluster name> tag with a value of shared or owned and the kubernetes.io/role/internal-elb tag signifying it should be used for ALBs Additionally, there must be at least 2 subnets with unique availability zones as required by ALBs. Either tag subnets to meet this requirement or use the subnets annotation on the ingress resource to explicitly call out what subnets to use for ALB creation. The subnets that did resolve were [].
```

I had followed the docs at https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/docs/ingress-resources.md

I added the following tags to my subnets:
- `kubernetes.io/cluster/mycluster="owned"`
- `kubernetes.io/role/elb` (I tried setting it to `""`, and also `true`, figuring it can probably be anything)

I noticed someone in a totally different repo setting the same tag to "1", and now it appears to find my subnet.
https://github.com/kubernetes/kubernetes/issues/60152

I'm happy to submit a PR if that is the correct way to do it.

NickLarsenNZ:
Fixes #606 

_I'm yet to sign the CLA, and read through the contributing guide - but I can't right now_

### DIFF
--- a/docs/ingress-resources.md
+++ b/docs/ingress-resources.md
@@ -92,8 +92,8 @@ alb.ingress.kubernetes.io/actions.<ACTION NAME>
 
   - `kubernetes.io/cluster/$CLUSTER_NAME` where `$CLUSTER_NAME` is the same cluster name specified on the ingress controller. The value of this tag must be `shared` or `owned`.
 
-  - `kubernetes.io/role/internal-elb` should be set for internal load balancers.
-  - `kubernetes.io/role/elb` should be set for internet-facing load balancers.
+  - `kubernetes.io/role/internal-elb` should be set to `1` for internal load balancers.
+  - `kubernetes.io/role/elb` should be set to `1` for internet-facing load balancers.
 
   - After subnets matching the above 2 tags have been located, they are checked to ensure 2 or more are in unique AZs, otherwise the ALB will not be created. If 2 subnets share the same AZ, only 1 of the 2 is used.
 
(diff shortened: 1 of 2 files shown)


NickLarsenNZ:
ping

M00nF1sh:
Hi Nick, 
thanks for contributing 👍, actually both empty string and `1` works.
Could you update it to be like "should be set to `1` or leave empty"

cv:
+1 @M00nF1sh -- code here: https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/df5bc6afba621bc6848243ee66aa974da4f5a6d1/internal/aws/albrgt/rgt.go#L79-L88

M00nF1sh:
Yes 👍 , but the document is not aligned with the code. Since from the code, both '' and '1' works 😸
Not sure why previously the code allowed "" empty tag value.

NickLarsenNZ:
Apologies for the delay. I retested my setup and it does indeed work with an empty tag value as the code shows. I can't explain why it didn't work the first time.

Hoping to update the docs after dinner...

PR #610, opened 2018-09-12
Catching a nil pointer error

bigkraig:


bigkraig:
@M00nF1sh

internal/aws/albrgt/rgt.go
```diff
@@ -147,6 +147,35 @@ func (r *RGT) GetClusterResources() (*Resources, error) {
 		}
 	}
 
+	// Legacy deployments may not have the proper tags, and RGT doesn't allow you to use wildcards on names
+	p := request.Pagination{
+		EndPageOnSameToken: true,
+		NewRequest: func() (*request.Request, error) {
+			req, _ := r.GetResourcesRequest(&resourcegroupstaggingapi.GetResourcesInput{
```
cv: Couldn't this use `GetResourcesPages` too, like that call we changed a couple of days ago in #605?

bigkraig: Yeah we can. I went with the old method since we hadn’t merged that yet.

pkg/util/types/tags.go
```diff
@@ -64,14 +64,17 @@ func (t ELBv2Tags) ServiceNameAndPort() (name string, port intstr.IntOrString, e
...
+
+	if name == "" {
 		return "", intstr.IntOrString{}, fmt.Errorf("kubernetes.io/service-name tag is missing")
 	}
 
 	if v, ok := t.Get("kubernetes.io/service-port"); ok {
 		port = intstr.Parse(v)
 	} else {
-		return "", intstr.IntOrString{}, fmt.Errorf("kubernetes.io/service-port is missing")
+		port = intstr.Parse("1")
```
M00nF1sh: will this be an issue? i think it's better to keep the original one..

bigkraig: It was necessary to upgrade from an old version of the controller that didnt have the `kubernetes.io/service-port` tag yet. TBH now that we are using TG port 1, i need to look and see if this tag is still useful

M00nF1sh: the service-port is what as specified in ingress spec, e.g.(80/http), it's not same as the target port(port on pods) 😄

bigkraig: so the tag is still useful but if we come across a TG w/o it, we shouldn't abort the assembly by returning an error here. perhaps a `intstr.Parse("0")` to identify it as an invalid TG and force a recreation of it?

M00nF1sh: i think it’s fine to use port0, since the worse case is the targetgroup will be orphaned :smile:

Issue #559, opened 2018-08-15
Error adding targets to target group,InvalidInstanceID: There are multiple interfaces attached to instance


sashanka-kandala:
Deployed alb controller on AWS EKS.
Attached policy to worker nodes.
Used 2048-game example to test.
Below are the config files:
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: "2048-ingress"
  namespace: "2048-game"
  annotations:
    kubernetes.io/ingress.class: alb
    alb.ingress.kubernetes.io/scheme: internal
    alb.ingress.kubernetes.io/target-type: ip
    #alb.ingress.kubernetes.io/security-groups: sg-0941bddfa4d5a39a5
    alb.ingress.kubernetes.io/subnets: subnet-xxxx, subnet-xxxx
  labels:
    app: 2048-ingress
spec:
  rules:
  - http:
      paths:
      - path: /
        backend:
          serviceName: "service-2048"
          servicePort: 80

```
```
apiVersion: v1
kind: Service
metadata:
  name: "service-2048"
  namespace: "2048-game"
spec:
  ports:
  - port: 80
    targetPort: 80
    protocol: TCP
  selector:
    app: "2048"
```
**Getting below error:**

Event(v1.ObjectReference{Kind:"Ingress", Namespace:"2048-game", Name:"2048-ingress", UID:"7c48f85e-a0ad-11e8-99ae-061cda1a6546", APIVersion:"extensions/v1beta1", ResourceVersion:"133013", FieldPath:""}): type: 'Warning' reason: 'ERROR' Error adding targets to target group xxxxxxxxxx: InvalidInstanceID: There are multiple interfaces attached to instance 'i-xxxxxxx'. Please specify an interface ID for the operation instead.
```

And instance security groups are not attached to nodes.

**Comment by Kefa7y** [comment]
I get the same error in here as well.

**Comment by kalelc** [comment]
I have the same error.

**Comment by sa-mao** [comment]
I have the same error.

**Comment by kalelc** [comment]
I was able to fix this problem by leaving the worker node with one interface. 

I use terraform to create the Cluster; that configuration creates 2 ENI. I suppose that the problem is because Ingress cannot resolve which IP can use if the instance has 2.

Ingress Log:

With two ENI
```
[...]

PR #614, opened 2018-09-17
Refactor of securityGroup handling

M00nF1sh:
The purpose of this PR:
1. Refactor of securityGroup handling functionality
1. Fixing the multi-ENI assigned issue(#559)
1. Remove the unused targetPort in targetGroup

Tested cases:
- scale up / down Ingress with mode ip
- scale up / down Ingress with mode instance
- change listener-port for ingress with mode ip/instance
- change cidr for ingress with mode ip/instance
- change from managed sg to external sg for ingress with mode ip/instance
- change from external sg to managed sg for ingress with mode ip/instance
- delete ingress with mode ip/instance and sg managed/external


internal/alb/sg/instance_attachment.go
```diff
@@ -0,0 +1,193 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+func (controller *instanceAttachmentController) Delete(attachment *InstanceAttachment) error {
+	instanceENIs, err := controller.getClusterInstanceENIs()
+	if err != nil {
+		return fmt.Errorf("failed to get cluster enis, Error:%s", err.Error())
```
gyuho: nits: `s/Error:%s/Error: %v/`?

internal/alb/sg/instance_attachment.go
```diff
@@ -0,0 +1,193 @@
...
+		return nil
+	}
+	_, err := controller.ec2.ModifyNetworkInterfaceAttribute(&ec2.ModifyNetworkInterfaceAttributeInput{
+		NetworkInterfaceId: eni.NetworkInterfaceId,
+		Groups:             aws.StringSlice(desiredGroups),
+	})
+	return err
+}
+
+//findENIsSupportingTargets find the ID of ENIs that are used to supporting ingress traffic to targets
```
gyuho: `s///findENIsSupportingTargets/// findENIsSupportingTargets/`?

internal/alb/sg/instance_attachment.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+	}
+	return result
+}
+
+// getClusterInstanceENIs retrives all ENIs attached to instances indexed by instanceID
+func (controller *instanceAttachmentController) getClusterInstanceENIs() (map[string][]*ec2.InstanceNetworkInterface, error) {
+	instanceIDs, err := controller.getClusterInstanceIDs()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get instance IDs within cluster, Error:%s", err.Error())
```
gyuho: nits: `s/Error:%s/Error: %s/`?

internal/alb/sg/instance_attachment.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	instances, err := controller.ec2.GetInstancesByID(instanceIDs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get instances within cluster, Error:%s", err.Error())
```
gyuho: nits: `s/Error:%s/Error: %s/`?

internal/alb/sg/lb_attachment.go
```diff
@@ -0,0 +1,92 @@
...
+	return nil
+}
+
+func (controller *lbAttachmentController) Delete(attachment *LbAttachment) error {
+	loadBalancer, err := controller.elbv2.GetLoadBalancerByArn(attachment.LbArn)
+	if err != nil {
+		return err
+	}
+	if loadBalancer == nil {
+		return fmt.Errorf("loadBalancer %s doesn't exists", attachment.LbArn)
```
gyuho: nits: `s/exists/exist/`?

internal/alb/sg/security_group.go
```diff
@@ -0,0 +1,255 @@
...
+
+func (controller *securityGroupController) reconcileByModifySGInstance(group *SecurityGroup, instance *ec2.SecurityGroup) error {
+	if group.GroupID == nil {
+		group.GroupID = instance.GroupId
+	}
+	if group.GroupName == nil {
+		group.GroupName = instance.GroupName
+	}
+
+	permissionsToRevoke := diffIPPermissions(instance.IpPermissions, group.InboundPermissions)
```
gyuho: Can we document what variables mean here? It's not clear to me what the desired state is here (e.g. source is `instance`, and target is `group`). Are we trying to reconcile `instance` so that `group` matches the permissions of `instance`?

internal/alb/sg/security_group.go
```diff
@@ -0,0 +1,255 @@
...
+	if group.GroupID != nil {
+		instance, err := controller.ec2.GetSecurityGroupByID(*group.GroupID)
+		if err != nil {
+			return nil, err
+		}
+		if instance == nil {
+			return nil, fmt.Errorf("securityGroup %s doesn't exist", *group.GroupID)
+		}
+		return instance, nil
+	} else if group.GroupName != nil {
```
gyuho: No need `else` here? Just `if group.GroupName != nil {` in the next line?

internal/alb/sg/instance_attachment.go
leakingtapan: If there isn't any, we probably should set up golint to automate the checking of similar styling issue. Also this could be an issue tracked separately.

internal/alb/sg/association.go
```diff
@@ -0,0 +1,259 @@
...
+		return controller.reconcileWithExternalSGs(association)
+	}
+	return controller.reconcileWithManagedSGs(association)
+}
+
+func (controller *associationController) Delete(association *Association) error {
+	return controller.deletedManagedSGs(association)
+}
+
+func (controller *associationController) reconcileWithExternalSGs(association *Association) error {
```
leakingtapan: It will be helpful to add some docs here to explain what is external SG. Similar to the managed SG. Or add a link if there is any

internal/alb/sg/association.go
```diff
@@ -0,0 +1,259 @@
...
+					{
+						GroupId: lbSG.GroupID,
+					},
+				},
+			},
+		},
+	}
+	err := controller.sgController.Reconcile(instanceSG)
+	if err != nil {
+		return fmt.Errorf("failed to reconcile managed Instance securityGroup due to %s", err.Error())
```
leakingtapan: NIT. The `.Error()` at the end is redundant. It should still print the error message without calling the `Error()` method explicitly.

internal/alb/sg/association.go
```diff
@@ -0,0 +1,259 @@
...
+	return nil
+}
+
+func (controller *associationController) deleteManagedInstanceSG(association *Association) error {
+	instanceSGName := controller.namer.NameInstanceSG(association.LbID)
+	instanceSGID, err := controller.findSGIDByName(instanceSGName)
+	if err != nil {
+		return err
+	}
+	if instanceSGID == nil {
```
leakingtapan: Could `instanceSGID` be empty string? And is this a valid case if it happens?

internal/alb/sg/association.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	instanceSGAttachment := &InstanceAttachment{
+		GroupID: *instanceSGID,
+	}
+	err = controller.instanceAttachmentController.Delete(instanceSGAttachment)
+	if err != nil {
+		return err
+	}
+	instanceSG := &SecurityGroup{
+		GroupID: instanceSGID,
```
leakingtapan: Why `instanceSGID` is used as value here by used as pointer in
```
instanceSGAttachment := &InstanceAttachment{
		GroupID: *instanceSGID,
	}
```
Should we make it consistent?

internal/alb/sg/instance_attachment.go
```diff
@@ -0,0 +1,199 @@
...
+
+// InstanceAttachment represents the attachment of securityGroups to instance
+type InstanceAttachment struct {
+	GroupID string
+	Targets tg.TargetGroups
+}
+
+// InstanceAttachementController manages InstanceAttachment
+type InstanceAttachementController interface {
+	Reconcile(*InstanceAttachment) error
```
leakingtapan: Could you add some docs on what does Reconcile do? In general, all the public APIs should be documented (golint will complain once we add it). It is not urgent now, but could help a lot for other contributor to understand the code.

internal/alb/sg/instance_attachment.go
```diff
@@ -0,0 +1,199 @@
...
+	return result
+}
+
+// getClusterInstanceENIs retrives all ENIs attached to instances indexed by instanceID
+func (controller *instanceAttachmentController) getClusterInstanceENIs() (map[string][]*ec2.InstanceNetworkInterface, error) {
+	instanceIDs, err := controller.getClusterInstanceIDs()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get instance IDs within cluster, Error:%s", err.Error())
+	}
+	instances, err := controller.ec2.GetInstancesByID(instanceIDs)
```
leakingtapan: should be `GetInstancesByIDs`?

internal/alb/sg/lb_attachment.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// Delete ensures specified SecurityGroup don't exists in LoadBalancer, other sg are kept.
+	// If there are remaining sg, the default SG for vpc will be kept.
+	Delete(*LbAttachment) error
+}
+
+type lbAttachmentController struct {
+	elbv2  albelbv2.ELBV2API
+	ec2    *albec2.EC2
+	logger *log.Logger
```
leakingtapan: What's the reason why each controller has its own logger?

internal/alb/sg/namer.go
```diff
@@ -0,0 +1,26 @@
...
+// Namer can name securityGroup related resources.
+type Namer interface {
+	NameLbSG(loadBalancerID string) string
+	NameInstanceSG(loadBalancerID string) string
+}
+
+type namer struct{}
+
+func (namer *namer) NameLbSG(loadBalancerID string) string {
+	return fmt.Sprintf(lbSGNamePtn, loadBalancerID)
```
leakingtapan: NIT: feels its completely fine to inline `lbSGNamePtn` and `instanceSGNamePtn` to where it is used inside method if only one method uses it. It has the benefit that, during refactoring, the constant will be moved along with the method instead of lingering around.

internal/alb/sg/security_group.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	return controller.reconcileByNewSGInstance(group)
+}
+
+func (controller *securityGroupController) Delete(group *SecurityGroup) error {
+	if group.GroupID != nil {
+		controller.logger.Infof("deleting securityGroup %s", *group.GroupID)
+		err := controller.ec2.DeleteSecurityGroupByID(*group.GroupID)
+		if err != nil {
+			controller.logger.Warnf("%s", err)
```
leakingtapan: Redundant log?

internal/alb/sg/security_group.go
```diff
@@ -0,0 +1,267 @@
...
+			controller.logger.Warnf("%#v", err)
+			if awsErr, ok := err.(awserr.Error); ok {
+				controller.logger.Warnf("%#v", awsErr)
```
leakingtapan: Same above

internal/alb/sg/security_group.go
```diff
@@ -0,0 +1,267 @@
...
+	if len(diffIPRanges(target.IpRanges, source.IpRanges)) != 0 {
+		return false
+	}
+	if len(diffUserIDGroupPairs(source.UserIdGroupPairs, target.UserIdGroupPairs)) != 0 {
+		return false
+	}
+	if len(diffUserIDGroupPairs(target.UserIdGroupPairs, source.UserIdGroupPairs)) != 0 {
+		return false
+	}
+
```
leakingtapan: Should we also compare `Ipv6Ranges`?

internal/alb/sg/instance_attachment.go
bigkraig: actually it probably doesn't even need the `Error:`

M00nF1sh: i have already changed these to be "XXXX due to %s" 😄

internal/alb/sg/instance_attachment.go
M00nF1sh: fixed :D actually i have glint in my ide(vscode), not sure why it doesn't complain about this case

internal/alb/sg/security_group.go
M00nF1sh: added method doc to that method, group is our internal securityGroup definition, which contains the desired state. instance is the AWS securityGroup representation

internal/alb/sg/security_group.go
M00nF1sh: i explicitly used else if to express the idea that one of the branch should occur. (to avoid cases like someone modified the if groupId != nil branch, and assigned groupName there)

internal/alb/sg/association.go
M00nF1sh: added documentation to the code 😄

internal/alb/sg/association.go
M00nF1sh: i'll change the errors to be "XXX due to %v", err

internal/alb/sg/association.go
M00nF1sh: it might be empty string, since it could be deleted manually from aws console, or during previous reconcile process(and the controller crashed or errored for some reason after that)

internal/alb/sg/association.go
M00nF1sh: instanceSGID i retrieved maybe nil, so it retrieved as *string. (Hope we got optional support in golang to better express the idea) 
however, in the API of my code, i deliberately make the inputs to be string if it's required and cannot be nil. 

I may have named the variable in `old styles` like pInstanceSGID, but it's unnecessary with modern IDE support 😄

internal/alb/sg/instance_attachment.go
M00nF1sh: added docs on these methods

internal/alb/sg/instance_attachment.go
M00nF1sh: i think it's better to remain as GetInstancesByID, since the "IDs" is already expressed by the strong typed arguments.
Actually, i'd prefer to named it as GetInstances if golang supports method overload

internal/alb/sg/lb_attachment.go
M00nF1sh: actually i don't need them, since golang's log already contains the file information, do you have any recommendations like use global logger?

bigkraig: there should be a logger created for each ingress resource and then its been saved in the dependent structs for ease of use. the ingress logger includes the ingress name and namespace when printing.

internal/alb/lb/loadbalancer.go
```diff
@@ -572,6 +441,13 @@ func (l *LoadBalancer) modify(rOpts *ReconcileOptions) error {
 
 // delete Deletes the load balancer from AWS.
 func (l *LoadBalancer) delete(rOpts *ReconcileOptions) error {
+	l.deleted = true
```
bigkraig: I think we need to move this to the end, the deleted flag should only be set once the deletion is confirmed, was there a particular reason for moving it up here?

internal/alb/lb/loadbalancer.go
```diff
@@ -572,6 +441,13 @@ func (l *LoadBalancer) modify(rOpts *ReconcileOptions) error {
...
+
+	l.sgAssociation.LbArn = *l.lb.current.LoadBalancerArn
```
bigkraig: use aws.StringValue?

internal/alb/sg/namer.go
M00nF1sh: yeah..this part can be inlined in the association that created it.
I created a separate component here because the naming(and also tagging?) of AWS resources are important and should have consist rule/patterns.
The ultimate goal is to have a simple component generating names/tag for all aws resources we created. I think it's better than distribute them around different modules(hard to apply a consist rules to them)

internal/alb/sg/security_group.go
M00nF1sh: it could be in future, if we used userIdGroupsPairs.. currently, i'd rather not write test cases for that 😄.
This refactor is just a start point to make code modular and ez to change, we can add ipv6 support easily in the future

internal/ingress/controller/controller.go
```diff
@@ -50,11 +50,11 @@ func (c *ALBController) syncIngress(interface{}) error {
 	c.runningConfig.Ingresses = newIngresses
 
 	// Reconcile the states
-	removedIngresses.Reconcile(c.metricCollector)
+	removedIngresses.Reconcile(c.metricCollector, c.sgAssociationController)
```
bigkraig: should we use a ReconcileOpts like we do for other Reconcile funcs?

internal/aws/albelbv2/dummy.go
```diff
@@ -103,6 +103,11 @@ func (d *Dummy) DeleteRule(in *elbv2.DeleteRuleInput) (*elbv2.DeleteRuleOutput,
 	return d.outputs["DeleteRuleOutput"].(*elbv2.DeleteRuleOutput), d.outputs.error("DeleteRuleError")
 }
 
+// SetField ...
```
bigkraig: GetLoadBalancerByArn not SetField

internal/alb/lb/loadbalancer.go
M00nF1sh: i search the code, it's not used in other places(only by my code).

in the end of reconcile part, my code will check l.deleted to choose whether to do "reconcile on association" or not.
If i don't mark it as deleted(maybe shouldBeDeleted) from the beginning, and the delete failed due to low-level errors, my code will tried to reconcile deleted lb(e.g. attach sgs)

internal/aws/albec2/ec2.go
```diff
@@ -297,403 +170,97 @@ func (e *EC2) GetSecurityGroups(names []*string) (sgs []*string, err error) {
...
+		return nil, nil
 	}
+	return securityGroups[0], nil
+}
 
-	// attempt to resolve instance sg
-	instanceSGName := fmt.Sprintf("%s-%s", instSpecifierTag, *sgName)
-	inInstance := &ec2.DescribeSecurityGroupsInput{
+// GetSecurityGroupByName retrives securityGroup by vpcID and securityGroupName(SecurityGroup names within vpc are unique)
+func (e *EC2) GetSecurityGroupByName(vpcID string, sgName string) (*ec2.SecurityGroup, error) {
```
bigkraig: Lets change `sgName` to `groupName` since its a `group-name` filter

internal/alb/tg/targetgroup.go
```diff
@@ -68,7 +65,7 @@ func NewDesiredTargetGroup(o *NewDesiredTargetGroupOptions) *TargetGroup {
 		ID:         id,
 		SvcName:    o.SvcName,
 		SvcPort:    o.SvcPort,
-		TargetPort: o.TargetPort,
+		TargetType: *o.Annotations.TargetGroup.TargetType,
```
bigkraig: use aws.StringValue here?

internal/alb/sg/namer.go
```diff
@@ -0,0 +1,26 @@
+package sg
+
+import (
+	"fmt"
+)
+
+const (
+	lbSGNamePtn       = "%s"
```
bigkraig: `lb-%s` for consistency?

internal/alb/sg/security_group.go
```diff
@@ -0,0 +1,257 @@
...
+	InboundPermissions []*ec2.IpPermission
+}
+
+// SecurityGroupController manages SecurityGroups
+type SecurityGroupController interface {
+	// Reconcile ensures the securityGroup exists and match the specification.
+	// Field GroupID or GroupName will be populated if unspecified.
+	Reconcile(*SecurityGroup) error
+
+	// Delete ensures the securityGroup does not exists.
```
bigkraig: /exists/exist/

internal/alb/sg/security_group.go
bigkraig: i have used switch statements to get that point across

internal/aws/albec2/ec2.go
bigkraig: Also, vpcid is basically fixed for the controller, so we could resolve the vpcid in this func and save passing it in as a parameter everywhere its called

internal/alb/sg/security_group.go
```diff
@@ -0,0 +1,257 @@
...
+	})
+	if err != nil {
+		return err
+	}
+	controller.logger.Infof("created new securityGroup: %s", *group.GroupID)
+
+	return nil
+}
+
+func (controller *securityGroupController) reconcileByModifySGInstance(group *SecurityGroup, instance *ec2.SecurityGroup) error {
```
bigkraig: This func is missing reconciling tags

internal/alb/sg/namer.go
bigkraig: not renaming to avoid SG swap and orphaning old SGs

internal/alb/sg/security_group.go
M00nF1sh: yeah... the tags is not used currently. I plan to add that feature into the spec of securityGroup maybe in other pr 😅

internal/alb/sg/association.go
```diff
@@ -0,0 +1,266 @@
...
+	// Reconcile ensured the securityGroups in AWS matches the state specified by assocation.
+	Reconcile(*Association) error
+	
+	// Delete ensures the securityGroups created by ingress controller for specified LbID doesn't exists.
+	Delete(*Association) error
+}
+
+// NewAssociationController constructs a new association controller
+func NewAssociationController(store store.Storer, ec2 *albec2.EC2, elbv2 albelbv2.ELBV2API) AssociationController {
+	logger := log.New("sg.association")
```
bigkraig: gotta figure out some way to use a logger that has the ingress id instead of `sg.association`

internal/alb/sg/association.go
bigkraig: external is probably just a bad name, they are 'externally managed' i guess? maybe we call the security groups managed and unmanaged?

internal/alb/sg/association.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+func (controller *associationController) reconcileWithExternalSGs(association *Association) error {
+	lbSGAttachment := &LbAttachment{
+		GroupIDs: association.ExternalSGIDs,
+		LbArn:    association.LbArn,
+	}
+	err := controller.lbAttachmentController.Reconcile(lbSGAttachment)
+	if err != nil {
+		return fmt.Errorf("failed to reconcile external LoadBalancer securityGroup due to %s", err.Error())
```
bigkraig: i think the word `external` here or even `unmanaged` if we go with that will be confusing to people, maybe just remove the specifics completely?

internal/alb/sg/association.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	err = controller.deletedManagedSGs(association)
```
bigkraig: I think we need another one of these or just something to delete associations, we do not want to delete the externally manasged SG

internal/alb/sg/lb_attachment.go
M00nF1sh: fixed

internal/alb/sg/association.go
bigkraig: actually maybe this isnt needed at all for an externally managed sg, the only responsibility of the controller is to manage the SG attachmetns to the LB and instances right?

internal/alb/sg/association.go
M00nF1sh: i'll change them to be externalManagedSGIDs 😄

internal/alb/sg/association.go
bigkraig: `if aws.StringValue(instanceSGID) == ""` ?

internal/alb/sg/association.go
```diff
@@ -0,0 +1,266 @@
...
+		GroupID: instanceSGID,
+	}
+	err = controller.sgController.Delete(instanceSG)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (controller *associationController) findSGIDByName(sgName string) (*string, error) {
```
bigkraig: is this better off in albec2?

internal/alb/sg/namer.go
M00nF1sh: misunderstanded the comment, will move the constant into the method

internal/alb/sg/instance_attachment.go
```diff
@@ -0,0 +1,210 @@
...
+					break
+				}
+			}
+		}
+	}
+	return result
+}
+
+// getClusterInstanceENIs retrives all ENIs attached to instances indexed by instanceID
+func (controller *instanceAttachmentController) getClusterInstanceENIs() (map[string][]*ec2.InstanceNetworkInterface, error) {
```
bigkraig: should this move to albec2?

internal/alb/sg/instance_attachment.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	result := make(map[string][]*ec2.InstanceNetworkInterface)
+	for _, instance := range instances {
+		result[aws.StringValue(instance.InstanceId)] = instance.NetworkInterfaces
+	}
+	return result, nil
+}
+
+// getClusterInstanceIDs retrives the aws instanceIDs in k8s cluster
+func (controller *instanceAttachmentController) getClusterInstanceIDs() (result []string, err error) {
```
bigkraig: I think this is eligible to be moved into `store`.

M00nF1sh: agreed, will move it there 👍 .

However, i think the store have do too much things. 
There should be a separate module to hold it, but anyway it doesn't belong to instance_attachment.
I'll move it to store first(though it's not ideal to me =.=)

internal/alb/sg/instance_attachment.go
M00nF1sh: changed to due to %v

internal/alb/sg/security_group.go
M00nF1sh: changed to use switch instead

internal/alb/sg/association.go
M00nF1sh: it's better to test for nil as it is, since it's model `optional` concept 😄

internal/alb/sg/lb_attachment.go
M00nF1sh: yeah, there could be a way like passing a reconcileContext into reconcile options. or have each object contains that reconcile context like ingress name.

Will leave it as it is and change it in later PR when needed.

internal/alb/sg/security_group.go
M00nF1sh: removed, it's for debug purpose

internal/aws/albec2/ec2.go
M00nF1sh: vpc was kept as it is, since it's should be ec2 API's argument. maybe there will be multiple-vpc k8s cluster in the future(though a lot of things will break).
For now, i think it's better to resolve VPC during controller startup and saved as an configuration

internal/alb/sg/security_group.go
M00nF1sh: resolved

internal/alb/sg/association.go
M00nF1sh: maybe pass the ingress name into the object been reconciled.
anyway, will do it in separate PR

internal/alb/sg/association.go
M00nF1sh: decided to kept external after discussing with kraig offline

gyuho:
@M00nF1sh @bigkraig I manually tested this branch with internal e2e tests, and all look good.

Issue #597, opened 2018-09-10
Default backend behavior is different from other ingresses


lbernail:
We are currently using the `ip` target-type and this feature is great

During some testing, I noticed that when I try to access a route with no definition, I end up on one of the backend service and I do not get a 404.

Example:
```yaml
rules:
  - host: echov1.lbernail.us1.staging.dog
    http:
      paths:
      - path: /
        backend:
          serviceName: echov1
          servicePort: 80
  - host: echov2.lbernail.us1.staging.dog
    http:
      paths:
      - path: /
        backend:
          serviceName: echov2
          servicePort: 80
```

When I try to access a host without any rule, I would expect a 404, but I get to the first service (after checking, the default action for the listener is set to the first targetgroup).

After looking at the code it seems that the controller looks for a default backend configured 
with `ingress.spec.backend` and if there is none it uses the first valid backend: https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/27f2048c1af4fd5e6a7e31402701e2bddff66862/internal/albingress/albingresses.go#L231

Adding this to the ingress:
```yaml
  backend:
    serviceName: fixed-response-error
    servicePort: use-annotation
```

works great but I wonder if we shouldn't use this as a default if there is no default backend key in the spec

lbernail:
I'm going to take a shot at this issue
In terms of implementation I was thinking of adding a new `serviceName: default-http-backend` in `applyDefaults` without a service port and implement the default logic in `NewDesiredRule`

bigkraig:
Yeah, `applyDefaults` right now has the sole responsibility of setting the default backend to the first defined path, the specific thing we are getting rid of.

I'm not sure you need to get into the rules though, have a look at `Listener.defaultBackend`.

PR #615, opened 2018-09-17
Set the ingress default backend to a 404 response

bigkraig:
Should additionally resolve loop issues in #487

### DIFF
--- a/internal/alb/ls/listener.go
+++ b/internal/alb/ls/listener.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/aws/albrgt"
+	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/annotations/action"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/k8s"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -138,7 +139,7 @@ func NewCurrentListener(o *NewCurrentListenerOptions) (*Listener, error) {
 		}
 	} else {
 		defaultBackend = &extensions.IngressBackend{
-			ServicePort: intstr.FromString("use-annotation"),
+			ServicePort: intstr.FromString(action.UseActionAnnotation),
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


internal/alb/ls/types.go
```diff
@@ -33,3 +34,16 @@ type ReconcileOptions struct {
 	LoadBalancerArn *string
 	TargetGroups    tg.TargetGroups
 }
+
+const Default404 = "Default 404"
```
M00nF1sh: Will `Default-404` or `default404` be better? since it should be an [serviceName](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names), use 'Default 404' might be broken once k8s core validated the serviceName in ingress spec 😸

internal/alb/ls/listener.go
```diff
@@ -150,34 +151,49 @@ func NewCurrentListener(o *NewCurrentListenerOptions) (*Listener, error) {
 	}, nil
 }
 
+func (l *Listener) resolveDefaultBackend(rOpts *ReconcileOptions) (*elbv2.Action, error) {
+	if l.defaultBackend.ServicePort.String() == action.UseActionAnnotation {
+		if l.defaultBackend.ServiceName == Default404 {
```
M00nF1sh: i think it might be better to move this logic into where we build annotations for ingress. (e.g. if there user doesn't specify default404 action, we will insert the default one into parsed annotation, otherwise, you use user specified one), so that customer can override our 404 action 😄

internal/alb/ls/types.go
bigkraig: well since its injected after the ingress has been parsed, i think that is ok. i specifically wanted something that would not be possible for a user to enter into an ingress resource :)

M00nF1sh: got it 😄

bigkraig:
Also resolves #597

lbernail:
👍

internal/alb/ls/listener.go
stafot: @bigkraig @M00nF1sh I have exactly this issue. I want to override the default 404 action with a different one and/or produce a different error code? Is there a way to do it? I didn't find any related annotation or any other way.

PR #618, opened 2018-09-18
Generate target group IDs with hashed LB scheme

cv:
This way we can have two instances of alb-ingress-controller running, one taking care of public ingresses, and another taking care of internal ones.


cv:
…hang on a sec. Hashing the lb scheme is actually a bad idea. For my use case, simply setting the `--alb-name-prefix` flag in the command arguments yields more readable names, and solves the issue of not having distinct objects across alb-ingress-controller instances.

I'll keep the refactoring and test, but revert the behavior back to what it was before.

cv:
Closed in favor of #620.

PR #634, opened 2018-09-21
Session level API caching

bigkraig:


M00nF1sh:
Nice work! 👍 
Remaining question before putting the merge label 😸 .
1. should we leverage [LayeredCache](https://github.com/karlseguin/ccache), instead of sync.Map?(however, there will be no changes in this PR)

1. should we directly embed AWS API in our wrapper API interfaces? It's convenient, but we are locked to specific aws go sdk version (unrelated to this PR, but just wondering what should the correct abstraction level in our AWS API wrapper, and whether we should do cache there, since caching on session level is a "kind of hacky")

1. should we indeed need caching, once we implemented "only refresh specific ingress when changes happen"?

bigkraig:
1. That's a good idea. Can backlog it on the caching package.
2. I need to be convinced on having an opinion on this one way or the other.. I'm not really worried about being locked to the AWS SDK version, maybe I should be? 
3. We will probably still need it for larger deployments. The API throttles at relatively low rates and is troublesome to track down what is making API calls when you hit the limit. Anything we can do to reduce superfluous queries is good in my book.

M00nF1sh:
For #2, actually i'm ok with both solution if we adding more integration test cases for the caching(just incase aws break the internal implementation some how).

For #3, i'm afraid our local caching might cause problems. E.g. our ingress controller wants to update securityGroups on nodes, but it overrides changes make by other components/user due to there is new API like addSecurityGroup, and aws doesn't support compareAndSet.

PR #650, opened 2018-09-27
WIP: Created a controller to manage ALB attributes

bigkraig:


internal/alb/lb/attributes.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/aws/aws-sdk-go/service/elbv2/elbv2iface"
+	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/albctx"
+	"github.com/kubernetes-sigs/aws-alb-ingress-controller/pkg/util/log"
+	api "k8s.io/api/core/v1"
+)
+
+const (
+	DeletionProtectionEnabledString = "deletion_protection.enabled"
```
M00nF1sh: Maybe like DeletionProtectionEnabledKey and DefaultDeletionProtectionEnabled ? 
but current workings for me too 😸 , just nitpicking...

internal/alb/lb/attributes.go
```diff
@@ -0,0 +1,204 @@
...
+	a = &Attributes{
+		DeletionProtectionEnabled: DeletionProtectionEnabled,
+		AccessLogsS3Enabled:       AccessLogsS3Enabled,
+		AccessLogsS3Bucket:        AccessLogsS3Bucket,
+		AccessLogsS3Prefix:        AccessLogsS3Prefix,
+		IdleTimeoutTimeoutSeconds: IdleTimeoutTimeoutSeconds,
+		RoutingHTTP2Enabled:       RoutingHTTP2Enabled,
+	}
+	for _, attr := range attrs {
+		switch aws.StringValue(attr.Key) {
```
M00nF1sh: could be switch attrKey := aws.StringValue(attr.Key); attrKey{...

so avoids lot's of aws.StringValue(attr.Key) in the error report...
nitpicking too 😸

internal/alb/lb/attributes.go
```diff
@@ -0,0 +1,204 @@
...
+			if a.IdleTimeoutTimeoutSeconds < 1 || a.IdleTimeoutTimeoutSeconds > 4000 {
+				return a, fmt.Errorf("%s must be within 1-4000 seconds", aws.StringValue(attr.Key))
+			}
+		case RoutingHTTP2EnabledString:
+			a.RoutingHTTP2Enabled, err = strconv.ParseBool(aws.StringValue(attr.Value))
+			if err != nil {
+				return a, fmt.Errorf("invalid load balancer attribute value %s=%s", aws.StringValue(attr.Key), aws.StringValue(attr.Value))
+			}
+		default:
+			return a, fmt.Errorf("invalid load balancer attribute %s", aws.StringValue(attr.Key))
```
M00nF1sh: i think it should be a warning via log here. Or add another params to the NewAttribute to indicate whether we are build from user's annotation or elbv2's response.

Since AWS elbv2 might adds new default attributes to lb, and our controller will error. We should be forward compatible too

internal/alb/lb/attributes.go
```diff
@@ -0,0 +1,204 @@
...
+func (c *attributesController) Reconcile(ctx context.Context, desired *Attributes) error {
+	raw, err := c.elbv2.DescribeLoadBalancerAttributes(&elbv2.DescribeLoadBalancerAttributesInput{
+		LoadBalancerArn: aws.String(desired.LbArn),
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to retrieve attributes from ELBV2 in AWS: %s", err.Error())
+	}
+
+	current, _ := NewAttributes(raw.Attributes)
```
M00nF1sh: ignored error isn't good enough
suppose a default new attribute get added in aws and appears first in attribute list.
All the attributes we get in current will be default inited, and will cause we always modify attributes over and over...

internal/alb/lb/attributes.go
```diff
@@ -0,0 +1,204 @@
...
+			return fmt.Errorf("failed modifying attributes: %s", err)
+		}
+
+	}
+	return nil
+}
+
+// attributesChangeSet returns a list of elbv2.LoadBalancerAttribute required to change a into b
+func attributesChangeSet(a, b *Attributes) (changeSet []*elbv2.LoadBalancerAttribute, ok bool) {
+	if a.DeletionProtectionEnabled != b.DeletionProtectionEnabled && b.DeletionProtectionEnabled != DeletionProtectionEnabled {
```
M00nF1sh: is the default compare desired?
I mean if user  changed b.DeletionProtectionEnabled  to true(via anno), and then change it back to false.
The a. DeletionProtectionEnabled(on alb) will always stays in true.

If we want to model the concept of optional attribute(like unset), we should use pointers to better model it. e.g. DeletionProtectionEnabled: *bool, which is why aws using *string instead of string. Hope golang got optional[bool] :D

internal/alb/lb/attributes.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	if a.RoutingHTTP2Enabled != b.RoutingHTTP2Enabled && b.RoutingHTTP2Enabled != RoutingHTTP2Enabled {
+		changeSet = append(changeSet, &elbv2.LoadBalancerAttribute{
+			Key:   aws.String(RoutingHTTP2EnabledString),
+			Value: aws.String(fmt.Sprintf("%v", b.RoutingHTTP2Enabled)),
+		})
+	}
+
+	if len(changeSet) > 0 {
```
M00nF1sh: i think we don't need the ok here, we can test for len(changeSet) that used the result of attributesChangeSet.
which is more clear to express the logic ("if changeSet is not empty, then..")
😸

internal/alb/lb/attributes.go
bigkraig: either way the error should be pushed up to fail the reconcile and it will be added with an event and error log there

PR #653, opened 2018-09-28
Added a tag controller

bigkraig:


internal/alb/tags/tags.go
```diff
@@ -0,0 +1,161 @@
...
+	t := &Tags{Tags: make(map[string]string)}
+	for i := range m {
+		for k, v := range m[i] {
+			t.Tags[k] = v
+		}
+	}
+	return t
+}
+
+func (t *Tags) Copy() *Tags {
```
M00nF1sh: i think can be replaced with NewTags(t.Tags) 😸

internal/alb/tags/tags.go
```diff
@@ -0,0 +1,161 @@
...
+func (c *tagsController) reconcileElasticloadbalancing(ctx context.Context, desired *Tags) error {
+	current := NewTags()
+
+	resp, err := c.elbv2.DescribeTags(&elbv2.DescribeTagsInput{ResourceArns: []*string{aws.String(desired.Arn)}})
+	if err != nil {
+		return err
+	}
+
+	for _, tagDescription := range resp.TagDescriptions {
+		if aws.StringValue(tagDescription.ResourceArn) != desired.Arn {
```
M00nF1sh: is this possible? If not, I'd rather don't leave it here..

internal/alb/tags/tags.go
```diff
@@ -0,0 +1,161 @@
...
+	modify, remove := tagsChangeSet(current, desired)
+
+	if len(modify) > 0 {
+		albctx.GetLogger(ctx).Infof("Modifying tags on %v to %v", desired.Arn, log.Prettify(modify))
+
+		addParams := &elbv2.AddTagsInput{
+			ResourceArns: []*string{aws.String(desired.Arn)},
+			Tags:         mapToELBV2(modify),
+		}
+		if _, err := c.elbv2.AddTags(addParams); err != nil {
```
M00nF1sh: resource tagging api have similar api, but in service agnostic way

internal/alb/tags/tags.go
```diff
@@ -0,0 +1,161 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// tagsChangeSet compares b to a, returning a map of tags to add/change to a and a list of tags to remove from a
+func tagsChangeSet(a, b *Tags) (map[string]string, []string) {
+	modify := make(map[string]string)
```
M00nF1sh: this aligns with current behavior iirc.

But should we be more restrictive.
e.g. only maintain tags begin with specific prefix like alb.k8s.io. (I believe customers want to add their own custom tags too).
Similar situation can apply to securityGroup rules, i'm considering use some fixed prefix as description, and only manage rules with such prefix, which will allow customers add their own custom rules

M00nF1sh: as long as we support custom tags via annotation, this should be fine for now 😄

PR #664, opened 2018-10-04
Rules controller

bigkraig:


internal/alb/rs/rules.go
```diff
@@ -3,202 +3,339 @@ package rs
...
+		Ingress: ingress,
+	}
+}
 
-		if *r.Actions[0].Type == elbv2.ActionTypeEnumForward {
-			i, tg := o.TargetGroups.FindCurrentByARN(*r.Actions[0].TargetGroupArn)
-			if i < 0 {
-				return nil, fmt.Errorf("failed to find a target group associated with a rule. This should not be possible. Rule: %s, ARN: %s", awsutil.Prettify(r.RuleArn), *r.Actions[0].TargetGroupArn)
+func (rs *Rules) TargetGroupArns() (result []string) {
+	for _, rule := range rs.Rules {
```
M00nF1sh: i don't think we need to reply these arns to know which target group we need.
Since when we build these target groups, we already know it. (this can be changed later when we refactoring more)

internal/alb/rs/rules.go
```diff
@@ -3,202 +3,339 @@ package rs
...
 		}
+	}
+	return output, nil
+}
 
-		for _, rule := range r {
-			if rule.rs.current == nil {
-				continue TG
+func (c *rulesController) getCurrentRules(listenerArn string) (results []*Rule, err error) {
+	p := request.Pagination{
```
M00nF1sh: I'd rather not deal with terrible aws pagination api in application logic(which makes the unit test hard).
Instead, i'd prefer to create a wrapper api in elbv2 that returns a list of rules directly.

internal/alb/rs/rules.go
```diff
@@ -3,202 +3,339 @@ package rs
...
-TG:
-	for _, t := range tgs {
-		used := false
+func (c *rulesController) getDesiredRules(ingress *extensions.Ingress, targetGroups tg.TargetGroups) ([]*Rule, error) {
+	var output []*Rule
 
-		arn := t.CurrentARN()
-		if arn == nil {
-			continue
+	currentPriority := 1
```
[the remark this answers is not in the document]
bigkraig: managed via the default action on the listener. has to be done there because the listener needs the default rule at creation :+1:

M00nF1sh:
i guess you have to close & reopen this pr to trigger travis again 😸

bigkraig:
@M00nF1sh I rebased it on master but its going to need another lgtm

M00nF1sh:
/lgtm
😄

PR #670, opened 2018-10-10
Fixes a problem where we inject a bad path rule that nobody asked for.

bigkraig:


internal/ingress/controller/store/store.go
```diff
@@ -519,16 +519,10 @@ func (s k8sStore) ListIngresses() []*extensions.Ingress {
 			continue
 		}
 
-		for ri, rule := range ing.Spec.Rules {
+		for _, rule := range ing.Spec.Rules {
```
M00nF1sh: i guess we don't need this for loop=.=

bigkraig: if we get rid of it we will need to go through the rest of the code and check that rule.HTTP isn't nil, because it can be undefined

Issue #681, opened 2018-10-17
Managed SG not being associated with ENI's.


timmyers:
The controller is successfully provisioning and configuring all resources, including managed security groups, but is not associating the instance security group with the ENI's.

This in EKS using `'alb.ingress.kubernetes.io/target-type': 'ip'`

From the debugging I have performed, it seems that in the `Reconcile()` of `instance_attachment.go`, the `supportingENIs` are not being successfully found.

The `attachment *InstanceAttachment` parameter passed in has the following value:
```
{"GroupID":"sg-052bf7e4b2694b2b5","Targets":[{"ID":"98993eff-ff32dd8d315234ffa7a","SvcName":"nginx-75n7ln4m","SvcPort":80,"TargetType":"ip"}]}
```

Which seems to be missing the detailed target information that is needed to find the supporting ENI's.  The target group is indeed getting all of the appropriate targets registered to it.

timmyers:
I really thought I was `quay.io/coreos/alb-ingress-controller:1.0-beta.7` when encountering this, but I think I was accidentally on `quay.io/coreos/alb-ingress-controller:latest`.

This seems to work okay on `1.0-beta.7`, closing.

PR #684, opened 2018-10-19
Return targets details in targets Reconcile() function.

timmyers:
This fixes a bug #681 where managed SG was not getting correctly associated to supporting ENI's because the target details were not available..


timmyers:
CLA test.

PR #694, opened 2018-10-25
Refactor AWS client

bigkraig:


bigkraig:
the only place left with the global `Cloudsvc` is in `loadbalancer/main.go` now

internal/ingress/metric/main_test.go
```diff
@@ -0,0 +1 @@
+package metric
```
M00nF1sh: what's this used for?

bigkraig: makes coveralls count the package in our overall test coverage

PR #697, opened 2018-10-26
Some refactors and code cleanup

bigkraig:


internal/aws/ec2.go
```diff
@@ -63,7 +64,13 @@ type EC2API interface {
 	DeleteSecurityGroupByID(string) error
 
 	// ClusterSubnets returns the subnets that are tagged for the cluster
-	ClusterSubnets(scheme *string) (util.Subnets, error)
+	ClusterSubnets(scheme string) ([]string, error)
+
+	// ResolveSecurityGroupNames returns the security group ids for a list of security group names and ids
+	ResolveSecurityGroupNames(context.Context, []string) ([]string, error)
```
M00nF1sh: Did the resolveSecurityGroupNames and ResolveSubnets belong to CloudSvc?
My current mental model for cloudSvc is enhancement for aws-sdk-go, which provides better usable API without application specific api.(which should potentially  be used in other controllers as well).
e.g. for ResolveSubnets, it embed ingress controller specific implementation details, like ingress schema, tagged subnets. I'd preferrer declare this method in alb/lb package.

For the GetSecurityGroups, I'd think we can rename to like GetSecurityGroupsByName or GetSecurityGroupsByNames, which returns []*ec2.SecurityGroup or []ec2.SecurityGroup, or map[string]ec2.SecurityGroup, which seems more general

internal/aws/ec2.go
```diff
@@ -175,6 +186,70 @@ func (c *Cloud) GetSecurityGroupByID(groupID string) (*ec2.SecurityGroup, error)
 	return securityGroups[0], nil
 }
 
+func (c *Cloud) ResolveSecurityGroupNames(ctx context.Context, in []string) ([]string, error) {
```
M00nF1sh: I'd preferrer clear name like in -> sgIDOrNames   similar to the intOrString in k8s core.

internal/alb/lb/loadbalancer.go
```diff
@@ -321,3 +334,143 @@ func (controller *defaultController) validateLBConfig(ctx context.Context, ingre
...
+			if aws.StringValue(tag.Key) == key {
+				p := strings.Split(arn, "/")
+				subnetID := p[len(p)-1]
+				subnetIds = append(subnetIds, subnetID)
+			}
+		}
+	}
+
+	if len(subnetIds) == 0 {
+		sort.Strings(out)
```
M00nF1sh: seems unneeded 😄

bigkraig: the sorting helps when doing comparisons for changes

M00nF1sh: but seems out is empty at this branch 😅

bigkraig: now i see :)

PR #710, opened 2018-11-01
Fix typo

xichengliudui:
Fix typo


xichengliudui:
If don't need to change it, please let me know

M00nF1sh:
omm....i still think it should be `their` instead of `there`... 
@bigkraig you are the authority here 🤣

bigkraig:
Right, I think this is the proper 'their', since its being used to show possession of the values. Thanks for checking though! 👍

PR #711, opened 2018-11-01
Optionally apply default tags to all resources.

### LINKED ISSUE #187: Allow default set of AWS Tags to be specified on the controller
[enhancement]
It would be convenient if I could specify a default set of Tags on the controller, such that all ALBs it creates would be given those tags (some of which might be overridden by an individual Ingress object's annotations).

jmcarp:
If the approach makes sense, I'll add tests and docs.

[Fixes #187]


bigkraig:
:+1: this looks like a good approach to me

M00nF1sh:
👍 looks good to me too.(we can add support to tag securityGroup in similar way)

M00nF1sh:
@jmcarp 
Hi, this changes looks good to me, would you help adding some tests and docs, so we can merge this 👍

jmcarp:
Added tests and docs.

M00nF1sh:
/lgtm
Thanks for make this contribution 👍

PR #723, opened 2018-11-07
[WIP]resolve vpcID and region dynamically

### LINKED ISSUE #687: Fails without access to metadata server *and instance id*
EC2Metadatasvc is used to get the VPC ID in a round-about-fashion (metadata service -> instance id -> vpc id).

This broke it for us, because we have the metadata service firewalled off from pods, to prevent privilege escalation attacks (and were supplying a secret and name to the controller).

Please consider using the ExternalID field of the node descriptor instead, which should be readily accessible to a service account such as the ingress controller requires

M00nF1sh:
Changes:
1. Refactor the option/configs handling, to make settings more coherence
1. resolve vpcID by ec2Metadata alone(ec2.DescribeInstances are not needed)
1. Add vpcID and region as cloud configuration. Which will be resolved from ec2Metadata by default.(user can manually specify it as flags or environment variable to break the dependency of ec2Metadata)
1. fixes #687 

Tests done:
- power ingress successfully.

Pending work:
- [ ] fix lint & tests


cmd/options.go
```diff
@@ -64,126 +57,83 @@ type Options struct {
...
-	}
-	if options.config.ALBNamePrefix == "" {
-		options.config.ALBNamePrefix = generateALBNamePrefix(options.config.ClusterName)
-	}
-
-	// check port collisions
-	if !ing_net.IsPortAvailable(options.HealthzPort) {
+func (options *Options) Validate() error {
+	if !net.IsPortAvailable(options.HealthzPort) {
 		return fmt.Errorf("port %v is alreadt in use. Please check the flag --healthz-port", options.HealthzPort)
```
bigkraig: s/alreadt/already/

PR #732, opened 2018-11-14
Allow the target service to be type LoadBalancer or type NodePort

hatmatter:
LoadBalancer types also bind to a NodePort in order to route traffic into the cluster so we should be able to support both types here, not just NodePort.


hatmatter:
CLA should be taken care of, my organization  is approved and I have signed in to the CNCF site

M00nF1sh:
@hatmatter 
Thanks for contributing 👍 
have you setup email in your commit?

hatmatter:
Ok, that should be all straightened out now I think?

Issue #573, opened 2018-08-22
How do I create rule without `path-pattern` ?


sawanoboly:
I thought that using this plugin, I would like to create a rule with only the host header. The definition is as follows.

```
spec:
  backend:
    serviceName: fallback-nginx
    servicePort: 80
  rules:
    - host: myapp1.example.com
      http:
        paths:
          - backend:
              serviceName: myapp1
              servicePort: 80
```

However, path-pattern '/' is automatically added. I want to exclude the path-pattern from the rule, is it possible?

```
$ aws elbv2 describe-rules  --rule-arns arn:aws:elasticloadbalancing:us-west-2:xxxxxx:listener-rule/app/xxxxxx
{
    "Rules": [
        {
            "RuleArn": "arn:aws:elasticloadbalancing:us-west-2:xxxxxx:listener-rule/app/xxxxxx",
            "Priority": "1",
            "Conditions": [
                {
                    "Field": "path-pattern",
                    "Values": [
                        "/"
                    ]
                },
                {
                    "Field": "host-header",
                    "Values": [
                        "myapp1.example.com"
                    ]
                }
            ],
            "Actions": [
                {
                    "Type": "forward",
                    "TargetGroupArn": "arn:aws:elasticloadbalancing:us-west-2:xxxxxx:listener-rule/app/xxxxxx"
                }
            ],
            "IsDefault": false
        }
    ]
}
```

bigkraig:
It's not possible, the ALB requires that a rule matches a path. Does this create an issue or is it just unexpected?

sawanoboly:
https://docs.aws.amazon.com/ja_jp/elasticloadbalancing/latest/APIReference/API_RuleCondition.html

ALB allows host-header only rule, path-pattern could omitted.

![ec2 management console 2018-09-06 02-56-03](https://user-images.githubusercontent.com/272388/45111663-7b751100-b180-11e8-8e73-4ec7e9094964.png)

Current situation If we want to use host name based routing using this plugin, we need to set `/*` to path-pattern for all. This means that we can only use half the maximum number of if matches.

rabbitfang:
I am also being affected by this issue. It appears this bug was introduced in #476 (first released as `1.0-beta.5`) in this repo; kubernetes/ingress-nginx#2244 introduced the behavior for the nginx ingress controller, which works for nginx because the paths in nginx are prefixes, not exact matches or globs.

[...]

PR #736, opened 2018-11-16
fix empty path handling

M00nF1sh:
Changes done:
1. When ingress path is empty, we'll make the rule to be '/*', which aligns the ingress spec of "catch all traffic"
- This is required since there should be at least one condition for ALB rules.
- This is required if you modify the rules with specify condition, the condition will not be override. (e.g. remove host/path from ingress spec, the host/path condition won't be modified in ALB rule).

Tests done:
1. Create ingress without path => ALB rule with "/*" are created.
1. 
    - Create ingress with hostname/path => ALB rule with host/condition are created.
    - Remove host/path from ingress => ALB rule modified to have only "/*" path condition

### DIFF
--- a/internal/alb/rs/rules.go
+++ b/internal/alb/rs/rules.go
@@ -158,6 +158,8 @@ func (c *defaultController) getDesiredRules(listener *elbv2.Listener, ingress *e
 
 			if path.Path != "" {
 				elbRule.Conditions = append(elbRule.Conditions, condition("path-pattern", path.Path))
+			} else {
+				elbRule.Conditions = append(elbRule.Conditions, condition("path-pattern", "/*"))
 			}
 
 			if createsRedirectLoop(listener, elbRule) {


sawanoboly:
@M00nF1sh This change conflicts with Issue below. Please consider about Hostname only entry. 🙏 

https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/573#issuecomment-453812104

M00nF1sh:
> @M00nF1sh This change conflicts with Issue below. Please consider about Hostname only entry. 🙏
> 
> [#573 (comment)](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/573#issuecomment-453812104)

@sawanoboly , thanks for sharing the use case 👍 
I'll provide a fix today to change the behavior to only maps to "/*" when host condition is not specified.(it's required to have at least one rule)

PR #754, opened 2018-11-28
cognito/oidc auth support

M00nF1sh:
# Proposal for Auth Support
## Requirement
1. Support auth provider of  OIDC and AWS Cognito
1. Support configure auth per rule(host/path)
1. Support load OIDC clientSecrets from kubernetes secrets

### Alternative1: Use dedicated auth annotation (prefer this with 👍   )
We can use dedicated set of annotations to provide authentication support.

| annotation             | description                               | example        |
| ------------------ | ----------------------------  |------------- |
|auth-type               | authentication type(required for express non-auth while  specify default auth cfg at controller level)|oidc\|cognito\|none|
| auth-idp-cognito | Cognito specific configuration |'{'UserPoolArn': 'xxx', 'UserPoolClientId': 'xxx',  'UserPoolDomain':'xxx'}' |
| auth-idp-oidc        | OIDC specific configuration     |'{'Issuer':'xxx', 'ClientID':'xxx', 'ClientSecret':'xxx', 'AuthorizationEndpoint':'xxx', 'TokenEndpoint':'xxx', 'UserInfoEndpoint':'xxx'} |
|auth-scope| The set of user claims to be requested from the IdP| openid|
|auth-session-cookie| name of session cookie| AWSELBAuthSessionCookie|
|auth-session-timeout|time of session|86400|

#### How to configure auth per rule?
The above set of annotations can be applied to both ingress / service. If it's applied to ingress, all host/path under that ingress will be impacted. If it's applied to service, only backend to such service will be impacted.
However, the above approach still have below limitations:
* We cannot use different auth config for same service under different ingress(If we don't want to apply global auth annotation to these ingresses)
* We cannot use different auth config for same service but different port(if we don't want to apply global auth annotation to these ingresses)
* We cannot specify auth config for redirect/fixed-response action introduced by the `use-annotation` targets(if we don't want to apply global auth annotation to these ingresses)

The above limitation come from a fundamental problem that we cannot customize ingress behavior for per rule(cannot specify annotations that only apply to specific rules). This can be solved by introduce support for `ingress group` as describe in https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/688

#### Pros
* Users don't need to deal with the complicated action annotation configuration
* User can apply global auth config at ingress level
* Common auth config like scope/session-cookie that are shared by OIDC or Cognito is abstracted out
* We can limiting the action annotation usage to only for redirect/fixed-response
#### Cons
* The above limitation exists until we implemented `ingress group` support.

### Alternative2: Extend current [`alb.ingress.kubernetes.io/actions`](https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/ingress/annotation/#actions) annotation. (prefer this with 😃  )
The `actions` annotation is introduce to support redirect and fixed-response action and it's an single entry. However in ALB API, the actions is the combination of `auth` action(optional) and response action(forward to targetGroup/redirect/fixed-response).
To extend this, we need to allow specify an array of actions(while also accept an single action).
Also, we need to represent `forward to targetGroup action`(need to encode serviceName and servicePort in this action config)

An sample action that enables OIDC support can be like below:
```yaml
alb.ingress.kubernetes.io/actions: >-
[{
          "Type": "authenticate-oidc",
          "AuthenticateOidcConfig": {
              "Issuer": "https://idp-issuer.com",
              "AuthorizationEndpoint": "https://authorization-endpoint.com",
              "TokenEndpoint": "https://token-endpoint.com",
              "UserInfoEndpoint": "https://user-info-endpoint.com",
              "ClientId": "abcdefghijklmnopqrstuvwxyz123456789",
              "ClientSecretSecretName": "",     // -< name of kubernetes secret, we'll load the content from it.
              "SessionCookieName": "my-cookie",
              "SessionTimeout": 3600,
              "Scope": "email",
              "AuthenticationRequestExtraParams": {
                  "display": "page",
                  "prompt": "login"
              },
              "OnUnauthenticatedRequest": "deny"
          },
}, {
       "Type": "forward",
       "ServiceName": "xxx",
       "ServicePort": "xxxx"
}]
```

#### Pros:
* It's an extension to existing action annotation
#### Cons:
* The action annotation become more complicated.
* The action annotation won't be a direct mapping to ALB actions, since 
   * we need to include serviceName & servicePort in forward action. And the targetGroup component should honor it when creating targetGroups.
   * introduce of ClientSecretSecretName which will be the k8s secret name to load clientSecret.

### Alternative3:(prefer this with ❤️ )
Help contribute your preferred alternative as comments .

# References
https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/592


hluchej:
Hi, the proposal looks nice and I voted for Alternative 1.

I am just thinking about a possible extension of this proposal regarding injection of the OIDC client secret.

We are currently using AWS Secrets Manager for many purposes and we would likely prefer to fetch the secret value from there instead of from K8s Secret. In our applications, we are not really forced to use k8s secrets because we can provide just ARN of the secret in Secrets Manager via e.g. env property and apps can fetch the value from there.

We cannot use this approach with ingress controller because there are no "our" services where we would implement the fetching so the fetching has to be part of the ingress controller.

So I was thinking about extending the _auth-oidc-cfg_ annotation. Example:
```
{
'Issuer':'xxx', 
'ClientID':'xxx', 
'ClientSecretSecretsManagerId':'xxx', 
'AuthorizationEndpoint':'xxx', 
'TokenEndpoint':'xxx', 
'UserInfoEndpoint':'xxx'
}
```

where _ClientSecretSecretsManagerId_ is basically this: https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_GetSecretValue.html#SecretsManager-GetSecretValue-request-SecretId - therefore can contain full ARN of the secret or "friendly name" of the secret.
Obviously only one of the _ClientSecretSecretManagerId_ and _ClientSecretSecretName_ can be specified, otherwise configuration error should be raised.

**Pros**:

- Adds option to use AWS Secrets Manager for OIDC credentials storage

**Cons**:

- additional complexity

What do you think?

M00nF1sh:
The PR should be done by tomorrow, and i have some concerns around the usage of 
ClientSecretSecretName. Since this introduced a `privilege escalation`, someone creating ingress can stolen secrets that don't belong to him=.=

There is two way for this:
1. Via documentation to tell users to only grant controller permission to read specific secrets. 
1. Using a central managed secret(configurable via controller flag), so only certain k8s admin can manage that secret. The format of that secret can be a map of clientId: clientSecret.

Which one do you guys prefers 😄

JakubJecminek:
+1 for option 1

M00nF1sh:
current code works with some imperfection:
* the listener actions will be updated every time when sync events happens(ingress/service changes or  
every 60minutes by default). This is due to we cannot retrieve clientSecret via API on aws, so we cannot compare).

I can change it to be:
1. when compare the actions, don't check for clientSecret. (users will need to delete the auth-type annotation and add it back, if only clientSecret are changed).

To avoid such change
Another change could be record the desired status in last sync, and compare them, and only sync when anything changed. But this will a a little more complicated.

hluchej:
Hi, I am not a security expert so I cannot judge regarding forcing the SecretNamespace...

> I can change it to be: when compare the actions, don't check for clientSecret. (users will need to delete the auth-type annotation and add it back, if only clientSecret are changed)

I would not like that change. Seems to require external action which is error prone and it will be really hard to realize why the change of clientSecret did not propagate... We sometimes regenerate secrets (e.g. because employee offboarding) and clientId does not change in these cases. So at least Okta does not create a new clientId when you want to rotate the secret.

M00nF1sh:
@hluchej Thanks for your comments, these points are pretty valuable. I'll add watch on the secrets referenced by ingresses to trigger sync when they changed.

goruha:
@M00nF1sh when do you expect to have the PR merged?
I'd like to test it asap )

M00nF1sh:
> @M00nF1sh when do you expect to have the PR merged?
> I'd like to test it asap )

Hi, it's currently under an internal sec review, i'll update the status here once it's finished. (it should went through smoothly)

M00nF1sh:
> @M00nF1sh it would be perfect.
> I would be happy to help you with beta testing.

Hi, sorry for the late reply, just get back from holiday trip 😃 
I have pushed the changes to https://github.com/M00nF1sh/aws-alb-ingress-controller/tree/cognito-support.

Sample ingress I have tested:
```
kind: Deployment
metadata:
  namespace: testcase
  name: echoserver
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: echoserver
    spec:
      containers:
      - name: echoserver
        image: gcr.io/google_containers/echoserver:1.4
        imagePullPolicy: Always
        ports:
        - containerPort: 8080
---
apiVersion: v1
kind: Service
metadata:
  namespace: testcase
  name: echoserver
spec:
  ports:
    - port: 80
      targetPort: 8080
      protocol: TCP
  type: NodePort
  selector:
    app: echoserver
---
apiVersion: v1
kind: Secret
metadata:
  namespace: testcase
  name: odic-secret
data:
  clientId: xxxx
  clientSecret: xxxxx
---
kind: Ingress
metadata:
  namespace: testcase
  name: echoserver
  annotations:
    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP":80},{"HTTPS":443}]'
    alb.ingress.kubernetes.io/certificate-arn: xxxxx
   alb.ingress.kubernetes.io/actions.ssl-redirect: '{"Type": "redirect", "RedirectConfig": { "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}'
    alb.ingress.kubernetes.io/auth-type: oidc
    alb.ingress.kubernetes.io/auth-idp-oidc: '{"Issuer":"https://albingress.auth0.com/","AuthorizationEndpoint":"https://albingress.auth0.com/authorize","TokenEndpoint":"https://albingress.auth0.com/oauth/token","UserInfoEndpoint":"https://albingress.auth0.com/userinfo","SecretName":"odic-secret"}'
spec:
  rules:
    - http:
        paths:
           - backend:
               serviceName: ssl-redirect
               servicePort: use-annotation
          - path: /normal-path
            backend:
              serviceName: echoserver
              servicePort: 80
```

Some tip for testing,
1.  clientId&clientSecret can be generated via `echo -n "clientIdFromOIDC" | base64`

goruha:
@M00nF1sh thanks. Will try on this week

M00nF1sh:
> @M00nF1sh thanks. Will try on this week

@goruha Have you got time to test it out?
I have tested both OIDC & Cognito works 😄 , and plans to merge it and release a new version.

goruha:
@M00nF1sh testing now. 
Will report feedback today.

goruha:
@M00nF1sh 
I have 
```
This site can’t be reached The webpage at https://example.com/ might be temporarily down or it may have moved permanently to a new web address.
ERR_SPDY_PROTOCOL_ERROR
```
 In browser.
Did I config OIDC app wrong?

M00nF1sh:
> @M00nF1sh
> I have
> 
> ```
> This site can’t be reached The webpage at https://example.com/ might be temporarily down or it may have moved permanently to a new web address.
> ERR_SPDY_PROTOCOL_ERROR
> ```
> In browser.
> Did I config OIDC app wrong?

it's strange to get SPDY Error..what's your configuration looks like?
For my auth0 based OIDC, it's works fine
```yaml
alb.ingress.kubernetes.io/auth-type: cognito
alb.ingress.kubernetes.io/auth-idp-oidc: '{"Issuer":"https://albingress.auth0.com/","AuthorizationEndpoint":"https://albingress.auth0.com/authorize","TokenEndpoint":"https://albingress.auth0.com/oauth/token","UserInfoEndpoint":"https://albingress.auth0.com/userinfo","SecretName":"odic-secret"}'
```
```yaml
apiVersion: v1
kind: Secret
metadata:
  namespace: testcase
  name: odic-secret
data:
  clientId: N014SzFiY0RSbmZyblFBd05CRGdyTWlVN0J2TG9kVXY=
  clientSecret: TnlubjBxVGxkc3FYV2ZucFgxNW1pYVpadXhsQWYtM1gzY3JwXzA4bDlOcGZ4YkdQVTZCd2oyZnlnN1hLZWpLdQ==
```
The secret(clientId&clientSecret) is generated by `echo -n "xxxxx" | base64`

M00nF1sh:
> Hey, I stumbled accross this because I wanted to do Cognito auth on my ingress alb, and after fighting with all the other issues with an alb ingress I can confirm that Cognito also looks good, I get a 401, don't really have the time right now to actually get past it, but in the AWS console it looks good

My working cognito setup is like below:
```yaml
alb.ingress.kubernetes.io/auth-idp-cognito: '{"UserPoolArn":"arn:aws:cognito-idp:us-west-2:xxxx:userpool/xxxxx", "UserPoolClientId":"xxxxx", "UserPoolDomain":"m00nf1sh"}'
```

AdeOpe:
hi,
may i please confirm which release this code change is in?
thanks

AdeOpe:
Hi,
i just tried this from your repo, its seems to work good. thank you :)

Is there a way to specify:
- Path-A to have oidc
- and Path B to not authenticate.

E.g. grafana has direct integration with oidc & oauth2 so there's no need to challenge before landing on the grafana page.
whilst other paths, i want to protect before landing on the destination

(apologies, its probably something i have missed)
any ideas?

M00nF1sh:
> Hi,
> i just tried this from your repo, its seems to work good. thank you :)
> 
> Is there a way to specify:
> 
> * Path-A to have oidc
> * and Path B to not authenticate.
> 
> E.g. grafana has direct integration with oidc & oauth2 so there's no need to challenge before landing on the grafana page.
> whilst other paths, i want to protect before landing on the destination
> 
> (apologies, its probably something i have missed)
> any ideas?

Hi,
@AdeOpe these annotations can be applied to service objects as well.
so if your ingress looks like 
```
/path1: service1
/path2: service2
```
You can add these auth annotation on service1, to get /path1 authenticated while path2 not 😄

AdeOpe:
> > Hi,
> > i just tried this from your repo, its seems to work good. thank you :)
> > Is there a way to specify:
> > 
> > * Path-A to have oidc
> > * and Path B to not authenticate.
> > 
> > E.g. grafana has direct integration with oidc & oauth2 so there's no need to challenge before landing on the grafana page.
> > whilst other paths, i want to protect before landing on the destination
> > (apologies, its probably something i have missed)
> > any ideas?
> 
> Hi,
> @AdeOpe these annotations can be applied to service objects as well.
> so if your ingress looks like
> 
> ```
> /path1: service1
> /path2: service2
> ```
> You can add these auth annotation on service1, to get /path1 authenticated while path2 not 😄


thanks for the reply @M00nF1sh , this should like what i am looking for.

I just need some direction with the implementation.
sorry to be a pain, but this is the part where i am confused. 

can this all be done directly in the ingress.yaml?
or shall i add the annotation directly to the service1.yaml?

e.g. so i currently have the following in my ingress.yaml

      - path: /path1/*
        backend:
          serviceName: app-1-service
          servicePort: 80
      - path: /path2/*
        backend:
          serviceName: app-2-service
          servicePort: 80

with the above and with oidc everything path is now integrated with oidc.
how do instruct the ingress.yaml to only use oidc for path1 and not path2?

M00nF1sh:
@AdeOpe 
Hi, currently, you have to add these annotations to the service.yaml of app-1-service.

AdeOpe:
ah ok got it. I will give it a try :)
thank you again

AdeOpe:
@M00nF1sh - i am just reporting back after my test.
this works brilliantly.
I decided to use it on the service.yaml instead of the ingress.yaml

thank you again 👍

AdeOpe:
Hi @M00nF1sh ,
can you please confirm if you have uploaded this?

I get the following error with v1.1.0 & v1.1.1

  Warning  Failed                 10s              kubelet, ip-10-1-6-205.eu-west-1.compute.internal  Failed to pull image "894847497797.dkr.ecr.us-west-2.amazonaws.com/aws-alb-ingress-controller:v1.1.0": rpc error: code = Unknown desc = Error response from daemon: manifest for 894847497797.dkr.ecr.us-west-2.amazonaws.com/aws-alb-ingress-controller:v1.1.0 not found

M00nF1sh:
> Hi @M00nF1sh ,
> can you please confirm if you have uploaded this?
> 
> I get the following error with v1.1.0 & v1.1.1
> 
> Warning Failed 10s kubelet, ip-10-1-6-205.eu-west-1.compute.internal Failed to pull image "894847497797.dkr.ecr.us-west-2.amazonaws.com/aws-alb-ingress-controller:v1.1.0": rpc error: code = Unknown desc = Error response from daemon: manifest for 894847497797.dkr.ecr.us-west-2.amazonaws.com/aws-alb-ingress-controller:v1.1.0 not found

@AdeOpe , sorry for the confusing.
The image have moved to docker.io/amazon/aws-alb-ingress-controller:v1.1.1 since v1.0.1. 
(The helm e2e tests cannot be run with ecr images, so we moved to company's dockerhub).

AdeOpe:
ah great!
thanks @M00nF1sh

ajaydevtron:
Hello @M00nF1sh 
Getting the RBAC issue while implanting  the same for get oidc secret which having client id and secret .

`" cannot get resource \"secrets\" in API group \"\" in the namespace \"xyz\""}`

ALB controller version(deployed through helm chart) :- **v2.8.1** , Does we need to modify the cluster role accordingly ?

PR #796, opened 2019-01-02
update docs

sdarwin:
When following the echoserver walkthrough some small changes were required to get it running. See commit for details.

### DIFF
--- a/docs/examples/external-dns.yaml
+++ b/docs/examples/external-dns.yaml
@@ -1,8 +1,43 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: external-dns
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  name: external-dns
+rules:
+- apiGroups: [""]
+  resources: ["services"]
+  verbs: ["get","watch","list"]
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get","watch","list"]
+- apiGroups: ["extensions"] 
+  resources: ["ingresses"] 
+  verbs: ["get","watch","list"]
+- apiGroups: [""]
+  resources: ["nodes"]
+  verbs: ["list"]
+---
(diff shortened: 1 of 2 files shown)


docs/examples/external-dns-rbac.yaml
```diff
@@ -0,0 +1,61 @@
+apiVersion: v1
```
M00nF1sh: Hi, i'd suggest we separate this into two file. rbac.yaml and deployment.yaml.
So we don't need to maintain two separate external-dns deployment configuration.
Maybe inside a separate folder like docs/example/external-dns/rbac.yaml  and docs/example/external-dns/deployment.yaml ?

docs/examples/external-dns-rbac.yaml
```diff
@@ -0,0 +1,61 @@
...
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        app: external-dns
+    spec:
+      serviceAccountName: external-dns
+      containers:
+      - name: external-dns
+        image: registry.opensource.zalan.do/teapot/external-dns:latest
```
M00nF1sh: I'd suggest ping the version to a specific version((which is current latest) instead of using "latest", the upgrade in external-dns might broken the flags or changed behavior.

docs/examples/external-dns-rbac.yaml
sdarwin: The deployments are not 100% identical.   RBAC sets serviceAccountName: external-dns
Solution: How about only including the newer RBAC version now?  Add a comment "for non-RBAC deployments refer to the external-dns tutorial", where that code is still available.

M00nF1sh: agree, let's delete the non-rbac version 👍

sdarwin: updated

docs/examples/external-dns-rbac.yaml
sdarwin: changed

Issue #799, opened 2019-01-03
Healthcheck for service on a secondary port when using NodePort


backjo:
On a service, we can currently specify "alb.ingress.kubernetes.io/healthcheck-port" to have the healthcheck for the generated TargetGroup use a secondary port. It would be nice to have some way to configure this to accept a port name, so that the ALB Ingress Controller could lookup the NodePort for that name and use it

M00nF1sh:
seems like a legitimate use case. The only troubling part i can think of is it currently support an special port called "traffic-port", if some user do have a port called "traffic-port" on their service(unlikely this will happen though), we need to be clear about the behavior in this case and document it.
/kind feature

backjo:
Implemented in #802

PR #802, opened 2019-01-05
Resolve NodePort / TargetPort when a port name is passed in as a service health check port

backjo:
First attempt at implementing #799

### DIFF
--- a/internal/alb/tg/targetgroup.go
+++ b/internal/alb/tg/targetgroup.go
@@ -3,20 +3,22 @@ package tg
 import (
 	"context"
 	"fmt"
-
-	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/albctx"
-
-	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/k8s"
+	"strconv"
 
 	"github.com/aws/aws-sdk-go/service/elbv2"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/alb/tags"
+	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/albctx"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/aws"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/annotations"
+	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/annotations/healthcheck"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/backend"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/controller/store"
+	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/k8s"
 	util "github.com/kubernetes-sigs/aws-alb-ingress-controller/pkg/util/types"
+	"github.com/pkg/errors"
 	extensions "k8s.io/api/extensions/v1beta1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // The port used when creating targetGroup serves as a default value for targets registered without port specified.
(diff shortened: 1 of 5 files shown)


internal/alb/tg/targetgroup.go
```diff
@@ -150,6 +156,33 @@ func (controller *defaultController) reconcileTGInstance(ctx context.Context, in
 	return instance, nil
 }
 
+// resolveServiceHealthCheckPort checks if the service-port annotation is a string. If so, it tries to look up a port with the same name
+// on the service and use that port's NodePort as the health check port.
+func resolveServiceHealthCheckPort(store store.Storer, namespace string, serviceName string, servicePortAnnotation intstr.IntOrString) (* string) {
```
M00nF1sh: 1. I'd prefer to make this an member function of controller, so we don't need to pass singleton object(store) around.
1. I'd prefer to always return string instead of *string, when there is no "optional" semantic. (Ideally, every *xxx needs to be null-checked, otherwise, it should be xxx).

internal/alb/tg/targetgroup.go
```diff
@@ -150,6 +156,33 @@ func (controller *defaultController) reconcileTGInstance(ctx context.Context, in
...
+	servicePort := servicePortAnnotation.String()
+
+	//If the annotation uses the default port ("traffic-port"), do not try to look up a port by that name.
+	if servicePortAnnotation.Type == intstr.String && servicePortAnnotation.StrVal != healthcheck.DefaultPort {
```
M00nF1sh: 1. i think we can create an utility function under internal/k8s package.(which can be leveraged by https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/aeb86a388d1e565cc6629a2b1d23fba3dcbdbc47/internal/ingress/backend/endpoint.go#L114 as well)
Something like 
```lookupServicePort(service corev1.Service, port intstr.IntOrString) (corev1.ServicePort error)```sounds good to me =.=(don't make store part of it)

File: internal/alb/tg/targetgroup.go [hunk 3]
```diff
@@ -150,6 +156,33 @@ func (controller *defaultController) reconcileTGInstance(ctx context.Context, in
...
+		serviceKey := namespace + "/" + serviceName
+
+		service, err := store.GetService(serviceKey)
+		if err != nil {
+			return &servicePort
+		}
+
+		for _, p := range service.Spec.Ports {
+			if p.Name == servicePortAnnotation.StrVal && p.NodePort != 0 {
```
M00nF1sh: this only works under instance mode, for ip mode(which targets pods directly), it should be the`targetPort`

backjo:
@M00nF1sh addressed your initial comments

internal/alb/tg/targetgroup.go
```diff
@@ -97,20 +101,21 @@ func (controller *defaultController) Reconcile(ctx context.Context, ingress *ext
...
+
 	return TargetGroup{
 		Arn:        tgArn,
 		TargetType: targetType,
 		Targets:    tgTargets.Targets,
 	}, nil
 }
 
-func (controller *defaultController) newTGInstance(ctx context.Context, name string, serviceAnnos *annotations.Service) (*elbv2.TargetGroup, error) {
+func (controller *defaultController) newTGInstance(ctx context.Context, name string, serviceAnnos *annotations.Service, healthCheckPort *string) (*elbv2.TargetGroup, error) {
```
M00nF1sh: better to use string as argument, and pass to underline api with aws.String(healthCheckPort)

internal/alb/tg/targetgroup.go
```diff
@@ -128,14 +133,15 @@ func (controller *defaultController) newTGInstance(ctx context.Context, name str
 	return tgInstance, nil
 }
 
-func (controller *defaultController) reconcileTGInstance(ctx context.Context, instance *elbv2.TargetGroup, serviceAnnos *annotations.Service) (*elbv2.TargetGroup, error) {
+func (controller *defaultController) reconcileTGInstance(ctx context.Context, instance *elbv2.TargetGroup, serviceAnnos *annotations.Service, healthCheckPort *string) (*elbv2.TargetGroup, error) {
```
M00nF1sh: better to use string as argument, and pass to underline api with aws.String(healthCheckPort)

internal/alb/tg/targetgroup.go
```diff
@@ -150,6 +156,37 @@ func (controller *defaultController) reconcileTGInstance(ctx context.Context, in
...
+		if err != nil {
+			return servicePort
+		}
+
+		resolvedServicePort, err := k8s.LookupServicePort(service, servicePortAnnotation)
+		if err != nil {
+			return servicePort
+		}
+
+		if resolvedServicePort.Name == servicePortAnnotation.StrVal && resolvedServicePort.NodePort != 0 {
```
M00nF1sh: 1. the resolvedServicePort.Name == servicePortAnnotation.StrVal is unneeded, which is already expressed in k8s.LookupServicePort.

1. the logic nest order should be changed to below, since mode IP don't require nodePort
```
if targetType == elbv2.TargetTypeEnumInstance {
    if resolvedServicePort.NodePort == 0 {
        return "", errors.New("service must be type NodePort for named healthCheckPort")
    }
   return strconv.Itoa(int(resolvedServicePort.NodePort)), nil
}
return resolvedServicePort.TargetPort.String()
```

M00nF1sh:
overall lgtm, added some final set comments 😅 ...i think we can keep the the limitation of "cannot resolve namedPort on pods) and document it somewhere =.=.

M00nF1sh:
perfect! Thanks for contributing 🤣 

/lgtm
/approve

PR #808, opened 2019-01-10
Fix link for BUILDING.md in README.md

seanvm:
Changes:
* Fixes the link to `BUILDING.md` within `README.md`. Link now points to the live doc site.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ To get started with the controller, see our [walkthrough](https://kubernetes-sig
 
 ## Building
 
-For details on building this project, see [BUILDING.md](/docs/BUILDING.md).
+For details on building this project, see our [building guide](https://kubernetes-sigs.github.io/aws-alb-ingress-controller/BUILDING/).
 
 ## Community, discussion, contribution, and support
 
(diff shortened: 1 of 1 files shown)


M00nF1sh:
Thanks for fixing this.
This will break the live Docs website =.= need to find a way to mitigate it

seanvm:
@M00nF1sh - I'll update the link to work with the doc site.

M00nF1sh:
> @M00nF1sh - I'll update the link to work with the doc site.

Hi, sorry for not been clearly enough...this won't work either, since the live doc site is generated by this   README as well.

The simplest fix is:
1. keep the root/README.MD as it is, and create an symbol link for BUILD.MD into doc/BUILDING.md (ln -s docs/BUILDING.md  BUILDING.md).  (so both live docs website and this readme in root works).

Another 2nd fix is:
1. modify the /root/README.MD to refers docs/BUILDING.MD
1. create an real docs/index.md(maybe different content) instead of symbol link to /root/README.MD, which refers BUILDING.MD using relative path(BUILDING.MD)

Another 3rd fix is:
1. update mkdocs.yaml to change docs_dir to root dir instead of default "docs".
1. symbol link root/index.md to root/README.MD
1. update all links in docs/ to use "/docs/xxx" instead of "xxx".

At this point, i'm not sure which one works best=.= I'd like the 2nd fix personally, what do you think

PR #816, opened 2019-01-17
Add kustomization.yaml

### LINKED ISSUE #815: Add kustomization.yaml
Would you be amenable to adding a `kustomization.yaml` file to enable installation via [kustomize](https://github.com/kubernetes-sigs/kustomize)? It could live alongside the existing manifests in `docs/examples`. The manifests themselves would remain (mostly) the same.

mgoodness:
Added a `kustomization.yaml` file to enable installation via [kustomize](https://github.com/kubernetes-sigs/kustomize). Also tweaked the existing manifests to make the container & service account names consistent with the other resources.

Finally, removed or commented fields in the Deployment that are defaulted by the cluster (e.g. replicas, dnsPolicy, restartPolicy) or deprecated (serviceAccount). While not strictly necessary, it does make it somewhat easier to create kustomize overlays.

Closes #815

### DIFF
--- a/docs/examples/alb-ingress-controller.yaml
+++ b/docs/examples/alb-ingress-controller.yaml
@@ -21,7 +21,8 @@ spec:
         app: alb-ingress-controller
     spec:
       containers:
-        - args:
+        - name: alb-ingress-controller
+          # args:
             # Limit the namespace where this ALB Ingress Controller deployment will
             # resolve ingress resources. If left commented, all namespaces are used.
             # - --watch-namespace=your-k8s-namespace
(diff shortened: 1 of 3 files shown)


docs/examples/alb-ingress-controller.yaml
```diff
@@ -12,42 +12,37 @@ metadata:
...
+          # args:
             # Limit the namespace where this ALB Ingress Controller deployment will
             # resolve ingress resources. If left commented, all namespaces are used.
             # - --watch-namespace=your-k8s-namespace
 
             # Setting the ingress-class flag below ensures that only ingress resources with the
             # annotation kubernetes.io/ingress.class: "alb" are respected by the controller. You may
             # choose any class you'd like for this controller to respect.
-            - --ingress-class=alb
+            # - --ingress-class=alb
```
M00nF1sh: i think this should remain uncommented.
The behavior of "ingress-class" parameter is slightly tweaked since v1.0.0. If it's left unspecified, the controller will target ingress `without ingress.class` or alb. It's better to remain restricted to `alb` for example setup.

M00nF1sh:
Thanks so much for doing this 👍 
overall lgtm with minimal comments.

docs/examples/alb-ingress-controller.yaml
mgoodness: Heh, that happens to be the one thing I was on the fence about. No problem!

PR #825, opened 2019-01-23
Fix error message for internet LB

gonfva:
When using an internet-facing, the error message shows the wrong tag. The correct tag is defined in line 383, variable key.


M00nF1sh:
Thanks for helping 👍 

/lgtm
/approve

PR #826, opened 2019-01-23
Small clarification

gonfva:
The template for ingress-controller has a name already defined. If you already has your subnets tagged with a different cluster-name, you will find the error "retrieval of subnets failed to resolve 2 qualified subnet" and it's not clear why. This clarification in the docs may help.


M00nF1sh:
Thanks for helping 😄 

I have always considered to make clusterName auto-discoverable, a lot ppl get trapped by it.

PR #827, opened 2019-01-24
Update kustomize

mgoodness:
* Replaced the `app` label on all resources with `app.kubernetes.io/name` as [recommended](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) by sig-apps
* Bumped version since I was in there anyway


M00nF1sh:
🤣 Thanks for helping 👍 

/lgtm
/approve

PR #830, opened 2019-01-25
Extra parameter to force usage of only one ALB per cluster

marcosdiez:
This PR makes the server support one extra command line parameter: `--force-alb-name=XXXXXX`

When this is used, the server will always use this ALB all the time, no matter what.
That means it will create it if necessary and it will use the existing one otherwise. If one combines with the annotation to keep the ALB, then it will just live forever.

Also, it only cares about the rules it needs to create and to modify in the ALB.
That means one can create other rules manually and as long as they don't conflict with your kubernetes annotations, they will be untouched.

Currently it has two known issues: 
- it will probably break if one adds more 50 rules in a ALB ( aws alb limitation )
- it will *NOT* delete any rule, even if it created in the past and the rule became unnecessary/obsolete.

I do plan to address the above issue in the future, but I am pretty sure even with them this is a step forward.

One can try it using the following docker image: `docker.io/marcosdiez/aws-alb-ingress-controller:v1.1.0-single-alb`

Fixes https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/724 and https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/688 and https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/298 and https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/228

Extra explanation on how to use it:

`kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/aws-alb-ingress-controller/v1.1.0/docs/examples/rbac-role.yaml`

Download `https://raw.githubusercontent.com/kubernetes-sigs/aws-alb-ingress-controller/v1.1.0/docs/examples/alb-ingress-controller.yaml`

- change the image to
`image: docker.io/marcosdiez/aws-alb-ingress-controller:v1.1.0-single-alb`

- on the `args` section, add a new line with the load balancer name. something like `- --force-alb-name=k8s-alb`

- follow the rest of the instructions here: https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/walkthrough/echoserver/

### DIFF
--- a/docs/examples/iam-policy.json
+++ b/docs/examples/iam-policy.json
@@ -36,6 +36,7 @@
     {
       "Effect": "Allow",
       "Action": [
+        "elasticloadbalancing:AddListenerCertificates",
         "elasticloadbalancing:AddTags",
         "elasticloadbalancing:CreateListener",
         "elasticloadbalancing:CreateLoadBalancer",
@@ -46,6 +47,7 @@
         "elasticloadbalancing:DeleteRule",
         "elasticloadbalancing:DeleteTargetGroup",
         "elasticloadbalancing:DeregisterTargets",
+        "elasticloadbalancing:DescribeListenerCertificates",
         "elasticloadbalancing:DescribeListeners",
         "elasticloadbalancing:DescribeLoadBalancers",
         "elasticloadbalancing:DescribeLoadBalancerAttributes",
@@ -61,6 +63,7 @@
         "elasticloadbalancing:ModifyTargetGroup",
         "elasticloadbalancing:ModifyTargetGroupAttributes",
         "elasticloadbalancing:RegisterTargets",
+        "elasticloadbalancing:RemoveListenerCertificates",
         "elasticloadbalancing:RemoveTags",
         "elasticloadbalancing:SetIpAddressType",
         "elasticloadbalancing:SetSecurityGroups",


marcosdiez:
Just for the record, I just signed the CLA

christopherhein:
@marcosdiez - this is very cool. 

> it will NOT delete any rule, even if it created in the past and the rule became unnecessary/obsolete.

Does this mean if there a conflict the new rule will take precedence or does the old one?

M00nF1sh:
@marcosdiez 
Thanks so much for contributing 👍 .

However, i'm a little opposed to this approach given it's limitations.
Actually, this will cause more troubles, like when you have listeners with different listen ports.

BTW, i'm currently working on a different solution called ingress group to solve this problem based a different approach.

The basic idea is add another annotation called ingressGroup, and ingress belongs to same group are merged and reconciled together(result in single alb)
This approach give you the flexibility on how to merge ingresses. e.g.
- you can set default ingressGroup at controller level, to use single ALB for all ingress cluster-wise.
- you can set default ingressGroup at namespace level, to use single ALB for all ingress within namespace.
- you can set same ingressGroup for  all ingresses of your micro-service application across namespaces, to use single ALB.
- you can old way one ALB ingress.

Also, this helps solve the pain-point of path / port based configurations. e.g.
- different ingress rule for different port on same alb.
- different auth setting for different path.(cannot achieved for now if your service have multiple ports)

marcosdiez:
@christopherheim my code adds new rules and modifies existing ones, no matter if it created it or not (it does not track which rules it created). So conflicting rules will be overwritten.

marcosdiez:
@M00nF1sh your proposal is clearly better then mine. It's much more flexible without any added downsides. It's properly designed and now that I know about it, I also believe merging this PR is not a good idea. 

The problem is that your ETA is middle march and I can't wait for it. 
So I suggest we leave this PR open until yours is ready, so people who can't wait can have an alternative.

Porting from one container to the other is trivial. Just adding the annotations you described and removing the `--force-alb-name` parameter.

The only issue with this approach is that it has the same disadvantages as mine:

- it's not clear what happens after the 50th ALB rule
- it's not clear in your description if it deletes rules it has created and are not needed anymore

Both are solvable. Do you have any plans on them ?

By the way, nice design. I couldn't do better.

M00nF1sh:
@marcosdiez 
Thanks a lot for the comments 🤣 .
I can speed up my implementation and deliver a draft version before that(without unit test cases, which is the most time consuming part =.=)
BTW,
1. the rules limit should be able to be increased via aws support ticket. Also, if users can manually split ingress into groups to sastify the default rule limit. (smarter variants exists like  auto-split into groups, but i don't think users will want them since how the split happens is un-predictable from user's pov).
2. Yes, the rules that are not needed will be automatically deleted.(the code base are been refactored to support it 😄 )

marcosdiez:
Small status update: I just found a bug that was causing issues with two ingress rules with the same name on different namespaces. It has  been fixed. The container image has been updated as well with the same name. The old broken on has been renamed to `broken`.

This is clearly not a best practice but assigning versions to a fork that has it's days counted does not worth the trouble.

marcosdiez:
Due to reasons beyond my control, I had to withdraw my pull request. If they get solved, I'll put it back when possible. Sorry.

oba11:
Was looking forward to this PR being merged 😢 . @marcosdiez please can I ask why the pr was withdrawn even after it has been approved?

tonidas:
What happened @marcosdiez ? It was already approved :(

Do you authorize anyone to use your PR to merge it?

sc-chad:
Do either of these solutions solve the problem of if you have target groups that have different health endpoints?

Right now the healthcheck path is in the annotation, what would take this a HUGE step forward is to move it to the backend parameter.  So we could do something like:
```
spec:
  rules:
  - http:
      paths:
      - backend:
          serviceName: redirect
          servicePort: use-annotation
        path: /*
  - host: myhost.example.com
    http:
      paths:
      - path: /*
        backend:
           serviceName: "some-web"
           servicePort: 80
           healthcheck: /health
```

With this I could specify multiple hosts using different target groups.  Is this something you are accounting for @M00nF1sh ?

M00nF1sh:
@sc-chad ,
Currently you can already add the healthcheck related annotations on your service objects(which can achieve your seetings), which only affects settings for corresponding backend(and override settings on ingress objects)

sc-chad:
@M00nF1sh do you have an example of this?  Would love to be able to do that.

sc-chad:
Nevermind, figured out what you were saying.  In case anyone else comes across this.  You need to add:

```
  annotations:
    alb.ingress.kubernetes.io/healthcheck-path: /
```
To your service that you want to attach to the ALB.  Thanks @M00nF1sh for the awesome module!

frederiksf:
When will this feature be documented and generally available?

yanivpaz:
@sc-chad  it will be  great if you can document it here :-)

sc-chad:
I'll see what I can do!

sc-chad:
https://github.com/kubernetes-sigs/aws-alb-ingress-controller/pull/861

benderillo:
@M00nF1sh Where can we track the work you mentioned for the IngressGroups? What is general status and ETA?

M00nF1sh:
@benderillo 
I'm going through a internal design review today, and will release an working alpha version soon.

rverma-nikiai:
@M00nF1sh this is great news. Awaiting eagerly.

tdmalone:
Just in case anyone didn't know - the Ingress Groups work is being discussed at https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/914

PR #851, opened 2019-02-14
Auto-add certs from ACM by hostname if none are specified via annotation

khyew:
(I'm picking up a PR from @cv that was lost when the master branch of this project was re-based: https://github.com/kubernetes-sigs/aws-alb-ingress-controller/pull/600)

This change adds a fallback behaviour to the controller's handling of TLS certificates. If the ingress spec has a HTTPS listener but no `alb.ingress.kubernetes.io/certificate-arn` annotation is found, then the controller will attempt to attach certificates from ACM that best match what is required by the listener. Matching is done via the listeners' `spec.rules[].host` entries, as well as any hostnames in `spec.tls.hosts[]`. Wildcard certificates are correctly matched.

e.g.
If I have successfully issued a cert for `*.cv.dev.example.com` and for `website.qifan.example.com`, then the following ingress spec will attach both certs to all listeners of the ALB:
```
kind: Ingress
metadata:
  annotations:
    alb.ingress.kubernetes.io/listen-ports: '[{"HTTPS": 443}]'
  name: echoserver
  namespace: echoserver
spec:
  tls:
  - hosts:
    - echoserver.cv.dev.example.com
  rules:
  - host: website.qifan.dev.example.com
    http:
      paths:
      - backend:
          serviceName: echoserver
          servicePort: 80
        path: /
  ...
```

Potential TODOs:
* When multiple certs are attached to a listener, the correct certificate will be used via SNI. Clients that do not support SNI will use the default cert, which is chosen arbitrarily. The controller should instead try to attach the "best" cert as the default for a particular listener.

* ALBs only support 25 certificate attachments per load balancer (https://aws.amazon.com/blogs/aws/new-application-load-balancer-sni/). There is no specific messaging for users whose ingress specs exceed this limitation.

### DIFF
--- a/internal/alb/ls/listener.go
+++ b/internal/alb/ls/listener.go
@@ -232,6 +232,8 @@ func (controller *defaultController) buildListenerConfig(ctx context.Context, op
 				return config, errors.Errorf("missing certificates annotation %v could not find any matching certificates from ACM to auto-load",
 					parser.GetAnnotationWithPrefix(AnnotationCertificateARN))
 			}
+
+			albctx.GetLogger(ctx).Infof("Auto-detected and added %d certificates to listener", len(certs))
 			certificateARNs = certs
 		}
 		config.DefaultCertificate = []*elbv2.Certificate{
@@ -277,7 +279,7 @@ func inferCertARNs(acmsvc aws.ACMAPI, ingress *extensions.Ingress, logger *log.L
 	for _, c := range certs {
 		for _, h := range ingressHosts {
 			if domainMatchesHost(aws.StringValue(c.DomainName), h) {
-				logger.Debugf("Domain name '%s', matches TLS host '%v', adding to Listener", aws.StringValue(c.DomainName), h)
+				logger.Infof("Domain name '%s', matches TLS host '%v', adding to Listener", aws.StringValue(c.DomainName), h)
 				if !seen[aws.StringValue(c.CertificateArn)] {
 					certArns = append(certArns, aws.StringValue(c.CertificateArn))
 					seen[aws.StringValue(c.CertificateArn)] = true
--- a/internal/alb/ls/listener.go
+++ b/internal/alb/ls/listener.go
@@ -228,6 +228,10 @@ func (controller *defaultController) buildListenerConfig(ctx context.Context, op
 				return config, errors.Errorf("missing certificates annotation %v and could not auto-load certificates from ACM: %v",
 					parser.GetAnnotationWithPrefix(AnnotationCertificateARN), err)
 			}
+			if len(certs) == 0 {
+				return config, errors.Errorf("missing certificates annotation %v could not find any matching certificates from ACM to auto-load",
+					parser.GetAnnotationWithPrefix(AnnotationCertificateARN))
+			}
 			certificateARNs = certs
 		}
 		config.DefaultCertificate = []*elbv2.Certificate{
--- a/internal/alb/ls/listener.go
+++ b/internal/alb/ls/listener.go
@@ -3,6 +3,7 @@ package ls
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 
@@ -12,6 +13,7 @@ import (
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/auth"
 
 	"github.com/aws/aws-sdk-go/aws/awsutil"
+	"github.com/aws/aws-sdk-go/service/acm"
 	"github.com/aws/aws-sdk-go/service/elbv2"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/alb/tg"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/albctx"
@@ -19,6 +21,7 @@ import (
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/annotations"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/annotations/action"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/annotations/loadbalancer"
+	"github.com/kubernetes-sigs/aws-alb-ingress-controller/pkg/util/log"
 	util "github.com/kubernetes-sigs/aws-alb-ingress-controller/pkg/util/types"
 	extensions "k8s.io/api/extensions/v1beta1"
 )
@@ -220,7 +223,12 @@ func (controller *defaultController) buildListenerConfig(ctx context.Context, op
 		var certificateARNs []string
 		_ = annotations.LoadStringSliceAnnotation(AnnotationCertificateARN, &certificateARNs, options.Ingress.Annotations)
 		if len(certificateARNs) == 0 {
-			return config, errors.Errorf("annotation %v must be specified for https listener", parser.GetAnnotationWithPrefix(AnnotationCertificateARN))
+			certs, err := inferCertARNs(controller.cloud, options.Ingress, albctx.GetLogger(ctx))
+			if err != nil {
+				return config, errors.Errorf("missing certificates annotation %v and could not auto-load certificates from ACM: %v",
+					parser.GetAnnotationWithPrefix(AnnotationCertificateARN), err)
+			}
+			certificateARNs = certs
 		}
 		config.DefaultCertificate = []*elbv2.Certificate{
 			{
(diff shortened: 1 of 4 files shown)


khyew:
I have signed the CNCF CLA. Mister k8s-ci-robot, please re-verify me.

internal/alb/ls/listener.go
```diff
@@ -250,3 +264,76 @@ func (controller *defaultController) buildDefaultActions(ctx context.Context, op
...
+func domainMatchesHost(domainName string, tlsHost string) bool {
+	if strings.HasPrefix(domainName, "*.") {
+		ds := strings.Split(domainName, ".")
+		hs := strings.Split(tlsHost, ".")
+
+		if len(ds) != len(hs) {
+			return false
+		}
+
+		for i, dp := range ds {
```
M00nF1sh: i think it for-loop can be simplied to  ```return reflect.DeepEqual(ds[1:],hs[1:])```

internal/alb/ls/listener.go
```diff
@@ -250,3 +264,76 @@ func (controller *defaultController) buildDefaultActions(ctx context.Context, op
...
+			}
+		}
+		return true
+	}
+
+	return domainName == tlsHost
+}
+
+func uniqueHosts(ingress *extensions.Ingress) []string {
+	var result []string
```
M00nF1sh: I think all these `set` operations via `seen` trick can be simplified via sets.NewString()

internal/alb/ls/listener.go
```diff
@@ -250,3 +264,76 @@ func (controller *defaultController) buildDefaultActions(ctx context.Context, op
 	}
 	return buildActions(ctx, authCfg, options.IngressAnnos, backend, options.TGGroup)
 }
+
+// inferCertARNs retrieves a set of certificates from ACM that matches the ingress' hosts list
```
M00nF1sh: it will be better to make inferCertArns an member function of controller, so it don't need to pass singtons like (acmsvc aws.ACMAPI) around.
Alternatively, i sugges make the signature below:
```func (controller *defaultController) inferCertARNs(ctx context.Context, ingress *extension.Ingress)```
So 
1.  the log object can be extracted within inferCertARNs.
1. We can pass context to ListCertificates to carry timeout info in the future.

**Comment by M00nF1sh** [comment]
Thanks so much for make this change 😄 
Overall lgtm with minimal comments.

There also should be works to be do in future PRs:
1. documentation for this feature
1. resolve conflicts when there are multiple certificates for same domain.

File: internal/alb/ls/listener.go (hunk 1 above)
Comment (khyew) [reply internal/alb/ls/listener.go]: That would cause it to erroneously return true for the following trivial case:
```
ds := []string{"google", "com"}
hs := []string{"facebook", "com"}
```

File: internal/alb/ls/listener.go (hunk 2 above)
Comment (khyew) [reply internal/alb/ls/listener.go]: Gotcha, fixed.

**Comment by khyew** [comment]
@M00nF1sh I was in the middle of addressing the PR fixes that you proposed. Do you still want me to submit them (by cutting a branch and opening a new PR)?

**Comment by M00nF1sh** [comment]
> @M00nF1sh I was in the middle of addressing the PR fixes that you proposed. Do you still want me to submit them (by cutting a branch and opening a new PR)?

That would be awesome 😄 . I merged this since it's just some nitpicks.

**Comment by khyew** [comment]
Unfortunately I'm having difficulty building the project now. `make clean && make` errors out with module/dependency related issues. Any tips? Building on OS X.

```
CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -installsuffix cgo -ldflags '-s -w -X github.com/kubernetes-sigs/aws-alb-ingress-controller/version.COMMIT=git-f3ae919b -X github.com/kubernetes-sigs/aws-alb-ingress-controller/version.RELEASE=v1.1.0 -X github.com/kubernetes-sigs/aws-alb-ingress-controller/version.REPO=git@github.com:kubernetes-sigs/aws-alb-ingress-controller.git' -o server ./cmd
go: finding github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi latest
go: finding github.com/aws/aws-sdk-go/service/iam/iamiface latest
go: finding github.com/aws/aws-sdk-go/service/ec2/ec2iface latest
go: finding github.com/aws/aws-sdk-go/service/wafregional/wafregionaliface latest
go: finding github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi/resourcegroupstaggingapiiface latest
go: finding github.com/aws/aws-sdk-go/service/elbv2/elbv2iface latest
go: finding github.com/aws/aws-sdk-go/service/iam latest
go: finding github.com/aws/aws-sdk-go/service/acm latest
go: finding github.com/aws/aws-sdk-go/internal/shareddefaults latest
go: finding github.com/aws/aws-sdk-go/service latest
go: finding github.com/aws/aws-sdk-go/service/ec2 latest
go: finding github.com/aws/aws-sdk-go/internal/sdkuri latest
go: finding github.com/aws/aws-sdk-go/service/wafregional latest
go: finding github.com/aws/aws-sdk-go/service/elbv2 latest
go: finding github.com/aws/aws-sdk-go/internal latest
go: finding github.com/aws/aws-sdk-go/service/sts latest
go: finding github.com/aws/aws-sdk-go/internal/ini latest
go: finding github.com/aws/aws-sdk-go/service/waf latest
go: finding github.com/aws/aws-sdk-go/internal/sdkio latest
go: finding github.com/aws/aws-sdk-go/service/acm/acmiface latest
go: finding github.com/aws/aws-sdk-go/internal/sdkrand latest
internal/aws/acm.go:10:2: unknown import path "github.com/aws/aws-sdk-go/service/acm": cannot find module providing package github.com/aws/aws-sdk-go/service/acm
internal/aws/cloud.go:9:2: unknown import path "github.com/aws/aws-sdk-go/service/acm/acmiface": cannot find module providing package github.com/aws/aws-sdk-go/service/acm/acmiface
cmd/main.go:29:2: unknown import path "github.com/aws/aws-sdk-go/service/ec2": cannot find module providing package github.com/aws/aws-sdk-go/service/ec2
internal/aws/cloud.go:11:2: unknown import path "github.com/aws/aws-sdk-go/service/ec2/ec2iface": cannot find module providing package github.com/aws/aws-sdk-go/service/ec2/ec2iface
internal/aws/cloud.go:12:2: unknown import path "github.com/aws/aws-sdk-go/service/elbv2": cannot find module providing package github.com/aws/aws-sdk-go/service/elbv2
internal/aws/cloud.go:13:2: unknown import path "github.com/aws/aws-sdk-go/service/elbv2/elbv2iface": cannot find module providing package github.com/aws/aws-sdk-go/service/elbv2/elbv2iface
internal/aws/cloud.go:14:2: unknown import path "github.com/aws/aws-sdk-go/service/iam": cannot find module providing package github.com/aws/aws-sdk-go/service/iam
internal/aws/cloud.go:15:2: unknown import path "github.com/aws/aws-sdk-go/service/iam/iamiface": cannot find module providing package github.com/aws/aws-sdk-go/service/iam/iamiface
cmd/main.go:30:2: unknown import path "github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi": cannot find module providing package github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi
internal/aws/cloud.go:17:2: unknown import path "github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi/resourcegroupstaggingapiiface": cannot find module providing package github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi/resourcegroupstaggingapiiface
../../../../pkg/mod/github.com/aws/aws-sdk-go@v1.17.1/aws/credentials/stscreds/assume_role_provider.go:89:2: unknown import path "github.com/aws/aws-sdk-go/service/sts": cannot find module providing package github.com/aws/aws-sdk-go/service/sts
internal/aws/wafregional.go:8:2: unknown import path "github.com/aws/aws-sdk-go/service/waf": cannot find module providing package github.com/aws/aws-sdk-go/service/waf
internal/aws/cloud.go:18:2: unknown import path "github.com/aws/aws-sdk-go/service/wafregional": cannot find module providing package github.com/aws/aws-sdk-go/service/wafregional
internal/aws/cloud.go:19:2: unknown import path "github.com/aws/aws-sdk-go/service/wafregional/wafregionaliface": cannot find module providing package github.com/aws/aws-sdk-go/service/wafregional/wafregionaliface
make: *** [server] Error 1
```

internal/alb/ls/listener.go
M00nF1sh: i don't think so, since you already have two precondition before:

1. strings.HasPrefix(domainName, "*."), ensures ds begins with "*",
1. if len(ds) != len(hs), ensures hs is at least 1. (so hs[1:] is safe).

use reflect.DeepEqual(ds[1:], hs[1:]) after the two condition seems easier to read 🤣

M00nF1sh:
umm, are you based on the new master branch(i bump dependencies two commits ago), and using a go version that supports modules?

khyew:
`go version go1.11.3 darwin/amd64`

Module support shouldn't be a problem. Worked fine before I pulled master to get up-to-date today.

khyew:
Hold on, I'm going to `git bisect` to narrow down the offending commit

M00nF1sh:
i think you can try do the following:
1. delete the vendor directory in repo(if any)
2. delete everything from  ~/go/pkg/  (module caches)

khyew:
There is no vendor directory in the repo. I deleted the package cache but that didn't resolve the problem.

According to the results of the bisect, commit a28d0d01400a557eaa1c1d78f18abb022c44cb7f is the first to introduce the issues so I assume it's related to bumping dependency versions. When I try to build from that commit, I get the same error that I posted. When I build from the previous commit, it succeeds without issues.

khyew:
Update: deleting the _actual_ module cache directory worked. My $GOPATH is set to `~` so for me it was `~/pkg` and not `~/go/pkg`. I did a `rm -rf ~/go/pkg` and it didn't report any errors (deleting a non-existing file with `-f` is considered a success). Instead I should have either checked manually or done a `rm -rf $GOPATH/pkg`.

M00nF1sh:
There is some bugs in go's module support. Seems multiple cocurrent gets via go module migth corrupt the module directory. I guess that's what you encountered 😄

internal/alb/ls/listener.go
khyew: oh whoops, I missed the pre-condition. Good catch.

khyew: Oh I see, `strings.Split("*.", ".")` actually evaluates to `["*", ""]` which is size 2, so if `ds` and `hs` are equal then they are both at least size 2 and `ds[1:]` and `hs[1:]` will be safe operations.

PR #861, opened 2019-02-20
adding additional examples on how to use annotations and healthchecks

sc-chad:
I added two real examples on how we use the ALB Ingress controller both on internal application load balancers and external application load balancers.

I was unsure on where you would like to link to these examples, but please link to or use however you would like.


sc-chad:
Signed teh CLA.

M00nF1sh:
This is awesome! I'll take a deeper look later today 👍

docs/examples/rails-api/configmap.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-map
+  labels:
+    application: web
```
M00nF1sh: better to be `app.kubernetes.io/name: rails-api`, list of recommended labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

docs/examples/rails-api/configmap.yaml
```diff
@@ -0,0 +1,19 @@
...
+    namespace: some-name
```
M00nF1sh: better to be an explicit name like `rails-api-example`, wonder whether we should use kustomize to layout these examples

docs/examples/rails-api/ingress-external.yaml
```diff
@@ -0,0 +1,37 @@
...
+  name: ingress-external
+  namespace: some-namespace
+  annotations:
+    kubernetes.io/ingress.class: alb
+    alb.ingress.kubernetes.io/actions.redirect: '{"Type": "redirect", "RedirectConfig": { "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}'
+    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP":80,"HTTPS": 443}]'
+    alb.ingress.kubernetes.io/certificate-arn: acm:certificate:arn
+    alb.ingress.kubernetes.io/scheme: internet-facing
+    alb.ingress.kubernetes.io/subnets: 123456789,95481321,321457987
+    alb.ingress.kubernetes.io/security-groups: sg-23157498
```
M00nF1sh: `alb.ingress.kubernetes.io/security-groups`, `alb.ingress.kubernetes.io/security-subnets ` `alb.ingress.kubernetes.io/ssl-policy` should be removed. These three all have sensible default&auto-discover functionality.
Our goal is to make number annotations be as less as possible.

M00nF1sh:
@sc-chad 
Overall looks good to me with nit picks. 
But i'm kind of resist to merge it as examples, since it's not a working examples(e.g. stuffs like external DB dependencies), and lacks a workthrough accompany with it.
But the knowledges are valuable, like how to do external-DNS, HPA, etc. I think it's worth to break this example into smaller pieces, with each one focus on specific tasks.

docs/examples/rails-api/ingress-external.yaml
sc-chad: I added them because it wasn't intuitive that they were auto discoverable when we first walking through the examples.  I can add some commenting around these to show they are optional and should be auto discovered if you've taken the appropriate steps.

docs/examples/rails-api/configmap.yaml
sc-chad: I've had issues with lookups on name versus app.kubernetes.io/name.  It's probably my own understanding and how to use them.  I do need to fix the labels on this config.  Will add a PR.

sc-chad:
Added in some commenting and removed dependencies.

sc-chad:
@M00nF1sh Anything else you want me to cover here?

docs/examples/rails-api/ingress-external.yaml
bigkraig: What if they were included but commented out?

dennisotugo:
@M00nF1sh 
 @bigkraig

docs/examples/rails-api/ingress-external.yaml
sc-chad: Done.

docs/examples/rails-api/web-external.yaml
```diff
@@ -0,0 +1,124 @@
...
+  labels:
+    application: web-external
+    namespace: some-namespace
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1beta2
+    kind: Deployment
+    name: web-external
+  minReplicas: 1
+  maxReplicas: 1
```
zakkg3: Why HPA with min 1 max 1? it will recreate the pod if it reaches 50% cpu?

sc-chad: To reduce the resources required to demo, absolutely could be 1min 2 max

PR #864, opened 2019-02-20
Refactor/PR comments for auto-cert matching + documentation

khyew:
* Addressed PR comments from previously-approved (and merged) PR https://github.com/kubernetes-sigs/aws-alb-ingress-controller/pull/851

* Use DeepEquals instead of a loop to match wildcard hostnames

* Use k8s `sets` package instead of string-boolean maps as sets for removing duplicate hosts

* Made `inferCertArns` a member function of the `defaultController` struct to avoid having to pass singletons around

* Added documentation and examples for auto-cert matching

### DIFF
--- a/docs/guide/ingress/annotation.md
+++ b/docs/guide/ingress/annotation.md
@@ -340,6 +340,58 @@ SSL support can be controlled with following annotations:
             alb.ingress.kubernetes.io/certificate-arn: arn:aws:acm:us-west-2:xxxxx:certificate/cert1,arn:aws:acm:us-west-2:xxxxx:certificate/cert2,arn:aws:acm:us-west-2:xxxxx:certificate/cert3
             ```
 
+    !!!tip
+        If the `alb.ingress.kubernetes.io/certificate-arn` annotation is not specified, the controller will attempt to add certificates to listeners that require it by matching available certs from ACM with the `host` field in each listener's ingress rule.
+
+    !!!example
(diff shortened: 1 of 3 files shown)


M00nF1sh:
awesome!
/ok-to-test

khyew:
I... don't understand. The unit test passes locally.

khyew:
Force-pushing after a rebase onto `master` to see if things are out of sync from my local repo...

khyew:
@M00nF1sh looks like the unit tests pass now (with no changes other than a rebase and forced re-push). Can you approve again when you get a chance?

M00nF1sh:
umm..that should indicate some flakes happend..wait me double check your test cases

M00nF1sh:
Hi,
It's because you are returning sets.String.unsortedList() from inferCertCRN & uniqueHosts.(the order of these returned [] is pure random), and in the test comparsion, order matters when compare [].

Fix:  
1. return sets.String instead of []string, and change the expectations in test cases to sets.String{} too.
1. return sets.String.List() (i'm ok with this since this have little performance drawbacks)

khyew:
oh wow yeah, that totally slipped my mind

khyew:
I actually would have been in favour of changing the _tests_ to not expect the results to come in a specific order, because the function itself makes no guarantee of specific ordering and this couples the tests a bit too tightly with the current implementation. But in this case since the fix takes literally 5 seconds, I'm just going to convert the calls from UnsortedList() to List().

Issue #865, opened 2019-02-21
Working with shared subnet aws-alb-ingress-controller can't create ingress


rverma-nikiai:
Steps to recreate
1. Share public and private subnet using a different account.

2. Create eks cluster using terraform-aws-eks (https://github.com/terraform-aws-modules/terraform-aws-eks/). Select all 4 nodes for control pane.

3. Create a worker asg in private subnet.

4. Deploy aws-alb-ingress-controller (1.1.0) using the helm

5. create ingress with annotations 
```
 alb.ingress.kubernetes.io/scheme: internet-facing
 alb.ingress.kubernetes.io/security-groups: 'cluster_sg,worker_sg,https_sg'
 kubernetes.io/ingress.class: alb
 alb.ingress.kubernetes.io/listen-ports: '[{"HTTP":80,"HTTPS": 443}]'
 alb.ingress.kubernetes.io/subnets: 'public_subnet1,public_subnet2'
```
Observe exception as

```
E0221 10:28:14.476654       1 loadbalancer.go:195] echoserver/echoserver: 	status code: 400, request id: 64d66e20-35c3-11e9-93fe-d10ee5c193ac
E0221 10:28:14.477021       1 :0] kubebuilder/controller "msg"="Reconciler error" "error"="failed to create LoadBalancer due to ValidationError: A security group must be specified for creating a loadbalancer in a VPC not owned by the AWS account\n\tstatus code: 400, request id: 64d66e20-35c3-11e9-93fe-d10ee5c193ac"  "controller"="alb-ingress-controller" "request"={"Namespace":"echoserver","Name":"echoserver"}

```

rverma-nikiai:
@M00nF1sh were you able to address this in your last release? As mentioned I was able to create the load balancer by adding the unofficial variable but there is no target group launched.

PR #867, opened 2019-02-22
Create loadbalancer.newLBinstance doesn't honour security group annotations

rverma-nikiai:
Create loadbalancer.newLBinstance doesn't honour security group annotations, leading to #865

### DIFF
--- a/internal/alb/lb/loadbalancer.go
+++ b/internal/alb/lb/loadbalancer.go
@@ -62,10 +62,11 @@ type loadBalancerConfig struct {
 	Name string
 	Tags map[string]string
 
-	Type          *string
-	Scheme        *string
-	IpAddressType *string
-	Subnets       []string
+	Type           *string
+	Scheme         *string
+	IpAddressType  *string
+	Subnets        []string
+	SecurityGroups []string
 }
 
 type defaultController struct {
@@ -184,12 +185,13 @@ func (controller *defaultController) ensureLBInstance(ctx context.Context, lbCon
 func (controller *defaultController) newLBInstance(ctx context.Context, lbConfig *loadBalancerConfig) (*elbv2.LoadBalancer, error) {
 	albctx.GetLogger(ctx).Infof("creating LoadBalancer %v", lbConfig.Name)
 	resp, err := controller.cloud.CreateLoadBalancerWithContext(ctx, &elbv2.CreateLoadBalancerInput{
-		Name:          aws.String(lbConfig.Name),
-		Type:          lbConfig.Type,
-		Scheme:        lbConfig.Scheme,
-		IpAddressType: lbConfig.IpAddressType,
-		Subnets:       aws.StringSlice(lbConfig.Subnets),
-		Tags:          tags.ConvertToELBV2(lbConfig.Tags),
+		Name:           aws.String(lbConfig.Name),
+		Type:           lbConfig.Type,
+		Scheme:         lbConfig.Scheme,
+		IpAddressType:  lbConfig.IpAddressType,
+		Subnets:        aws.StringSlice(lbConfig.Subnets),
+		Tags:           tags.ConvertToELBV2(lbConfig.Tags),
+		SecurityGroups: aws.StringSlice(lbConfig.SecurityGroups),
 	})
 	if err != nil {
 		albctx.GetLogger(ctx).Errorf("failed to create LoadBalancer %v due to %v", lbConfig.Name, err)
--- a/internal/alb/lb/loadbalancer.go
+++ b/internal/alb/lb/loadbalancer.go
@@ -3,6 +3,7 @@ package lb
 import (
 	"context"
 	"fmt"
+	"os"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


rverma-nikiai:
I signed it

rverma-nikiai:
This alone is not sufficient but I can implement rest of logic after a discussion.

M00nF1sh:
Hi @rverma-nikiai 
Thanks for contributing this.
I personally haven't tried vpc sharing and the mechanism behind that.
In my understanding, vpc(is owned) by account1, where (subnet1) is owned by account2, and our ALB ingress controller is runnings on worker nodes in subnet1, am I correct?

Have you tested whether you will running into problem with this patch when you delete the LB? [This part of code](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/internal/alb/sg/lb_attachment.go#L42) will try to set the securityGroup on ALB to be the default SG of VPC(in another account), i'm not sure whether there will be error(I suppose there will be, since normally when create ALB without sg specified, it will use VPC's default SG)

M00nF1sh:
Update:
I didn't noticed you don't have securityGroup available in lb package. You will need to move the logic of  resolveSecurityGroupIDs out of sg/association.go to lb/loadbalancer.go, and pass externalSG as parameter to sg.AssoicationController

rverma-nikiai:
@M00nF1sh I unintentionally confused you. It's actually not a shared vpc but a shared subnet. AWS doesn't allow you to share a vpc. If you share a subnet than its vpc get automatically shared(can't edit apart from adding tags).
I haven't debugged much but found that the createLoadbalancer request is having below body
```
{
  IpAddressType: "ipv4",
  Name: "ff987766-echoserver-echose-2ad7",
  Scheme: "internet-facing",
  Subnets: [
    "subnet-03f458f395230b30d",
    "subnet-0ca65fb76018ab187"
  ],
  Tags: [
    {
      Key: "kubernetes.io/cluster/audit_niki_ai",
      Value: "owned"
    },
    {
      Key: "kubernetes.io/namespace",
      Value: "echoserver"
    },
    {
      Key: "kubernetes.io/ingress-name",
      Value: "echoserver"
    }
  ],
  Type: "application"
}```
The security group is missing from post request. Since the default value can't be used (we can't access the default security group from another account), this failed with corresponding reason as you assumed also.
What I was wondering if rather than resolving the default sg at each ingress level, I believe it will make more sense to allow the use of another security group as default. Something like a parameter as `--default-sg=sg-xxx` makes more sense to me. If you think its fine then what's left is adding logic to controller.getDefaultSecurityGroupID() to use env variable first, if absent try for default.

**Comment by M00nF1sh** [comment]
@rverma-nikiai 
From this blog, "https://aws.amazon.com/blogs/networking-and-content-delivery/vpc-sharing-a-new-approach-to-multiple-accounts-and-vpc-management/", it said "VPC sharing participants can reference security group IDs of each other.", would you help test whether explicitly set the sgID to the defaultSG of VPC when creating LoadBalancer works? I'll test it as well once i got my second account.

BTW,
The default SG is actually unnecessarily needed, it used it as an trick to group the logic of create/attach securityGroup into a single method(association.Reconcile). so it works like
* createLB with `defaultSG` ---> create `LBSG` -> attach `LBSG` to LB --> detach `LBSG` from LB(reset to defaultSG) ---> delete `LBSG` ---> delete LB.

However, ideally we don't need the defaultSG If we reorder to process to be below(though the sg related logic is more coupled around LB)
* create `LBSG` ---> createLB with `LBSG` --> attach `LBSG`/externalLBSG to LB --> delete LB --> delete `LBSG`.
I'm working on a CRD based new version, if you can wait a bit, i'll include this logic. otherwise, i think an undocumented environment variable like VPC_DEFAULT_SG_ID can surve as an temp fix for your case 😄

**Comment by rverma-nikiai** [comment]
This is good, I will wait for your fix. When you release an alpha version
let me know the branch I will deploy and test this :).

**Comment by rverma-nikiai** [comment]
@M00nF1sh found that waf support having some issues while running in a region where waf regional services are not available. E.g. ap-south-1. Getting exception as 
```
E0223 13:02:59.540879       1 :0] kubebuilder/controller "msg"="Reconciler error" "error"="error getting web acl for load balancer arn:aws:elasticloadbalancing:ap-south-1:614927918057:loadbalancer/app/ff987766-analytics-superse-acf8/b39236ab7a2497eb: RequestError: send request failed\ncaused by: Post https://waf-regional.ap-south-1.amazonaws.com/: dial tcp: lookup waf-regional.ap-south-1.amazonaws.com on 172.20.0.10:53: no such host"  "controller"="alb-ingress-controller" "request"={"Namespace":"analytics","Name":"superset"}
```
Waf-regional is not available in ap-south-1 and we are using waf global.

rverma-nikiai:
@M00nF1sh with the above fix the load balancer get created but still there is no target group.

rverma-nikiai:
@M00nF1sh were you able to address this in your last release? As mentioned I was able to create the load balancer by adding the unofficial variable but there is no target group launched. Is this something expected?

JasonCarter80:
Is there an ETA on this?

PR #875, opened 2019-02-26
Add template for Cognito integrated ALB ingress

procyclinsur:
### Requirement:
Cognito integration is not well documented in official capacity, and a template would at very least help users configure aws-alb-ingress-controller for a Cognito ALB ingress with considerably less effort.

**This PR:**
- [x] Cognito integrated ALB ingress template

### DIFF
--- a/docs/examples/cognito-ingress-template.yaml
+++ b/docs/examples/cognito-ingress-template.yaml
@@ -1,0 +1,68 @@
+# Requirements:
+#
+#    - For this template, Cognito should have the following basic settings:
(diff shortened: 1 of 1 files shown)


procyclinsur:
Currently this has the template in `docs/examples/` should we create a `templates` folder somewhere for this purpose?

M00nF1sh:
This looks good to me. 
Not sure where to put this, but i think this worth as an workthrough or task section. But i think it's good to merge into examples for now.

/ok-to-test
/lgtm
/approve

PR #877, opened 2019-02-26
Add support for pod IP targets outside of the VPC

### LINKED ISSUE #783: Kops / kubenet pod ips are not registered with AZ "all" to targetGroup anymore
Hi, 

I'm running a Kops-based cluster with Kubenet configuration where pods get IPs from `nonMasqueradeCIDR`, which is not listed in the VPC config. In this configuration Kops adds direct routes to the VPC routing table so that the k8s nodes can access pods and services on other nodes.  

With `alb-ingress-controller:1.0-beta.6` this worked well; with the ALB being deployed to the same subnet and `target-type: ip`, it could register pod endpoint to the ALB. 

With the upgrade to v1.0.1 (same with v.1.0.0) this raises an error:

```
kubebuilder/controller "msg"="Reconciler error" "error"="failed to reconcile targetGroups due to failed to reconcile targetGroup targets due to ValidationError: The Availability Zone for IP address '100.96.7.56' must be 'all' for Application Load Balancer target groups, when not within the VPC\n\tstatus code: 400, request id: 83067527-fec9-11e8-88c7-15463d1956f9"  "Controller"="alb-ingress-controller" "Request"={"Namespace":"<snipped>","Name":"<snipped>"} 
```

Ingress config: 

```
  annotations:
    alb.ingress.kubernetes.io/certificate-arn: arn:aws:acm:<arn snipped>
    alb.ingress.kubernetes.io/healthcheck-path: /healthz
    alb.ingress.kubernetes.io/subnets: subnet-XXXX,subnet-YYYYY
    alb.ingress.kubernetes.io/success-codes: "200"
```

relevant  KOPS settings: 

```
  kubernetesVersion: 1.10.11
  networkCIDR: 10.10.0.0/16
  networkID: vpc-xxxxx
  networking:
    kubenet: {}
  nonMasqueradeCIDR: 100.64.0.0/10
```

### LINKED ISSUE #858: Support explicit "all" availability zone for IP targets
For IP targets in a TargetGroup, each target can specify whether it should receive traffic from the ALB nodes in its own AZ or from ALB nodes in all AZs. See the [AvailabilityZone field here](https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_TargetDescription.html).

I would like to specify a value of `all` for one of my ingress' services. The specific struct that needs modifying is located here:

https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/63e5be04f16a562cd839cf2d4520486de99ffe9b/internal/ingress/backend/endpoint.go#L104-L107

I'm just not sure how you would recommend exposing this to the user. A new `alb.ingress.kubernetes.io/availability-zone` annotation on the ingress and service? I could see how it could be easily confused with the `alb.ingress.kubernetes.io/subnets` annotation, so we should document it carefully.

I can open a PR to support this if you can provide guidance on the best way to make this setting configurable.

rifelpet:
As discussed in #858 this adds support for pod IP targets outside of the ALB's VPC by specifying the target's `AvailabilityZone` of `all`. Originally implemented in #468

closes #783
closes #858

I've tested this on my cluster using `--aws-vpc-id` and it appears to work correctly:
```
I0226 23:05:46.256829       1 api.go:2231] Request: elasticloadbalancing/DescribeTargetHealth, Payload: {  TargetGroupArn: "arn:aws:elasticloadbalancing:us-east-1:000000000000:targetgroup/7d116f13-62358f9ba102d9c729f/158773697cb693ed"}
I0226 23:05:46.280342       1 api.go:2231] Response: elasticloadbalancing/DescribeTargetHealth, Body: {}
I0226 23:05:46.280386       1 targets.go:79] test/my-ingress: Adding targets to arn:aws:elasticloadbalancing:us-east-1:000000000000:targetgroup/7d116f13-62358f9ba102d9c729f/158773697cb693ed: 1.2.3.4:80, 1.2.3.5:80
I0226 23:05:46.280719       1 api.go:2830] Request: elasticloadbalancing/RegisterTargets, Payload: {  TargetGroupArn: "arn:aws:elasticloadbalancing:us-east-1:000000000000:targetgroup/7d116f13-62358f9ba102d9c729f/158773697cb693ed",  Targets: [{      AvailabilityZone: "all",      Id: "1.2.3.4",      Port: 80    },{      AvailabilityZone: "all",      Id: "1.2.3.5",      Port: 80    }]}
I0226 23:05:46.632281       1 api.go:2830] Response: elasticloadbalancing/RegisterTargets, Body: {}
```

### DIFF
--- a/internal/alb/tg/targets.go
+++ b/internal/alb/tg/targets.go
@@ -64,6 +64,12 @@ func (c *targetsController) Reconcile(ctx context.Context, t *Targets) error {
 	if err != nil {
 		return err
 	}
+	if t.TargetType != elbv2.TargetTypeEnumInstance {
+		err = aws.PopulateTargetAZ(ctx, c.cloud, desired)
+		if err != nil {
+			return err
+		}
+	}
 	current, err := c.getCurrentTargets(ctx, t.TgArn)
 	if err != nil {
 		return err
--- a/internal/alb/tg/targets.go
+++ b/internal/alb/tg/targets.go
@@ -66,7 +66,7 @@ func (c *targetsController) Reconcile(ctx context.Context, t *Targets) error {
 		return err
 	}
 	if t.TargetType == elbv2.TargetTypeEnumIp {
-		err = populateTargetAZ(ctx, c.cloud, desired)
+		err = c.populateTargetAZ(ctx, desired)
 		if err != nil {
 			return err
 		}
@@ -126,6 +126,34 @@ func (c *targetsController) getCurrentTargets(ctx context.Context, TgArn string)
 	return current, nil
 }
 
+func (c *targetsController) populateTargetAZ(ctx context.Context, a []*elbv2.TargetDescription) error {
+	vpc, err := c.cloud.GetVpcWithContext(ctx)
+	if err != nil {
+		return err
+	}
+	cidrBlocks := make([]*net.IPNet, 0)
+	for _, cidrBlockAssociation := range vpc.CidrBlockAssociationSet {
+		_, ipv4Net, err := net.ParseCIDR(*cidrBlockAssociation.CidrBlock)
+		if err != nil {
+			return err
+		}
+		cidrBlocks = append(cidrBlocks, ipv4Net)
+	}
+	for i := range a {
+		inVPC := false
+		for _, cidrBlock := range cidrBlocks {
+			if cidrBlock.Contains(net.ParseIP(*a[i].Id)) {
+				inVPC = true
+				break
+			}
+		}
+		if !inVPC {
+			a[i].AvailabilityZone = aws.String("all")
+		}
+	}
+	return nil
+}
+
 // targetChangeSets compares b to a, returning a list of targets to add and remove from a to match b
 func targetChangeSets(current, desired []*elbv2.TargetDescription) (add []*elbv2.TargetDescription, remove []*elbv2.TargetDescription) {
 	currentMap := map[string]bool{}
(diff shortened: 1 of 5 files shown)


internal/alb/tg/targets.go
```diff
@@ -64,6 +64,12 @@ func (c *targetsController) Reconcile(ctx context.Context, t *Targets) error {
 	if err != nil {
 		return err
 	}
+	if t.TargetType != elbv2.TargetTypeEnumInstance {
```
M00nF1sh: I prefer to use t.TargetType == elbv2.TargetTypeEnumIP to be explicit 😄

internal/alb/tg/targets.go
```diff
@@ -64,6 +64,12 @@ func (c *targetsController) Reconcile(ctx context.Context, t *Targets) error {
...
+		err = aws.PopulateTargetAZ(ctx, c.cloud, desired)
```
M00nF1sh: can we move the PopulateTargetAZ to be member func of this TargetsController?
It doesn't fits into elbv2 since it involves multiple aws services.

internal/aws/elbv2.go
```diff
@@ -322,3 +323,23 @@ func (c *Cloud) describeTargetGroupsHelper(input *elbv2.DescribeTargetGroupsInpu
...
 }
+
+// PopulateTargetAZ populates the AvailabilityZone field of the provided TargetDescriptions
+func PopulateTargetAZ(ctx context.Context, c CloudAPI, a []*elbv2.TargetDescription) error {
+	vpc, err := c.GetVpcWithContext(ctx)
+	if err != nil {
+		return err
+	}
+
+	_, ipv4Net, err := net.ParseCIDR(*vpc.CidrBlock)
```
M00nF1sh: we should use vpc.CidrBlockAssociationSet instead of vpc.CidrBlock. (some vpc have multiple CIDR blocks)

rifelpet:
I addressed the feedback. I can squash my commits if desired.

internal/alb/tg/targets.go
```diff
@@ -157,3 +164,32 @@ func tdsString(tds []*elbv2.TargetDescription) string {
...
+
+	for i := range a {
+		inVPC := false
+		for _, cidrBlock := range cidrBlocks {
+			if cidrBlock.Contains(net.ParseIP(*a[i].Id)) {
+				inVPC = true
+				break
+			}
+		}
+		if inVPC {
```
M00nF1sh: umm, i suppose it should be `if !inVPC`?

internal/alb/tg/targets.go
```diff
@@ -157,3 +164,32 @@ func tdsString(tds []*elbv2.TargetDescription) string {
 	}
 	return strings.Join(s, ", ")
 }
+
+func populateTargetAZ(ctx context.Context, c aws.CloudAPI, a []*elbv2.TargetDescription) error {
```
M00nF1sh: nitpick:
I prefer func (c *targetsController) populateTargetAZ(ctx context.Context, a []*elbv2.TargetDescription) {
    //use c.cloud when reference cloud 😄 
}

internal/alb/tg/targets.go
rifelpet: Oops, you're right. I'll get that fixed and investigate why the unit tests didn't fail.

rifelpet: I figured out why the unit test incorrectly failed and fixed both the test and the conditional in `populateTargetAZ`

internal/alb/tg/targets.go
rifelpet: Added `targetsController` as a receiver for `populateTargetAZ` and moved the function definition up with the other `targetsController` functions.

M00nF1sh:
Perfect 😄 
/lgtm
/approve

PR #900, opened 2019-03-21
Scopes must be space-separated list

cwegrzyn:
If you try to use commas as indicated in the documentation, you get an error similar to the following:

```
kubebuilder/controller "msg"="Reconciler error" "error"="failed to reconcile listeners due to failed to reconcile listener due to InvalidLoadBalancerAction: A scope cannot contain a comma. Separate multiple scopes using spaces\n\tstatus code: 400, request id: faaec23d-4bed-11e9-a3d9-25197b2b13db"
```


cwegrzyn:
I signed it

PR #920, opened 2019-04-14
ignore unreachable rules in alb listener

eddycharly:
I implemented a simple detection mechanism of unreachable rules in alb listener.
This prevents adding rules that will never be reached.
For example, in case of an ssl redirect, the current implementation added one that handled the redirect and another rule that forwarded to the target group, though the second rule could never be reached.


eddycharly:
@bigkraig @M00nF1sh CLA signed

M00nF1sh:
@eddycharly 
Thanks so much for helping! Really appreciate this work 👍 .
However, I'd prefer not to merge this due to the following considerations:
1. This will get really complicated when we supports [header/verb/query based routing](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/909).
1. This is an hack(the http->https redirection is an hack too) to solve the problem of different rule for different port. Personally I'd like to use ingress group(different rule for different port in different ingress spec, and use same group to aggregate to one alb) or explicit annotation like `alb.ingress.kubernetes.io/redirect-to-https`

eddycharly:
I didn’t know about the new routing strategy, I agree it will be difficult to implement such a logic.

I knew it was just a workaround and as with the ssl redirect hack there’s a lot of corner cases and side effects. 

In the end the ingress group gives more control to the end user (it’s not perfect though as it can generate conflicts when combining the ingresses together). 

Any idea if it will be available soon ?

M00nF1sh:
@eddycharly 
There is an alpha version available here: https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/914
Needs some time to add UT & cleanup some code before release a stable version.
But it works good in my local testing env 😄

eddycharly:
@M00nF1sh
Thanks, I’ll give it a try !

PR #923, opened 2019-04-18
enhance certificate auto-discover functionality

### LINKED ISSUE #896: Fallback automatic certificate assignment not working
Based on this PR:
https://github.com/kubernetes-sigs/aws-alb-ingress-controller/pull/851

I would expect to be able to pass in this configuration for an ingress and have certificate assignment work using wildcards, provided there is an ACM cert that matches:

```yaml
kind: Ingress
metadata:
  name: test-app
  labels:
    app: test-app
  annotations:
    alb.ingress.kubernetes.io/listen-ports: '[{"HTTPS":443}]'
    alb.ingress.kubernetes.io/security-groups: sg-name
    alb.ingress.kubernetes.io/ssl-policy: ELBSecurityPolicy-TLS-1-2-2017-01
    alb.ingress.kubernetes.io/healthcheck-path: "/health"
    alb.ingress.kubernetes.io/target-group-attributes: stickiness.enabled=true,stickiness.lb_cookie.duration_seconds=604800,deregistration_delay.timeout_seconds=0
spec:
  rules:
  - http:
      paths:
        - path: /*
          backend:
            serviceName: test-app
            servicePort: 443
  tls:
  - hosts:
    - test-app.foobar.com
```

In this case, a wildcard certificate for `*.foobar.com` exists. However, when a Helm release is pushed using this configuration (no certificate annotation, with the inclusion of the `spec.tls.hosts` section), the controller logs present these errors:

```bash
E0314 17:56:38.698337       1 :0] kubebuilder/controller "msg"="Reconciler error" "error"="failed to reconcile listeners due to failed to build listener config due to missing certificates annotation alb.ingress.kubernetes.io/certificate-arn could not find any matching certificates from ACM to auto-load"  "controller"="alb-ingress-controller" "request"={"Namespace":"default","Name":"test-app"}
```

Here is the version information. We are using the official chart but specifying the latest release tag:

```bash
-------------------------------------------------------------------------------
AWS ALB Ingress controller
  Release:    v1.1.1
  Build:      git-87bef3eb
  Repository: https://github.com/kubernetes-sigs/aws-alb-ingress-controller
-------------------------------------------------------------------------------
```

M00nF1sh:
Fix #896 
Action Required:
1. If customers are relying on certificate auto-discover feature, the following changes are made:
    1.  If multiple certificates were found for a host, an error will be issued. (previously, it will be silently ignored if at least one certificate are found)
    1.  If none certificate were found for a host, an error will be issued. (previously, it will be silently ignored if at least one certificate are found for other hosts)

Changes Done:
1. domains in SAN section of the certificate will be matched too
2. If multiple certificates were found for a host, an error will be issued.
3. If none certificate were found for a host, an error will be issued.

Note: 
1. AWS automatically includes the CommonName of the certificate into SAN as well.
2. create an certificate with SAN:
```openssl req -new -x509 -nodes -newkey rsa:2048 -sha256 -subj '/CN=m00nf1sh.io' -days 3650 -extensions SAN -config <(cat /etc/ssl/openssl.cnf <(printf "[SAN]\nsubjectAltName=DNS:example.com,DNS:www.example.com")) -out cert.pem -keyout key.pem```

### DIFF
--- a/internal/alb/ls/listener.go
+++ b/internal/alb/ls/listener.go
@@ -266,29 +266,57 @@ func (controller *defaultController) buildDefaultActions(ctx context.Context, op
 }
 
 // inferCertARNs retrieves a set of certificates from ACM that matches the ingress' hosts list
+// If multiple or none certificate were found for specific host, an error will be issued.
 func (controller *defaultController) inferCertARNs(ctx context.Context, ingress *extensions.Ingress) ([]string, error) {
 	var ingressHosts = uniqueHosts(ingress)
-	certArns := sets.NewString()
+	if len(ingressHosts) == 0 {
+		return nil, nil
+	}
 
-	logger := albctx.GetLogger(ctx)
-
-	certs, err := controller.cloud.ListCertificates([]string{acm.CertificateStatusIssued})
+	domainsByCertArn, err := controller.loadDomainsByCertArn(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	for _, c := range certs {
-		for _, h := range ingressHosts {
-			if domainMatchesHost(aws.StringValue(c.DomainName), h) {
-				logger.Infof("Domain name '%s', matches TLS host '%v', adding to Listener", aws.StringValue(c.DomainName), h)
-				certArns.Insert(aws.StringValue(c.CertificateArn))
-			} else {
-				logger.Debugf("Ignoring domain name '%s', doesn't match '%s'", aws.StringValue(c.DomainName), h)
+	certArns := sets.NewString()
+	for host := range ingressHosts {
+		certArnsForHost := sets.NewString()
+		for certArn, domains := range domainsByCertArn {
+			for _, domain := range domains {
+				if domainMatchesHost(domain, host) {
+					certArnsForHost.Insert(certArn)
+					break
+				}
 			}
 		}
+		if len(certArnsForHost) > 1 {
+			return nil, errors.Errorf("multiple certificate found for host: %s, certARNs: %v", host, certArnsForHost.List())
+		}
+		if len(certArnsForHost) == 0 {
+			return nil, errors.Errorf("none certificate found for host: %s", host)
+		}
+		certArns = certArns.Union(certArnsForHost)
 	}
-
 	return certArns.List(), nil
+}
+
+func (controller *defaultController) loadDomainsByCertArn(ctx context.Context) (map[string][]string, error) {
+	certSummaries, err := controller.cloud.ListCertificates(ctx, &acm.ListCertificatesInput{
+		CertificateStatuses: aws.StringSlice([]string{acm.CertificateStatusIssued}),
+	})
+	if err != nil {
+		return nil, err
+	}
+	domainsByCertArn := make(map[string][]string, len(certSummaries))
+	for _, certSummary := range certSummaries {
+		certArn := aws.StringValue(certSummary.CertificateArn)
+		certDetail, err := controller.cloud.DescribeCertificate(ctx, certArn)
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: internal/alb/ls/listener.go [hunk 1]
```diff
@@ -266,31 +266,59 @@ func (controller *defaultController) buildDefaultActions(ctx context.Context, op
...
+		if len(certArnsForHost) == 0 {
+			return nil, errors.Errorf("none certificate found for host: %s", host)
+		}
+		certArns = certArns.Union(certArnsForHost)
 	}
-
 	return certArns.List(), nil
 }
 
+func (controller *defaultController) loadDomainsByCertArn(ctx context.Context) (map[string][]string, error) {
```
leakingtapan: The name is confusing, from the code, it is returning a map from `certArn` to list of `SAN`s. Should we name it as `loadCertArnToSANs`?

internal/alb/ls/listener.go
```diff
@@ -266,31 +266,59 @@ func (controller *defaultController) buildDefaultActions(ctx context.Context, op
...
+	certSummaries, err := controller.cloud.ListCertificates(ctx, &acm.ListCertificatesInput{
+		CertificateStatuses: aws.StringSlice([]string{acm.CertificateStatusIssued}),
+	})
+	if err != nil {
+		return nil, err
+	}
+	domainsByCertArn := make(map[string][]string, len(certSummaries))
+	for _, certSummary := range certSummaries {
+		certArn := aws.StringValue(certSummary.CertificateArn)
+		certDetail, err := controller.cloud.DescribeCertificate(ctx, certArn)
```
devkid: @M00nF1sh This is a **really** expensive operation if you have many certificates in ACM. This will get called once for **each** certificate in ACM, for **each** listener in **each** controller-controlled ALB. Let's say you have 50 ingresses, 2 listeners each and 100 certs in ACM, this will perform 20k API operations in every reconcilation loop iteration.

In our cluster, I rolled back the update to a master build because the controller didn't finish the initial reconcilation after 10 minutes.

M00nF1sh: @devkid 
Hi, this is indeed heavy for such scenario. 
But list each certificate is unavoidable to support  SubjectAlternativeNames. There are some optimizations i can think of like.
1. do this once for all listeners for a single ingress. 
2. do this with a global cache that invalidates itself every X minutes. (but i don't think its that useful since ppl shouldn't not create/delete ingress frequently. This might be useful for current version since every pod relocate will trigger reconcile, and in V2, pod relocate only triggers targetGroup updates).
3. When doing ListCertificates, if all the domain name is found, use these certificates. Otherwise, describe these certificates to find SubjectAlternativeName.
4. The ideal solution should be provided by ACM with the ability to filter cert by domain/SAN in listCertificate API.

What do you think :D

devkid: 1. My first thought was also to use a cache for this. Is it necessary to invalidate it, though? As far as I can tell by the API documentation, it's not possible to update a certificate after its creation, so it would be sufficient to handle the case that certificates are deleted from ACM. (I guess you could easily notice this when the controller tries to use a certificate ARN and it gets an error.)
2. Also thought about this, but then you will not be able to throw an error if another certificate also has the desired domain name in its SANs. Not sure if this is important, though.
3. Either that or add `DescribeCertificates` which returns a list with all details of the certificates. I don't see why it would be a problem providing this as there doesn't seem to be any dynamic data returned from `DescribeCertificate` which would justify having separate `List` and `Describe` functions.

M00nF1sh: @devkid created an PR to fix this: https://github.com/kubernetes-sigs/aws-alb-ingress-controller/pull/1025

PR #927, opened 2019-04-22
Add healthcheck-enabled annotation and make the default `false`

devkid:
This adds support for the [`HealthCheckEnabled` setting of a target group](https://docs.aws.amazon.com/sdk-for-go/api/service/elbv2/#CreateTargetGroupInput).

I would suggest to set this flag to `false` by default as I would assume that usually you would configure a `readinessProbe` in the pod, then the ALB health check would be redundant. I'm aware that this would be a breaking change for some users and should be clearly noted in the release notes.

If the maintainers have a different opinion on this, I'll change the default it in the pull request of course.

### DIFF
--- a/docs/guide/ingress/annotation.md
+++ b/docs/guide/ingress/annotation.md
@@ -24,6 +24,7 @@ You can add kubernetes annotations to ingress and service objects to customize t
 |[alb.ingress.kubernetes.io/auth-type](#auth-type)|none\|oidc\|cognito|none|ingress,service|
 |[alb.ingress.kubernetes.io/backend-protocol](#backend-protocol)|HTTP \| HTTPS|HTTP|ingress,service|
 |[alb.ingress.kubernetes.io/certificate-arn](#certificate-arn)|stringList|N/A|ingress|
+|[alb.ingress.kubernetes.io/healthcheck-enabled](#healthcheck-enabled)|bool|false|ingress,service|
 |[alb.ingress.kubernetes.io/healthcheck-interval-seconds](#healthcheck-interval-seconds)|integer|'15'|ingress,service|
 |[alb.ingress.kubernetes.io/healthcheck-path](#healthcheck-path)|string|/|ingress,service|
 |[alb.ingress.kubernetes.io/healthcheck-port](#healthcheck-port)|integer \| traffic-port|traffic-port|ingress,service|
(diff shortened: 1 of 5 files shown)


M00nF1sh:
Hi, Thanks for making this change.
The aws documentation seems a little confusing, currently the HealthCheckEnabled parameter is only meant for lambda targetGroup, and can only be set to true for normal instance/ip targetGroups.

devkid:
So is it confusing or wrong?

> If the target type is instance or ip, the default is true. If the target type is lambda, the default is false.

This only states which default values are used for each target type.

M00nF1sh:
> So is it confusing or wrong?
> 
> > If the target type is instance or ip, the default is true. If the target type is lambda, the default is false.
> 
> This only states which default values are used for each target type.

Yes, i'd say it's missing some information here.(Actually I remember there are some docs that said healthCheckEnabled is only for lambda TG, but i cannot find it now). 
BTW, i tested disable HealthCheck for TG from aws cli, and got expected error like `when calling the ModifyTargetGroup operation: Health check enabled must be true for target groups with target type 'instance'`.  I'll sync with ALB team to check whether they can expand the disableHealthCheck functionality for instance/IP mode as well.

devkid:
Just as an additional side node on why this feature would be useful:

Usually, the ALB ingress controller is really fast in (de)registering targets with target groups. However, having a `RollingUpdate` deployment is still problematic if you can't disable the ALB health check because it can take longer for the target health check in the ALB (even with settings for lowest possible delay) to turn healthy than it takes kubernetes to replace one pod after another: This is especially a problem if you run a deployment with only two replicas and max surge=0, max unavailable=1; on a deployment update, one pod terminates (and is deleted within seconds) and gets removed from the ALB, a new pod comes up and becomes ready within seconds, which triggers the removal of the remaining old pod (while the one ready replacement pod is not yet healthy in the ALB because it's waiting for the health check to turn green – so in the mean time there is no healthy target behind the ALB).

BrianChristie:
@devkid if I understand your last message properly it sounds like it is describing the issue that is also discussed in https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/905 and the linked issues.

devkid:
@BrianChristie Yes, this is exactly what I was describing. I still think having the possibility to disable the ALB health check would make sense.

devkid:
@M00nF1sh any update on this?

M00nF1sh:
@devkid
Hi, i have already forward this feature request to ALB team last month. Will follow up with them whether it's on roadmap or not.

devkid:
@M00nF1sh ping?

dudicoco:
@M00nF1sh any updates? 
This is a really important feature, it would be much more efficient to have the healthchecks on the readinessProbe side.

dudicoco:
@devkid why was this closed?

devkid:
@dudicoco Because I waited unsuccessfully over a year for an answer from the maintainer, so I consider this a dead end and wanted the issue to vanish from my list of open issues :man_shrugging:

PR #947, opened 2019-06-02
Optional host rule

dblooman:
We have a scenario where we do not want to use the host checking rule, but this is on by default when using a hostname in ingress.  This PR allows for disabling of the host validation rule.

### DIFF
--- a/docs/guide/ingress/annotation.md
+++ b/docs/guide/ingress/annotation.md
@@ -24,6 +24,7 @@ You can add kubernetes annotations to ingress and service objects to customize t
 |[alb.ingress.kubernetes.io/auth-type](#auth-type)|none\|oidc\|cognito|none|ingress,service|
 |[alb.ingress.kubernetes.io/backend-protocol](#backend-protocol)|HTTP \| HTTPS|HTTP|ingress,service|
 |[alb.ingress.kubernetes.io/certificate-arn](#certificate-arn)|stringList|N/A|ingress|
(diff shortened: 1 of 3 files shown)


M00nF1sh:
Hi, would you help share the usecase of disabling it?
If it's because of external-DNS, they have a mechanism of use hosts from annotations.

From ingress's perspective, we don't like such feature that violates ingress spec.

dblooman:
When using an ALB behind a CDN where the CDN uses a different certificate.  We use a different domain for ALB's with ACM managed certificates, with our CDN matching routes that forward traffic to different ALB's.  I don't see this as breaking ingress spec as by default host validation isn't on, its something that is explicitly added as a rule by the controller.  I will have a second look at external DNS

M00nF1sh:
Hi, for such usecase, I think the ideal solution should be:
1. remove the Host condition from ingress rule spec.
2. use annotations on Ingress to configure external-dns if you used  it(https://github.com/kubernetes-incubator/external-dns/pull/545)
3. If you relies on alb-ingress-controller's cert-discover feature, you can specify the host in `ingress.spec.tls section` instead of `ingress.spec.rules`)
The reason I against adding this feature is that its an maintenance burden that have to keep in mind when doing feature upgrades while the original problem can be addressed in other ways above.

dblooman:
Hi @M00nF1sh, your approach worked great, thanks.

PR #955, opened 2019-06-18
Pod readiness gates

### LINKED ISSUE #905: Support Custom Pod Status for PodReadinessGate to Block Premature Pod Termination
Presently when using IP target mode on the ALB, it's difficult to guarantee zero downtime rolling updates. This is because there is additional state in the ALB Target Group which is not propagated back to Kubernetes: `initial`,`draining`, and `healthy`. 

This was discussed in:  https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/660 & https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/814

This issue is to discuss adding support to the ALB controller for setting a custom Pod Status which mirrors the ALB TargetGroup. This can then be used with `PodReadinessGate` and `MaxUnavailable` / `PodDisruptionBudget` to ensure a rolling update does not proceed faster than the new Pods becoming healthy on the ALB.

- KEP: kubernetes/enhancements#580
- Docs: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate

devkid:
This adds a feature to set the status of pod readiness gates on pods that are registered with an ALB. See added documentation for details of implementation.

Todo:
- [x] fix up unit tests, add test cases for pod readiness gates
- [x] test manually

This closes #905.


lelikg:
+1 on this feature, we'd love to get it

devkid:
I have continued working on this. It's currently in a usable state and tested manually. I will follow up with unit tests.

devkid:
@M00nF1sh ready for review

M00nF1sh:
@devkid 
Thanks sooo much for implementing this. I'll dive into the code today. some high-level thoughts would like to discuss:
1. should we simplify to use a single readiness gate instead of targetGroup?  (personally I favor your current approach)
   pros: easy for user usage before we have mutating webhook to automatically inject it
   cons: harder to implement. Needs to bookkeeping all targetGroup status.
   cons: cannot opt-out readiness gate for specific targetGroup.(not sure whether there is a use case for this)
2. should we have a timeout setting like 5 minutes. If pod didn't become healthy within 5 minute, unblock the readiness gate. (to prevent deployment lock for wrong settings, e.g. security groups).

docs/guide/ingress/pod-conditions.md
```diff
@@ -0,0 +1,84 @@
...
+If your pod is part of multiple ingresses / target groups and you want to make sure your pod is `Healthy` in all of them before it is marked as `Ready`, add one `readinessGate` per ingress.
+
+
+## <a name="annotations">Ingress annotations</a>
+
+The following annotations can be used on the `Ingress` to control the reconcilation behavior:
+
+* `alb.ingress.kubernetes.io/target-health-reconciliation-strategy`: can be either `initial` (default) or `continuous`
+  * `initial`: pod condition statuses are only reconiliated as long as the pod is unready; once it becomes ready, reconciliatio is stopped – it is started again if the pod becomes unready during its runtime
+  * `continuous`: pod condition statuses are reconciled as long as the ingress / target group exists; use with care as this can potentially cause a lot of AWS API calls if there are many target groups
```
M00nF1sh: what's the use case for continuous? seems there is no Kubernetes mechanism to leverage it. The deployment. minReadySeconds seems only consider pod's container Ready.
I'd rather avoid unnecessary functionality/configuration if there is no good use case for it.

internal/alb/tg/targethealth.go
```diff
@@ -0,0 +1,358 @@
...
+}
+
+// From the given targets, only returns the ones that have a readiness gate for the given ingress/service and whose pod conditions actually need to be reconciled.
+func (c *targetHealthController) filterTargetsNeedingReconciliation(conditionType api.PodConditionType, t *Targets, desiredTargets []*elbv2.TargetDescription) ([]*elbv2.TargetDescription, error) {
+	targetsToReconcile := []*elbv2.TargetDescription{}
+	if len(desiredTargets) == 0 {
+		return targetsToReconcile, nil
+	}
+
+	strategy, err := ingressTargetHealthReconciliationStrategy(t.Ingress)
```
M00nF1sh: if we really want to support a reconcile strategy(to expose pod's status to user?), it should be derived from Ingress & Service, while annotation on services takes priority. since it's an attribute related to targetGroups.

docs/guide/ingress/pod-conditions.md
devkid: I don't have a strong opinion on this, we could drop the continuous thingy altogether if you think it's a bad idea, I just thought some people would want to have this feedback into the cluster.

devkid:
>     1. should we simplify to use a single readiness gate instead of per targetGroup?  (personally I favor your current approach)
I think we should start with the per-target group approach. We can add the per-ingress functionality later if required.

>     2. should we have a timeout setting like 5 minutes. If pod didn't become healthy within 5 minute, unblock the readiness gate. (to prevent deployment lock for wrong settings, e.g. security groups).
I don't think we should have a timeout for this. The whole idea of readiness gates is that they *should* block deployments if the pods don't get registered in the load balancer (for whatever reason). Progressing with deployment when the readiness gate is not ok yet will most likely bring the service down eventually (e.g. when – for any reason – the ALB ingress controller is down).

docs/guide/ingress/pod-conditions.md
M00nF1sh: would help to make that change? we can drop the entire `target-health-reconciliation-strategy`

M00nF1sh:
overall looks good to me 👍 just we should remove the `target-health-reconciliation-strategy`.

devkid:
@M00nF1sh removed the strategy and interval. It uses now `healthCheckInterval * healthCheckThreshold` for the `time.After` and afterwards only `healthCheckInterval`.

devkid:
@M00nF1sh is there anything else that prevents from merging? Or do you need to coordinate this with the v2 work?

M00nF1sh:
@devkid 
There isn't, and we should get this into v1.1.6 🤣 . Let me do a test for the code logic and ship this by EOD.

internal/alb/tg/targethealth.go
```diff
@@ -0,0 +1,339 @@
...
+		return healthcheck.DefaultIntervalSeconds * targetgroup.DefaultHealthyThresholdCount
+	}
+	return healthcheck.DefaultIntervalSeconds
+}
+
+// PodConditionTypeForIngressBackend returns the PodConditionType that is associated with the given ingress and backend
+func podConditionTypeForIngressBackend(ingress *extensions.Ingress, backend *extensions.IngressBackend) api.PodConditionType {
+	return api.PodConditionType(fmt.Sprintf(
+		"target-health.%s/%s_%s_%s",
+		parser.AnnotationsPrefix,
```
M00nF1sh: i'd prefer to change this to hard-coded `alb.ingress.kubernetes.io` or better `alb.ingress.k8s.aws` (aligns with v2).

the parser.AnnotationsPrefix is dedicated for annotations, and people can override it with commandline arguments.

internal/alb/tg/targethealth.go
```diff
@@ -0,0 +1,339 @@
...
+
+	tgWatch, ok := c.tgWatches[t.TgArn]
+	if ok {
+		// targetGroupWatch for this target group already exists
+		if len(targetsToReconcile) == 0 {
+			tgWatch.cancel()
+			delete(c.tgWatches, t.TgArn)
+		} else {
+			tgWatch.mux.Lock()
+			tgWatch.ingress = t.Ingress
```
M00nF1sh: what's the purpose here. why only ingress & targets been updated but not backend?

internal/ingress/backend/endpoint.go
```diff
@@ -54,6 +58,55 @@ func (resolver *endpointResolver) Resolve(ingress *extensions.Ingress, backend *
...
+			// servicePort.Name is optional if there is only one port
+			if servicePort.Name != "" && servicePort.Name != epPort.Name {
+				continue
+			}
+			for _, epAddr := range append(epSubset.Addresses, epSubset.NotReadyAddresses...) {
+				if epAddr.TargetRef == nil || epAddr.TargetRef.Kind != "Pod" {
+					continue
+				}
+
+				pod, ok := podMap[epAddr.TargetRef.Name]
```
M00nF1sh: why we don't get to pod directly using store(or better controller-runtime client/cache) directly. It will be in-memory lookup and avoid the dependency of service selector.

internal/ingress/backend/endpoint.go
```diff
@@ -93,14 +146,72 @@ func (resolver *endpointResolver) resolveIP(ingress *extensions.Ingress, backend
...
 			if servicePort.Name != "" && servicePort.Name != epPort.Name {
 				continue
 			}
-			for _, epAddr := range epSubset.Addresses {
+			addresses := epSubset.Addresses
+			if service.Spec.PublishNotReadyAddresses {
+				addresses = append(addresses, epSubset.NotReadyAddresses...)
+			} else {
+				// if `PublishNotReadyAddresses` is not set, we need to loop over all unready pods to check if the ALB readiness gate is the only condition preventing the pod from being ready;
+				// if this is the case, we return the pod as a desired target although its not in `Addresses`
```
M00nF1sh: this is a hacky logic. If there are another similar controller using same logic (only try to reconcile condition if it's the only pending condition), it will be dead lock..

internal/ingress/backend/endpoint.go
```diff
@@ -93,14 +146,72 @@ func (resolver *endpointResolver) resolveIP(ingress *extensions.Ingress, backend
...
 	var result []*elbv2.TargetDescription
 	for _, epSubset := range eps.Subsets {
 		for _, epPort := range epSubset.Ports {
 			// servicePort.Name is optional if there is only one port
 			if servicePort.Name != "" && servicePort.Name != epPort.Name {
 				continue
 			}
-			for _, epAddr := range epSubset.Addresses {
+			addresses := epSubset.Addresses
+			if service.Spec.PublishNotReadyAddresses {
```
M00nF1sh: i don't think PublishNotReadyAddresses is used for this case.

internal/alb/tg/targethealth.go
```diff
@@ -0,0 +1,339 @@
...
+		tgWatch = &targetGroupWatch{
+			ingress:            t.Ingress,
+			backend:            t.Backend,
+			targetsToReconcile: targetsToReconcile,
+			cancel:             cancel,
+		}
+		c.tgWatches[t.TgArn] = tgWatch
+
+		// start watching target health in target group and updating pod condition status
+		go c.reconcilePodConditionsLoop(ctx, t.TgArn, conditionType, tgWatch)
```
M00nF1sh: this is a bit hacky but works. (it replies on pod been updated and re-trigger the reconcile logic, thus triggers the `if len(targetsToReconcile) == 0`  branch. It should works under current implementation. 
But in my opinion, a better safe approach is like:
```
if len(newTargets) != 0 {
      tgWatch = c.tgWatches[t.TgArn]
      addNewTargetsIntoTgWatch(tgWatch, newTargets)
      if tgWatch.CoroutingCompleted() {
          go tgWatch.NewCoroutingRunIntoCompletionUponTracingNewTargetsHealthy()
      }
}
```

internal/alb/tg/targethealth.go
```diff
@@ -0,0 +1,339 @@
...
+			logger.Infof("Stopping reconciliation of pod condition status for target group: %v", tgArn)
+			return
+		}
+	}
+}
+
+// For each given pod, checks for the health status of the corresponding target in the target group and adds/updates a pod condition that can be used for pod readiness gates.
+func (c *targetHealthController) reconcilePodConditions(ctx context.Context, tgArn string, conditionType api.PodConditionType, ingress *extensions.Ingress, backend *extensions.IngressBackend, targetsToReconcile []*elbv2.TargetDescription) error {
+	in := &elbv2.DescribeTargetHealthInput{TargetGroupArn: aws.String(tgArn)}
+	resp, err := c.cloud.DescribeTargetHealthWithContext(ctx, in)
```
M00nF1sh: DescribeTargetHealthInput supports passing targets as argument, which can safe calls & data transfer when there are lots of pod backends

internal/alb/tg/targethealth.go
devkid: That's actually a left-over. The ingress update was so that changes to the target health-specific annotations would be detected while the go routine was running, but since these annotations are gone now, this is not needed anymore. (I guess updates to the annotations for health check interval / health threshold count shouldn't matter.)

internal/ingress/backend/endpoint.go
devkid: > If there are another similar controller using same logic (only try to reconcile condition if it's the only pending condition), it will be dead-lock..

That's a good point. But what if you have other readiness gates that need to be fulfilled until a pod really gets ready to accept traffic? If you ignore those, the pod will be registered and healthy in the load balancer maybe without a necessary side condition (represented by the other readiness gate) being fulfilled.

Right now, we don't have other readiness gates, but that might change in the near future and these would need to be fulfilled before the pod gets registered. How about for now I change the code to only look at `ContainersReady` and later on we could introduce another annotation like `<prefix>/wait-for-pod-conditions: '["readiness-gate-1", "readiness-gate-2"]'` to wait for specific conditions before registering the pod? Would that make sense to you?

internal/ingress/backend/endpoint.go
devkid: I see, according to the documentation it's mainly meant for DNS. We don't need this at the moement, I just thought it _might_ be useful for _someone_. I can remove it if you want.

internal/alb/tg/targethealth.go
devkid: I don't think I understand what you mean exactly. You mentioned the `if len(targetsToReconcile) == 0` branch but your code is for the `if len(newTargets) != 0` branch?

internal/alb/tg/targethealth.go
devkid: So `target-health.alb.ingress.k8s.aws`? Is the `alb.ingress.k8s.aws` part already decided upon or might this change in the future?

internal/ingress/backend/endpoint.go
devkid: Sounds good, will change it accordingly.

internal/alb/tg/targethealth.go
devkid: I see, I didn't notice that it had a parameter for targets. Will change it.

internal/alb/tg/targethealth.go
M00nF1sh: yes, we plan to use `alb.ingress.k8s.aws` for ingress related. more specifically, we plan to use `k8s.aws` as domain suffix for our CRDs as well.

internal/alb/tg/targethealth.go
M00nF1sh: then we may consider make tgWatch depends on `healthCheckInterval` instead of both Ingress & Backend.

internal/ingress/backend/endpoint.go
M00nF1sh: yeah, i think it's better to remove it since it's not our use case.

internal/alb/tg/targethealth.go
M00nF1sh: sorry, i meant that the current code will keep looping and calls describeTargets until we receive a new pod event and triggers an new reconcile(which may not happen in cases like Cx modified the ALB annotation to be malformed).

I have write a pseudo code below: 
TargetHealthController.AddTargets will only be called when adding new targets to targetGroup, which will find the corresponding tgWatch and call tgWatch.AddTargets
```
type TargetHealthController interface {
    AddTargets(ctx context.Context, t* Targets, targets []*elbv2.TargetDescription) error
    RemoveTargets(ctx context.Context, t* Targets, targets []]*elbv2.TargetDescription) error
	StopReconciling(ctx context.Context, tgArn)
}

type TGWatch interface {
    AddTargets(targets []]*elbv2.TargetDescription) error
    RemoveTargets(targets []]*elbv2.TargetDescription) error
    Stop() error
}

type tgWatch struct {
    lbArn string
    targets []*elbv2.TargetDescription
    healthCheckInterval int64
    cloud aws.cloud
    cancel func()

    addTargetsChan chan []*elbv2.TargetDescription
    removeTargetsChan chan []*elbv2.TargetDescription
}

func NewTGWatch(ctx context.Context, lbArn string, targets []*elbv2.TargetDescription, healthCheckInterval int64) *tgWatch {
    ctx, cancel = ctx.WithCancel()
    w := &tgWatch {
        lbArn: lbArn,
        targets: targets,
        healthCheckInterval: healthCheckInterval
        cancel: cancel
    }
    go w.run(ctx)
    return w
}

func (w *tgWatch) AddTargets(targets []*elbv2.TargetDescription) error {
    w.addTargetsChan <- targets
}

func (w *tgWatch) RemoveTargets(targets []]*elbv2.TargetDescription) error {
    w.removeTargetsChan <- targets
}

func (w *tgWatch) run(ctx context.Context) error {
    for {
        var startPull <-chan time.Time
        if len(w.targets) != 0 {
            startPull = time.After(w.healthCheckInterval)
        }
        switch {
            case <-startPull:   # will be ignored if w.targets is empty since switch won't select nil channel
                w.pullOnce()
            case targets := <-addTargetsChan:
                // modify w.targets by adding targets into it, targets already exists don't need to be added.
                addTargetsHandlingDuplications(w.targets, targets)
            case targets :=  <-removeTargetsChan:
                removeTargetsHandlingDuplications(w.targets, targets)
            case <-ctx.Done():
                return
        }
    }
}

func (w *tgWatch) pullOnce(ctx context.Context) {
    targetHealths := w.cloud.DescribeTargetHealth(lbArn, w.targets)
    healthyTargets := filterHealthyTargets(targetHealths)
    w.RemoveTargets(ctx, healthyTargets)

    removePodReadinessGateFromTargets(healthyTargets)
}
```

devkid: Unfortunately the add/remove part will not really work. If the ingress controller gets killed for any reason after targets have been added to the target group and before the pod condition has been set to `True`, the pod condition would never get `True`. That's why we **must** look at all the desired targets (not only the ones that will be added by a reconciliation loop iteration) and afterwards filter the pods that still need to have their target health checked.

But what we could do instead is to stop the go routine and remove it from the map from within the loop if there are no targets to reconcile anymore.

internal/alb/tg/targethealth.go
devkid: Not sure what you mean, it needs both ingress and backend, they just don't need to be updated while the go routine is running.

M00nF1sh: i didn't notice that ingress & backend is used in logs, we can keep current behavior and skip the update

internal/alb/tg/targethealth.go
M00nF1sh: Good point, I haven't considered the controller restart case.
Yeah, we either can stop the go routing or make the loop no-op or block on channels. The point is we should keep track of targets that needs to be checked(and remove targets already turns healthy), are only make the APICall for these targets need to be checked in every loop.

internal/ingress/backend/endpoint.go
M00nF1sh: yeah, make sense to me :D

internal/alb/tg/targethealth.go
devkid: It's also used for retrieving the pods corresponding to the targets.

internal/alb/tg/targethealth.go
devkid: @M00nF1sh I changed the communication pattern between the `SyncTargetsForReconciliation` and the `reconcilePodConditionsLoop` to use channels instead of mutexes.

internal/ingress/backend/endpoint.go
```diff
@@ -54,6 +56,50 @@ func (resolver *endpointResolver) Resolve(ingress *extensions.Ingress, backend *
...
+				if epAddr.TargetRef == nil || epAddr.TargetRef.Kind != "Pod" {
+					continue
+				}
+
+				podKey := ingress.Namespace + "/" + epAddr.TargetRef.Name
+				pod, err := resolver.store.GetPod(podKey)
+				if err != nil {
+					continue
+				}
+
```
nirnanaaa: @M00nF1sh @devkid  shouldn't we also check if DeletionTimestamp is != nil? because pods, that are in state terminating can still be found in the store, but should already be drained, right?

devkid: Pods in `Terminating` state appear neither in `Addresses` nor in `NotReadyAddresses`.

nirnanaaa: :man_shrugging: for us they do, that's why I mentioned it here.

devkid: That's weird. I actually thought about implementing the check for `DeletionTimestamp != nil` but I verified on our end that terminating pods do *not* appear in `NotReadyAddresses`. Feel free to add the check. I'll have another look at this on our end as well next week.

PR #958, opened 2019-06-21
Added docs for creating DNS pointing from ingress object

luvpreetsingh:
### Description:
Added docs for creating DNS pointing from ingress object

### Motivation:
I had set up the external-dns but I was having trouble finding the docs on how to add the dns entry using ingress/service/external-dns for a particular service in kubernetes. I then had to go through a lot of blogs/articles and at one place I finally found that it can be done using annotations. I think that it should be described in the docs that after setting up the external-dns how one can use it to create dns record-sets for their apps. Maybe I didn't look thoroughly enough, if so, please point in the right direction :) Also, please excuse any English grammar mistakes, if any, it's not my first language.

### Testing:
I did not do any testing as it was only documentation changes.

### DIFF
--- a/docs/guide/external-dns/setup.md
+++ b/docs/guide/external-dns/setup.md
@@ -46,9 +46,9 @@ Adequate roles and policies must be configured in AWS and available to the node(
 ## Usage:
 To create a record set in the subdomain, from your ingress which has been created by the ingress-controller, simply add the following annotation in the ingress object specification,
 
-  annotations:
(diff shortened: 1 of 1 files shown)


M00nF1sh:
Thanks so much for making contributions 🚀

luvpreetsingh:
can someone please help with this failed test? I have no idea of it whatsoever.

luvpreetsingh:
@M00nF1sh can it be merged if you approve of this?

PR #978, opened 2019-07-26
Add automountServiceAccountToken to example

maur1th:
I ran into this issue: https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/965

Adding `automountServiceAccountToken: true` to the example deployment yaml fixed it.

### DIFF
--- a/docs/examples/alb-ingress-controller.yaml
+++ b/docs/examples/alb-ingress-controller.yaml
@@ -20,6 +20,7 @@ spec:
       labels:
         app.kubernetes.io/name: alb-ingress-controller
     spec:
+      automountServiceAccountToken: true
       containers:
         - name: alb-ingress-controller
           args:


leakingtapan:
ok-to-test

maur1th:
Tests appear to pass now. Could I get a review please? 🙏

M00nF1sh:
@maur1th Hi, would you help get me understand why this is needed...since automountServiceAccountToken should be default true in our serviceAccount's yaml...

dthvt:
I don't know enough k8s to explain why it's needed, but I also ran into this as an issue while deploying alb-ingress-controller.  Adding it allowed the controller pod to deploy successfully.

maur1th:
Just the same here. We do our K8s deployments through the Terraform Kubernetes provider though so it might come from this.

PR #990, opened 2019-08-08
added authorization bypass annotation

tipok:
Added an authorization bypass annotation, which allows creating rules without authorization.

It will be probably possible to achieve the same behavior once 1.2.0 is released.


tipok:
signed CLA

M00nF1sh:
Hi Anton, thanks so much for this PR. But i don't think it's a good idea to merge this....since we'll going to support advanced routing soon(route by header/verb/etc..), and this didn't play well with it.
As an work around before V2(IngressGroup), the auth annotations can be applied to Service object as well(which takes preference over annotation on Ingress). So you can add the auth annotation to Service when u want to have authentication(you can create another service if both auth/no-auth have same pod backend).

tipok:
Hi,
thank you for the update. Looking forward to the new release.

regorsmitz:
> As a workaround before V2(IngressGroup), the auth annotations can be applied to Service object as well(which takes preference over annotation on Ingress). So you can add the auth annotation to Service when u want to have authentication(you can create another service if both auth/no-auth have same pod backend).

This might be too far in the weeds, but if we are using Istio, and our Ingress directs to a Service via an Ingress Gateway and VirtualService, is there any way to take advantage of this auth annotation behavior? I tested by naively adding this annotation: `alb.ingress.kubernetes.io/auth-on-unauthenticated-request: allow` to my service (after traffic direction by Istio), and it seemed to have no effect.

Issue #887, opened 2019-03-07
Load balancer security group not being created with IPv6 ingress rules


orrc:
Creating an ALB ingress with the following annotations, almost everything appears to be set up correctly — an ALB with internet-facing, dualstack configuration is set up, along with the correct target groups.

However, the associated security group (labelled "managed LoadBalancer securityGroup by ALB Ingress Controller"), only creates two rules: one for each port (80, 443), with the IPv4 CIDR: `0.0.0.0/0`.  IPv6 clients are unable to connect unless I manually update the security group with a `::/0` rule for each port.

```yaml
kubernetes.io/ingress.class: alb

alb.ingress.kubernetes.io/scheme: internet-facing
alb.ingress.kubernetes.io/ip-address-type: dualstack
alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS": 443}]'

alb.ingress.kubernetes.io/actions.redirect-to-https: '{"Type": "redirect", "RedirectConfig":{"Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}'

alb.ingress.kubernetes.io/target-type: ip

alb.ingress.kubernetes.io/ssl-policy: ELBSecurityPolicy-TLS-1-1-2017-01
alb.ingress.kubernetes.io/certificate-arn: arn:aws:acm:<snip>
```

I see that the list of CIDRs to add to the security group comes from the `alb.ingress.kubernetes.io/inbound-cidrs` and defaults to `0.0.0.0/0`, but when I try to add `::/0` for IPv6 to this list, it's [rejected](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/cc1c59719630076243ac177a5f5c0df8d69970d3/internal/ingress/annotations/loadbalancer/main.go#L214-L224).

Am I missing something in my configuration to allow IPv6 to work when the ingress controller creates ALBs?

<details>
<summary>Logs when creating a new ingress</summary>

```
09:19:55.054314       1 association.go:224] kube-system/test-ingress: creating securityGroup 1173f105-kubesystem-testin-8be6:managed LoadBalancer securityGroup by ALB Ingress Controller
[...]

PR #991, opened 2019-08-08
IPv6 Support in Inbound CIDR Annotation

csnitker:
1. Added support for IPv6 CIDR blocks in the inbound CIDRs annotation
2. Added default CIDR (`::/0`) when ip address type is set to dualstack
 
This fixes: #887

### DIFF
--- a/internal/ingress/annotations/loadbalancer/main.go
+++ b/internal/ingress/annotations/loadbalancer/main.go
@@ -214,25 +214,25 @@ func parseCidrs(ing parser.AnnotationInterface) (v4CIDRs, v6CIDRs []string, err 
 	}
 
 	for _, inboundCidr := range cidrConfig {
-		ip, _, err := net.ParseCIDR(inboundCidr)
+		_, _, err := net.ParseCIDR(inboundCidr)
 		if err != nil {
 			return v4CIDRs, v6CIDRs, err
 		}
 
-		if ip.To4() == nil {
+		if strings.Contains(inboundCidr, ":") {
 			v6CIDRs = append(v6CIDRs, inboundCidr)
 		} else {
 			v4CIDRs = append(v4CIDRs, inboundCidr)
 		}
 	}
 
-	if len(v4CIDRs) == 0 {
+	if len(v4CIDRs) == 0 && len(v6CIDRs) == 0 {
 		v4CIDRs = append(v4CIDRs, "0.0.0.0/0")
-	}
 
-	addrType, _ := parser.GetStringAnnotation("ip-address-type", ing)
-	if addrType != nil && *addrType == elbv2.IpAddressTypeDualstack && len(v6CIDRs) == 0 {
-		v6CIDRs = append(v6CIDRs, "::/0")
+		addrType, _ := parser.GetStringAnnotation("ip-address-type", ing)
+		if addrType != nil && *addrType == elbv2.IpAddressTypeDualstack {
+			v6CIDRs = append(v6CIDRs, "::/0")
+		}
 	}
 
 	return v4CIDRs, v6CIDRs, nil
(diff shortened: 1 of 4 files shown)


csnitker:
@M00nF1sh or @micahhausler - Any chance I could get a review and an ok for testing?

internal/ingress/annotations/loadbalancer/main.go
```diff
@@ -214,18 +216,26 @@ func parseCidrs(ing parser.AnnotationInterface) (out []string, err error) {
 	for _, inboundCidr := range cidrConfig {
 		ip, _, err := net.ParseCIDR(inboundCidr)
 		if err != nil {
-			return out, err
+			return v4CIDRs, v6CIDRs, err
 		}
 
 		if ip.To4() == nil {
-			return out, fmt.Errorf("CIDR must use an IPv4 address: %v", inboundCidr)
```
M00nF1sh: ip.To4() is not enough, ip.To4() for "::/0" will return non-nil, and cause SG update fails.
Example checks in V2 branch: https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/v2/pkg/build/lb_security_group.go#L145

internal/ingress/annotations/loadbalancer/main.go
```diff
@@ -214,18 +216,26 @@ func parseCidrs(ing parser.AnnotationInterface) (out []string, err error) {
...
+			v6CIDRs = append(v6CIDRs, inboundCidr)
+		} else {
+			v4CIDRs = append(v4CIDRs, inboundCidr)
 		}
-		out = append(out, inboundCidr)
 	}
-	if len(out) == 0 {
-		out = append(out, "0.0.0.0/0")
```
M00nF1sh: would we change the logic to align with V2: https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/v2/pkg/build/lb_security_group.go#L166
Basically only append v4/v6 default settings when customer haven't explicitly configured any cidrs.
If the customer only configured v6 cidr but didn't specify "dualstack", it's typically due an error(where customer expected only v6 works), and we shouldn't append default 0.0.0.0/0 in such case)

M00nF1sh:
Thanks for making contributions :D left a few tiny comments :D

M00nF1sh:
@csnitker 
Hi, would you help change sg/security_group.go as well, currently it doesn't support ipv6 permissions.
The e2e test is failing with below msg:
```
ingress-7696/multi-path-echo: granting inbound permissions to securityGroup sg-0a930ff60cad5318e: [{    FromPort: 80,    IpProtocol: "tcp",    IpRanges: [{        CidrIp: "0.0.0.0/0",        Description: "Allow ingress on port 80 from 0.0.0.0/0"      }],    Ipv6Ranges: [],    ToPort: 80  }]
E0816 23:07:36.507947       1 controller.go:217] kubebuilder/controller "msg"="Reconciler error" "error"="failed to reconcile securityGroup associations due to failed to reconcile managed LoadBalancer securityGroup due to failed to grant inbound permissions due to InvalidParameterValue: missing mandatory parameter: exactly one of remote-security-group, remote-ip-range, remote-ipv6-range, or prefix-list-id must be present\n\tstatus code: 400, request id: 9bc39a33-f294-43b2-a8a9-b807c5c35960"  "controller"="alb-ingress-controller" "request"={"Namespace":"ingress-7696","Name":"multi-path-echo"}
```

M00nF1sh:
@csnitker 
Looks like it's still failing..i'll update the e2e test case to have a timeout instead of forever wait...

As of now, I think the easiest fix is to add IpV4 and IpV6 as separate ec2.IpPermission Object(above the line of `inboundPermissions = append(inboundPermissions, permission)`) That's because ec2 expects each ec2.IpPermission contains either ipv4 or ipv6 permission.
Alternatively, we'll need to implement https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/f8556d824c158201ecbda09bf25c4dc2ae610896/pkg/deploy/lb_security_group.go#L279 in securityGroup.Reconcile.

csnitker:
@M00nF1sh - Looks like that was the problem, the tests appear to be passing this time

PR #1023, opened 2019-09-13
[WIP] Upgrade controller-runtime

rifelpet:
We've been experiencing some controller bugs where the ALB ingress controller isn't detecting certain changes to ingress annotations. We'll update an ingress annotation and the ALB ingress controller isn't notified about it until the ingress is then updated a second time.

I'm going to try upgrading the controller runtime (and as a result, client-go) to see if that helps. There are breaking changes in the Informers (https://github.com/kubernetes-sigs/controller-runtime/pull/267) that I haven't had an opportunity to fix. If someone wants to step in and fix the remaining issues, feel free. Otherwise I'm hoping to get to it sometime soon.


M00nF1sh:
i think it might due to the annotation is parsed separately than the reconcile. (this problem don't exists in V2 though :D). a simple fix(without hard changes) should be force parse annotations before each reconcile :D

rifelpet:
Based on the log output and my understanding of how controller-runtime works, it seemed to me that the [EventHandler](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/internal/ingress/controller/store/store.go#L163-L198) wasn't being called during the first ingress update. I may be able to test your v2 implementation soon. It looks like your v2 still uses the controller-runtime 0.1.0, so I should be able to tell whether this issue is fixed or not 👍 regardless I think it would be good to get onto CR 0.2.0

PR #1039, opened 2019-09-24
Try to prevent and don't count predictable AWS API errors

### LINKED ISSUE #1008: AWS API errors reported for "LoadBalancerNotFoundError" that occurs while checking if a load balancer already exists before trying to create it
If a new ingress is created, the ALB ingress controller first checks if an ALB with the generated name already exists before creating it. This check will obviously return an error that the ALB does not exist. However, this is printed as an error on stdout:

```
I0829 11:36:16.126978       1 request_pagination.go:105] Request: elasticloadbalancing/DescribeLoadBalancers, Payload: {  Names: ["my-alb-name"]}
E0829 11:36:16.214391       1 request_pagination.go:105] Failed request: elasticloadbalancing/DescribeLoadBalancers, Payload: {  Names: ["my-alb-name"]}, Error: LoadBalancerNotFound: Load balancers '[my-alb-name]' not found
I0829 11:36:16.214417       1 loadbalancer.go:185] my-namespace/my-ingress-name: creating LoadBalancer my-alb-name
I0829 11:36:16.214706       1 api.go:474] Request: elasticloadbalancing/CreateLoadBalancer, Payload: {  IpAddressType: "ipv4",  Name: "my-alb-name",  Scheme: "internal",  Subnets: ["subnet-09eb234da1baf588b","subnet-927bf4e5","subnet-e618bbe9"],  Tags: [{      Key: "kubernetes.io/namespace",      Value: "my-namespace"    },{      Key: "kubernetes.io/ingress-name",      Value: "my-ingress-name"    },{      Key: "kubernetes.io/cluster/my-cluster",      Value: "owned"    }],  Type: "application"}
I0829 11:36:16.954966       1 api.go:474] Response: elasticloadbalancing/CreateLoadBalancer, Body: {  LoadBalancers: [{      AvailabilityZones: [{          SubnetId: "subnet-09eb234da1baf588b",          ZoneName: "us-east-1b"        },{          SubnetId: "subnet-927bf4e5",          ZoneName: "us-east-1c"        },{          SubnetId: "subnet-e618bbe9",          ZoneName: "us-east-1f"        }],      CanonicalHostedZoneId: "Z35SXDOTRQ7X7K",      CreatedTime: 2019-08-29 11:36:16.74 +0000 UTC,      DNSName: "internal-my-alb-name-2067509925.us-east-1.elb.amazonaws.com",      IpAddressType: "ipv4",      LoadBalancerArn: "arn:aws:elasticloadbalancing:us-east-1:1234567890:loadbalancer/app/my-alb-name/ac55786ca4867b4c",      LoadBalancerName: "my-alb-name",      Scheme: "internal",      SecurityGroups: ["sg-2f834e4b"],      State: {        Code: "provisioning"      },      Type: "application",      VpcId: "vpc-12347890"    }]}
I0829 11:36:16.954999       1 loadbalancer.go:201] my-namespace/my-ingress-name: LoadBalancer my-alb-name created, ARN: arn:aws:elasticloadbalancing:us-east-1:1234567890:loadbalancer/app/my-alb-name/ac55786ca4867b4c
```

and also counts towards the prometheus counter metric `aws_alb_ingress_controller_aws_api_errors` and disturbs our monitoring.

devkid:
This inserts a delay after creating an ALB listener because sometimes
interacting with a listener directly after creating it will produce
"listener not found" errors.

It also adds a context.Context value that is passed to Describe calls
where it is a normal situation that they report a "not found" error if
the function is called prior to creating the ALB/target group.

This closes #1008.

### DIFF
--- a/internal/aws/session.go
+++ b/internal/aws/session.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"regexp"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/request"
@@ -32,11 +33,14 @@ func NewSession(awsconfig *aws.Config, AWSDebug bool, mc metric.Collector) *sess
 		}
 	})
 
+	notFoundRegex := regexp.MustCompile("^[A-za-z]+NotFound")
 	session.Handlers.Complete.PushFront(func(r *request.Request) {
 		if r.Error != nil {
-			mc.IncAPIErrorCount(prometheus.Labels{"service": r.ClientInfo.ServiceName, "operation": r.Operation.Name})
-			if AWSDebug {
-				glog.ErrorDepth(4, fmt.Sprintf("Failed request: %s/%s, Payload: %s, Error: %s", r.ClientInfo.ServiceName, r.Operation.Name, log.Prettify(r.Params), r.Error))
+			if value, ok := r.Context().Value("report-not-found-error").(bool); !notFoundRegex.MatchString(r.Error.Error()) && !ok || value {
+				mc.IncAPIErrorCount(prometheus.Labels{"service": r.ClientInfo.ServiceName, "operation": r.Operation.Name})
+				if AWSDebug {
+					glog.ErrorDepth(4, fmt.Sprintf("Failed request: %s/%s, Payload: %s, Error: %s", r.ClientInfo.ServiceName, r.Operation.Name, log.Prettify(r.Params), r.Error))
+				}
 			}
 		} else {
 			if AWSDebug {
--- a/internal/alb/ls/listener.go
+++ b/internal/alb/ls/listener.go
@@ -3,6 +3,7 @@ package ls
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 
@@ -119,6 +120,10 @@ func (controller *defaultController) newLSInstance(ctx context.Context, lbArn st
 	if err != nil {
 		return nil, err
 	}
+
+	// due to EC2's eventual consistency, sometimes we get "not found" errors when trying to interact with a listener (e.g. Describe) directly after creating it
+	time.Sleep(1)
+
 	return resp.Listeners[0], nil
 }
 
--- a/internal/aws/elbv2.go
+++ b/internal/aws/elbv2.go
@@ -222,7 +222,7 @@ func (c *Cloud) DeleteListenersByArn(ctx context.Context, lsArn string) error {
 }
 
 func (c *Cloud) GetLoadBalancerByArn(ctx context.Context, arn string) (*elbv2.LoadBalancer, error) {
-	loadBalancers, err := c.describeLoadBalancersHelper(&elbv2.DescribeLoadBalancersInput{
+	loadBalancers, err := c.describeLoadBalancersHelper(ctx, &elbv2.DescribeLoadBalancersInput{
 		LoadBalancerArns: []*string{aws.String(arn)},
 	})
 	if err != nil {
@@ -235,7 +235,7 @@ func (c *Cloud) GetLoadBalancerByArn(ctx context.Context, arn string) (*elbv2.Lo
 }
 
 func (c *Cloud) GetLoadBalancerByName(ctx context.Context, name string) (*elbv2.LoadBalancer, error) {
-	loadBalancers, err := c.describeLoadBalancersHelper(&elbv2.DescribeLoadBalancersInput{
+	loadBalancers, err := c.describeLoadBalancersHelper(context.WithValue(ctx, "report-not-found-error", false), &elbv2.DescribeLoadBalancersInput{
 		Names: []*string{aws.String(name)},
 	})
 	if err != nil {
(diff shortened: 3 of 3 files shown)


devkid:
@M00nF1sh any chance to get this reviewed?

M00nF1sh:
Hi,
Thanks for making this change.
1. I don't think the 'report-not-found-error' metrics is needed. it's too intrusive(also, not every aws service will have the same error message).
2. alternatively to`time.Sleep(1)`, I think it's better to use describeListener to pull listener until it exists. [wait.PollImmediateUntil](https://godoc.org/k8s.io/apimachinery/pkg/util/wait#PollImmediateUntil) with stopCh be a 5 minute upper bound should be good.

devkid:
Thanks for review!

>     1. I don't think the 'report-not-found-error' metrics is needed. it's too intrusive(also, not every aws service will have the same error message).

Hm I'm not entirely sure what you mean. `report-not-found-error` is not a metric, but is just a random string that I made up that can be added to a context when invoking any AWS function that ends with `WithContext` (which is most I believe?) and that is evaluated in the request handlers installed in `NewSession`. The advantage of passing this in the context is that it's easy to add to the existing code (because there is no other obvious way to pass additional metadata per request that says "if you get a 'NotFound' error, don't count and don't report it" to the handlers).

>     2. alternatively to`time.Sleep(1)`, I think it's better to use describeListener to pull listener until it exists. [wait.PollImmediateUntil](https://godoc.org/k8s.io/apimachinery/pkg/util/wait#PollImmediateUntil) with stopCh be a 5 minute upper bound should be good.

Thanks, will check this out. I don't have a real high-level overview of the code, is the code path to `newLSInstance` running in a separate go routine or will it block the process when I wait for an extended period of time at this point?

M00nF1sh:
Hi, 
1. i think it's better not to have the change with "report-not-found-error". it's a bit too hacky..we need to find better ways to deal with this. (and TBH, i think it's ok to report not found error, since it's indeed an error from api).
2. it's a blocking operation in newLSInstance. Which keep check whether the listener is ready using describeListener with ARN at fixed period and with a maximum 5 minute timeout(maybe 1 minute?)

devkid:
> 1. i think it's better not to have the change with "report-not-found-error". it's a bit too hacky..we need to find better ways to deal with this. (and TBH, i think it's ok to report not found error, since it's indeed an error from api).

Can you suggest an alternative? Would `count-not-found-error` be better (so still report it, but don't increase the counter in the prometheus metrics)? I'm still not convinced that it makes sense to actually report those errors; the fact that there is an error in the first place is only due to limitation of the API not providing a function to verify if an entity exists without throwing an error.

devkid:
@M00nF1sh ping?

internal/aws/session.go
```diff
@@ -32,11 +33,14 @@ func NewSession(awsconfig *aws.Config, AWSDebug bool, mc metric.Collector) *sess
...
 	session.Handlers.Complete.PushFront(func(r *request.Request) {
 		if r.Error != nil {
-			mc.IncAPIErrorCount(prometheus.Labels{"service": r.ClientInfo.ServiceName, "operation": r.Operation.Name})
-			if AWSDebug {
-				glog.ErrorDepth(4, fmt.Sprintf("Failed request: %s/%s, Payload: %s, Error: %s", r.ClientInfo.ServiceName, r.Operation.Name, log.Prettify(r.Params), r.Error))
+			if value, ok := r.Context().Value("report-not-found-error").(bool); !notFoundRegex.MatchString(r.Error.Error()) && !ok || value {
+				mc.IncAPIErrorCount(prometheus.Labels{"service": r.ClientInfo.ServiceName, "operation": r.Operation.Name})
+				if AWSDebug {
+					glog.ErrorDepth(4, fmt.Sprintf("Failed request: %s/%s, Payload: %s, Error: %s", r.ClientInfo.ServiceName, r.Operation.Name, log.Prettify(r.Params), r.Error))
+				}
```
alejandrox1: This is an odd bit. Let's say that there really is an issue creating an ALB, the error wouldn't properly be sent out unless `AWSDebug` was true, which is not optimal - if something is broken or mis-configured then a user would want to know right away.

M00nF1sh:
> Jumping into the conversation: I don't think this is an optimal solution.
> An ALB not found can be a real error and this should be surfaced as soon as possible by default.
> Looking at #1008 the one big issue that stands out is the increase in the prometheus counter metric but that can be dealt with in prometheus.

nice point on deal it within prometheus :D we shouldn't hide these errors within controller in a tricky way.

alejandrox1:
Still thinking about this one.
Given that this is a controller, it is assumed that infrastructure may fail at some point so a non-zero amount of failures is okay.
It is only the case when failures start piling up that one should be alerted and take manual action.

If we were inclined to prevent the "not found" error from occurring before creating an ALB a more appropriate approach, I think, would be to somehow check that the Ingress resource is "new". But even that is really tricky.

devkid:
>  An ALB not found can be a real error and this should be surfaced as soon as possible by default.

But the error is **not** silently swallowed with this change. It is still handled and logged in the functions that call the helper functions if they expect the resources to already exist and they don't.

> Looking at #1008 the one big issue that stands out is the increase in the prometheus counter metric but that can be dealt with in prometheus.

Not really. In our dev environment, we sometimes create multiple ingresses at once (e.g. multiple deployments of a single application with a parameter sweep or so). With the current state, each ingress creation yields at least one increase in the error counter. This is indistinguishable from "real" errors.

internal/aws/session.go
devkid: Like I said below, if this is an actual error, it will surface in the function that actually initiated the API call. Also, the `if AWSDebug { ... }` was there before my change.

Issue #325, opened 2018-02-07
How to exclude nodes from specific target groups?


blytheaw:
I have a use case to isolate a workload on a specific node in my cluster. This one node does work that utilizes a lot of bandwidth, and I would like to directly send requests to this node rather than proxy through the other nodes in my cluster.

By default, the alb-ingress-controller adds all nodes to all target groups (excluding masters). I haven't been able to find a way to exclude particular nodes or control which target groups send traffic to which Kubernetes nodes. Is this possible?

colinmorelli:
Perhaps this is a separate issue, but it would be great if this controller supported the `alpha.service-controller.kubernetes.io/exclude-balancer` annotation that kubernetes uses by default to exclude certain nodes from being included in `LoadBalancer` services.

ewbankkit:
[Virtual Kubelet](https://github.com/virtual-kubelet/virtual-kubelet) is an example of a node that [sets](https://github.com/virtual-kubelet/virtual-kubelet/blob/8f6b4b9e209c2597e844ea60b5401804f48ed621/vkubelet/vkubelet.go#L151) `alpha.service-controller.kubernetes.io/exclude-balancer` to `true`.

Currently if I run a VK and then attempt to create an Ingress I get errors like:

```
  Warning  ERROR   56s   ingress-controller  Error registering targets to target group xyz-31823-HTTP-0008dc6: ValidationError: Instance ID 'virtual-kubelet' is not valid
           status code: 400, request id: 9167a633-5525-11e8-96c0-dbae4da6c542
```

bigkraig:
I've added support for `alpha.service-controller.kubernetes.io/exclude-balancer` in #403

bigkraig:
Please test with `1.0-alpha.9` and let me know how it goes

PR #1042, opened 2019-09-30
Allow node-labels annotation.

### LINKED ISSUE #580: Do not add all worker nodes to target group
> All workers are added to the ALB TargetGroup but only the workers that have the pod running locally will pass the healthchecks.

Hi again! I'm back to using `target-type: instance` because I simply cannot get amazon-vpc-cni to work reliably (ugh, super frustrated by that).

Can an annotation be added that says only add instances to the target group that have some label?

For example...
```
alb.ingress.kubernetes.io/node-labels: kops.k8s.io/instancegroup=web,foo=bar
```

We have hundreds of nodes, but only about 6 of them will have pods that the ingress is forwarding to.

Thanks!

shrinandj:
This commit allows the ingress object to specify an annotation about which nodes should
be in the target group. The annotation is of the format:

`alb.ingress.kubernetes.io/node-labels: kops.k8s.io/instancegroup=web,foo=bar`

The implementation checks the labels on the node against the key-value pairs specified in the
above annotation. If they match, the node is part of the target group. Otherwise not.

If the `alb.ingress.kubernetes.io/node-labels` annotation does not exist, the original
behavior stays; i.e. all nodes are selected.

Testing Done:

1. Verified that adding no annotation retains existing behavior.
2. Verified that adding a single key-value pair works. The correct set of nodes are selected.
3. Verified that adding multiple key-value pairs works as expected. The correct set of nodes are selected.
4. Verified that removing a key-value pair from the annotation works as expected. The appropriate nodes are removed from the target-group.
5. Verified that removing the entire annotation works as expected. All nodes get added to the target-group.

### DIFF
--- a/internal/ingress/controller/store/store.go
+++ b/internal/ingress/controller/store/store.go
@@ -62,6 +62,9 @@ type Storer interface {
 
 	// GetNodeInstanceID gets the instance id of node
 	GetNodeInstanceID(node *corev1.Node) (string, error)
+
+	// IsNodeSelected returns whether a node has the correct node labels so that it can be added to the target group.
+	IsNodeSelected(nodeLabelAnnotation string, node *corev1.Node) (bool, error)
 }
 
 // Informer defines the required SharedIndexInformers that interact with the API server.
@@ -82,6 +85,12 @@ type Lister struct {
 	Pod               PodLister
 	IngressAnnotation IngressAnnotationsLister
 	ServiceAnnotation ServiceAnnotationsLister
+}
+
+// NodeLabel stores the key and value of the label specified in the node.
+type NodeLabel struct {
+	Key   string
+	Value string
 }
 
 // NotExistsError is returned when an object does not exist in a local store.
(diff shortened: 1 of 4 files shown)


M00nF1sh:
thanks for making this change :D. API looks good to me :D

shrinandj:
`ServiceNodeExclusion` is already implemented in the aws-alb-ingress-controller, right?

https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/325
https://github.com/kubernetes-sigs/aws-alb-ingress-controller/pull/403/files

If the node has a label "alpha.service-controller.kubernetes.io/exclude-balancer", it will be removed from the target-group.

M00nF1sh:
> #325

Yes. it's still there: https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/internal/ingress/annotations/class/main.go#L49
=.= i just messed up in my mind. (the pending work i recorded in my mind is to change to rely on k8s node status instead of ec2 api call for filter out unhealthy nodes).

shrinandj:
How do I get the do-not-merge/invalid-commit-message label removed?

M00nF1sh:
@shrinandj i think you can remove the "Closes sentence" from commit message by `git commit --amend` and then force push to same branch :D

shrinandj:
I pushed a commit without the `Closes` keyword. Looks like the PR needs a new `lgtm` label now. 

Sorry for the back-n-forth :(.

internal/ingress/controller/store/store.go
```diff
@@ -301,6 +310,51 @@ func (s k8sStore) GetServiceEndpoints(key string) (*corev1.Endpoints, error) {
 	return s.listers.Endpoint.ByKey(key)
 }
 
+func (s *k8sStore) annotationToNodeLabelObj(nodeLabelAnnotation string) *[]NodeLabel {
```
M00nF1sh: 1. it should return []NodeLabel instead of *[]NodeLabel to ease GC pressure.  (minimal comment)

internal/ingress/controller/store/store.go
```diff
@@ -301,6 +310,51 @@ func (s k8sStore) GetServiceEndpoints(key string) (*corev1.Endpoints, error) {
...
+}
+
+func (s *k8sStore) compareNodeLabels(node *corev1.Node, nodeLabelsObj *[]NodeLabel) (bool, error) {
+	for _, nodeLabel := range *nodeLabelsObj {
+		val, ok := node.ObjectMeta.Labels[nodeLabel.Key]
+		if !ok {
+			continue
+		}
+
+		if val == nodeLabel.Value {
```
M00nF1sh: the current logic imposes that the nodeSelector is in OR condition(e.g. one label match is selected), which is counterintuitive and doesn't match kubernetes's label selector's behavior(AND condition)

internal/ingress/controller/store/store.go
```diff
@@ -301,6 +310,51 @@ func (s k8sStore) GetServiceEndpoints(key string) (*corev1.Endpoints, error) {
 	return s.listers.Endpoint.ByKey(key)
 }
 
+func (s *k8sStore) annotationToNodeLabelObj(nodeLabelAnnotation string) *[]NodeLabel {
+	var nodeLabels []NodeLabel
+	cleaned := strings.Replace(nodeLabelAnnotation, " ", "", -1)
```
M00nF1sh: 1. The type matches our stringMap type in [our doc](https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/ingress/annotation/).
I'd like to use existing logic to maintain same behavior among other annotations:
    1. support  for customize annotation prefix via [--alb-name-prefix](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/internal/ingress/controller/config/config.go#L75), which allows you to use another prefix instead of 'alb.ingress.kubernetes.io'
    2. maintain same behavior on how to parse annotations. e.g. 'a=b c=d' will be parsed differently under current implementation.

I'd suggest add an method with syntax `func LoadStringMapAnnotation(annotation string, value *map[string]string, annotations ...map[string]string) bool {` [here](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/internal/ingress/annotations/annotations.go)
Also, see [sample code](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/internal/ingress/annotations/tags/main.go#L41) here for implementation same behavior.

2. The currently solution requires the annotation to be set on ingress only. But ideally this should be allows to set on either ingress or service and the annotation on service takes priority. (since the backend(targetGroup) differs by service).
See sample code [here](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/b3de77759486af1bfb9d75960942723544437c06/internal/ingress/auth/auth.go#L101) on how to handle it

nustiueudinastea:
@shrinandj @M00nF1sh this PR was never merged right? Is there an alternative to the functionality implemented in this PR? It would be very useful.

M00nF1sh:
@nustiueudinastea 
Yeah..it's not merged due to rebase needed. @shrinandj I'll be happy to merge this into old master branch if you are still interested.
We also plan to support this in new main branch(aws-load-balancer-controller), tracked here: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1575. (the implementation is ready but we haven't exposed it via annotation)

shrinandj:
Merging in the old master branch would be great.

BTW, it wasn't closed because the rebase wasn't done. There were additional comments about having the annotation work for both `ingress` or `service` objects which is complicated.

PR #1067, opened 2019-11-08
fixup docs for load-balancer-attributes

tghaas:
There was a space missing between the key and value and this
made it unclear what was the key and what was the value.


M00nF1sh:
thanks for making this change
/ok-to-test
/approve

M00nF1sh:
/retest
looks like prow changed their go version again...

PR #1087, opened 2019-12-01
Load balancing algorithm for target groups support

lukeprentice202:
ALB's now support an additional algorithm, set at the target group level. This PR adds support for this.

Fixes https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/1082

https://aws.amazon.com/about-aws/whats-new/2019/11/application-load-balancer-now-supports-least-outstanding-requests-algorithm-for-load-balancing-requests/


lukeprentice202:
I signed it

PR #1088, opened 2019-12-02
Advanced routing/weighted loadbalancing support

M00nF1sh:
# Supports Advanced routing:  fix https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/909

We will add a new annotation called `alb.ingress.kubernetes.io/conditions.<use-annotation-service>` to let customer specify routing conditions in additional to original host+path condition.
```
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: simple-ingress-example
  annotations:
    alb.ingress.kubernetes.io/conditions.advanced-routing: '[{"Field": "http-request-method", "HttpRequestMethodConfig": {"Values" : ["POST"]}}, {"Field": "query-string", "QueryStringConfig": {"Values" : [{"Key": "user", "Value": "m00nf1sh"}]}}]'
spec:
  rules:
  - host: m00nf1sh.example.com
    http:
      paths:
      - path: /foo
        backend:
          serviceName: advanced-routing
          servicePort: use-annotation
```
The one-line json for 'advanced-routing' above expands to
```
[
   {
      "Field": "http-request-method",
      "HttpRequestMethodConfig": {
         "Values": [
            "POST"
         ]
      }
   },
   {
      "Field": "query-string",
      "QueryStringConfig": {
         "Values": [
            {
               "Key": "user",
               "Value": "m00nf1sh"
            }
         ]
      }
   }
]
```
By merging in original HOST+PATH condition in Ingress, the result routing condition matches an HTTP request of "POST m00nf1sh.example.com/foo?user=m00nf1sh". 

Note:
1. The conditions JSON configuration schema is an close match of ALB's condition API to allow us keep up with future ALB enhancements quickly: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-listenerrule-conditions.html
1. The original HTTP+PATH condition in Ingress will be merged into the conditions specified via annotation. E.g. If customer also specified HOST conditions via annotation, either the HOST in Ingress or HOST in annotation will be matched.  This is intended to honer the Ingress specification.
1. when magic string `use-annotation` is specified, the `alb.ingress.kubernetes.io/conditions.<use-annotation-service>` will be an optional annotation while `alb.ingress.kubernetes.io/actions.<use-annotation-service>` is an required annotation. both annotation can be used together.

#  Supports weighted loadBalancing: fix https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/1078

We extend our original `alb.ingress.kubernetes.io/actions.<use-annotation-service>` annotation to supported added [ForwardConfig](https://github.com/boto/botocore/blob/develop/botocore/data/elbv2/2015-12-01/service-2.json#L1697). Sample Ingress below:
```
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: simple-ingress-example
  annotations:
    alb.ingress.kubernetes.io/actions.weighted-routing: '{"Type":"forward","ForwardConfig":{"TargetGroupStickinessConfig":{"Enabled":"True","DurationSeconds":100},"TargetGroups":[{"Weight":20,"ServiceName":"Service1","ServicePort":"80"},{"Weight":80,"TargetGroupArn":"arn:aws:******/external-legecy-tg"}]}}'
spec:
  rules:
  - host: m00nf1sh.example.com
    http:
      paths:
      - path: /foo
        backend:
          serviceName: weighted-routing
          servicePort: use-annotation
```
The one-line json for 'weighted-routing' above expands to
```
{ 
   "Type":"forward",
   "ForwardConfig":{ 
      "TargetGroupStickinessConfig":{ 
         "Enabled":"True",
         "DurationSeconds":100
      },
      "TargetGroups":[ 
         { 
            "Weight":20,
            "ServiceName":"service1",
            "ServicePort":"80"
         },
         { 
            "Weight":80,
            "TargetGroupArn":"arn:aws:******/external-legecy-tg"
         }
      ]
   }
}
```
The above configuration result in HTTP request to "m00nf1sh.example.com/foo" been routed to service1:80 with 20% possibility and an external-provisioned target group("arn:aws:******/external-legecy-tg") with 80% possibility. The routing decision will be stick per client for 100 second.

Note:
1. The actions JSON configuration schema is a close match of ALB's action API to allow us keep up with future ALB enhancements quickly.
1. The TargetGroups within ForwardConfig have been extended to allow use both Kubernetes Service and TargetGroupArn. Customers can configure backend with Kubernetes Service only / TargetGroupArn only / mix of two(for migration use case)
1. **limitation**: the auth annotation on service if not honored if it's using use-annotation forward-action.It's intentionally not supported since the ALB's API can only configure auth per rule instead of per targetGroup(in weighted routing). So auth actions shouldn't be bound to services). Use cases like have different auth configuration based on path along with use advanced-routing/weighted-routing have to wait for ingressGroup(in V2).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/kubernetes-sigs/aws-alb-ingress-controller
 require (
 	github.com/appscode/jsonpatch v0.0.0-20190108182946-7c0e3b262f30 // indirect
 	github.com/aws/aws-k8s-tester/e2e/tester v0.0.0-20190907061006-260b0e114d90
-	github.com/aws/aws-sdk-go v1.23.21
+	github.com/aws/aws-sdk-go v1.27.3
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/go-logr/glogr v0.1.0
 	github.com/go-logr/logr v0.1.0 // indirect
@@ -31,7 +31,7 @@ require (
 	github.com/prometheus/common v0.4.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.4.0
-	github.com/ticketmaster/aws-sdk-go-cache v0.0.0-20180926195306-58922816129c
+	github.com/ticketmaster/aws-sdk-go-cache v0.0.0-20180926195306-58922816129c // indirect
 	golang.org/x/oauth2 v0.0.0-20190212230446-3e8b2be13635 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	k8s.io/api v0.0.0-20181213150558-05914d821849
--- a/internal/alb/ls/rules.go
+++ b/internal/alb/ls/rules.go
@@ -3,9 +3,11 @@ package ls
 import (
 	"context"
 	"fmt"
-	"reflect"
-	"sort"
 	"strconv"
+
+	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/annotations/conditions"
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/intstr"
 
 	"github.com/aws/aws-sdk-go/service/elbv2"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/alb/tg"
@@ -135,7 +137,7 @@ func (c *rulesController) getDesiredRules(ctx context.Context, listener *elbv2.L
 			if err != nil {
 				return nil, err
 			}
-			elbConditions := buildConditions(ctx, ingressRule, path)
+			elbConditions := buildConditions(ctx, ingressAnnos, ingressRule, path)
 			elbRule := elbv2.Rule{
 				IsDefault:  aws.Bool(false),
 				Priority:   aws.String(strconv.Itoa(nextPriority)),
@@ -172,22 +174,26 @@ func (c *rulesController) getCurrentRules(ctx context.Context, listenerArn strin
 
 // buildActions will build listener rule actions for specific authCfg and backend
 func buildActions(ctx context.Context, authCfg auth.Config, ingressAnnos *annotations.Ingress, backend extensions.IngressBackend, tgGroup tg.TargetGroupGroup) ([]*elbv2.Action, error) {
-	var actions []*elbv2.Action
+	var elbActions []*elbv2.Action
 
 	// Handle auth actions
 	authAction := buildAuthAction(ctx, authCfg)
 	if authAction != nil {
-		actions = append(actions, authAction)
+		elbActions = append(elbActions, authAction)
 	}
 
 	// Handle backend actions
 	if action.Use(backend.ServicePort.String()) {
 		// backend is based on annotation
-		backendAction, err := ingressAnnos.Action.GetAction(backend.ServiceName)
+		annotationAction, err := ingressAnnos.Action.GetAction(backend.ServiceName)
 		if err != nil {
 			return nil, err
 		}
-		actions = append(actions, &backendAction)
+		annotationELBAction, err := buildAnnotationAction(ctx, annotationAction, tgGroup)
+		if err != nil {
+			return nil, err
+		}
+		elbActions = append(elbActions, annotationELBAction)
 	} else {
 		// backend is based on service
 		targetGroup, ok := tgGroup.TGByBackend[backend]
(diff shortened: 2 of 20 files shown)


internal/alb/ls/rules_comp.go
```diff
@@ -0,0 +1,185 @@
...
+			}
+		}
+		if !found {
+			return false
+		}
+	}
+	return true
+}
+
+func sliceMatches(listA interface{}, listB interface{}, comparator func(interface{}, interface{}) bool) bool {
```
mogren: This function looks fairly slow, how often are the rules checked?

internal/alb/tg/targetgroup_group.go
```diff
@@ -47,18 +51,20 @@ var _ GroupController = (*defaultGroupController)(nil)
...
 	tgController Controller
 }
 
 func (controller *defaultGroupController) Reconcile(ctx context.Context, ingress *extensions.Ingress) (TargetGroupGroup, error) {
 	tgByBackend := make(map[extensions.IngressBackend]TargetGroup)
-	var err error
-	for _, backend := range controller.extractIngressBackends(ingress) {
-		if action.Use(backend.ServicePort.String()) {
-			continue
-		}
```
mogren: Ah, nice moving this check into `extractTargetGroupBackends`

internal/ingress/annotations/action/main_test.go
```diff
@@ -15,45 +16,163 @@ type mockBackend struct {
...
+	//ai, err := NewParser(mockBackend{}).Parse(ing)
+	//if err != nil {
+	//	t.Error(err)
+	//	return
+	//}
+	//
+	//a, ok := ai.(*Config)
+	//if !ok {
+	//	t.Errorf("expected a Config type")
+	//}
```
mogren: Nit: Remove?

internal/ingress/annotations/action/types.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+	case elbv2.ActionTypeEnumRedirect:
+		if a.RedirectConfig == nil {
+			return errors.New("missing RedirectConfig")
+		}
+		if err := a.RedirectConfig.validate(); err != nil {
+			return errors.Wrap(err, "invalid RedirectConfig")
+		}
+	case elbv2.ActionTypeEnumForward:
+		if (a.TargetGroupArn != nil) == (a.ForwardConfig != nil) {
```
mogren: xor in go always look weird to me. 😃

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/kubernetes-sigs/aws-alb-ingress-controller
 require (
 	github.com/appscode/jsonpatch v0.0.0-20190108182946-7c0e3b262f30 // indirect
 	github.com/aws/aws-k8s-tester/e2e/tester v0.0.0-20190907061006-260b0e114d90
-	github.com/aws/aws-sdk-go v1.23.21
+	github.com/aws/aws-sdk-go v1.25.43
```
mogren: What in v1.25.43 is it that we need?

internal/alb/ls/rules.go
```diff
@@ -291,70 +460,65 @@ func rulesChangeSets(current, desired []elbv2.Rule) (add []elbv2.Rule, modify []
...
 		}
-		if host != nil && aws.StringValue(rc.Host) != aws.StringValue(host) && aws.StringValue(rc.Host) != "#{host}" {
-			return false
+
+		if len(hosts) != 0 && aws.StringValue(rc.Host) != "#{host}" {
+			hostMatches := false
+			for _, host := range hosts {
+				if aws.StringValue(rc.Host) == host {
+					hostMatches = true
+					break
```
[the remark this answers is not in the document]
M00nF1sh: added comments. the logic is if none of the hostMatches, it's won't be redirect loop(just from this function's purpose perspective).

The createsRedirectLoop is a hacky way to allow HTTP->HTTPS redirection, it should be deprecated and replace with an more specific annotation.

internal/alb/ls/rules.go
```diff
@@ -291,70 +460,65 @@ func rulesChangeSets(current, desired []elbv2.Rule) (add []elbv2.Rule, modify []
...
-			if aws.StringValue(c.Field) == "host-header" {
-				host = c.Values[0]
-			}
-			if aws.StringValue(c.Field) == "path-pattern" {
-				path = c.Values[0]
+			switch aws.StringValue(c.Field) {
+			case conditions.FieldHostHeader:
+				hosts = append(hosts, aws.StringValueSlice(c.HostHeaderConfig.Values)...)
+			case conditions.FieldPathPattern:
+				paths = append(paths, aws.StringValueSlice(c.PathPatternConfig.Values)...)
```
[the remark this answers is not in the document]
M00nF1sh: this is adding host/path conditions from annotation(in addition to host/path condition from ingress itself).
Added more test cases  :D

internal/alb/ls/rules_comp.go
M00nF1sh: the rule check currently triggers on pod relocate events(maybe batched) in v1 version.
This won't be bottleneck since the rules are pretty small. 
In fact, the code logic is much more clean and can be even faster than other comparsion method like sort+compare/hash+compare, given the two list is pretty small and highly likely equal.

go.mod
M00nF1sh: the weighted-routing API(unfortunately, imds v2 is added in same revision). I updated it to a latest version.

internal/alb/ls/rules_comp.go
mogren: Makes sense, thanks for the clarification.

internal/ingress/annotations/conditions/types.go
```diff
@@ -0,0 +1,255 @@
...
+	FieldPathPattern       = "path-pattern"
+	FieldHTTPHeader        = "http-header"
+	FieldHTTPRequestMethod = "http-request-method"
+	FieldQueryString       = "query-string"
+	FieldSourceIP          = "source-ip"
+)
+
+// Information about a host header condition.
+type HostHeaderConditionConfig struct {
+	// One or more host names. The maximum size of each name is 128 characters.
```
mogren: Nit: Should it say "maximum size" or "maximum length"?

internal/ingress/annotations/conditions/types.go
```diff
@@ -0,0 +1,255 @@
...
+func (c *PathPatternConditionConfig) validate() error {
+	if len(c.Values) == 0 {
+		return errors.New("Values cannot be empty")
+	}
+	return nil
+}
+
+// Information about a key/value pair.
+type QueryStringKeyValuePair struct {
+	// The key. You can omit the key.
```
mogren: What does it mean if you omit the key?
`// The key, optional. If omitted, XXXX???`

internal/ingress/annotations/conditions/types.go
```diff
@@ -0,0 +1,255 @@
...
+//
+// You can use this condition to route based on the IP address of the source
+// that connects to the load balancer. If a client is behind a proxy, this is
+// the IP address of the proxy not the IP address of the client.
+type SourceIpConditionConfig struct {
+	// One or more source IP addresses, in CIDR format. You can use both IPv4 and
+	// IPv6 addresses. Wildcards are not supported.
+	//
+	// If you specify multiple addresses, the condition is satisfied if the source
+	// IP address of the request matches one of the CIDR blocks. This condition
```
mogren: Nit: `// IP address of the request matches any of the CIDR blocks.`

internal/ingress/annotations/conditions/types.go
M00nF1sh: Aha, both applies a guess :D.
In fact, the struct and comments is directly copied from AWSSDKGo. (instead of using it directly, i create my own copy to 
1. avoid been broken when upgrades SDK
2. add extensions, like allow ServiceName/ServicePort instead of TargetGroupArn in forwardConfig
)

internal/ingress/annotations/conditions/types.go
M00nF1sh: it should be any param equal to valueXXX. (however, my tests didn't seems so, looks like an bug in ALB side).
The comments is copied directly from awsSDKGO :D

internal/ingress/annotations/conditions/types.go
M00nF1sh: The comments is copied directly from awsSDKGO :D

internal/ingress/annotations/action/types.go
M00nF1sh: 🤣

tschmittni:
awesome, thank you for the work! Could you cut us a new release with this feature?

PR #1095, opened 2019-12-06
update to latest version

8398a7:
Fargate was supported by EKS and I wanted to use it.
At that time, since an old version of information exists and is confused, a PR is issued.

Also, I was worried that the yaml versions of some documents are old, but should I update them too?

For example, deployment versions.
Ingress was also an old notation, but if you modify it, it may affect rbac.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@
 
 all: container
 
-TAG?=v1.1.3
+TAG?=v1.1.4
 PREFIX?=amazon/aws-alb-ingress-controller
 ARCH?=amd64
 OS?=linux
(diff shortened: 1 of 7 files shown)


8398a7:
Changing the makefile tag may have a big impact.
I would like to remove that range if necessary.

However, it was confusing that it was displayed as v1.1.3 when I moved with the v1.1.4 image and looked at the log.

M00nF1sh:
thanks for making this PR.
I think we should upgrade the deployments to like apps/v1. Will do in follow up PRs.

8398a7:
Thank you for the review.
I updated deployments to apps/v1 with #1096 .

PR #1104, opened 2019-12-12
docs: update example for ExternalDNS

vutny:
* fix `kubectl logs` example command since `external-dns` being deployed in
  `default` namespace with provided sample manifest
* update the container args example
* update the logs output example and add the recordset creation log


vutny:
I signed it

vutny:
Thank you, @M00nF1sh !

PR #1109, opened 2019-12-19
Rewrite GetClusterSubnets() using EC2 specific API

lanSs:
This PR moves `rgt.GetClusterSubnets()` to `ec2.GetClusterSubnets()` and updates the `GetClusterSubnets()` function to use `EC2 specific API` instead of `Resource Groups Tagging API`.

Since there is no VPC endpoint for  `Resource Groups Tagging API`, this change will make the controller be able to run inside an internal VPC subnet without `internet gateway` or `nat gateway`.
Please let me know if I need to update anything.

Thanks,

### DIFF
--- a/internal/aws/ec2.go
+++ b/internal/aws/ec2.go
@@ -44,6 +44,9 @@ type EC2API interface {
 	// GetSecurityGroupsByName retrieves securityGroups by securityGroupName(SecurityGroup names within vpc are unique)
 	GetSecurityGroupsByName(context.Context, []string) ([]*ec2.SecurityGroup, error)
 
+	// GetClusterSubnets retrieves the subnets associated with the cluster, by matching tags
+	GetClusterSubnets(string) ([]*ec2.Subnet, error)
+
 	// DeleteSecurityGroupByID delete securityGroup by securityGroupID
 	DeleteSecurityGroupByID(context.Context, string) error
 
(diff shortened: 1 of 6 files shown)


internal/aws/ec2.go
```diff
@@ -272,6 +303,28 @@ func (c *Cloud) describeSecurityGroupsHelper(params *ec2.DescribeSecurityGroupsI
...
+		result = append(result, output.Subnets...)
+		return true
+	})
+
+	return result, err
+}
+
+// hasClusterTagHelper is a helper to find if a Subnet has the correct ELB tag
+func (c *Cloud) hasClusterTagHelper(params []*ec2.Tag) bool {
+	for _, tag := range params {
```
M00nF1sh: Hi, this logic should be:
1. If it's an internet-facing LB, looks for `kubernetes.io/role/elb` tag.
2. If it's an internal LB, looks for `kubernetes.io/role/internal-elb` tag.
Which aligns  with k8s, and provide an way to denote public/private subnet

lanSs: Thank you for the review,
I have updated the code with the new logic, I also updated the test case to handle this.

internal/aws/ec2.go
```diff
@@ -164,6 +167,34 @@ func (c *Cloud) GetSubnetsByNameOrID(ctx context.Context, nameOrIDs []string) (s
...
+	if err != nil {
+		return nil, err
+	}
+
+	if subnets == nil {
+		return nil, nil
+	}
+
+	for _, s := range subnets {
+		if c.hasClusterTagHelper(s.Tags, tagSubnetType) {
```
M00nF1sh: I think if we put this as a tag:filter alone with existing `tag:kubernetes.io/cluster/` filter, it will be more clear and simple :D

M00nF1sh:
@lanSs 
This overall looks good to me and should works.
Just one nit before merge, would you help test change to 
```
in := &ec2.DescribeSubnetsInput{Filters: []*ec2.Filter{
		{
			Name:   aws.String("tag:kubernetes.io/cluster/" + c.clusterName),
			Values: aws.StringSlice([]string{"owned", "shared"}),
		},
                {
			Name:   aws.String("tag: + tagSubnetType),
			Values: aws.StringSlice([]string{"", "1"}),
		},
	}}
```
I believe it should works and more efficient if you have a lot subnets in your VPC. (we had customers with 100+ subnets) :D

lanSs:
@M00nF1sh 
Thank you very much for spending time reviewing it. Your suggestion totally make sense and I have updated the code and the test suite.

Sorry, between my first version and your last review, I should have revisited the `hasClusterTagHelper`, having the `tagSubnetType` as parameter make the whole functions easier and we don't need the helper function anymore :)

PR #1122, opened 2020-01-06
Docs: Duplicate port warning.

nitrag:
Notate that duplicate ports are not allowed.

### DIFF
--- a/docs/guide/ingress/annotation.md
+++ b/docs/guide/ingress/annotation.md
@@ -57,6 +57,9 @@ Traffic Listening can be controlled with following annotations:
         ```
         alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS": 443}, {"HTTP": 8080}, {"HTTPS": 8443}]'
         ```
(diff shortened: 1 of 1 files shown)


alejandrox1:
Hi @nitrag thank you for raising this issue.
before anything else: you should sign the CLA so you can contribute to Kubernetes projects :smile: 

follow up question: why did you consider the warning was necessary?

M00nF1sh:
Hi, @nitrag 
Thanks for contributing.
I think Jorge's point is good...i don't think this worth an warning as well :D...

nitrag:
My use case was forwarding **either** HTTP or HTTPS, which is an AWS limitation.

Without any log errors or a note in the docs I didn't realize that was the problem with my configuration and only realized my error when trying to make the change in the AWS dashboard and it showed me the error there. I figured adding the warning could save someone 1-2 hours troubleshooting in the future.

bhagwat070919:
Hi,

I like the idea of providing enough information about the failure and saving developer's time.

PR #1126, opened 2020-01-13
Add support for enabling AWS Shield Advanced protection

hhamalai:
When subscribed to AWS Shield Advanced service, the Shield protection must be turned on for each Application Load Balancer resource. While AWS offers automation to do this via services like Firewall Manager, their use cases are limited by their constraints (e.g., impossible to utilize rate-based WAF rules [1]).

The proposed change adds an annotation for the ingress resource, which can is used to turn on the Shield Advanced protection for the load balancers created by aws-alb-ingress-controller. 

[1]: https://docs.aws.amazon.com/waf/latest/developerguide/ddos-get-started-rate-based-rules.html


M00nF1sh:
Wow, thanks so much for contributing! 
/ok-to-test

internal/aws/cloud.go
```diff
@@ -77,6 +81,7 @@ func New(cfg CloudConfig, clusterName string, mc metric.Collector) (CloudAPI, er
 		ec2.New(awsSession, regionCfg),
 		elbv2.New(awsSession, regionCfg),
 		iam.New(awsSession, regionCfg),
+		shield.New(awsSession, &aws.Config{Region: aws.String("us-east-1")}),
```
mogren: Any reason this should be hard-coded to `"us-east-1"` and not just be `regionCfg`?

mogren: Ah, thanks! I was looking for that link in the documentation, since I saw that the service itself was available in a lot more regions.

internal/alb/lb/loadbalancer.go
```diff
@@ -255,6 +259,34 @@ func (controller *defaultController) reconcileLBInstance(ctx context.Context, in
...
+func (controller *defaultController) reconcileShield(ctx context.Context, lbArn string, enableProtection bool) error {
+	available, err := controller.cloud.ShieldAvailable(ctx)
+	if err != nil {
+		return fmt.Errorf("unable to get shield advanced subscription state %v", err)
+	}
+	if enableProtection && !available {
+		return fmt.Errorf("unable to create shield advanced protection for loadBalancer %v, shield advanced subscription is not active", lbArn)
+	}
+
+	protection, _ := controller.cloud.GetProtection(ctx, aws.String(lbArn))
```
n1koo: `isProtectionEnabled`? Also if this errored out you will enabled protection which might not be wanted in a transient error for get?

internal/aws/shield.go
```diff
@@ -0,0 +1,53 @@
...
+	GetSubscriptionStatus(ctx context.Context) (*shield.GetSubscriptionStateOutput, error)
+	GetProtection(ctx context.Context, resourceArn *string) (*shield.Protection, error)
+	CreateProtection(ctx context.Context, resourceArn *string) (*shield.CreateProtectionOutput, error)
+	DeleteProtection(ctx context.Context, protectionID *string) (*shield.DeleteProtectionOutput, error)
+}
+
+func (c *Cloud) ShieldAvailable(ctx context.Context) (bool, error) {
+	status, err := c.GetSubscriptionStatus(ctx)
+	if err != nil {
+		return false, err
```
n1koo: We don't check error type here. In happy case it seems to be not found, but what if its something else?

internal/aws/shield_test.go
```diff
@@ -0,0 +1,177 @@
...
+	"errors"
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/shield"
+	"github.com/kubernetes-sigs/aws-alb-ingress-controller/mocks"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCloud_GetSubscriptionStatus(t *testing.T) {
```
n1koo: Tests in general look good, but the things I pointed out earlier come from not handling unexpected errors. Might be worth to add a transient error to each scenario just to see flow goes as expected still

internal/alb/lb/loadbalancer.go
hhamalai: protection is either existing protection object or nil if there are currently no protection for the given load balancer resource. Added check for the not-found path, and error checks for failures.

internal/aws/shield.go
hhamalai: The only service specific GetSubscriptionState error case seems to be InternalErrorException. Added extra test cases.

hhamalai:
Hi, can @bigkraig give any feedback how to take this proposed change further? Should try to drag more review feedback or assign this into some person to review?

M00nF1sh:
I'll test this out today

hhamalai:
Thanks for the great feedback, I'm happy to implement your suggestions and ping you once ready.

hhamalai:
@M00nF1sh I've implemented & tested your proposed changes. Regarding the last one, it's now as you suggested, i.e. when the annotation is missing no Shield Advanced API calls are made. This means, as you noticed, that when annotation is absent, when it previously was present with shield protection enabled, the protection is left present. 

Other than leaving it as is, one option would be to go through the current protections in the cache (no API calls made before and will be made if the cache is empty). If there are cache items, then these would be tried to remove. Few issues with this:
1) Cache expiration creates a requirement for another kind of cache
2) The cache usage to track protections creates persistence, if alb-ingress-controller pod is destroyed the cache is lost that the protections are left active as orphans.

hhamalai:
@M00nF1sh Any comments on the suggested changes / how to push this forward?

M00nF1sh:
> @M00nF1sh Any comments on the suggested changes / how to push this forward?

Thanks for making these changes, i think it's good enough to merge for now.
For the annotation remove issue, i think we should train Cx to explicitly disable shield instead of remove annotations, maybe via docs.

geofflancaster:
Does this work for NLBs as well?

hhamalai:
> Does this work for NLBs as well?

@geofflancaster , aws-alb-ingress-controller project is about ALBs, not NLBs. If you need NLBs with Shield Advanced enabled, you can use Elastic IPs, enable Shield Advanced protection on those IPs, and associate the IPs with your NLB.

christopher-wong:
Now that the load balancer controller supports NLBs, is there an NLB annotation to enable Shield Advanced?

hhamalai:
> Now that the load balancer controller supports NLBs, is there an NLB annotation to enable Shield Advanced?

AFAIK Shield Advanced cannot be enabled directly on NLB, but  for Elastic IPs attached to NLB instance with
`service.beta.kubernetes.io/aws-load-balancer-eip-allocations` annotation. These EIPs are not allocated by load balancer controller at the moment, so assigning shield protections to these EIPs should likely also be externally managed.

abebars:
@hhamalai Since the Controller allows you to pass these EIPs as an option parameter, Can we have `service.beta.kubernetes.io/aws-load-balancer-eip-allocations` as a prerequisite to `service.beta.kubernetes.io/shield-advanced-protection:  'true' ` and in that case we can make the correlation between the resources?

PR #1128, opened 2020-01-14
adding cache around WAF API usage

M00nF1sh:
Adding caches for WAF API usage.
The WAF API have a 1 TPS limitation, adding client-side cache to mitigate it.
The default cache timeout is 10 minutes, as long as user didn't manually adjust webACL on ALB from console/CLI, there won't be any cache issue.


Jeffwan:
Seems tests have been fixed. 

/lgtm

PR #1129, opened 2020-01-14
ignore fargate nodes for instance target type

### LINKED ISSUE #1113: Mixed alb.ingress.kubernetes.io/target-type ingresses with fargate causes spurious errors.
When you have a cluster with both EC2 and fargate, and you are using ingresses that have `alb.ingress.kubernetes.io/target-type: instance' you will get these errors in the log:

```
E1223 21:38:11.141147       1 controller.go:217] kubebuilder/controller "msg"="Reconciler error" "error"="failed to reconcile targetGroups due to failed to reconcile targetGroup targets due to Unable to DescribeInstanceStatus on fargate-ip-10-20-206-16.ec2.internal: InvalidInstanceID.Malformed: Invalid id: \"fargate-ip-10-20-206-16.ec2.internal\"\n\tstatus code: 400, request id: f0f2c7d0-9fbf-4a9b-a7fd-3ad76e288cdf"  "controller"="alb-ingress-controller" "request"={"Namespace":"keystone","Name":"public"}
```

Both ALBs seem to work fine (though I haven't tested what happens if you add another EC2 instance).  It would be nice for the controller to ignore the fargate nodes when looking at 'instance' ALBs.

M00nF1sh:
Ignore fargate nodes when using instance target-type.
This allows to use instance target-type under mixed setup(both EC2 pods & fargate pods)
Fixes #1113


casret:
Thanks!

Issue #1100, opened 2019-12-11
Removal of AWS Cache renders the controller un-usable on large clusters


jescarri:
Controller Version: v1.1.3
Cluster Size: 240+ Nodes.
Ec2 instances get replaced fairly regular, (within hours).
Alb ingresses count: 300+

This commit: https://github.com/kubernetes-sigs/aws-alb-ingress-controller/commit/b5e33240ef6fe417fd9c09e43e87b0f2bf54f868#diff-4ac4ee94afffa06e29cb914abb118917 

Removal of cache made the alb-ingress-controller un-usable, because every time a new instance comes up or whenever a tg update is required, for each instance * each tg a few ec2Describe calls are required.

We get throttled, and reconcile can take up to 30+ Minutes.

No new updates are processed or new ingresses added.

I have an internal fork with the cache added and it's working as expected.

Is there a way to re-enable cache or enable it via a flag?, I can submit a PR for that.
Thanks!

gaganapplatix:
Basically the biggest culprit is the call IsNodeHealthy, for every instance addition to a target, it does a GetInstanceStatus for all targets on that TG. When we run a cluster with 450 ingress with 200 nodes the 200 calls in IsNodeHealthy cause a 10 second delay to reconcile each ingress which adds up to 4500 seconds just to do instance adds. Since each ingress is sequentially reconciled. The solution is of course a cache for the instance health since this doesn't change that much. I mentioned in the issue which lead to the cache removal that we have PR'ed into the cache code stuff that removes the go routine leaks (there may be other things necessary) but yeah two things needed are:

1) AWS caching
2) More parallel reconciles

M00nF1sh:
@gaganapplatix 
I believe the IsNodeHealthy should be replaced with k8s's node ready status instead of making AWS API calls.

gaganapplatix:
Yeah, that would be reasonable, since that moves the work of keeping if at target node is healthy or not up to kubernetes control plane (kubelet + NodeLifecycleController) which is appropriate since the decision to add a Node as target only happens until after Node object shows up in kubernetes and is only removed on removal of Node from kubernetes, the healthy check should be left to Node Object (kubernetes control plane as well). Also IsNodeHealthy is only called on endpoint.Resolve which I believe only called from

[...]

PR #1154, opened 2020-02-06
Add back aws api cache

jescarri:
This commits reverts changes made on commit b5e33240ef6fe417fd9c09e43e87b0f2bf54f868
but gives the users the ability to enable the cache and set the cache TTL, by default
is turned off.

Related to: #1100


jescarri:
With cache enabled:

image

image


With cache disabled, no metrics for cache shown.

Issue #853, opened 2019-02-15
HTTPS only rules


JakubJecminek:
Hello,
would it be possible to specify that certain rules should be crated only for https listener? Currently all rules are propagated to both http and https listener. And since ingress controller is creating rules in order in witch they are defined in ingress yaml object, you can potentially make mistake by specifying rule some service before actual rule for http -> https redirect:

```yaml
  "spec": {
    "rules": [
      {
        "host": "test.mydomain.com",
        "http": {
          "paths": [
            {
              "path": "/my-app/*",
              "backend": {
                "serviceName": "my-app",
                "servicePort": 80
              }
            },
            {
        "http": {
          "paths": [
            {
              "path": "/*",
              "backend": {
                "serviceName": "redirect",
                "servicePort": "use-annotation"
              }
            }
          ]
        }
      },
          ]
        }
      }]}
```

Thus alb rule for redirect will be created as second and my-app can be accessed both through http and https. 

Optimal situation would be that if redirect rule is present in configuration then all other rules are created only in https listener. Or there could be some new annotation that would specify if rule is supposed to be applied to http/https listener or both.

With best regards
Jakub

hluchej:
+1

One problem is the wrong order of redirect rule which is definitely a security risk and error-prone.

The other problem is "a review". A security guy looks at ALB configuration and HTTP listener rules which are supposed to do just redirect. Instead, it shows a ton of rules which are meant only for HTTPS listener - that triggers some worry there.

Feels like the ingress controller does more "magic" than would be necessary. The AWS Console UX considers the listeners as separate while the ingress updates them "as one".

M00nF1sh:
Hi, the ingress group i'm working on can solve this problem like below. For such cases, you can create two seperate ingress like below:
```
kind: Ingress
  annotations:
    kubernetes.io/ingress.class: alb
    alb.ingress.kubernetes.io/group: my-app
    alb.ingress.kubernetes.io/listen-ports: 'HTTPS:443'
    alb.ingress.kubernetes.io/scheme: internet-facing
spec:
  rules:
    - http:
        paths:
          - path: /https-only
            backend:
              serviceName: echoserver
              servicePort: 443
---
kind: Ingress
  annotations:
    kubernetes.io/ingress.class: alb
    alb.ingress.kubernetes.io/group: my-app
     alb.ingress.kubernetes.io/listen-ports: 'HTTP:80'
[...]

PR #1162, opened 2020-02-11
Detect unconditional redirects and ignore any rules defined afterwards

tomfotherby:
(_this is my first PR for GoLang code and for Kubernetes-related projects, so bear with me if I make mistakes or faux pas_)

ALBs only support 100 rules and we have hit the limit in our project -  we are looking for a way to reduce rules and have identified that the aws alb controller is creating some unneeded ALB rules.

We use a annotation to redirect port 80 requests to port 443. This is the annotation documentation: https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/docs/guide/tasks/ssl_redirect.md

If a unconditional redirect rule is present in the configuration then all following rules within the same listener are not needed. This PR changes things so they are skipped thereby conserving rules.

For the biggest benefit, it is best to define the redirect rule first in the config. Because only subsequent rules are skipped. This is in line with the ssl-redirect annotation which [says](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/docs/guide/tasks/ssl_redirect.md):
> the ssl-redirect action must be be first rule(which will be evaluated first by ALB)

---

I have tested on our project, with success. But please let me know if I can make any improvements.

If this is not something that fits with the direction of the project, no problem, please close this PR. I only created it in case it's useful - I'm 100% happy to close if not the case.

---

Relvant GitHub issue https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/853#issuecomment-578547468

### DIFF
--- a/internal/alb/ls/rules.go
+++ b/internal/alb/ls/rules.go
@@ -128,7 +128,13 @@ func (c *rulesController) getDesiredRules(ctx context.Context, listener *elbv2.L
 			continue
 		}
 
+		seenUnconditionalRedirect := false
+
 		for _, path := range ingressRule.HTTP.Paths {
+			if seenUnconditionalRedirect {
+				// Ignore rules that follow a unconditional redirect, they are moot
+				continue
+			}
 			authCfg, err := c.authModule.NewConfig(ctx, ingress, path.Backend, aws.StringValue(listener.Protocol))
 			if err != nil {
 				return nil, err
@@ -146,6 +152,8 @@ func (c *rulesController) getDesiredRules(ctx context.Context, listener *elbv2.L
 			}
 			if createsRedirectLoop(listener, elbRule) {
 				continue
+			} else if isUnconditionalRedirect(listener, elbRule) {
+				seenUnconditionalRedirect = true
 			}
 			output = append(output, elbRule)
 			nextPriority++
@@ -525,6 +533,45 @@ func createsRedirectLoop(listener *elbv2.Listener, r elbv2.Rule) bool {
 		if aws.StringValue(rc.Protocol) != "#{protocol}" && aws.StringValue(rc.Protocol) != aws.StringValue(listener.Protocol) {
 			return false
 		}
+		return true
+	}
+	return false
+}
+
+// isUnconditionalRedirect checks whether specified rule always redirects
+// We consider the rule is a unconditional redirect if
+// 1) The Path condition is nil, or at least one Path condition is /*
+// 2) All other rule conditions are nil (ignoring the Host condition).
+// 3) RedirectConfig is not nil.
+func isUnconditionalRedirect(listener *elbv2.Listener, r elbv2.Rule) bool {
+	for _, action := range r.Actions {
+		rc := action.RedirectConfig
+		if rc == nil {
+			continue
+		}
+
+		var paths []string
+		for _, c := range r.Conditions {
+			switch aws.StringValue(c.Field) {
+			case conditions.FieldPathPattern:
+				paths = append(paths, aws.StringValueSlice(c.PathPatternConfig.Values)...)
+			case conditions.FieldHTTPRequestMethod, conditions.FieldSourceIP, conditions.FieldHTTPHeader, conditions.FieldQueryString:
+				// If there are any conditions, then the redirect is not unconditional
+				return false
+			}
+		}
+
+		if len(paths) != 0 {
+			// ALB path conditions are ORed, so if any of them are a wildcard, the redirect is unconditional
+			for _, path := range paths {
+				if path == "/*" {
+					return true
+				}
+			}
+			// The redirect isn't unconditional if none of the path conditions are a wildcard
+			return false
+		}
+
 		return true
 	}
 	return false
--- a/internal/alb/ls/rules_test.go
+++ b/internal/alb/ls/rules_test.go
@@ -2878,6 +2878,200 @@ func Test_createsRedirectLoop(t *testing.T) {
 	}
 }
 
+func Test_isUnconditionalRedirect(t *testing.T) {
+	for _, tc := range []struct {
+		name     string
+		listener elbv2.Listener
+		rule     elbv2.Rule
+
+		expected bool
+	}{
+		{
+			name:     "No RedirectConfig",
+			listener: elbv2.Listener{Protocol: aws.String("HTTP"), Port: aws.Int64(80)},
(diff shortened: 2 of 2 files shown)


internal/alb/ls/rules.go
```diff
@@ -529,3 +537,53 @@ func createsRedirectLoop(listener *elbv2.Listener, r elbv2.Rule) bool {
...
+            case conditions.FieldPathPattern:
+                paths = append(paths, aws.StringValueSlice(c.PathPatternConfig.Values)...)
+            case conditions.FieldHTTPRequestMethod:
+                return false
+            case conditions.FieldSourceIP:
+                return false
+            case conditions.FieldHTTPHeader:
+                return false
+            case conditions.FieldQueryString:
+                return false
```
mogren: In go you could use a comma separated list for this:
```go
case conditions.FieldHTTPRequestMethod, conditions.FieldSourceIP, conditions.FieldHTTPHeader, conditions.FieldQueryString:
    return false
```

tomfotherby: Thank you @mogren - I pushed a update with your suggestion included.

tomfotherby:
We have been running a ALB ingress controller that uses the code in this PR at https://peopleperhour.com in production successfully. It freed up > 40 ALB rules for us. The ALB gets about 110 requests per second:

Selection_311

internal/alb/ls/rules.go
```diff
@@ -529,3 +537,48 @@ func createsRedirectLoop(listener *elbv2.Listener, r elbv2.Rule) bool {
...
+                // If there are any conditions, then the redirect is not unconditional
+                return false
+            }
+        }
+
+        if len(paths) == 0 {
+            return true
+        }
+        if len(paths) != 0 {
+            wildcardPathMatch := false
```
M00nF1sh: can be simplified to  🤣 
```
for _, path := range paths {
    if path == "/*" {
          wildcardPathMatch = true
          return true
    }
 }
```

M00nF1sh:
would you help fix the lint..i'll clean up some vpcs in our test account to make e2e pass :D

tomfotherby:
> would you help fix the lint..

Yes, I'll fix this. If I have understood it, I should have used `gofmt -w yourcode.go`. Leason learnt.

internal/alb/ls/rules.go
tomfotherby: How did I miss that! Thanks! I'll fix.

tomfotherby:
I pushed a fix which I hope will pass the linter :crossed_fingers: . I also added more unit tests to gain more confidence the code works as expected and refactored given the suggestion in https://github.com/kubernetes-sigs/aws-alb-ingress-controller/pull/1162#discussion_r379190396

M00nF1sh:
thanks 🤣 
/approve
/lgtm

tdmalone:
Thanks so much for this @tomfotherby and for merging it @M00nF1sh. We're really keen to use this and just wondering if there is a target date for the next release containing it? Thanks!

AndrewBoklashko:
Hi @tomfotherby @M00nF1sh

I just tried v1.1.6 with this fix. Is this intended to work across multiple hosts? For example:
``` yaml
spec:
  rules:
    - http:
        paths:
          - path: /*
            backend:
              serviceName: ssl-redirect
              servicePort: use-annotation
    - host: hostA
      http:
        paths:
          - path: /pathA/*
            backend:
              serviceName: serviceA
              servicePort: 80
    - host: hostB
      http:
        paths:      
          - path: /*
            backend:
              serviceName: serviceB
              servicePort: 80
```

I believe that if the redirect rule does not have host specified, it should be considered as unconditional redirect as well. So in this case `hostA` and `hostB` rules should be skipped.

tomfotherby:
@AndrewBoklashko - Good point. I agree with you. Alas, currently the controller processes each rule in turn and the short circuit mechanism that avoids unneeded rules only operates in the scope of the individual rule.

In my understanding, but please correct me if I'm wrong, the bug you found isn't causing active harm, i.e. it's not causing needed rules to go missing, but it's not operating as well as it could (i.e. it's not doing what it says on the tin). Is that correct?

---

I appologise for this mistake. I hadn't considered "hostless rules" even though I see [the example](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/docs/guide/tasks/ssl_redirect.md) does not have a host specified. Thanks for bringing it up - I will work on a patch because we would also benefit from this. I can't promise speed though, I think it will be a level-up in complexity from the current solution and there's rather a lot going on presently...

AndrewBoklashko:
> In my understanding, but please correct me if I'm wrong, the bug you found isn't causing active harm, i.e. it's not causing needed rules to go missing, but it's not operating as well as it could (i.e. it's not doing what it says on the tin). Is that correct?

That is correct.

> I appologise for this mistake. I hadn't considered "hostless rules" even though I see the example does not have a host specified. Thanks for bringing it up - I will work on a patch because we would also benefit from this. I can't promise speed though, I think it will be a level-up in complexity from the current solution and there's rather a lot going on presently...

All good, thanks for the follow up. Will be looking to try a patch, but no rush - we can live with how it works now too.

tdmalone:
hi @tomfotherby, @AndrewBoklashko, I just thought I'd follow up and see if either you got any further on the above ^^

Is this being tracked under #853? If not, would you like me to create a new issue for it?

tomfotherby:
@tdmalone - At my company we don't currently use "hostless rules" in our ALB ingress spec. So to set expectations, I'm not going to be working on an enhancement for the foreseeable future. I would find it hard to test things without a real use-case in my aws account. My PR unfortunately had a bug (raised in https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/1274 and fixed in https://github.com/kubernetes-sigs/aws-alb-ingress-controller/pull/1286) which revealed to me the sharp edges and how important it would be to test well. With any luck, other folk will get involved.

> Is this being tracked under #853? If not, would you like me to create a new issue for it?

I would say #853 tracks it.

tdmalone:
No worries @tomfotherby, thanks for the reply. I might see if I can refactor to hostless rules and/or try to dig into the code here to see if I can make something work (not a Go programmer, but I could take a stab!)

gpolicante:
resolved to version 1.16 https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/1274

PR #1166, opened 2020-02-14
Ingress controller V2 skeleton

M00nF1sh:
## Summary
The skeleton of Ingress Controller V2 that add build-in support for IngressGroup.
based on prototype implementation in https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/914, added full-UT coverage 🤣 

## Design
1. Ingress belongs to same group will be reconciled together, which results in a single LoadBalancer(ALB).
    * We use annotation `alb.ingress.kubernetes.io/group.name` to identify groupName. All Ingress have same group name will be considered to be in same group(we call it *explicit group*)
    * To normalize the handling logic, Ingresses without such annotation is considered to be in *implicit group* of its own.
2. The rules on LoadBalancer have evaluation orders. We use the following way to order rules by multiple Ingresses in a group.
    * by default rules belongs to different Ingresses will be ordered by Ingresses fully-qualified name.
    * We use annotation `alb.ingress.kubernetes.io/group.order` to explicitly specify order. Rules for Ingresses with explicit order will be ordered based on this order.
   * Ingresses with explicit specified order will be ordered after Ingresses without explicit order. (Since most order specific rule is an catch all rule)
3. We use finalizer to identity group ownership. Each group will have their own finalizer.
    * explicit group: `alb.ingress.k8s.aws/group-name`
    * implicit group: `alb.ingress.k8s.aws/namespace.ingress-name`
   
    This allows us to 
    * hold Ingress deletion before the rule is cleaned up on LoadBalancer.
    * have a nice & clean way to handle group change events.

## Next steps:
1. The Ingress within a group will be translated by an "compile module" into in-memory data models(which closely matches ALB's data structure). It opens door for future implementations like:
     * compile module for V2 Ingress: https://github.com/kubernetes-sigs/service-apis
     * compile module for k8s Service for AWS NLB.  (NLB using the same data structure as ALB, they can share the same deployment module below. we need a standalone controller to support NLB's ip mode due to cloudProvider's interface limitations)
2. The  in-memory data models will be deployed by an "deployment module" into AWS. initial version will be AWSSDK based. It opens door for future implementations like:
    * CFN based deployment module. (currently CFN lacks functionalities like weighted target group support)
    * "service-operator" based module. (we can deploy our models into ALB specific CRDs, and let these CRD's controller handle the rest) 

## Test done:
1. Full UT
2. Manually E2E tests for cases like: 
       * normal implicit group
       * normal explicit group
       * change an Ingress without group join an explicit group. (by add `alb.ingress.kubernetes.io/group.name` annotation)
       * change an Ingress in an explicit group into it's own implicit group. (by remove `alb.ingress.kubernetes.io/group.name` annotation)
       * change an Ingress in an explicit group to another explicit group. (by modify `alb.ingress.kubernetes.io/group.name` annotation)
       * delete implicit group
       * delete a member in explicit group
       * delete all members in explicit group


controllers/ingress/eventhandlers/ingress_events.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (h *enqueueRequestsForIngressEvent) Update(e event.UpdateEvent, queue workqueue.RateLimitingInterface) {
+	ingOld := e.ObjectOld.(*networking.Ingress)
+	ingNew := e.ObjectNew.(*networking.Ingress)
+	if equality.Semantic.DeepEqual(ingOld.Annotations, ingNew.Annotations) &&
+		equality.Semantic.DeepEqual(ingOld.Spec, ingNew.Spec) &&
+		equality.Semantic.DeepEqual(ingOld.DeletionTimestamp.IsZero(), ingNew.DeletionTimestamp.IsZero()) {
+		logger.V(1).Info("ignoring unchanged Ingress Update event", "event", e)
+		return
+	}
```
sramabad1: Comment here to explain the changes that we are interested in handling (Annotations, Spec and deletion case with differing timestamp ) would help the reader.

PR #1167, opened 2020-02-18
Update API group of Ingress to networking.k8s.io

headcr4sh:
Starting with Kubernetes 1.16 Ingress in the extensions/v1beta API has been deprecated.
Ingress in the extensions/v1beta1 API will no longer be served starting with Kubernetes 1.20.
Since Kubernetes 1.14 there is a new apiGroup which should be used instead.

See https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/ for further details.

[The Kubernetes project maintains release branches for the most recent three minor releases](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md), therefore it seems legit to drop support for Kubernetes 1.13 with this change.


headcr4sh:
Mh. Something went wrong here. I suppose, I'll have to double-check what happened.
Looks like the Go module system has fooled me one more time... ;-)

M00nF1sh:
:D Hi, thanks for implementing this.
We do have to support eks 1.12~1.13 for a few months... We'll have to inject a client-side translating layer in client-go for this..(i'm experimenting this in v2 currently)

headcr4sh:
Ok. Then I'll close this PR and wait until 1.13 is reeeeaally not necessary to be supported anymore. ;-)

PR #1172, opened 2020-02-26
rely on node condition instead of EC2 and add cordon node support

### LINKED ISSUE #1161: When nodes are cordoned they are not removed from the target group
I am running eks 1.14 and deployments with the ingress controller seem to work ok but when I drain a node or cordon a node it stays in the target group meaning its still routable via the ALB.

This has the effect that when running a load test and then draining a node I see 504
 errors when traffic is sent to the node and it is in the process of being terminated.

Is this the desired behaviour or a bug?

M00nF1sh:
Change node pool selection logic to match [Kubernetes logic](https://github.com/kubernetes/kubernetes/blob/b6b494b4484b51df8dc6b692fab234573da30ab4/pkg/controller/service/controller.go#L605): 
  * Remove the dependency of EC2:DescribeInstance for node healthCheck. (instead, node condition is checked)
  * Check for whether node are unschedulable.(cordoned node will be unschedulable)
  * Check for whether `node.kubernetes.io/exclude-from-external-load-balancers` annotation presents
  * Check for whether `alpha.service-controller.kubernetes.io/exclude-balancer` annotation presents(previously checked the value equals to 'true')

# Test done:
1. cordon node
2. uncordon node
3. delete node
4. delete cordoned node

Fix #1161

### DIFF
--- a/internal/aws/ec2.go
+++ b/internal/aws/ec2.go
@@ -29,9 +29,6 @@ type EC2API interface {
 	// StatusEC2 validates EC2 connectivity
 	StatusEC2() func() error
 
-	// IsNodeHealthy returns true if the node is ready
-	IsNodeHealthy(string) (bool, error)
-
 	// GetInstancesByIDs retrieves ec2 instances by slice of instanceID
 	GetInstancesByIDs([]string) ([]*ec2.Instance, error)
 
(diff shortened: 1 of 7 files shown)


karolinepauls:
What are the implications when cordoning all nodes and relying on node autoscaling (reacting to kubectl drain or kubectl rollout restart on all workloads) to provision new nodes and later delete the cordoned nodes when they become unused?

Will the cordoned nodes be instantly removed from the ALB targets?

Issue #1163, opened 2020-02-12
When running by target-type:ip on EKS, ingress targetgroup changes occur due to the increase or decrease of unrelated nodes ASG


cw-sakamoto:
We create ingress with target-type:ip on EKS, alb-ingress-controller modify target group when the increase or decrease of unrelated nodes(pod controlled taint and affinity) .

Is this the expected behavior?
The log of alb-ingress-controller with `-v = 4 ` is as follows.

```
I0212 00:54:45.133850       1 listener.go:162] util: DeepEqual([{    ForwardConfig: {      TargetGroupStickinessConfig: {        Enabled: false      },      TargetGroups: [{          TargetGroupArn: "arn:aws:elasticloadbalancing:ap-northeast-1:249XXXXXXXX:targetgroup/df1XXXX-1cf6af38ccfdbc5994f/e2fdb9XXXXX769a1",          Weight: 1        }]    },    Order: 1,    TargetGroupArn: "arn:aws:elasticloadbalancing:ap-northeast-1:249XXXXXXXX:targetgroup/df1XXXX-1cf6af38ccfdbc5994f/e2fdb9XXXXX769a1",    Type: "forward"  }], [{    ForwardConfig: {      TargetGroupStickinessConfig: {        Enabled: false      },      TargetGroups: [{          TargetGroupArn: "arn:aws:elasticloadbalancing:ap-northeast-1:249XXXXXXXX:targetgroup/df1XXXX-1cf6af38ccfdbc5994f/e2fdb9XXXXX769a1",          Weight: 1        }]    },    Order: 1,    Type: "forward"  }]) found inequality
I0212 00:54:45.133904       1 log.go:30] test-web/test-web-php: listener defaultActions needs modification: [{
I0212 00:54:45.133911       1 log.go:30] test-web/test-web-php:     ForwardConfig: {
I0212 00:54:45.133915       1 log.go:30] test-web/test-web-php:       TargetGroupStickinessConfig: {
I0212 00:54:45.133919       1 log.go:30] test-web/test-web-php:         Enabled: false
I0212 00:54:45.133923       1 log.go:30] test-web/test-web-php:       },
I0212 00:54:45.133928       1 log.go:30] test-web/test-web-php:       TargetGroups: [{
[...]

Issue #1103, opened 2019-12-12
Ingress controller reconciles all endpoints sequentially causing delay that results in 502s


krunalnsoni:
When a pod is deleted from k8s, ingress controller receives endpoint change event and reconciles all endpoints sequentially. As number of ingress resources are added, the delay for reconciliation with k8s cluster gets longer.  This is very easy to reproduce because endpoints are reconciled in alphabetical order.  Updating endpoint for last service will take the longest.

In our case, we have 8 ingress resources. It takes at least 15 seconds for AWS ingress controller to reconcile with k8s cluster. Once pod has been terminated, this results in `502`. 

The workaround is to add preStop delay to the pod.  We have added 30 seconds delay to resolve this issue. However, this is not scalable as more ingress resources are being added to the cluster.

Controller should only reconcile endpoint being changed to improve efficiency. If not, endpoints could be reconciled in parallel.

There is already a `TODO` for this.
https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/31aa413a6b63ebc4e30b400f94613fda485c0a2b/internal/ingress/controller/handlers/endpoints.go#L52

gaganapplatix:
You need the readiness gate if you are on flat networking and Pods are direct targets for the ALB. There is a PR sitting around which needs to be merged for this

#955

Without that the deployment controller will move forward with its rolling updates regardless of whether target groups are up to date with regard to the new pods.

PR #1173, opened 2020-02-26
optimize endpoint reconcile

OmerKahani:
Checks if the endpoint is of a service that the ingress serve
related to #1103 #1163


OmerKahani:
@m00nf1sh Hi, I want to add this check, we had a problem where unrelated pod triggers a lot of reconciles, resulting in API Request Throttling. 
before I start with implementing the check, do you think this is the right place to put the function?

M00nF1sh:
Hi thanks for taking this 🤣.
i think the check can just be put inside the `EnqueueRequestsForEndpointsEvent` file.

OmerKahani:
There is a typo in the branch name :(
closing this PR in favor for #1179 (same code)

PR #1176, opened 2020-02-27
Typo in the "alb-ingress-controller.yaml" if env variables are to be used

FlorianOtel:
If the access key / secret key are to be used, the "env:" line must be uncommented as well -- otherwise the resulting YAML is not valid.


M00nF1sh:
Thanks for fixing this. (Don't know it works without list items :D)
/ok-to-test
/approve
/lgtm

PR #1178, opened 2020-02-29
Check node health using k8s API instead of ec2 API

### LINKED ISSUE #1100: Removal of AWS Cache renders the controller un-usable on large clusters

JacobHenner:
Frequent calls to ec2's DescribeInstanceStatus function may result in
AWS API throttling - especially on large clusters. Instead, query
Kubernetes to determine if a node is healthy.

Fixes #1100 (the underlying issue, not caching).

### DIFF
--- a/internal/ingress/backend/endpoint.go
+++ b/internal/ingress/backend/endpoint.go
@@ -69,15 +69,16 @@ func (resolver *endpointResolver) resolveInstance(ingress *extensions.Ingress, b
 		instanceID, err := resolver.store.GetNodeInstanceID(node)
 		if err != nil {
 			return nil, err
-		} else if healthy, err := resolver.cloud.IsNodeHealthy(instanceID); err != nil {
-			return nil, err
-		} else if !healthy {
-			continue
 		}
-		result = append(result, &elbv2.TargetDescription{
-			Id:   aws.String(instanceID),
-			Port: aws.Int64(int64(nodePort)),
-		})
+		for _, nodeCondition := range node.Status.Conditions {
+			if nodeCondition.Type == corev1.NodeReady && nodeCondition.Status == corev1.ConditionTrue {
+				result = append(result, &elbv2.TargetDescription{
+					Id:   aws.String(instanceID),
+					Port: aws.Int64(int64(nodePort)),
+				})
+			}
+		}
+
 	}
 	return result, nil
 }


internal/ingress/backend/endpoint.go
```diff
@@ -69,15 +69,16 @@ func (resolver *endpointResolver) resolveInstance(ingress *extensions.Ingress, b
...
-			return nil, err
-		} else if !healthy {
-			continue
 		}
-		result = append(result, &elbv2.TargetDescription{
-			Id:   aws.String(instanceID),
-			Port: aws.Int64(int64(nodePort)),
-		})
+		for _, nodeCondition := range node.Status.Conditions {
+			if nodeCondition.Type == corev1.NodeReady && nodeCondition.Status == corev1.ConditionTrue {
```
jaypipes: Might be nice (in a future PR, not necessary in this one) to add a `nodeIsReady()` function that hides the above two lines. That would make this function easier to read:

```go
        for _, node := range resolver.store.ListNodes() {
		instanceID, err := resolver.store.GetNodeInstanceID(node)
		if err != nil {
			return nil, err
                }
                if nodeIsReady(node) {
		        result = append(result, &elbv2.TargetDescription{
		                Id:   aws.String(instanceID),
			        Port: aws.Int64(int64(nodePort)),
		        })
                }
        }
```

M00nF1sh:
actually I already have one open...https://github.com/kubernetes-sigs/aws-alb-ingress-controller/pull/1172  🤣 
Close this one since mine covers this and have additional features(only reconcile when node matters changes)

PR #1179, opened 2020-03-01
Optimize endpoint reconcile

### LINKED ISSUE #1103: Ingress controller reconciles all endpoints sequentially causing delay that results in 502s

OmerKahani:
replace #1173 (fix typo in branch name)
closes #1103 #1163

On an endpoint event, check the ingress in the namespace and reconsile only the one that are related to the endpoint.


internal/ingress/controller/handlers/endpoints.go
```diff
@@ -48,23 +49,42 @@ func (h *EnqueueRequestsForEndpointsEvent) Delete(e event.DeleteEvent, queue wor
...
+		for _, service := range services {
+			if service.ServiceName == endpoints.Name {
+				glog.Infof(ingress.Name)
+				queue.Add(reconcile.Request{
+					NamespacedName: types.NamespacedName{
+						Namespace: ingress.Namespace,
+						Name:      ingress.Name,
+					},
+				})
+				continue ingress
```
devkid: Why not just `break`? You would end the `for _, service := range services` loop and start over at line 60.

internal/ingress/controller/handlers/endpoints.go
```diff
@@ -48,23 +49,42 @@ func (h *EnqueueRequestsForEndpointsEvent) Delete(e event.DeleteEvent, queue wor
 func (h *EnqueueRequestsForEndpointsEvent) Generic(event.GenericEvent, workqueue.RateLimitingInterface) {
 }
 
-//TODO: this can be further optimized to only included ingresses referenced this endpoints(service) :D
+//TODO: this can be further optimized to only reconcile the target group referenced by the endpoints(service) :D
```
devkid: The comment should be removed as this is implemented now.

devkid:
@OmerKahani I like your implementation better than mine :slightly_smiling_face:

internal/ingress/controller/handlers/endpoints.go
```diff
@@ -48,23 +49,42 @@ func (h *EnqueueRequestsForEndpointsEvent) Delete(e event.DeleteEvent, queue wor
...
+				NamespacedName: types.NamespacedName{
+					Namespace: ingress.Namespace,
+					Name:      ingress.Name,
+				},
+			})
+		}
+
+		for _, service := range services {
+			if service.ServiceName == endpoints.Name {
+				glog.Infof(ingress.Name)
```
M00nF1sh: we should remove this log for prod, or using `debug` with sensible message

internal/ingress/controller/handlers/endpoints.go
M00nF1sh: yeah, we should use break instead of goto :D

internal/ingress/controller/handlers/endpoints.go
```diff
@@ -48,23 +49,42 @@ func (h *EnqueueRequestsForEndpointsEvent) Delete(e event.DeleteEvent, queue wor
...
 			continue
 		}
-		queue.Add(reconcile.Request{
-			NamespacedName: types.NamespacedName{
-				Namespace: ingress.Namespace,
-				Name:      ingress.Name,
-			},
-		})
+
+		services, err := tg.ExtractTargetGroupBackends(&ingress)
```
M00nF1sh: nit, variableName should be `backends` instead of `services` xD

internal/ingress/controller/handlers/endpoints.go
OmerKahani: I used it for testing, and forgot to delete it - deleted

internal/ingress/controller/handlers/endpoints.go
OmerKahani: Hi, there is still a way to optimize this function. Currently, we reconcile the whole ingress, we could only reconcile the target group associate with the backend.

internal/ingress/controller/handlers/endpoints.go
```diff
@@ -60,11 +61,28 @@ func (h *EnqueueRequestsForEndpointsEvent) enqueueImpactedIngresses(endpoints *c
...
+
+		backends, err := tg.ExtractTargetGroupBackends(&ingress)
+		if err != nil {
+			glog.Errorf("Failed to extract backend services from ingress: %v, reconcile the ingress. error: %e", ingress.Name, err)
+			queue.Add(reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Namespace: ingress.Namespace,
+					Name:      ingress.Name,
+				},
+			})
```
M00nF1sh: it's nice to have a break here. since we shouldn't access backends anymore. (even it will be nil if we returned an err, but it's tied to the implementation of ExtractTargetGroupBackends)

M00nF1sh:
@OmerKahani Hi, i think this is good to merge. Just one nit, we should added a break inside the `if err != nil` check
/ok-to-test

sonalkr132:
Thank you so much for fixing this :sparkling_heart:

PR #1198, opened 2020-03-19
support to use regional STS when using iam-for-service-accounts

M00nF1sh:
Support to use regional STS when using iam-for-service-accounts.

This will support china regions and switch to use regional STS for global regions as well.
Background:
   IAM for Service Accounts will use the region configured in AWS session.


SaranBalaji90:
lgtm
/approve

PR #1206, opened 2020-03-26
Make options consistent

tonglil:


M00nF1sh:
thanks for fixing this. 
/approved

Issue #1209, opened 2020-03-30
Restarts in loop on large pod specs


bpineau:
Kubernetes 1.14 (and corresponding apimachinery release) integrates a patch [bumping max objects size](https://github.com/kubernetes/kubernetes/commit/886400f62a9552ddb405a147fa14d8b0c0a7081f) supported by serializers from 1MB to 16MB.

Clients using older versions and running on k8s >= 1.14 might be served objects largers than they can support. This was unlikely to happen when aws-alb-ingress-controller was watching ingresses and services, but we're seeing such large pods hitting the new alb controller's pod informer:

The symptoms are controller crashs, preceded by logs like:
```
E0330 14:06:27.573394       1 streamwatcher.go:109] Unable to decode an event from the watch stream: object to decode was longer than maximum allowed size
W0330 14:06:27.573481       1 reflector.go:270] pkg/mod/k8s.io/client-go@v0.0.0-20181213151034-8d9ed539ba31/tools/cache/reflector.go:95: watch of *v1.Pod ended with: very short watch: pkg/mod/k8s.io/client-go@v0.0.0-20181213151034-8d9ed539ba31/tools/cache/reflector.go:95: Unexpected watch close - watch lasted less than a second and no items received
```

I verified that bumping aws-alb-ingress-controller apimachinery dep to 1.14.0 (from 1.13.1) solves the issue.

```
diff --git go.mod go.mod
index b8e40bc6..314958f0 100644
--- go.mod
+++ go.mod
@@ -35,7 +35,7 @@ require (
        golang.org/x/oauth2 v0.0.0-20190212230446-3e8b2be13635 // indirect
        gopkg.in/inf.v0 v0.9.1 // indirect
        k8s.io/api v0.0.0-20181213150558-05914d821849
-       k8s.io/apimachinery v0.0.0-20181127025237-2b1284ed4c93
+       k8s.io/apimachinery v0.0.0-20190313205120-d7deff9243b1
        k8s.io/apiserver v0.0.0-20190214201149-f9f16382a346
        k8s.io/client-go v0.0.0-20181213151034-8d9ed539ba31
        k8s.io/klog v0.2.0
diff --git go.sum go.sum
index a3da5bf8..fc1e38be 100644
--- go.sum
+++ go.sum
[...]

PR #1213, opened 2020-04-03
don't crash on large objets served by recent k8s-api-server

bpineau:
Per #1209 : now that aws-alb-controller has informers on pods, it's
much more likely to be served an object too large for < v1.14
apimachinery serializers.

Bump apimachinery dep to v1.14.0 solves the issue.


M00nF1sh:
/lgtm
/approve
thanks for fix this :D

Issue #1205, opened 2020-03-25
Target does not get registered if using readinessGate


vedatappsamurai:
I've upgraded to v1.1.6 to make use of the pod readiness gate feature to reduce 502/504's during HPA scales. Then I've proceeded to update my deployment per [this](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/docs/guide/ingress/pod-conditions.md) document.

After I updated my deployment to have the readiness gate, the first pod that had the readinessGate spec got registered by the controller just fine with following status:
```
status:
  conditions:
  - lastProbeTime: "2020-03-25T07:17:09Z"
    lastTransitionTime: "2020-03-25T07:13:48Z"
    status: "True"
    type: target-health.alb.ingress.k8s.aws/<ingress>_<service>_<service-port>
```

The following pod of the same deployment, had its status stuck like this:
```
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2020-03-25T07:10:29Z"
    message: corresponding condition of pod readiness gate "target-health.alb.ingress.k8s.aws/<ingress>_<service>_<service-port>"
      does not exist.
    reason: ReadinessGatesNotReady
    status: "False"
    type: Ready
```
I've waited for a good 5 minutes to see that pod get registered into the target group, but it did not. I could reproduce this error by performing multiple rollovers for the same deployment. However if I delete the controller pod and let it restart, it recognizes these pods and registers them into the target group. After the registrations at this point, it stops to register pods again.

Is there anything wrong with the approach I took? Or is this an issue at controllers end?

ivanmp91:
I'm not sure if it could be the same issue, but at least looks like I'm experimenting a similar behaviour. I also started to test the new 1.1.6 version in order to use pod readiness gate and for me it's happening the other way around, when a new pod is launched using readiness gate, it takes around 5-10 minutes to update the pod readiness gate status and register the new pod to the target group. Also I've found some cases where the pod never gets updated having to delete the pod to create a new one. At the time when a new pod is launched I got exactly the same status reported in this issue:

```
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2020-03-25T11:13:10Z"
    status: "True"
    type: Initialized
  - lastProbeTime: null
    lastTransitionTime: "2020-03-25T11:13:10Z"
[...]

PR #1214, opened 2020-04-06
React on pod events for readiness gates

bpineau:
We're relying on endpoints events to re-trigger reconciliations during
rollouts, and we're considering pod's containers status (eg. are all pod's
containers ContainerReady?) to either act upon, or swallow those events.

Both (pods changes, ep changes) can be out of sync. For instance: a starting
pod whose containers aren't all ContainerReady might have its addresses
registered to an ep subset's NotReadyAddresses, kicking a reconcile event
which won't propagate to the TargetGroups (+ conditions updates) since the pod
is evaluated as not ready. Further pod changes won't kick an endpoint change
(due to readiness gates the pod's address will stay in NotReadyAddresses
until we do something). As probably seen in  #1205 .

In order to react reliably on pods changes, we have to hook in a pod watch.
Doing so is slightly expensive as we have to map pod -> [service ->]
endpoint -> ingress on pods events, though we limit the search to pod's ns
(services can only reference pods from their own ns, and ingress services from
their ns).

We might optimize that by adding a pod indexer (if supported by controller-runtime).

### DIFF
--- a/internal/ingress/backend/endpoint.go
+++ b/internal/ingress/backend/endpoint.go
@@ -182,15 +182,10 @@ func (resolver *endpointResolver) resolveIP(ingress *extensions.Ingress, backend
 					continue
 				}
 
-				// check if all containers are ready
-				for _, condition := range pod.Status.Conditions {
-					if condition.Type == api.ContainersReady {
-						if condition.Status == api.ConditionTrue {
-							addresses = append(addresses, epAddr)
-						}
-						break
-					}
+				if IsPodSuitableAsIPTarget(pod) {
+					addresses = append(addresses, epAddr)
 				}
+
 			}
 			for _, epAddr := range addresses {
 				result = append(result, &elbv2.TargetDescription{
@@ -221,6 +216,17 @@ func IsNodeSuitableAsTrafficProxy(node *corev1.Node) bool {
 	for _, cond := range node.Status.Conditions {
 		if cond.Type == corev1.NodeReady && cond.Status == corev1.ConditionTrue {
 			return true
+		}
+	}
+	return false
+}
+
+// IsPodSuitableAsIPTarget check whether pod is suitable as a TargetGroup's target
+// (currently tested: are all pod's containers ready?).
+func IsPodSuitableAsIPTarget(pod *corev1.Pod) bool {
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == api.ContainersReady {
+			return condition.Status == api.ConditionTrue
 		}
 	}
 	return false
--- a/internal/ingress/controller/handlers/pod.go
+++ b/internal/ingress/controller/handlers/pod.go
@@ -2,11 +2,11 @@ package handlers
 
 import (
 	"context"
-	"reflect"
 
 	"github.com/golang/glog"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/alb/tg"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/annotations/class"
+	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/backend"
 	corev1 "k8s.io/api/core/v1"
 	extensions "k8s.io/api/extensions/v1beta1"
 	"k8s.io/apimachinery/pkg/types"
@@ -27,14 +27,17 @@ type EnqueueRequestsForPodsEvent struct {
 
 // Create is called in response to an create event - e.g. Pod Creation.
 func (h *EnqueueRequestsForPodsEvent) Create(e event.CreateEvent, queue workqueue.RateLimitingInterface) {
-	h.enqueueImpactedIngresses(e.Object.(*corev1.Pod), queue)
 }
 
 // Update is called in response to an update event -  e.g. Pod Updated.
 func (h *EnqueueRequestsForPodsEvent) Update(e event.UpdateEvent, queue workqueue.RateLimitingInterface) {
 	podOld := e.ObjectOld.(*corev1.Pod)
 	podNew := e.ObjectNew.(*corev1.Pod)
-	if !reflect.DeepEqual(podOld, podNew) {
+
+	// we only enqueue reconcile events for pods whose containers changed state
+	// (ContainersReady vs not ContainersReady).
+	if backend.IsPodSuitableAsIPTarget(podNew) != backend.IsPodSuitableAsIPTarget(podOld) {
+		// ... and only for pods referenced by an endpoint backing an ingress:
 		h.enqueueImpactedIngresses(podNew, queue)
 	}
 }
@@ -96,6 +99,7 @@ func (h *EnqueueRequestsForPodsEvent) enqueueImpactedIngresses(pod *corev1.Pod, 
 						Name:      ingress.Name,
 					},
 				})
+				break
 			}
 		}
 	}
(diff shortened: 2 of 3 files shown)


internal/ingress/controller/handlers/pod.go
```diff
@@ -0,0 +1,118 @@
...
+				})
+			}
+		}
+	}
+}
+
+func (h *EnqueueRequestsForPodsEvent) isPodInEndpoint(pod *corev1.Pod, endpoint *corev1.Endpoints) bool {
+	for _, sub := range endpoint.Subsets {
+		for _, addr := range sub.Addresses {
+			if addr.IP == pod.Status.PodIP {
```
nirnanaaa, line 107: maybe use targetRef instead, like here: https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/b8630a71bea087bc6c91249f9a550025ad24cc93/internal/ingress/backend/endpoint.go#L91 instead of matching against IPs. also filter for endpoints that don't have an IP or target ref

internal/ingress/controller/handlers/pod.go
```diff
@@ -0,0 +1,118 @@
...
+				return true
+			}
+		}
+		for _, addr := range sub.NotReadyAddresses {
+			if addr.IP == pod.Status.PodIP {
```
nirnanaaa, line 112: same here

internal/ingress/controller/handlers/pod.go
```diff
@@ -0,0 +1,118 @@
...
+				Namespace: ingress.Namespace,
+				Name:      backend.ServiceName,
+			}
+			if err = h.Cache.Get(context.Background(), nspname, endpoint); err != nil {
+				glog.Errorf("failed to fetch enpoint %s backing ingress %s/%s, ignoring",
+					backend.ServiceName, ingress.Namespace, ingress.Name)
+				continue
+			}
+
+			if h.isPodInEndpoint(pod, endpoint) {
```
devkid, line 92: Would it make sense to only reconcile the ingress if the `ContainersReady` condition is true to avoid unnecessary reconciliation iterations (and thus many AWS API calls) on frequent pod status updates during pod startup?

nirnanaaa: @devkid the point of reconciliations IMO is that you don't have to rely on state comparison to make a decision. 

In our test we didn't had any issues yet with AWS API limits.

Also we shouldn't rely on the fact that right now only containers ready is the important and driving factor for registering pods in the endpoints table. Upstream k8s could change or extend that logic any time and it would break the implementation.

bpineau: I agree with @nirnanaaa, that would specialise the watch by embedding expectations on implementations details of an unrelated part in the code base (which might change); eg. setting a trap for futur maintainers.

nirnanaaa:
@M00nF1sh or @bigkraig could you please have a look - would be much appreciated.

internal/ingress/controller/handlers/pod.go
M00nF1sh: @nirnanaaa 
I think we need to have the ContainerReady check. we have the check to decide whether to register the Pod to ALB anyway. Under current architecture, the reconcile will call AWS APIs to check every ALB settings before reconcile targetGroups, which is inefficient and cause troubles if there are many ingresses.

I think it's better to have below changes:
1. have a common function defined in backends package like "IsPodSuitableAsIPTarget", which checks containersReady
2. change https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/internal/ingress/backend/endpoint.go#L186 to call the function above for decide whether pod in NotReadyAddresses should be registered.
3. only enqueue Ingress for pod Update event, when PodInEndpoint(podNew)  && IsPodSuitableAsIPTarget(podNew) != IsPodSuitableAsIPTarget(podOld)

nirnanaaa: @M00nF1sh does that really justify adding all the complex logic to have oldRef stored, the handling if no old ref was found in store etc ? I don't think the comparison would be a huge deal other than duplicating logic.

nirnanaaa: Usually I've seen that not that many pod updates are triggered other than when the containers are ready. We're currently running this patch in prod.

M00nF1sh: @nirnanaaa i think it helps to reduce reconcile calls under current architecture. and podOld and podNew are guarantee to exists(by informer). It's a optimization other than requirement.
In the future, we can let user control the "IsPodSuitableAsIPTarget" logic, e.g. check for additional readinessGates before register pod to ALB.

nirnanaaa: Understood, thanks for clarifying. Would it might make sense to first check if that's an actual problem ? Theoretically there aren't that many pod events happening and we would save time by not overengineering a non existing problem.

nirnanaaa: Understood, thanks for clarifying. Would it might make sense to first check if that's an actual problem ? Theoretically there aren't that many pod events happening and we would save time by not overengineering a non existing problem.

M00nF1sh: @nirnanaaa I already got a lot issues caused by the controller generating too many reconcile s and cause AWS api throttling :D. (and that's my motivation to separate targetReconcile from ALB reconcile in v2).

nirnanaaa: Fully understood. Sorry if I sound rough -  I just want to find the best possible solution. The ingress controller codebase is already full of checks left and right. So keeping a clean base is also something important ;)

M00nF1sh: @nirnanaaa :D sure, i think it's pretty clean too, and makes the precondition of whether to add pod into ALB explicit. :D

bpineau: @M00nF1sh pushed a commit doing exactly that

M00nF1sh: thanks, ready to go :D, just  one nit, we need a "break" after queue.Add :D

bpineau: Thanks, added the `break` (optimize the loop but precludes pods being part of several readiness gate backed ingresses - which I guess was a shady corner case wrt readiness gates semantics anyway?).

devkid: @bpineau The break would only quit the loop that iterates over the backends, i.e. all other ingresses would still be processed. Enqueuing the ingress for reconciliation will reconcile all target groups anyhow.

internal/ingress/backend/endpoint.go
```diff
@@ -226,6 +221,19 @@ func IsNodeSuitableAsTrafficProxy(node *corev1.Node) bool {
 	return false
 }
 
+// IsPodSuitableAsIPTarget check whether pod is suitable as a TargetGroup's target
+// (currently tested: are all pod's containers ready?).
+func IsPodSuitableAsIPTarget(pod *corev1.Pod) bool {
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == api.ContainersReady {
+			if condition.Status == api.ConditionTrue {
```
devkid, line 229: This should be `return condition.Status == api.ConditionTrue` to quit whenever the `ContainersReady` condition was found.

internal/ingress/controller/handlers/pod.go
bpineau: @devkid you're right

M00nF1sh:
really awesome work :D, thanks sooo much!!
/lgtm
/approve

PR #1225, opened 2020-04-16
On node events, skip reconciling ingresses target type != "instance"

devkid:
This should help reduce the number of API calls.

### DIFF
--- a/internal/ingress/controller/handlers/node.go
+++ b/internal/ingress/controller/handlers/node.go
@@ -3,7 +3,9 @@ package handlers
 import (
 	"context"
 
+	"github.com/aws/aws-sdk-go/service/elbv2"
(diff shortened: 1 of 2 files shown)


internal/ingress/controller/handlers/node.go
```diff
@@ -67,11 +70,21 @@ func (h *EnqueueRequestsForNodeEvent) enqueueImpactedIngresses(queue workqueue.R
...
-				Name:      ingress.Name,
-			},
-		})
+
+		namespacedName := types.NamespacedName{
+			Namespace: ingress.Namespace,
+			Name:      ingress.Name,
+		}
+		key := namespacedName.String()
+		ingressAnnos, err := h.Store.GetIngressAnnotations(key)
```
M00nF1sh, line 79: the targetType annotation is supported on service as well. So you can even have different targetType in a single ingress.
We should check per service and only ignore node event if none of them are instance target.

PR #1233, opened 2020-04-22
Check for sufficient available free IP space before marking a subnet as useable.

cbajumpaa:
Fix constant errors like this when an IP space is low in a subnet.
```
InvalidSubnet: Not enough IP space available in subnet-XXXXXXXX. ELB requires at least 8 free IP addresses in each subnet.\n\tstatus code: 400
```

### DIFF
--- a/internal/alb/lb/loadbalancer.go
+++ b/internal/alb/lb/loadbalancer.go
@@ -384,7 +384,7 @@ func (controller *defaultController) clusterSubnets(ctx context.Context, scheme 
 	}
 
 	if len(out) < 2 {
-		return nil, fmt.Errorf(`failed to resolve 2 qualified subnet for ALB. Subnets must contains these tags: '%s/%s': ['shared' or 'owned'] and '%s': ['' or '1']. See https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/controller/config/#subnet-auto-discovery for more details. Resolved qualified subnets: '%s'`,
+		return nil, fmt.Errorf(`failed to resolve 2 qualified subnet with at least 8 free IP Addresses for ALB. Subnets must contains these tags: '%s/%s': ['shared' or 'owned'] and '%s': ['' or '1']. See https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/controller/config/#subnet-auto-discovery for more details. Resolved qualified subnets: '%s'`,
 			aws.TagNameCluster, controller.cloud.GetClusterName(), key, log.Prettify(out))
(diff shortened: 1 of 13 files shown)


internal/alb/lb/loadbalancer.go
```diff
@@ -395,11 +395,17 @@ func (controller *defaultController) clusterSubnets(ctx context.Context, scheme
...
+// Also determine if the subnet has sufficient free IP space available.
 func subnetIsUsable(new *ec2.Subnet, existing []*ec2.Subnet) bool {
 	for _, subnet := range existing {
 		if *new.AvailabilityZone == *subnet.AvailabilityZone {
 			return false
 		}
 	}
+
+	if *new.AvailableIpAddressCount < 8 {
+		return false
```
M00nF1sh, line 407: let's add additional message to 
```if len(out) < 2 {
		return nil, fmt.Errorf(`failed to resolve 2 qualified subnet for ALB. Subnets must contains these tags: '%s/%s': ['shared' or 'owned'] and '%s': ['' or '1']. See https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/controller/config/#subnet-auto-discovery for more details. Resolved qualified subnets: '%s'`,
			aws.TagNameCluster, controller.cloud.GetClusterName(), key, log.Prettify(out))
	}'
```
like "failed to resolve 2 qualified subnet with **at least 8 free IP Addresses** for ALB....."

internal/alb/lb/loadbalancer.go
```diff
@@ -395,11 +395,17 @@ func (controller *defaultController) clusterSubnets(ctx context.Context, scheme
...
 // the same availability zone.
+// Also determine if the subnet has sufficient free IP space available.
 func subnetIsUsable(new *ec2.Subnet, existing []*ec2.Subnet) bool {
 	for _, subnet := range existing {
 		if *new.AvailabilityZone == *subnet.AvailabilityZone {
 			return false
 		}
 	}
+
+	if *new.AvailableIpAddressCount < 8 {
```
M00nF1sh, line 406: better to use `aws.Int64Value(new.AvailableIpAddressCount) < 8` to be nil safe.

internal/alb/lb/loadbalancer.go
cbajumpaa: Done.

internal/alb/lb/loadbalancer.go
cbajumpaa: Done.

PR #1253, opened 2020-05-06
fix for pod condition type too long

### LINKED ISSUE #1217: readiness gate conditionType too long
we're starting to rollout the readiness gates on our cluster and have just come across a limitation with the way, readiness gates are set:

kubernetes imposes a 63 character limit on the length of readiness gates. Since the required string consists of `target-health.alb.ingress.k8s.aws/<name>_<name>_<port>` this is a problem for services, which rely on the name limit of 63 chars for service and ingress names.

from our logs:

```
Deployment.apps "some-lengthy-service-name-that-is-not-too-long" is invalid: spec.template.spec.readinessGates[0].conditionType: Invalid value: "target-health.alb.ingress.k8s.aws/some-lengthy-service-name-that-is-not-too-long_some-lengthy-service-name-that-is-not-too-long_http": name part must be no more than 63 characters
```

sandeepbhojwani:
-  existing pod condition type has the format "target-health.alb.ingress.k8s.aws/<INGRESS_NAME>_<SERVICE_NAME>_<SERVICE_PORT>". This can exceed the max. allowed length if the <INGRESS_NAME>_<SERVICE_NAME>_<SERVICE_PORT> part exceeds 63 characters. To work around this problem, added support for static pod condition type string. The string is "target-health.alb.ingress.k8s.aws/load-balancer-tg-ready"
- both old and new format are supported. Only one should be used at a time though.

Testing done
- created the required objects in a cluster and verified old format still works
- created the required objects in a cluster verified new format works as well
- verified a deployment transition from old format to new format

resolves #1217

### DIFF
--- a/internal/alb/tg/targethealth.go
+++ b/internal/alb/tg/targethealth.go
@@ -2,7 +2,6 @@ package tg
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"github.com/aws/aws-sdk-go/service/elbv2"
@@ -10,6 +9,7 @@ import (
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/aws"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/annotations/healthcheck"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/backend"
+	backendpkg "github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/backend"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/ingress/controller/store"
 	"github.com/kubernetes-sigs/aws-alb-ingress-controller/internal/k8s"
 	api "k8s.io/api/core/v1"
@@ -18,10 +18,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-)
-
-const (
-	staticConditionType = api.PodConditionType("target-health.alb.ingress.k8s.aws/load-balancer-any-tg-ready")
 )
 
 type targetGroupWatch struct {
@@ -79,21 +75,15 @@ type targetHealthController struct {
 
 // SyncTargetsForReconciliation starts a go routine for reconciling pod condition statuses for the given targets in the background until they are healthy in the target group
 func (c *targetHealthController) SyncTargetsForReconciliation(ctx context.Context, t *Targets, desiredTargets []*elbv2.TargetDescription) error {
-	conditionType := podConditionTypeForIngressBackend(t.Ingress, t.Backend)
-	targetsToReconcile, err := c.filterTargetsNeedingReconciliation(conditionType, t, desiredTargets)
+	readinessConditionTypes := []api.PodConditionType{
+		backendpkg.PodReadinessGateConditionType(t.Ingress, t.Backend),
+		backendpkg.AnyLBTGReadyConditionType,
+	}
+	targetsToReconcile, err := c.filterTargetsNeedingReconciliation(t, desiredTargets, readinessConditionTypes...)
 	if err != nil {
 		return err
 	}
-	// use static condition type if no targets to reconcile with the old type
-	useStaticConditionType := len(targetsToReconcile) == 0
-	if useStaticConditionType {
-		// new condition type uses a static string
-		conditionType = staticConditionType
-		targetsToReconcile, err = c.filterTargetsNeedingReconciliation(conditionType, t, desiredTargets)
-		if err != nil {
-			return err
-		}
-	}
+
 	// create, update or remove targetGroupWatch for this target group;
 	// a targetGroupWatch exists as long as there are targets in the target group whose pod condition statuses need to be reconciled;
 	// while the targetGroupWatch exists, a go routine regularly monitors the target health of the targets in the target group and updates the pod condition status for the corresponding pods
@@ -114,8 +104,7 @@ func (c *targetHealthController) SyncTargetsForReconciliation(ctx context.Contex
 		c.tgWatches[t.TgArn] = tgWatch
 
 		// start watching target health in target group and updating pod condition status
-		go c.reconcilePodConditionsLoop(ctx, t.TgArn, conditionType, tgWatch)
-
+		go c.reconcilePodConditionsLoop(ctx, t.TgArn, tgWatch, readinessConditionTypes...)
 	}
 	tgWatch.interval <- c.ingressTargetHealthReconciliationInterval(t.Backend.ServiceName, t.Ingress)
 	tgWatch.targetsToReconcile <- targetsToReconcile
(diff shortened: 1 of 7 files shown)


shaoxt:
@M00nF1sh  This fix is critical for using pod readiness gate.  Can you have a review on this?

shaoxt:
To use static string as condition Type also make it easier to avoid the following issue

https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/1241

M00nF1sh:
Hi, thanks for contributing.
This will only work correctly if **a pod is been used in a single TargetGroup.** (otherwise, the gate will be unblocked if pod is only registered in one targetGroup but not others).

I think better solution would be 
1. use a single pod readinessGate, but the controller tracks all the targetGroups pod should belong to(will be easy in v2 but hard to implement in this branch TBH), and unblocks when all targetGroup are healthy.
1. use multiple pod readinessGate, but hash the `<INGRESS_NAME><SERVICE_NAME><SERVICE_PORT>` part if it becomes too long

I'm happy to merge this change in, and we can extend the functionality in the future, just want to ensure you aware of the restriction above.

/ok-to-test

devkid:
I agree with @M00nF1sh. I would suggest to name the condition type something like `target-health.alb.ingress.k8s.aws/load-balancer-any-tg-ready` (or at least have `any` somewhere in the string) to stress that this will make the pod ready if it gets healthy in `any` target group. This way we could introduce `target-health.alb.ingress.k8s.aws/load-balancer-all-tgs-ready` later on.

sandeepbhojwani:
@devkid agree with your idea. I will change the name to `target-health.alb.ingress.k8s.aws/load-balancer-any-tg-ready`

internal/ingress/backend/endpoint.go
```diff
@@ -146,7 +146,8 @@ func (resolver *endpointResolver) resolveIP(ingress *extensions.Ingress, backend
 	}
 
 	conditionType := api.PodConditionType(fmt.Sprintf("target-health.alb.ingress.k8s.aws/%s_%s_%s", ingress.Name, backend.ServiceName, backend.ServicePort.String()))
-
+	// new condition type uses a static string
+	staticConditionType := api.PodConditionType("target-health.alb.ingress.k8s.aws/load-balancer-any-tg-ready")
```
devkid, line 150: Can you please make this a global const like in `targethealth.go`?

internal/alb/tg/targethealth.go
```diff
@@ -76,12 +80,20 @@ type targetHealthController struct {
...
 func (c *targetHealthController) SyncTargetsForReconciliation(ctx context.Context, t *Targets, desiredTargets []*elbv2.TargetDescription) error {
 	conditionType := podConditionTypeForIngressBackend(t.Ingress, t.Backend)
-
 	targetsToReconcile, err := c.filterTargetsNeedingReconciliation(conditionType, t, desiredTargets)
 	if err != nil {
 		return err
 	}
-
+	// use static condition type if no targets to reconcile with the old type
+	useStaticConditionType := len(targetsToReconcile) == 0
```
devkid, line 88: No need to have this in a variable if it's only used once, just put it in the `if` condition directly.

internal/alb/tg/targethealth.go
```diff
@@ -103,6 +115,7 @@ func (c *targetHealthController) SyncTargetsForReconciliation(ctx context.Contex
 
 		// start watching target health in target group and updating pod condition status
 		go c.reconcilePodConditionsLoop(ctx, t.TgArn, conditionType, tgWatch)
+
```
devkid, line 118: Unnecessary newline before closing bracket.

internal/alb/tg/targethealth.go
```diff
@@ -76,12 +80,20 @@ type targetHealthController struct {
 // SyncTargetsForReconciliation starts a go routine for reconciling pod condition statuses for the given targets in the background until they are healthy in the target group
 func (c *targetHealthController) SyncTargetsForReconciliation(ctx context.Context, t *Targets, desiredTargets []*elbv2.TargetDescription) error {
 	conditionType := podConditionTypeForIngressBackend(t.Ingress, t.Backend)
-
 	targetsToReconcile, err := c.filterTargetsNeedingReconciliation(conditionType, t, desiredTargets)
 	if err != nil {
 		return err
 	}
-
+	// use static condition type if no targets to reconcile with the old type
```
devkid, line 87: Please try to avoid using "old and new type" in the comments as this merely refers to the history of implementation and makes it more confusing to read.

internal/alb/tg/targethealth.go
devkid: Also, I think this might make trouble when mixing the per-ingress backend and static readiness gates (e.g. during a rolling deployment when someone wants to update from the old to the new format) because the way you have it, you will _only_ get pods with the old type or _only_ pods with the new type. Instead, I'd recommend to wrap the whole implementation of `SyncTargetsForReconciliation` into a loop so that it gets executed twice like
```go
for conditionType := range []api.PodConditionType{
    podConditionTypeForIngressBackend(t.Ingress, t.Backend),
    staticConditionType
} {
    // ...
}
```

sandeepbhojwani: Had tested rolling deployment with change from old to the new format and it worked fine. But makes sense to use a loop. Will make the change.

internal/ingress/backend/endpoint.go
sandeepbhojwani: will do.

internal/alb/tg/targethealth.go
sandeepbhojwani: ok.

internal/alb/tg/targethealth.go
sandeepbhojwani: @devkid Pod readiness condition failed to reconcile when i made it a loop. A deployment with single (old) pod readiness condition type failed to reconcile. Maybe because the watch is the same for the 2 condition types and the 2nd condition type in the loop removes the watch from the 1st condition type. Here's the code that didn't work..

```
func (c *targetHealthController) SyncTargetsForReconciliation(ctx context.Context, t *Targets, desiredTargets []*elbv2.TargetDescription) error {
	for _, conditionType := range []api.PodConditionType{
		podConditionTypeForIngressBackend(t.Ingress, t.Backend),
		staticConditionType,
	} {
		targetsToReconcile, err := c.filterTargetsNeedingReconciliation(conditionType, t, desiredTargets)
		if err != nil {
			return err
		}
		// create, update or remove targetGroupWatch for this target group;
		// a targetGroupWatch exists as long as there are targets in the target group whose pod condition statuses need to be reconciled;
		// while the targetGroupWatch exists, a go routine regularly monitors the target health of the targets in the target group and updates the pod condition status for the corresponding pods
		tgWatch, ok := c.tgWatches[t.TgArn]
		if ok {
			if len(targetsToReconcile) == 0 {
				tgWatch.cancel()
				delete(c.tgWatches, t.TgArn)
				return nil
			}
		} else {
			if len(targetsToReconcile) == 0 {
				return nil
			}

			// targetGroupWatch for this target group doesn't exist yet -> create it
			tgWatch, ctx = newTargetGroupWatch(ctx, t.Ingress, t.Backend)
			c.tgWatches[t.TgArn] = tgWatch

			// start watching target health in target group and updating pod condition status
			go c.reconcilePodConditionsLoop(ctx, t.TgArn, conditionType, tgWatch)

		}
		tgWatch.interval <- c.ingressTargetHealthReconciliationInterval(t.Backend.ServiceName, t.Ingress)
		tgWatch.targetsToReconcile <- targetsToReconcile
	}
	return nil
}
```

sandeepbhojwani: @devkid are you ok with not using a loop or do you have any suggestion on how to use it?

M00nF1sh:
i adjust this PR to make this two readinessGate to work together :D 
/ok-to-test

internal/alb/tg/targethealth.go
M00nF1sh: I adjusted the PR to work with both readiness :D

Issue #1245, opened 2020-05-04
Update docker image used in external-dns.yaml file


os-virtualretail:
On https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/master/docs/guide/external-dns/setup.md it states to use

https://raw.githubusercontent.com/kubernetes-sigs/aws-alb-ingress-controller/v1.1.7/docs/examples/external-dns.yaml

This external-dns.yaml contains 

**image: registry.opensource.zalan.do/teapot/external-dns:v0.5.9** 

this image is released on November 2018 and pretty old Latest is 0.7.1 ( https://github.com/kubernetes-sigs/external-dns/releases)

PR #1257, opened 2020-05-09
Updates the image version of examples/external-dns.yaml

literalice:
Fixes: #1245

As a test, I've confirmed that the Route53 hosted zone was successfully updated.


M00nF1sh:
thanks for making this change :D 
/lgtm
/approve

PR #1281, opened 2020-06-02
Update annotation.md

bgsilvait:
Adjusted the actions.forward-multiple-tg example. In front of arn-of-your-non-k8s-target-group there is a double ("), instead of single to open the value for the TG arn.


M00nF1sh:
/ok-to-test
/approve
thanks for fixing this :D

PR #1290, opened 2020-06-08
doc: workaround fix for deploying external-dns

### LINKED ISSUE #1289: doc: example 'external-dns' v1.1.8 does not work on IRSA
Since v1.1.8 the example of 'external-doc' uses `bitnami/external-dns` that uses user `1001` to run the external-dns process who does not has permission to read web identity token. It's also a known issue of https://github.com/aws/amazon-eks-pod-identity-webhook/issues/8 pending on upstream kubernetes.

There is a workaround
https://github.com/kubernetes-sigs/external-dns/pull/1185#issuecomment-530439786 to grant the permission.

zxkane:
make bitnami/external-dns working with IRSA

closes #1289

### DIFF
--- a/docs/examples/external-dns.yaml
+++ b/docs/examples/external-dns.yaml
@@ -50,6 +50,8 @@ spec:
         app: external-dns
     spec:
       serviceAccountName: external-dns
+      securityContext:
+        fsGroup: 65534
       containers:
       - name: external-dns
         image: bitnami/external-dns:0.7.1


docs/examples/external-dns.yaml
```diff
@@ -50,6 +50,8 @@ spec:
         app: external-dns
     spec:
       serviceAccountName: external-dns
+      securityContext:
+        fsGroup: 65534
```
M00nF1sh, line 54: would you help use `1337` to be consistent with our other examples(in other projects).
technically, any `fsGroups` works :D 

also you need to sign the CLA

docs/examples/external-dns.yaml
```diff
@@ -50,6 +50,8 @@ spec:
         app: external-dns
     spec:
       serviceAccountName: external-dns
+      securityContext:
+        fsGroup: 1337
```
sudo6si, line 54: The referenced external-dns issue PR recommends [`fsGroup: 65534`](https://github.com/kubernetes-sigs/external-dns/pull/1185/files#diff-89e5478cb30ea6ae2a798bde1f237a693ddeee6437f42db8be87b600b2032f4dR163) which would also make this example yaml match with the example yaml in their [AWS documentation](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/aws.md#manifest-for-clusters-with-rbac-enabled).

I don't think if it matters, but in case people want to use `istio` [their requirements](https://istio.io/latest/docs/ops/deployment/requirements/) specifically mention that "_Application UIDs: Ensure your pods do not run applications as a user with the user ID (UID) value of 1337._"

rothgar: The upstream external-dns from the kubernetes sig also recommends 65534 https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/aws.md#manifest-for-clusters-with-rbac-enabled

kishorj:
@zxkane Thanks for your PR. Could you port it to the main branch so that ti can be included in the v2 releases.

sudo6si:
@zxkane I would strongly urge you to update the example with `securityContext: { fsGroup: 65534 }` as is recommended by the upstream

zxkane:
> @zxkane Thanks for your PR. Could you port it to the main branch so that ti can be included in the v2 releases?

Will try it.

zxkane:
@kishorj let me know if you have any comment for this pr for main branch.

M00nF1sh (approved):
/lgtm
thanks for fixing this :D looks good to me.

PR #1299, opened 2020-06-17
Adds note about using named ports to Pod Conditions guide

jmriebold:
Current docs don't mention that when using named ports in a service with a `readinessGate` you must use the port _name_ instead of the number, resulting in endless `corresponding condition of pod readiness gate "[...]" does not exist.` errors, and pods that never become ready.


jmriebold:
CLA signed.

M00nF1sh:
Thanks so much for improving the doc 👍 
/lgtm
/approve

Issue #1313, opened 2020-07-09
Documentation : Reminder to check the target group type for the stickiness feature.


ipnextgen:
Could we add a reminder in the documentation to check the target group type when enabling stickiness? Not setting the target group type to IP is often a problem people face to have full pod stickiness.

I can do this simple PR if approved.

PR #1314, opened 2020-07-09
Documentation : Reminder to check the target group type for the stickiness feature. #1313

ipnextgen:
Add a reminder in the documentation to check the target group type when enabling stickiness? Not setting the target group type to IP is often a problem people face to have full pod stickiness.


ipnextgen:
Problem with the CLA, Ive signed it long time ago.

PR #1315, opened 2020-07-09
Add reminder to stickiness annotation to check target group type

ipnextgen:
Add a reminder in the documentation to check the target group type when enabling stickiness? Not setting the target group type to IP is often a problem people face to have full pod stickiness.


ipnextgen:
Resolves #1313

M00nF1sh:
/ok-to-test
Thanks for improve the documentation

PR #1342, opened 2020-08-10
add targets manager

M00nF1sh:
add targets manager, which is responsible for interacting with ELBV2 RegisterTarget/DeregisterTarget/DescribeTargetHealth API.
It offers below functionality:
1. provides a cached view of latest targets registered with targetGroup.
2. provides chunk functionality for RegisterTarget/DeregisterTarget operation.


pkg/targetgroupbinding/targets_manager.go
```diff
@@ -0,0 +1,276 @@
...
+	elbv2sdk "github.com/aws/aws-sdk-go/service/elbv2"
+	"k8s.io/apimachinery/pkg/util/cache"
+	"sigs.k8s.io/aws-alb-ingress-controller/pkg/aws/services"
+	"sync"
+	"time"
+)
+
+const (
+	defaultTargetsCacheTTL            = 10 * time.Minute
+	defaultRegisterTargetsChunkSize   = 200
```
kishorj, line 16: How did you determine this? Is it based on ELBv2 documentation? does the register and deregister chunk size have to be the same?

pkg/targetgroupbinding/targets_manager.go
```diff
@@ -0,0 +1,276 @@
...
+	for _, elem := range resp.TargetHealthDescriptions {
+		listedTargets = append(listedTargets, TargetInfo{
+			Target:       *elem.Target,
+			TargetHealth: elem.TargetHealth,
+		})
+	}
+	return listedTargets, nil
+}
+
+// recordSuccessfulDeregisterTargetsOperation will record a successful deregisterTarget operation
```
kishorj, line 194: nit: comment

pkg/targetgroupbinding/targets_manager.go
```diff
@@ -0,0 +1,276 @@
...
+
+// cache entry for targetsCache
+type targetsCacheItem struct {
+	// mutex protects below fields
+	mutex sync.RWMutex
+	// targets is the targets for TargetGroup
+	targets []TargetInfo
+}
+
+func (m *cachedTargetsManager) RegisterTargets(ctx context.Context, tgARN string, targets []elbv2sdk.TargetDescription) error {
```
kishorj, line 75: Plan to add UTs for this?

pkg/targetgroupbinding/targets_manager.go
```diff
@@ -0,0 +1,276 @@
...
+			Target:       *elem.Target,
+			TargetHealth: elem.TargetHealth,
+		})
+	}
+	return listedTargets, nil
+}
+
+// recordSuccessfulDeregisterTargetsOperation will record a successful deregisterTarget operation
+func (m *cachedTargetsManager) recordSuccessfulRegisterTargetsOperation(tgARN string, targets []elbv2sdk.TargetDescription) {
+	rawTargetsCacheItem, exists := m.targetsCache.Get(tgARN)
```
kishorj, line 196: does it not need the targetCacheMutex?

pkg/targetgroupbinding/targets_manager.go
```diff
@@ -0,0 +1,276 @@
...
+		deregisterTargetsChunkSize: defaultDeregisterTargetsChunkSize,
+	}
+}
+
+var _ TargetsManager = &cachedTargetsManager{}
+
+// an cached implementation for TargetsManager.
+// Targets for each TargetGroup will be refreshed per targetsCacheTTL.
+// When list Targets with RefreshTargets list Option set,
+// only targets with ongoing TargetHealth(unknown/initial/draining) TargetHealth will be refreshed.
```
kishorj, line 48: IsOngoing() returns true only for initial/draining. Which one is correct? or ongoing here means something else?

pkg/targetgroupbinding/targets_manager_types.go
```diff
@@ -0,0 +1,43 @@
...
+type TargetInfo struct {
+	// The target's description
+	Target elbv2sdk.TargetDescription
+
+	// The target's health information.
+	// If absent, the target's health information is unknown.
+	TargetHealth *elbv2sdk.TargetHealth
+}
+
+// IsOngoing returns whether target is in ongoing status
```
kishorj, line 19: What does ongoing mean?

pkg/targetgroupbinding/targets_manager.go
```diff
@@ -0,0 +1,276 @@
...
+	// targets is the targets for TargetGroup
+	targets []TargetInfo
+}
+
+func (m *cachedTargetsManager) RegisterTargets(ctx context.Context, tgARN string, targets []elbv2sdk.TargetDescription) error {
+	targetsChunks := chunkTargetDescriptions(targets, m.registerTargetsChunkSize)
+	for _, targetsChunk := range targetsChunks {
+		req := &elbv2sdk.RegisterTargetsInput{
+			TargetGroupArn: aws.String(tgARN),
+			Targets:        pointerizeTargetDescriptions(targetsChunk),
```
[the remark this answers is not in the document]
M00nF1sh: I'm aiming to reduce pointer usage as much as i can given how Golang escape analysis works. (one exception is for k8s types, where it's the pointer implements meta.Object interface)

Also, in future awsSDKGo versions, they will not use such [] to pointer as well.

pkg/targetgroupbinding/targets_manager.go
M00nF1sh: it's based on internal communication with elbv2 team. register/deregister batch size theoretically can be different.

pkg/targetgroupbinding/targets_manager.go
M00nF1sh: fixed :D

pkg/targetgroupbinding/targets_manager.go
M00nF1sh: added UTs for all these functions

pkg/targetgroupbinding/targets_manager.go
M00nF1sh: ideally, we don't need the targetCacheMutex here, since the cache access itself is already synchronized via mutex inside the cache implementation.
Also, for List/Register apicalls for the same tgARN won't happen concurrently, so we don't need to lock on targetCacheMutex here.

However, to make the interface itself secure without checking the implementation. I added a RLock here to make the Register/List for same tgARN safe in semantic too. (it hurts performance a little, ideally it should be a lock per targrtGroup).

pkg/targetgroupbinding/targets_manager.go
M00nF1sh: ongoing means targets in initial/draining. These targets's targetHealth might change, so need to be refreshed.

I changed the implementation to refresh "not healthy" targets instead of "ongoing" targets, to give "unhealthy" targets a chance to turn into "healthy".

pkg/targetgroupbinding/targets_manager_types.go
M00nF1sh: removed ongoing with isHealthy

PR #1343, opened 2020-08-10
V2 ga api el

M00nF1sh:
Add TargetType into TargetGroupBinding API.
The TargetType will be automatically set by mutatingWebhook if unset.

We add this into API types for two reason:
1. user can find the targetType via `kubectl get`
2. the model builders can set the TargetType for TargetGroups created by controller, so the TargetGroupBinding controller can avoid DescribeTargetGroup calls in it's reconcile loop.


apis/elbv2/v1alpha1/targetgroupbinding_types.go
```diff
@@ -102,6 +110,10 @@ type TargetGroupBindingSpec struct {
 	// targetGroupARN is the Amazon Resource Name (ARN) for the TargetGroup.
 	TargetGroupARN string `json:"targetGroupARN"`
 
+	// targetType is the TargetType of TargetGroup. If unspecified, it will be automatically inferred.
+	// +optional
+	TargetType *TargetType `json:"targetType,omitempty"`
```
kishorj, line 115: What will be the default value? Is it upto the model?

M00nF1sh: it will be automatically decided by a mutatingWebhook using DescribeTargetGroup API call if unset.
When model builders creates this TargetGroupBinding, it will set the TargetType to avoid the extra DescribeTargetGroup call.
When users creates this TargetGroupBinding without specify it, it will be set to the TargetType of TargetGroup by mutatingWebhook.

PR #1347, opened 2020-08-14
add resource model for securityGroup and Loadbalancer

M00nF1sh:
Add initial resource model for securityGroup and Loadbalancer.
Dependencies among resources will be captured via the token type.
sample usage
```
func modelBuilder() core.Stack {
	var stack core.Stack
	// stack := ....
	lbSG := ec2.NewSecurityGroup(stack, "lbSG", ec2.SecurityGroupSpec{
		GroupName:   "my-sg",
		Description: awssdk.String("managed by aws-load-balancer-controller"),
		Tags: map[string]string{
			"foo": "bar",
		},
	})
	alb := elbv2.NewLoadBalancer(stack, "lb", elbv2.LoadBalancerSpec{
		LoadBalancerName: "my-lb",
		SecurityGroups:   []core.StringToken{lbSG.GroupID(), core.LiteralStringToken("sg-xxxxxxx")},
		Tags: map[string]string{
			"foo": "bar",
		},
	})
	return stack
}

```

### DIFF
--- a/pkg/model/elbv2/load_balancer.go
+++ b/pkg/model/elbv2/load_balancer.go
@@ -1,0 +1,106 @@
+package elbv2
+
+import (
+	"sigs.k8s.io/aws-alb-ingress-controller/pkg/model/core"
+)
+
+var _ core.Resource = &LoadBalancer{}
+
+// LoadBalancer represents a ELBV2 LoadBalancer.
+type LoadBalancer struct {
+	id string
+
+	// desired state of LoadBalancer
+	spec LoadBalancerSpec `json:"spec"`
+
+	// observed state of LoadBalancer
+	// +optional
+	status *LoadBalancerStatus `json:"status,omitempty"`
+}
+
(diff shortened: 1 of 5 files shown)


pkg/model/elbv2/load_balancer.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// [Network Load Balancers] The private IPv4 address for an internal load balancer.
+	PrivateIPv4Address *string `json:"privateIPv4Address,omitempty"`
+
+	// The ID of the subnet.
+	SubnetID *string `json:"subnetID,omitempty"`
+}
+
+// LoadBalancerSpec defines the desired state of LoadBalancer
+type LoadBalancerSpec struct {
```
kishorj, line 75: Need LoadBalancerType to differentiate between NLB and ALB

M00nF1sh: thanks :D added type here :D

PR #1351, opened 2020-08-18
Annotations parser

kishorj:
Implement annotations parser for V2

### DIFF
--- a/pkg/annotations/parser.go
+++ b/pkg/annotations/parser.go
@@ -25,7 +25,7 @@ func WithExact() ParseOption {
 
 func WithAlternativePrefixes(prefixes ...string) ParseOption {
 	return func(opts *ParseOptions) {
-		opts.alternativePrefixes = prefixes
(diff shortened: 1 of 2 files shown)


pkg/annotations/parser.go
```diff
@@ -35,31 +69,100 @@ type suffixAnnotationParser struct {
...
+	if value != nil {
+		*value = keyValues
+	}
+	return true
+}
+
+// buildAnnotationKey returns list of full annotation keys based on suffix and parse options
+func (p *suffixAnnotationParser) buildAnnotationKeys(suffix string, opts ...ParseOption) []string {
+	keys := []string{}
+	exact := false
```
M00nF1sh, line 148: the idiomatic pattern is
```
opts :=  ParseOptions{
      exact: false, (optional since the zero value is false)
      alternativePrefixes: nil (optional since the zero value is nil)
}
for _, opt := range opts {
      opt(opts)
}
if opts.exact {
     ....
} else {
     ....
}
```

kishorj: ok

PR #1356, opened 2020-08-19
add TargetGroup types

M00nF1sh:
add TargetGroup types for v2 models

### DIFF
--- a/pkg/model/elbv2/target_group.go
+++ b/pkg/model/elbv2/target_group.go
@@ -1,0 +1,128 @@
+package elbv2
+
+import (
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/aws-alb-ingress-controller/pkg/model/core"
+)
+
+var _ core.Resource = &TargetGroup{}
+
+// TargetGroup represents a ELBV2 TargetGroup
+type TargetGroup struct {
+	// resource id
+	id string
+
+	// desired state of TargetGroup
+	spec TargetGroupSpec `json:"spec"`
+
+	// observed state of TargetGroup
+	// +optional
+	status *TargetGroupStatus `json:"status,omitempty"`
+}
+
+// NewLoadBalancer constructs new LoadBalancer resource.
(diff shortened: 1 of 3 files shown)


pkg/model/elbv2/target_group.go
```diff
@@ -0,0 +1,128 @@
...
+	// +optional
+	TimeoutSeconds *int64 `json:"timeoutSeconds,omitempty"`
+
+	// The number of consecutive health checks successes required before considering an unhealthy target healthy.
+	// +optional
+	HealthyThresholdCount *int64 `json:"healthyThresholdCount,omitempty"`
+
+	// The number of consecutive health check failures required before considering a target unhealthy.
+	// +optional
+	UnhealthyThresholdCount int64 `json:"unhealthyThresholdCount,omitempty"`
```
kishorj, line 84: Type is not consistent with other fields.

pkg/model/elbv2/target_group.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+// Information to use when checking for a successful response from a target.
+type HealthCheckMatcher struct {
+	// The HTTP codes.
+	HTTPCode string `json:"httpCode"`
+}
+
+// Configuration for TargetGroup's HealthCheck.
+type TargetGroupHealthCheckConfig struct {
```
kishorj, line 53: Any reason why all fields are pointers in this struct?

pkg/model/elbv2/target_group.go
M00nF1sh: opps, i missed this.

pkg/model/elbv2/target_group.go
M00nF1sh: 1. since they are optional fields in aws API as well. 
so i want to follow this principle around the model: if the type is optional in AWS API, make it optional, and the semantic is it will default to aws's default value server-side. and during reconcile, the settings won't be touched.

2. even the HealthCheck is required for ALB, i make `HealthCheckConfig *TargetGroupHealthCheckConfig` a pointer. So in the future, we can support lambda targets via annotations :D

PR #1359, opened 2020-08-20
On node events, reduce reconciling ingresses requests.

tcmichael:
The pr is same as #1225.
We manage node on demand and we have some fargate instances. So that there are many node events in our system.
Ingress controller receives node change event and reconciles all ingresses. We have about 100 ingresses in our kubernetes cluster. It take about 2-3mins for AWS ingress controller to reconcile with k8s cluster. Once the pod was deleted, it may not deregister targets in time, this results some 5xx(502 and 504)

### DIFF
--- a/internal/ingress/controller/controller.go
+++ b/internal/ingress/controller/controller.go
@@ -28,8 +28,12 @@ import (
 )
 
 func Initialize(config *config.Configuration, mgr manager.Manager, mc metric.Collector, cloud aws.CloudAPI) error {
+	store, err := store.New(mgr, config)
+	if err != nil {
+		return err
+	}
 	authModule := auth.NewModule(mgr.GetCache())
-	reconciler, err := newReconciler(config, mgr, mc, cloud, authModule)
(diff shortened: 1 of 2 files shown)


tcmichael:
BTW, `max-concurrent-reconciles` can also help to avoid some 5xx. Do you have any suggestion for the number of `max-concurrent-reconciles`? @M00nF1sh

M00nF1sh:
@tcmichael sorry for the late response, we were focusing on the v2 implementation.(I remember checked this pr but forget to comment somehow).
Do you still want this PR merged into the old master branch? (the v2 have this functionality built-in).
Some comments about this PR: it enqueues all NodePort service, and can be optimized to enqueue if the service have targetType: instance as well(similar to Ingress)

M00nF1sh:
also, for max-concurrent-reconciles, it should be adjusted according to api throttles you are facing, in v2, we have metrics which can be viewed via promethus. (and ideally you only need to adjust the targetgroupbinding-max-concurrent-reconciles only)

PR #1361, opened 2020-08-21
Model builder for NLB IP mode

kishorj:


pkg/annotations/constants.go
```diff
@@ -7,4 +7,35 @@ const (
...
+	ServiceAnnotationLoadBalancerSecurityGroups                = "aws-load-balancer-security-groups"
+	ServiceAnnotationLoadBalancerCertificate                   = "aws-load-balancer-ssl-cert"
+	ServiceAnnotationLoadBalancerSSLPorts                      = "aws-load-balancer-ssl-ports"
+	ServiceAnnotationLoadBalancerSSLNegotiationPolicy          = "aws-load-balancer-ssl-negotiation-policy"
+	ServiceAnnotationLoadBalancerBEProtocol                    = "aws-load-balancer-backend-protocol"
+	ServiceAnnotationLoadBalancerAdditionalTags                = "aws-load-balancer-additional-resource-tags"
+	ServiceAnnotationLoadBalancerHCHealthyThreshold            = "aws-load-balancer-healthcheck-healthy-threshold"
+	ServiceAnnotationLoadBalancerHCUnhealthyThreshold          = "aws-load-balancer-healthcheck-unhealthy-threshold"
+	ServiceAnnotationLoadBalancerHCTimeout                     = "aws-load-balancer-healthcheck-timeout"
+	ServiceAnnotationLoadBalancerHCInterval                    = "aws-load-balancer-healthcheck-interval"
```
M00nF1sh, line 35: let's make it consist, either HC or HealthCheck

pkg/model/elbv2/load_balancer.go
```diff
@@ -78,6 +78,23 @@ const (
...
+const (
+	DeletionProtectionEnabledKey = "deletion_protection.enabled"
+	AccessLogsS3EnabledKey       = "access_logs.s3.enabled"
+	AccessLogsS3BucketKey        = "access_logs.s3.bucket"
+	AccessLogsS3PrefixKey        = "access_logs.s3.prefix"
+	IdleTimeoutTimeoutSecondsKey = "idle_timeout.timeout_seconds"
+	RoutingHTTP2EnabledKey       = "routing.http2.enabled"
+	CrossZoneLoadBalancing       = "load_balancing.cross_zone.enabled"
+
+	DefaultDeletionProtectionEnabled = false
```
M00nF1sh, line 90: we shouldn't have these default value in the model package.
it should be inside relevant model builder

pkg/model/elbv2/target_group.go
```diff
@@ -5,6 +5,8 @@ import (
 	"sigs.k8s.io/aws-alb-ingress-controller/pkg/model/core"
 )
 
+const ProxyProtocolV2Enabled = "proxy_protocol_v2.enabled"
```
M00nF1sh, line 8: same comments as attributes for LoadBalancer

pkg/model/nlb/builder.go
```diff
@@ -0,0 +1,314 @@
+package nlb
+
```
M00nF1sh, line 2: i prefer the builder to be not inside model package.(which is only for model)

how about `/pkg/service/nlb_model_builder.go` or `/pkg/service/nlb/model_builder.go`

PR #1366, opened 2020-08-24
rewrite TargetGroup filter using ELBV2 tag operation

Fedomn:
This PR to support controller GC TargetGroup in the internal VPC that without internet access ability.

So i replace Resource Groups Tagging API to ELBV2 DescribeTargetGroups API in targetgroup_group GC function.

Same purpose as #1109 

Thanks


Fedomn:
@M00nF1sh any update about this PR, or i need to migrate this code to v2 first ?

M00nF1sh:
@Fedomn 
sorry for the late reply, we were actively working on v2 and haven't commented on this PR.
The problem with this solution is TargetGroups are created first before attach to LoadBalancers. If there are any error happens before creating listener/listener rules(e.g. a missing TLS certificate), these created TargetGroups will be leaked.

BTW, the v2 already supports private cluster. (it's using ELBv2's tag operation). Do you still want to improve this PR and get merged in v1 or migrate to use v2 😄 .

Fedomn:
@M00nF1sh thanks for your quick reply, i will improve this PR and use solution same as [ELBV2 tagging operations](https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/1387/files#diff-02e2dfc7a036f6d080a920261df60c92ed7cf0822417c2fa69e86251f218eb50R165)
I think this solution may solve the dangling TargetGroups issue

Fedomn:
just close this PR for v1.
because the v2 already supports private cluster.

PR #1368, opened 2020-08-24
e2e tests for k8s service

kishorj:
* e2e tests for NLB IP create/delete/traffic tests
* e2e tests for NLB IP with TLS annotations create/delete/traffic tests


kishorj:
Will base the changes on the new e2e framework.

PR #1373, opened 2020-08-27
reconciler and eventhandler framework for service objects

kishorj:
* reconciler and eventhandler for service objects
* port finalizer utils from aws-app-mesh-controller-for-k8s


controllers/service/service_controller.go
```diff
@@ -0,0 +1,99 @@
...
+	log              logr.Logger
+	annotationParser annotations.Parser
+	finalizerManager k8s.FinalizerManager
+}
+
+// +kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=services/status,verbs=get;update;patch
+func (r *ServiceReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	ctx := context.Background()
+	_ = r.log.WithValues("service", req.NamespacedName)
```
M00nF1sh, line 45: i think we can remove this line if the return value `_ = r.log.WithValues("service", req.NamespacedName)` is not used.

PR #1390, opened 2020-09-04
Fix security group reconciliation loop

glenonn:
See:
https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/1094

This fix combines ipv4 and ipv6 security group rules into one
IPPermission.


glenonn:
signed the CNCF CLA

PR #1399, opened 2020-09-09
add pod/node/service utility and endpoint resolver

M00nF1sh:
1. add pod/node/service utility that deal with object conditions
2. add endpoint resolver to resolve endpoints for services.

### DIFF
--- a/pkg/backend/endpoint_resolver.go
+++ b/pkg/backend/endpoint_resolver.go
@@ -1,0 +1,186 @@
+package backend
+
+import (
+	"context"
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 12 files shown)


pkg/backend/endpoint_resolver.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func extractNodeInstanceID(node *corev1.Node) (string, error) {
+	providerID := node.Spec.ProviderID
+	if providerID == "" {
+		return "", errors.Errorf("providerID is not specified for node: %s", node.Name)
+	}
+
+	p := strings.Split(providerID, "/")
+	return p[len(p)-1], nil
```
kishorj, line 169: what if providerID doesn't contain '/' character?

kishorj (approved):

pkg/backend/endpoint_resolver.go
M00nF1sh: if it don't contain '/', then full providerID is returned :D

PR #1400, opened 2020-09-09
Add cmdline flags aws-vpc-id and cluster-name

kishorj:
Add support for additional cmdline flags aws-vpc-id and k8s-cluster-name. Discover vpc-id from metadata.

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,20 +17,28 @@ limitations under the License.
 package main
 
 import (
-	"flag"
-	"os"
-	"sigs.k8s.io/aws-alb-ingress-controller/controllers/ingress"
-	"sigs.k8s.io/aws-alb-ingress-controller/controllers/service"
-
+	"github.com/spf13/pflag"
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -17,22 +17,31 @@ limitations under the License.
...
 
 	elbv2v1alpha1 "sigs.k8s.io/aws-alb-ingress-controller/apis/elbv2/v1alpha1"
 	elbv2controller "sigs.k8s.io/aws-alb-ingress-controller/controllers/elbv2"
 	// +kubebuilder:scaffold:imports
 )
 
+const (
+	flagMetricsAddr          = "metrics-addr"
+	flagEnableLeaderElection = "enable-leader-election"
+	flagK8sClusterName       = "k8s-cluster-name"
```
M00nF1sh, line 42: let's keep it  as `cluster-name` to remain compatible with old controller's flags

kishorj: done

M00nF1sh (approved):
/lgtm

PR #1401, opened 2020-09-10
Integrate service controller with model deployer

kishorj:
Verified SVC controller creates/updates/deletes Loadbalancer and target group resources.


M00nF1sh (approved):
/lgtm

PR #1421, opened 2020-09-15
add reconciler for securityGroup rules

M00nF1sh:
1.  networking.SecurityGroupInfo wraps the information about a securityGroup.
    * we expand the raw EC2IPPermission into multiple ones, so that each one contain only a single source.(ipRange/ipV6Range/etc).
    * for each expanded EC2IPPermission, we compute a set of labels from description(treat the description as comma-separated kv pairs). also, a special `raw/description` label is added for the raw description.
2.  networking.SecurityGroupManager provides cached operations for securityGroups. 
3. networking.SecurityGroupReconciler will reconcile permissions for securityGroups.
4. targetGroupBinding.NetworkingManager will manage the network settings for targetGroupBindings.


pkg/targetgroupbinding/networking_manager.go
```diff
@@ -2,36 +2,348 @@ package targetgroupbinding
...
+	for _, tgbNetworking := range tgbNetworkings {
+		for _, rule := range tgbNetworking.Ingress {
+			for _, port := range rule.Ports {
+				for _, peer := range rule.From {
+					ipPermission, err := m.computeDesiredIngressIPPermission(port, peer)
+					if err != nil {
+						return nil, err
+					}
+					containsPermission := false
+					for _, permission := range ipPermissions {
```
kishorj, line 192: what is the expected number of permissions in the worst case? 
This could cause a potential bottleneck.

pkg/targetgroupbinding/resource_manager.go
```diff
@@ -22,23 +23,29 @@ type ResourceManager interface {
 }
 
 // NewDefaultResourceManager constructs new defaultResourceManager.
-func NewDefaultResourceManager(k8sClient client.Client, elbv2Client services.ELBV2, logger logr.Logger) *defaultResourceManager {
+func NewDefaultResourceManager(k8sClient client.Client, elbv2Client services.ELBV2,
```
kishorj, line 26: Did you modify the main.go as well? if so please include it together.

pkg/targetgroupbinding/networking_manager.go
M00nF1sh: humm, one securityGroup only have at most 10X permissions.
calculations like these will never be bottleneck :D (networking is is)

pkg/targetgroupbinding/resource_manager.go
M00nF1sh: update the PR to modify main.go

kishorj (approved):

PR #1423, opened 2020-09-16
add networking config to NLB targetgroupbinding

kishorj:
Modify networking.DiscoverSubnets() to return []*ec2.Subnet
Add Subnet/Port to TargetGropuBinding for NLB
   Port is added so networking_manager is able to add the sg rules, will be removed later
Verified SG rules get added for the subnets


M00nF1sh (approved):
/lgtm

PR #1427, opened 2020-09-16
add support for pod mutator webhook

kishorj:
Add pod mutator webhook to inject readiness gate to the pod spec.
Mutator will remove any v1 readiness gates with prefixes target-health.alb.ingress.k8s.aws, so that existing pod sepc will be compatible with the v2 release. Any existing v2 readiness gates are also cleaned up to keep the readiness gates in sync with the targetgroup bindings configured.

Targetgroupbindings referring nonexistent services are ignored.

### DIFF
--- a/main.go
+++ b/main.go
@@ -28,9 +28,11 @@ import (
 	"sigs.k8s.io/aws-alb-ingress-controller/controllers/service"
 	"sigs.k8s.io/aws-alb-ingress-controller/pkg/aws"
 	"sigs.k8s.io/aws-alb-ingress-controller/pkg/aws/throttle"
+	podinjector "sigs.k8s.io/aws-alb-ingress-controller/pkg/inject"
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -136,6 +139,11 @@ func main() {
 		setupLog.Error(err, "Unable to create controller", "controller", "Service")
 		os.Exit(1)
 	}
+
+	if enablePodReadinessWebhook {
```
M00nF1sh, line 143: i think it's better to move such configuration enablePodReadinessWebhook  inside inject itself.(similar style as our appmesh's inject configurations like sidecar image).

also, the injector's setupWithManager should always run no matter it's enabled or not. since it setups the webhook servers, if it not running and our webhook's manifests is installed, pod creation requests will be blocked

kishorj: sounds fair. I will move it to inject.

PR #1431, opened 2020-09-17
adjust the deployment manifests according to best practices

M00nF1sh:
1. adjust the kustomize deployment YAMLs according to label/kustomize best practices. (also formatted these yamls)
   * used controller instead of manager
   * used `app.kubernetes.io/component` and `app.kubernetes.io/name`
   * removed `namespaces` from objects
2. adjust the RBAC permissions on controllers

# Test done:
deployed via new generated YAMLs


Dockerfile
```diff
@@ -11,17 +11,18 @@ RUN go mod download
 
 # Copy the go source
 COPY main.go main.go
-COPY api/ api/
+COPY apis/ apis/
 COPY controllers/ controllers/
+COPY pkg/ pkg/
```
kishorj, line 16: Also need 
COPY webhooks/ webhooks/

kishorj (approved):

Dockerfile
M00nF1sh: thanks 👍

kishorj (approved):

PR #1432, opened 2020-09-18
add support for specifying eip allocations

kishorj:
Add support for specifying EIP allocations via the annotation service.beta.kubernetes.io/aws-load-balancer-eip-allocations.
Tests
1. Verified using the following spec NLB got provisioned with the correct EIP allocation 
```yaml
kind: Service
apiVersion: v1
metadata:
  name: nlb-ip-svc
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-type: "nlb-ip"
    service.beta.kubernetes.io/aws-load-balancer-eip-allocations: "eipalloc-xxx, eipalloc-yyy"
spec:
```
2. Invalid allocation id results in error


M00nF1sh (approved):
/lgtm

PR #1445, opened 2020-09-25
 Add TargetGroupBinding networking rule for healthcheck port 

kishorj:
Refactor NLB model builder to use modelBuildTask to share state
Add TargetGroupBinding networking rule for healthcheck port


M00nF1sh (approved):
/lgtm

PR #1451, opened 2020-09-26
Make the docker container run as non-root & redact OIDC clientID and clientSecret

M00nF1sh:
1. Make the docker container run as non-root(1002 UID)
2. redact OIDC clientID and clientSecret

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,7 @@ RUN --mount=type=bind,target=. \
 
 FROM amazonlinux:2 as bin-unix
 COPY --from=build /out/controller /controller
+USER 1002
 ENTRYPOINT ["/controller"]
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -19,6 +19,7 @@ RUN --mount=type=bind,target=. \
 
 FROM amazonlinux:2 as bin-unix
 COPY --from=build /out/controller /controller
+USER 1002
```
kishorj, line 22: why 1002? does this id have to exist or any random id is ok?

kishorj (approved):

Dockerfile
M00nF1sh: random ID works too, i picked 1002 since it's same as distroless's non-root user

PR #1453, opened 2020-09-29
add sample IAM permission

M00nF1sh:
add sample IAM permission needed for the controller while Restricted the permissions where ever possible.

TODO: the permissions can be further scoped down when the vpcID / clusterName can be provided, we should document how to further scope down permissions.

### DIFF
--- a/docs/install/iam_policy.json
+++ b/docs/install/iam_policy.json
@@ -1,0 +1,226 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
+                "iam:CreateServiceLinkedRole",
+                "ec2:DescribeAccountAttributes",
+                "ec2:DescribeAddresses",
+                "ec2:DescribeInternetGateways",
+                "ec2:DescribeVpcs",
+                "ec2:DescribeSubnets",
+                "ec2:DescribeSecurityGroups",
+                "ec2:DescribeInstances",
+                "ec2:DescribeNetworkInterfaces",
+                "ec2:DescribeTags",
+                "elasticloadbalancing:DescribeLoadBalancers",
+                "elasticloadbalancing:DescribeLoadBalancerAttributes",
+                "elasticloadbalancing:DescribeListeners",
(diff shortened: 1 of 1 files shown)


astrived:
Do we have write up on how to further scope down when vpcID / clusterNam ?

docs/install/iam_policy.json
```diff
@@ -0,0 +1,226 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
+                "iam:CreateServiceLinkedRole",
+                "ec2:DescribeAccountAttributes",
+                "ec2:DescribeAddresses",
+                "ec2:DescribeInternetGateways",
```
astrived, line 10: Do we need IGW Describe ?

M00nF1sh: > Do we have write up on how to further scope down when vpcID / clusterName are provided ?

no, it's a TODO item.

M00nF1sh: > Do we need IGW Describe ?

It's need by ELB implicitly during CreateLoadBalancer call: https://docs.aws.amazon.com/elasticloadbalancing/latest/userguide/elb-api-permissions.html

astrived: LGTM

astrived (approved):

docs/install/iam_policy.json
```diff
@@ -0,0 +1,226 @@
...
+                "elasticloadbalancing:ModifyTargetGroup",
+                "elasticloadbalancing:ModifyTargetGroupAttributes",
+                "elasticloadbalancing:RegisterTargets",
+                "elasticloadbalancing:DeregisterTargets",
+                "elasticloadbalancing:DeleteTargetGroup"
+            ],
+            "Resource": "*",
+            "Condition": {
+                "Null": {
+                    "aws:ResourceTag/ingress.k8s.aws/cluster": "false"
```
kishorj, line 190: How does backwards compatibility work? Do we recommend these IAM permissions for new setups only?

kishorj (approved):

PR #1457, opened 2020-09-29
Change the format for NLB targetgroup naming

kishorj:
Use the format k8s-<namespace>-<name>-<uuid> to be consistent with the ingress target groups.


M00nF1sh (approved):
/lgtm

PR #1462, opened 2020-09-30
enable leader election by default

kishorj:
Enable leader election by default
Use aws-load-balancer-controller-leader as the default leader election id


M00nF1sh (approved):
/lgtm

PR #1492, opened 2020-10-13
Add support for specifying source ranges for NLB

kishorj:
The source IP gets preserved for UDP in IP mode, so the security group rules also need to provide access from the IP ranges outside of the VPC CIDRs. The source ranges is obtained from the service `Spec.LoadBalancerSrcRange`. If the `Spec.LoadBalancerSrcRange` not defined, it is read from the service annotation `service.beta.kubernetes.io/load-balancer-source-ranges`. The default range `0.0.0.0/0` gets used if the range is not configured.

### DIFF
--- a/pkg/service/nlb/builder.go
+++ b/pkg/service/nlb/builder.go
@@ -511,16 +511,16 @@ func (t *defaultModelBuildTask) buildPeersFromSourceRanges(_ context.Context) []
 func (t *defaultModelBuildTask) buildTargetGroupBindingNetworking(ctx context.Context, tgPort intstr.IntOrString, hcPort intstr.IntOrString,
 	tgProtocol corev1.Protocol, ec2Subnets []*ec2.Subnet) *elbv2model.TargetGroupBindingNetworking {
 	var fromVPC []elbv2model.NetworkingPeer
-	networkingProtocol := elbv2api.NetworkingProtocolTCP
-	if tgProtocol == corev1.ProtocolUDP {
-		networkingProtocol = elbv2api.NetworkingProtocolUDP
-	}
 	for _, subnet := range ec2Subnets {
 		fromVPC = append(fromVPC, elbv2model.NetworkingPeer{
 			IPBlock: &elbv2api.IPBlock{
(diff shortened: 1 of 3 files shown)


pkg/service/nlb/builder.go
```diff
@@ -486,41 +486,76 @@ func (t *defaultModelBuildTask) buildTargetGroupBinding(ctx context.Context, tar
...
 		Ingress: []elbv2model.NetworkingIngressRule{
 			{
-				From:  from,
-				Ports: ports,
+				From:  trafficSource,
+				Ports: trafficPorts,
 			},
 		},
 	}
+	if tgProtocol == corev1.ProtocolUDP || hcPort.String() != healthCheckPortTrafficPort && hcPort.IntValue() != tgPort.IntValue() {
```
M00nF1sh, line 543: let's change it to be `if tgProtocol == corev1.ProtocolUDP || (hcPort.String() != healthCheckPortTrafficPort && hcPort.IntValue() != tgPort.IntValue())`. 
Even we have operator proiority `&&` > `||`, adding `()` will make it clear to everyone

M00nF1sh (commented):
/lgtm

pkg/service/nlb/builder.go
```diff
@@ -486,41 +486,76 @@ func (t *defaultModelBuildTask) buildTargetGroupBinding(ctx context.Context, tar
...
+			},
+		})
+	}
+	return peers
+}
+
+func (t *defaultModelBuildTask) buildTargetGroupBindingNetworking(ctx context.Context, tgPort intstr.IntOrString, hcPort intstr.IntOrString,
 	tgProtocol corev1.Protocol, ec2Subnets []*ec2.Subnet) *elbv2model.TargetGroupBindingNetworking {
-	var from []elbv2model.NetworkingPeer
+	var fromVPC []elbv2model.NetworkingPeer
```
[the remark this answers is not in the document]
kishorj: updated

pkg/service/nlb/builder.go
kishorj: done

M00nF1sh (approved):

PR #1497, opened 2020-10-14
Fix addons_config default values

kishorj:
Fix https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/1496


M00nF1sh (approved):
/lgtm

PR #1500, opened 2020-10-14
Add e2e tests for NLB IP mode

kishorj:
Run e2e tests with the following arguments 
--kubeconfig=<kubeconfig_path> --cluster-name=<k8s cluster name> --aws-region=<aws region> --aws-vpc-id=<AWS VPC ID>

### DIFF
--- a/test/framework/utils/name.go
+++ b/test/framework/utils/name.go
@@ -14,5 +14,5 @@ func RandomDNS1123Label(length int) string {
 
 	labelBuf := make([]byte, seedLen*2)
 	hex.Encode(labelBuf, seedBuf)
-	return string(labelBuf[:length])
+	return "a" + string(labelBuf[:length-1])
 }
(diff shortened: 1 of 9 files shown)


test/framework/utils/name.go
```diff
@@ -14,5 +14,5 @@ func RandomDNS1123Label(length int) string {
 
 	labelBuf := make([]byte, seedLen*2)
 	hex.Encode(labelBuf, seedBuf)
-	return string(labelBuf[:length])
```
M00nF1sh, line 17: why do we change this RandomDNS1123Label? if we want a a prefix, we should use `fmt.Sprintf("a%v, RandomDNS1123Label(length-1))`

M00nF1sh (approved):
lgtm

test/framework/utils/name.go
kishorj: RFC 1123 compliant labels cannot begin with a numeric character. Prepended "a" to force compliance.

PR #1503, opened 2020-10-14
Fix NLB model for supporting multiple certificate ARNs

kishorj:
This bug was preventing from supporting multiple certificate ARNs in the NLB model.


M00nF1sh (approved):
/lgtm

PR #1522, opened 2020-10-19
Target group name based on target port

kishorj:
Use target port while computing target group name hash.
If target port is string, use 1 as the target port.


M00nF1sh (approved):
/lgtm

PR #1523, opened 2020-10-19
use numeric target group if available

kishorj:
For ALB target group port, use numeric value if available. If not, use 1.

### DIFF
--- a/pkg/ingress/model_build_target_group.go
+++ b/pkg/ingress/model_build_target_group.go
@@ -94,18 +94,6 @@ func (t *defaultModelBuildTask) buildTargetGroupBindingNetworking(_ context.Cont
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/ingress/model_build_target_group.go
```diff
@@ -118,10 +118,14 @@ func (t *defaultModelBuildTask) buildTargetGroupSpec(ctx context.Context,
 		return elbv2model.TargetGroupSpec{}, err
 	}
 	name := t.buildTargetGroupName(ctx, k8s.NamespacedName(ing), svc, port, targetType, tgProtocol)
+	targetGroupPort := 1
```
M00nF1sh, line 121: the port here is the servicePort's name or port.
we should use 
```
svcPort, err := k8s.LookupServicePort(svc, port)
targetPort := svcPort.TargetPort
```

M00nF1sh (approved):
/lgtm

PR #1525, opened 2020-10-20
Update slow_start.duration_seconds example to specify a valid value.

larosek:
Documentation is suggesting to set a 5 seconds slow_start.duration_seconds but this value is not valid in AWS.
See doc here: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#target-group-attributes


M00nF1sh:
@larosek 
Thanks for making this change, would you mind do this change against [main branch](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/tree/main). 
we are going to release a major new version and make [main branch] the default branch

PR #1526, opened 2020-10-20
Tighten RBAC permissions

kishorj:


controllers/elbv2/targetgroupbinding_controller.go
```diff
@@ -67,14 +67,14 @@ type targetGroupBindingReconciler struct {
 }
 
 // +kubebuilder:rbac:groups=elbv2.k8s.aws,resources=targetgroupbindings,verbs=get;list;watch;update;patch;create;delete
-// +kubebuilder:rbac:groups=elbv2.k8s.aws,resources=targetgroupbindings/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=elbv2.k8s.aws,resources=targetgroupbindings/status,verbs=update;patch
 // +kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch;update;patch
```
M00nF1sh, line 71: for pods, we should only need get/list/watch. but not update/patch. 
can we test whether without pod's update/patch, we can still patch pod status?

kishorj: sure

kishorj: don't need update/patch for pod for the status updates.

M00nF1sh (approved):
/lgtm

PR #1536, opened 2020-10-22
Update slow_start.duration_seconds example to specify a valid value.

larosek:
Documentation is suggesting to set a 5 seconds slow_start.duration_seconds but this value is not valid in AWS.
See doc here: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#target-group-attributes


M00nF1sh:
thanks dude

PR #1554, opened 2020-10-25
Undo resourceName restriction on leader election role

kishorj:
The prior `resourceName` constraint on the leader election role doesn't work as intended since we cannot restrict create request by `resourceName`. For create, this limitation is because the object name is not known at authorization time.

For further reference https://kubernetes.io/docs/reference/access-authn-authz/rbac/

### DIFF
--- a/config/rbac/leader_election_role.yaml
+++ b/config/rbac/leader_election_role.yaml
@@ -8,8 +8,6 @@ rules:
       - ""
     resources:
       - configmaps
-    resourceNames:
-      - aws-load-balancer-controller-leader
     verbs:
       - create
       - get
--- a/docs/install/v2_0_0_full.yaml
+++ b/docs/install/v2_0_0_full.yaml
@@ -259,8 +259,6 @@ metadata:
 rules:
   - apiGroups:
       - ""
-    resourceNames:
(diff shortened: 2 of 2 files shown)


config/rbac/leader_election_role.yaml
```diff
@@ -8,8 +8,6 @@ rules:
       - ""
```
fawadkhaliq, line 8: Keep the resource name for `get`, `patch`, `update`. Use this:
```
metadata:
  name: controller-leader-election-role
rules:
- apiGroups:
  - ""
  resources:
  - configmaps
  verbs:
  - list
  - watch
  - create
- apiGroups:
  - ""
  resources:
  - configmaps
  resourceNames:
  - aws-load-balancer-controller-leader
  verbs:
  - get
  - update
  - patch
```

M00nF1sh (approved):
lgtm

config/rbac/leader_election_role.yaml
M00nF1sh: @fawadkhaliq
I restricted them via https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/1555.

PR #1574, opened 2020-10-28
Update echo_server.md

posquit0:
I think `aws-load-balancer-controller` is better. 👀


M00nF1sh:
aha 😄 we merely copied this from old docs.

thanks 😄 
/lgtm

PR #1588, opened 2020-10-29
Remove circular redirect on docs page

rothgar:
I heard complaints from people (including myself) who clicked on this link from the live docs and frustrated when it links to itself.

The main repo README still has a link to the live docs page.


rothgar:
woops, I rebased master instead of main. I'll open a new PR

PR #1589, opened 2020-10-30
feat: add support to set ProtocolVersion based on annotation values

### LINKED ISSUE #1586: [v2] feature request: support ProtocolVersion on TargetGroups
Support for gRPC and HTTP/2 between ALB and targets was just announced https://aws.amazon.com/blogs/aws/new-application-load-balancer-support-for-end-to-end-http-2-and-grpc 🎉 . 

It would be useful to have an annotation - perhaps `alb.ingress.kubernetes.io/backend-protocol-version` - that allows service /ingress objects to specify the backend version that they wish to use.

backjo:
Closes #1586

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes-sigs/aws-load-balancer-controller/1589)

### DIFF
--- a/pkg/deploy/elbv2/target_group_manager_test.go
+++ b/pkg/deploy/elbv2/target_group_manager_test.go
@@ -394,6 +394,44 @@ func Test_buildSDKCreateTargetGroupInput(t *testing.T) {
 				TargetType:                 awssdk.String("ip"),
 			},
 		},
+		{
+			name: "standard case with protocol version",
+			args: args{
+				tgSpec: elbv2model.TargetGroupSpec{
+					Name:            "my-tg",
+					TargetType:      elbv2model.TargetTypeIP,
+					Port:            8080,
+					Protocol:        elbv2model.ProtocolHTTP,
+					ProtocolVersion: elbv2model.ProtocolVersionHTTP2,
+					HealthCheckConfig: &elbv2model.TargetGroupHealthCheckConfig{
+						Port:                    &port9090,
+						Protocol:                &protocolHTTP,
+						Path:                    awssdk.String("/healthcheck"),
+						Matcher:                 &elbv2model.HealthCheckMatcher{HTTPCode: "200"},
+						IntervalSeconds:         awssdk.Int64(10),
+						TimeoutSeconds:          awssdk.Int64(5),
+						HealthyThresholdCount:   awssdk.Int64(3),
+						UnhealthyThresholdCount: awssdk.Int64(2),
+					},
+				},
+			},
+			want: &elbv2sdk.CreateTargetGroupInput{
+				HealthCheckEnabled:         awssdk.Bool(true),
+				HealthCheckIntervalSeconds: awssdk.Int64(10),
+				HealthCheckPath:            awssdk.String("/healthcheck"),
+				HealthCheckPort:            awssdk.String("9090"),
+				HealthCheckProtocol:        awssdk.String("HTTP"),
+				HealthCheckTimeoutSeconds:  awssdk.Int64(5),
+				HealthyThresholdCount:      awssdk.Int64(3),
+				Matcher:                    &elbv2sdk.Matcher{HttpCode: awssdk.String("200")},
+				UnhealthyThresholdCount:    awssdk.Int64(2),
+				Name:                       awssdk.String("my-tg"),
+				Port:                       awssdk.Int64(8080),
+				Protocol:                   awssdk.String("HTTP"),
+				ProtocolVersion:            awssdk.String("HTTP2"),
+				TargetType:                 awssdk.String("ip"),
+			},
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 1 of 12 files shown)


pkg/ingress/model_build_target_group.go
```diff
@@ -184,6 +189,23 @@ func (t *defaultModelBuildTask) buildTargetGroupProtocol(_ context.Context, svcA
...
+	rawBackendProtocolVersion := string(t.defaultBackendProtocolVersion)
+	_ = t.annotationParser.ParseStringAnnotation(annotations.IngressSuffixBackendProtocolVersion, &rawBackendProtocolVersion, svcAndIngAnnotations)
+	switch rawBackendProtocolVersion {
+	case string(elbv2model.ProtocolVersionHTTP1):
+		return elbv2model.ProtocolVersionHTTP1, nil
+	case string(elbv2model.ProtocolVersionHTTP2):
+		return elbv2model.ProtocolVersionHTTP2, nil
+	case string(elbv2model.ProtocolVersionGRPC):
+		return elbv2model.ProtocolVersionGRPC, nil
+	case string(elbv2model.ProtocolVersionNone):
```
M00nF1sh, line 202: we shouldn't have protocolVersionNone. since there is no need to override it to be empty via annotation.
If this annotation presents, it should always be a valid protocol version.

pkg/ingress/model_builder.go
```diff
@@ -99,6 +99,7 @@ func (b *defaultModelBuilder) Build(ctx context.Context, ingGroup Group) (core.S
 		defaultSSLPolicy:                          "ELBSecurityPolicy-2016-08",
 		defaultTargetType:                         elbv2model.TargetTypeInstance,
 		defaultBackendProtocol:                    elbv2model.ProtocolHTTP,
+		defaultBackendProtocolVersion:             "",
```
M00nF1sh, line 102: i think we can set default protocol version to be HTTP1 here.

pkg/model/elbv2/listener.go
```diff
@@ -69,6 +69,15 @@ const (
 	ProtocolTCP_UDP Protocol = "TCP_UDP"
 )
 
+type ProtocolVersion string
+
+const (
+	ProtocolVersionHTTP1 ProtocolVersion = "HTTP1"
+	ProtocolVersionHTTP2 ProtocolVersion = "HTTP2"
+	ProtocolVersionGRPC  ProtocolVersion = "GRPC"
+	ProtocolVersionNone  ProtocolVersion = ""
```
M00nF1sh, line 78: i don't think we should have ProtocolVersionNone. let's make ProtocolVersion a `*ProtocolVersion` in the model since it's optional settings. (especially for tcp/tls case)

pkg/model/elbv2/target_group.go
```diff
@@ -121,6 +121,10 @@ type TargetGroupSpec struct {
 	// The protocol to use for routing traffic to the targets.
 	Protocol Protocol `json:"protocol"`
 
+	// The target group protocol version.
+	// +optional
+	ProtocolVersion ProtocolVersion `json:"protocolVersion,omitempty"`
```
M00nF1sh, line 126: let's make this to be `*ProtocolVersion`. all optional field should be *Type. even for types with sensible zero value.

M00nF1sh (changes requested):
Thanks for add support for this. overall looks good. changes needed in addition to the inline comments 😄 :
1. we should change the `buildTargetGroupName` in Ingress's model builder to have a protocolVersion passed-in, and use it as a hash component. (since protocolVersion is immutable, we need to create a targetGroup with different name if ppl changes it).

2. we should change the `isSDKTargetGroupRequiresReplacement` in  `deployment/elbv2/target_group_synthesizer.go` to be like 
```
if resTG.Spec.ProtocolVersion != nil {
		if awssdk.StringValue(resTG.Spec.ProtocolVersion) != awssdk.SrtingValue(sdkTG.TargetGroup.ProtocolVersion) {
			return true
		}
}
```

backjo:
sounds good - will tackle these shortly.

backjo:
Addressed the review comments

pkg/deploy/elbv2/target_group_manager_test.go
```diff
@@ -350,6 +350,7 @@ func Test_isSDKTargetGroupHealthCheckDrifted(t *testing.T) {
 func Test_buildSDKCreateTargetGroupInput(t *testing.T) {
 	port9090 := intstr.FromInt(9090)
 	protocolHTTP := elbv2model.ProtocolHTTP
+	procotolHTTP2 := elbv2model.ProtocolVersionHTTP2
```
M00nF1sh, line 353: typo, maybe protocolHTTP2 or better protocolVersionHTTP2

backjo: d'oh

M00nF1sh (approved):
/lgtm

M00nF1sh:
thanks soo much dude,  looks perfect now 👍

PR #1607, opened 2020-11-03
fix egrep script in walkthrough doc to correctly filter aws-load-balancer-controller

ysam12345:
### Description
Found the egrep script is missing apostrophe so it will fail to get the log of pod "aws-load-balancer-controller".
```
# original
kubectl logs -n kube-system $(kubectl get po -n kube-system | egrep -o aws-load-balancer-controller[a-zA-Z0-9-]+)

# fixed
kubectl logs -n kube-system $(kubectl get po -n kube-system | egrep -o 'aws-load-balancer-controller[a-zA-Z0-9-]+')
```
### Test
image

### DIFF
--- a/docs/guide/walkthrough/echo_server.md
+++ b/docs/guide/walkthrough/echo_server.md
@@ -45,7 +45,7 @@ In this walkthrough, you'll
 1.  Verify the deployment was successful and the controller started.
 
     ```bash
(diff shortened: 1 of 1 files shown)


ysam12345:
I signed it

kishorj (approved):

PR #1614, opened 2020-11-04
Clarify documentation on ALB sticky session support

otterley:
Sticky sessions are only supported when the target mode is `ip`.


M00nF1sh (approved):
/lgtm
thanks for improving this :D

PR #1624, opened 2020-11-05
Update service annotations documentation

otterley:
* Add a note regarding proxy-protocol annotation as to how to enable it
* Clarify string encoding rules
* Clean up Markdown table


kishorj (approved):
Thanks for the contribution.

PR #1625, opened 2020-11-06
Update prerequisites for NLB-IP mode

kishorj:
Mention that NLB-IP mode is also supported in EKS version 1.18 or later.


M00nF1sh (approved):
/lgtm

Issue #1611, opened 2020-11-04
[v2] Question: Additional Tags on Target Groups


alexandrud:
Hi,

In v1 there were a couple of tags set on target groups "kubernetes.io/service-port" and "kubernetes.io/service-name". This is no longer set in v2. For us it was useful because we have scripts that look at it and try to connect to the services in kubernetes to read swagger. Based on that info we configure WAFv2 rules.

Is there a way to re-add that tag on the target groups?

Thank you!

kishorj:
These are the new tags that we currently have 
```
ingress.k8s.aws/stack  <namespace/ingess-name>
ingress.k8s.aws/resource  <namespace/service_name:service_port>
elbv2.k8s.aws/cluster <cluster_name>
```

In addition you can also specify additional tags via the annotation on the ingress or service resources
`alb.ingress.kubernetes.io/tags`

taoxinyi:
@kishorj I tried to add my tags in `alb.ingress.kubernetes.io/tags` in my service file for `nlb-ip` but the target group still only has the 3 tags you mentioned.

kishorj:
The `alb.ingress.kubernetes.io/*` annotations applies for ALB only not NLB.

For NLB, this annotation `service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags` currently apples the tags to the load balancer resource only. I will put together a fix to apply this annotation to the target group resources as well.

taoxinyi:
@kishorj Thank you very much.

Issue #1570, opened 2020-10-27
[Feature Request] Support Target Group Attributes for NLB IP Mode


arhea:
The Ingress configuration supports the `alb.ingress.kubernetes.io/target-group-attributes` annotation which enables configuring `stickiness.enabled`, `stickiness.type`,  `deregistration_delay.timeout_seconds`, `load_balancing.algorithm.type`, `slow_start.duration_seconds`, `stickiness.lb_cookie.duration_seconds`, and `proxy_protocol_v2.enabled`. 

Some of these features are also available on the NLB such as configuring `stickiness.enabled`, `stickiness.type`,  `deregistration_delay.timeout_seconds`, and `proxy_protocol_v2.enabled`. It would be helpful to have an `service.beta.kubernetes.io/target-group-attributes` annotation for configuring these options on NLB IP based services.

There is a workaround to statically create the NLB and Target Group outside of Kubernetes then use the `TargetGroupBinding` CRD to link the Kubernetes service to the Target Group.

Key use cases include:
- sticky sessions for things like WebSocket based connections to pods
- sticky sessions for in memory session state

kishorj:
proxy_protocol_v2.enabled is already supported
For other attributes, we will add support for configuration via service annotations.
/kind feature

M00nF1sh:
Hi, TargetGroupsAttributes support is already released in v2.1.0. closing this

PR #1632, opened 2020-11-10
Add support for specifying NLB target group attributes

kishorj:
Add support for a new service annotation `service.beta.kubernetes.io/aws-load-balancer-target-group-attributes` of type stringMap to allow configuring of NLB target group attributes. Issue #1570, #1611.

Tests
- Verified that existing ingress/services got reconciled as expected
- Added the following annotation to the service and verified target group attributes got configured appropriately
  `service.beta.kubernetes.io/aws-load-balancer-target-group-attributes: stickiness.enabled=true,deregistration_delay.timeout_seconds=150,stickiness.type=source_ip`
- Verified the annotation  `service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: '*'` gets precedence over the attriibutes specified


elijah-roberts:
+1 for this feature. Our company is waiting on this feature to use the nlb-ip type lb. @M00nF1sh

pkg/service/model_build_load_balancer.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	return spec, nil
+}
+
+func (t *defaultModelBuildTask) buildLoadBalancerScheme(_ context.Context) (elbv2model.LoadBalancerScheme, error) {
+	scheme := elbv2model.LoadBalancerSchemeInternetFacing
+	internal := false
+	if _, err := t.annotationParser.ParseBoolAnnotation(annotations.SvcLBSuffixInternal, &internal, t.service.Annotations); err != nil {
+		return "", err
+	} else if internal {
```
M00nF1sh, line 67: i think remove this `else` seems more clear to me.

pkg/service/model_build_load_balancer.go
```diff
@@ -0,0 +1,152 @@
...
+		return nil, err
+	}
+	return tags, nil
+}
+
+func (t *defaultModelBuildTask) buildSubnetMappings(_ context.Context, ec2Subnets []*ec2.Subnet) ([]elbv2model.SubnetMapping, error) {
+	var eipAllocation []string
+	eipConfigured := t.annotationParser.ParseStringSliceAnnotation(annotations.SvcLBSuffixEIPAllocations, &eipAllocation, t.service.Annotations)
+	if eipConfigured && len(eipAllocation) != len(ec2Subnets) {
+		return []elbv2model.SubnetMapping{}, errors.Errorf("Error creating load balancer, number of EIP allocations (%d) and subnets (%d) must match", len(eipAllocation), len(ec2Subnets))
```
M00nF1sh, line 86: errors should always start with lower case. (they can be wrapped by code above call chain).
errors should only have necessary messages scoped to the function's context. 
e.g. it would be better to be simply `number of EIP allocations (%d) and subnets (%d) must match`.

pkg/service/model_build_load_balancer.go
```diff
@@ -0,0 +1,152 @@
...
+	return subnetMappings, nil
+}
+
+func (t *defaultModelBuildTask) buildLoadBalancerAttributes(_ context.Context) ([]elbv2model.LoadBalancerAttribute, error) {
+	var attrs []elbv2model.LoadBalancerAttribute
+	accessLogEnabled := t.defaultAccessLogS3Enabled
+	bucketName := t.defaultAccessLogsS3Bucket
+	bucketPrefix := t.defaultAccessLogsS3Prefix
+	if _, err := t.annotationParser.ParseBoolAnnotation(annotations.SvcLBSuffixAccessLogEnabled, &accessLogEnabled, t.service.Annotations); err != nil {
+		return attrs, err
```
M00nF1sh, line 107: to be consistent, let's return a empty value for error case, like `[]elbv2model.LoadBalancerAttribute{}` (which we did in later statements)

pkg/service/model_build_load_balancer.go
```diff
@@ -0,0 +1,152 @@
...
+	if accessLogEnabled {
+		t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixAccessLogS3BucketName, &bucketName, t.service.Annotations)
+		t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixAccessLogS3BucketPrefix, &bucketPrefix, t.service.Annotations)
+	}
+	crossZoneEnabled := t.defaultLoadBalancingCrossZoneEnabled
+	if _, err := t.annotationParser.ParseBoolAnnotation(annotations.SvcLBSuffixCrossZoneLoadBalancingEnabled, &crossZoneEnabled, t.service.Annotations); err != nil {
+		return []elbv2model.LoadBalancerAttribute{}, err
+	}
+
+	attrs = append(attrs, []elbv2model.LoadBalancerAttribute{
```
M00nF1sh, line 118: why not `attrs = []elbv2model.LoadBalancerAttribute{....}` since attrs is empty before append

pkg/service/model_build_listener.go
```diff
@@ -0,0 +1,110 @@
...
+		svcPort := intstr.FromInt(int(port.Port))
+		tgName := t.buildTargetGroupName(ctx, port.TargetPort, modelTargetType, tgProtocol, hc)
+		tgResId := t.buildTargetGroupResourceID(k8s.NamespacedName(t.service), svcPort)
+		targetPort := 1
+		if port.TargetPort.Type == intstr.Int {
+			targetPort = port.TargetPort.IntValue()
+		}
+		targetGroup, exists := targetGroupMap[port.TargetPort.String()]
+		if !exists {
+			targetGroup = elbv2model.NewTargetGroup(t.stack, tgResId, elbv2model.TargetGroupSpec{
```
M00nF1sh, line 64: this part logic should belong to model_build_targetgroup.go

pkg/service/model_build_target_group.go
```diff
@@ -0,0 +1,312 @@
...
+	elbv2api "sigs.k8s.io/aws-load-balancer-controller/apis/elbv2/v1beta1"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/annotations"
+	elbv2model "sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2"
+	"strconv"
+	"strings"
+)
+
+const (
+	TGAttrsProxyProtocolV2Enabled  = "proxy_protocol_v2.enabled"
+	TGAttrsPreserveClientIPEnabled = "preserve_client_ip.enabled"
```
M00nF1sh, line 25: seems this `preserve_client_ip ` feature isn't released before reinvent. better to remove it from the code.

pkg/service/model_build_target_group.go
```diff
@@ -0,0 +1,312 @@
...
+
+func (t *defaultModelBuildTask) buildTargetGroupHealthCheckUnhealthyThresholdCount(_ context.Context) (int64, error) {
+	unhealthyThresholdCount := t.defaultHealthCheckUnhealthyThreshold
+	if _, err := t.annotationParser.ParseInt64Annotation(annotations.SvcLBSuffixHCUnhealthyThreshold, &unhealthyThresholdCount, t.service.Annotations); err != nil {
+		return 0, err
+	}
+	return unhealthyThresholdCount, nil
+}
+
+func (t *defaultModelBuildTask) buildTargetHealthCheck(ctx context.Context) (*elbv2model.TargetGroupHealthCheckConfig, error) {
```
M00nF1sh, line 95: maybe `buildTargetGroupHealthCheckConfig` to align with Ingress' model builder :D

pkg/service/model_build_target_group.go
```diff
@@ -0,0 +1,312 @@
...
+		return nil, err
+	}
+	if rawAttributes == nil {
+		rawAttributes = make(map[string]string)
+	}
+	if _, ok := rawAttributes[TGAttrsProxyProtocolV2Enabled]; !ok {
+		rawAttributes[TGAttrsProxyProtocolV2Enabled] = "false"
+	}
+	proxyV2Annotation := ""
+	if t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixProxyProtocol, &proxyV2Annotation, t.service.Annotations) {
```
M00nF1sh, line 147: would be more clear and consist if we use `if exists := t.annotationParser.ParseStringAnnotation(xx); exists {`

pkg/service/model_build_target_group.go
```diff
@@ -0,0 +1,312 @@
...
+		if proxyV2Annotation != "*" {
+			return []elbv2model.TargetGroupAttribute{}, errors.Errorf("Invalid value %v for Load Balancer proxy protocol v2 annotation, only value currently supported is *", proxyV2Annotation)
```
M00nF1sh, line 149: errors should starts with lower case.

pkg/service/model_build_target_group.go
```diff
@@ -0,0 +1,312 @@
...
+func (t *defaultModelBuildTask) buildTargetGroupAttributes(_ context.Context) ([]elbv2model.TargetGroupAttribute, error) {
+	var rawAttributes map[string]string
+	if _, err := t.annotationParser.ParseStringMapAnnotation(annotations.SvcLBSuffixTargetGroupAttributes, &rawAttributes, t.service.Annotations); err != nil {
+		return nil, err
+	}
+	if rawAttributes == nil {
+		rawAttributes = make(map[string]string)
+	}
+	if _, ok := rawAttributes[TGAttrsProxyProtocolV2Enabled]; !ok {
+		rawAttributes[TGAttrsProxyProtocolV2Enabled] = "false"
```
M00nF1sh, line 144: why do we want to set it to false here. i think we can leave it unset, so AWS decides the default. or have a defaultProxyProtocolV2Enabled variable in modelBuildTask

pkg/service/model_build_load_balancer.go
kishorj: ok

pkg/service/model_build_load_balancer.go
kishorj: sure

pkg/service/model_build_load_balancer.go
kishorj: ok

pkg/service/model_build_listener.go
```diff
@@ -0,0 +1,110 @@
...
+		if tgProtocol != elbv2model.ProtocolUDP && certificateARNs != nil && (sslPortsSet.Len() == 0 || sslPortsSet.Has(port.Name) || sslPortsSet.Has(strconv.Itoa(int(port.Port)))) {
+			if backendProtocol == "ssl" {
+				tgProtocol = elbv2model.ProtocolTLS
+			}
+			listenerProtocol = elbv2model.ProtocolTLS
+		}
+
+		svcPort := intstr.FromInt(int(port.Port))
+		tgName := t.buildTargetGroupName(ctx, port.TargetPort, modelTargetType, tgProtocol, hc)
+		tgResId := t.buildTargetGroupResourceID(k8s.NamespacedName(t.service), svcPort)
```
M00nF1sh, line 57: 1.  let's move the target group related logic into model_build_target_group.go.
2. let's use tgResourceID to check whether we need to build a new target group or not. (we can save built targetGroup by resourceID), similar to what we have in Ingresses.
3. doing above with change the behavior in cases like two different service port refers to same targetPort. with this existing code, only one targetGroup will be created. after the change, two targetGroup will be created. However, i think the later case should be desired, as we might have annotations to set different tgAttributes per svc port.

pkg/service/model_build_load_balancer.go
kishorj: correct. It was remnant from previous refactoring.

pkg/service/model_build_listener.go
```diff
@@ -0,0 +1,110 @@
...
+		targetPort := 1
+		if port.TargetPort.Type == intstr.Int {
```
M00nF1sh, line 59: 1. let's have a method like `buildTargetGroupPort`. 
2. use the built targetGroupPort as hash component instead of `port.TargetPort` when construct tgName. since it's the actual immutable targetGroup attribute.

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/ingress/model_build_target_group.go#L126

pkg/service/model_build_listener.go
```diff
@@ -0,0 +1,110 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+	elbv2api "sigs.k8s.io/aws-load-balancer-controller/apis/elbv2/v1beta1"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/annotations"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/k8s"
+	elbv2model "sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2"
+	"strconv"
+)
+
+func (t *defaultModelBuildTask) buildListeners(ctx context.Context, ec2Subnets []*ec2.Subnet) error {
+	tgAttrs, err := t.buildTargetGroupAttributes(ctx)
```
M00nF1sh, line 18: tg handling should be moved to model_build_targetGroup

pkg/service/model_build_listener.go
kishorj: I have a work in progress to refactor the listener code, and would like to make it part of a different PR.

pkg/service/model_build_listener.go
```diff
@@ -0,0 +1,110 @@
...
+		return errors.Wrapf(err, "Unable to build target group attributes")
+	}
+	modelTargetType := t.buildTargetType(ctx)
+	apiTargetType := elbv2api.TargetType(modelTargetType)
+	preserveClientIP, err := t.buildPreserveClientIPFlag(ctx, apiTargetType)
+	if err != nil {
+		return errors.Wrapf(err, "Unable to get preserve Client IP configuration")
+	}
+	var certificateARNs []string
+	t.annotationParser.ParseStringSliceAnnotation(annotations.SvcLBSuffixSSLCertificate, &certificateARNs, t.service.Annotations)
```
M00nF1sh, line 29: 1. better to have methods like buildListenerCertificates.

pkg/service/model_build_target_group.go
kishorj: I'd prefer to put in place since beta customers can still utilize the feature. Unlike other attributes, this one has a dependency on how the security group rules are constructed.

pkg/service/model_build_listener.go
```diff
@@ -0,0 +1,110 @@
...
+	var sslPorts []string
+	t.annotationParser.ParseStringSliceAnnotation(annotations.SvcLBSuffixSSLPorts, &sslPorts, t.service.Annotations)
+	sslPortsSet := sets.NewString(sslPorts...)
+
+	backendProtocol := ""
+	t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixBEProtocol, &backendProtocol, t.service.Annotations)
+
+	targetGroupMap := map[string]*elbv2model.TargetGroup{}
+
+	for _, port := range t.service.Spec.Ports {
```
M00nF1sh, line 40: let's have a method of `buildListener`, which takes the port argument.

pkg/service/model_build_target_group.go
kishorj: sounds good.

pkg/service/model_build_target_group.go
kishorj: ok

pkg/service/model_build_target_group.go
kishorj: With the aws-load-balancer-proxy-protocol annotation, we need to make sure we disable this if the annotation gets deleted or not specified. I will add the defaultProxyProtocolV2Enabled in the modelBuildTask.

pkg/service/model_build_listener.go
kishorj: I had this refactoring as a separate change. I will merge with this PR.

pkg/service/model_build_target_group.go
elijah-roberts: I would +1 this... we are trying to get in on the beta to test this feature

pkg/service/model_build_load_balancer.go
kishorj: That means all errors should start with a lower case, or the top level errors still start with an upper case.

pkg/service/model_build_listener.go
kishorj: tgResourceID for service is namespace/name:svcPort, so there won't be duplicates. As per our offline discussion, we are choosing to create an individual target group per service port. This will allow us to have different configurations for each of the target groups in future.

pkg/service/model_build_listener.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	listenerResID := fmt.Sprintf("%v", port.Port)
+	ls := elbv2model.NewListener(t.stack, listenerResID, lsSpec)
+	return ls, nil
+}
+
+func (t *defaultModelBuildTask) buildListenerSpec(ctx context.Context, port corev1.ServicePort, cfg listenerConfig) (elbv2model.ListenerSpec, error) {
+	tgProtocol := elbv2model.Protocol(port.Protocol)
+	listenerProtocol := elbv2model.Protocol(port.Protocol)
+	if tgProtocol != elbv2model.ProtocolUDP && cfg.certificateARNs != nil && (cfg.tlsPortsSet.Len() == 0 ||
```
M00nF1sh, line 38: for empty check for lists, we should always use `len(cfg. certificateARNs) != 0`, which handles both cases.

pkg/service/model_build_listener.go
```diff
@@ -0,0 +1,137 @@
...
+				},
+			},
+		},
+	}
+}
+
+func (t *defaultModelBuildTask) buildSSLNegotiationPolicy(_ context.Context) *string {
+	var sslPolicy *string = nil
+	rawSslPolicyStr := ""
+	exists := t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixSSLNegotiationPolicy, &rawSslPolicyStr, t.service.Annotations)
```
M00nF1sh, line 87: nit, :D 
can be simplified with 
`if exists := t....; exists {
       ...
}`

pkg/service/model_build_listener.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (t *defaultModelBuildTask) buildBackendProtocol(_ context.Context) string {
+	rawBackendProtocol := ""
+	_ = t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixBEProtocol, &rawBackendProtocol, t.service.Annotations)
+	return rawBackendProtocol
+}
+
+type listenerConfig struct {
+	certificateARNs []string
```
M00nF1sh, line 113: i think we can use `certificateARNs` from listenerConfig, and use `certificates` instead when checking for listener/tg protocol

pkg/service/model_build_load_balancer.go
```diff
@@ -0,0 +1,157 @@
...
+	return spec, nil
+}
+
+func (t *defaultModelBuildTask) buildLoadBalancerScheme(_ context.Context) (elbv2model.LoadBalancerScheme, error) {
+	scheme := elbv2model.LoadBalancerSchemeInternetFacing
+	internal := false
+	if _, err := t.annotationParser.ParseBoolAnnotation(annotations.SvcLBSuffixInternal, &internal, t.service.Annotations); err != nil {
+		return "", err
+	}
+	if internal {
```
M00nF1sh, line 68: it would be more clear to me if it's like below. (the idea is to return asap)
```
if internal {
   return elbv2model.LoadBalancerSchemeInternal, nil
}
return elbv2model.LoadBalancerSchemeInternetFacing, nil
```

pkg/service/model_build_target_group.go
```diff
@@ -0,0 +1,392 @@
...
+	healthCheckPortTrafficPort     = "traffic-port"
+)
+
+func (t *defaultModelBuildTask) buildTargetGroup(ctx context.Context, port corev1.ServicePort, tgProtocol elbv2model.Protocol) (*elbv2model.TargetGroup, error) {
+	svcPort := intstr.FromInt(int(port.Port))
+	tgResourceID := t.buildTargetGroupResourceID(k8s.NamespacedName(t.service), svcPort)
+	if targetGroup, exists := t.tgByResID[tgResourceID]; exists {
+		return targetGroup, nil
+	}
+	healthCheckConfig, err := t.buildTargetHealthCheckConfig(ctx)
```
M00nF1sh, line 35: nit, should be `buildTargetGroupHealthCheckConfig`

pkg/service/model_build_target_group.go
```diff
@@ -0,0 +1,392 @@
...
+		return nil, err
+	}
+	if rawAttributes == nil {
+		rawAttributes = make(map[string]string)
+	}
+	if _, ok := rawAttributes[tgAttrsProxyProtocolV2Enabled]; !ok {
+		rawAttributes[tgAttrsProxyProtocolV2Enabled] = strconv.FormatBool(t.defaultProxyProtocolV2Enabled)
+	}
+	proxyV2Annotation := ""
+	exists := t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixProxyProtocol, &proxyV2Annotation, t.service.Annotations)
```
M00nF1sh, line 160: nit, better to be `if exists := ....; exists {}`.  do this will scope the exists variable inside the if statement.

pkg/service/model_build_target_group.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+	return attributes, nil
+}
+
+func (t *defaultModelBuildTask) buildPreserveClientIPFlag(_ context.Context, targetType elbv2model.TargetType) (bool, error) {
+	var rawAttributes map[string]string
+	if _, err := t.annotationParser.ParseStringMapAnnotation(annotations.SvcLBSuffixTargetGroupAttributes, &rawAttributes, t.service.Annotations); err != nil {
+		return false, err
+	}
+	if rawVal, ok := rawAttributes[tgAttrsPreserveClientIPEnabled]; ok {
```
M00nF1sh, line 182: i think it would be better to move this validation logic of bool type into buildTargetGroupAttributes. since it validate a targetGroup attribute.
and change this method to accept targetType and []elbv2model.TargetGroupAttribute

pkg/service/model_build_load_balancer.go
M00nF1sh: all errors will start with lower case :D

pkg/service/model_build_target_group.go
kishorj: I will modify this function to take []elbv2model.TargetGroupAttribute, and move the validation logic to buildTargetGroupAttributes. Since elbv2model.TargetGroupAttribute contains string key, value pairs, the strconv.ParseBool, and error return is still needed.

pkg/service/model_build_target_group.go
kishorj: correct, updated.

pkg/service/model_build_listener.go
kishorj: changed to certificates in the latest commit.

M00nF1sh (approved):
/lgtm

PR #1635, opened 2020-11-10
fix two targetGroup related bug

M00nF1sh:
## Changes done
1. remove TargetGroup's port from TargetGroup replacement criteria. so that TargetGroups can be reused if only port differs.
    * this allow us to migrate from AWSALBIngressController/AWSLoadBalancerController(v2.0.0) with 0 downtime.
2. change the port for instance TargetGroup to be nodePort for better UX. (the port don't matter)
3. fix an edge that when same service referenced by both name and port, only a single TargetGroup are created. Now two different targetGroup will be created.
    Details:
        1. in both AWSALBIngressController and AWSLoadBalancerController(v2.0.0), we used the backendPort specified as resourceID(so it can be either port name or port number). Ideally different resourceID should result in different TargetGroup.
        2. in v2.0.0, even the model contains two targetGroup, only a single targetGroup will be created since the TG name is same, and due to ELBV2's API's idempotency. This works fine, however, it relies on ELBv2 API's implementation details and hacky.
        3. an alternative design is to always use the numeric port as resourceID. and to be backwards-compatible for backend port referenced by name, we can add a logic in targetGroup's cleanup logic to ignore TargetGroups's arn been used. we decided not to take this approach as it's hacky. (also, in the future, we might have extra logic to compress TargetGroup usage across Ingresses, so no need to compress for this edge case this time).

### Pre steps:
1.  Install AWSALBIngressController with 0 replica
2. Install AWSLoadBalancerController with 1 replica
### compatibility test with AWSALBIngressController:
1. scale up AWSALBIngressController to 1 replica
2. launch Ingress with 4 path: `echo: http`, `echo: 80`, `echo-ip: http`, `echo-ip:80`. (`echo` is instance TargetType, `echo-ip` is ip TargetType, both `http` and `80` refers to same servicePort).
3. verify that traffic works, and 4 target group created.
4. scale down AWSALBIngressController to 0 replica
5. scale up AWSLoadBalancerController to 1 replica
6. verify the controller added `elbv2.k8s.aws/cluster":"m00nf1sh-dev` tags to resources, and kept old resources like LoadBalancer/TargetGroups.
### compatibility test with AWSLoadBalancerController:
1. change to AWSLoadBalancerController v2.0.0 and scale up AWSLoadBalancerController to 1 replica
2. launch Ingress with 4 path: `echo: http`, `echo: 80`, `echo-ip: http`, `echo-ip:80`. (`echo` is instance TargetType, `echo-ip` is ip TargetType, both `http` and `80` refers to same servicePort).
3. verify that traffic works, and only two target group created.
4. change to AWSLoadBalancerController to image built with this change and scale up AWSLoadBalancerController to 1 replica
5. verify that the controller added two new targetGroup and configure rules properly.

### DIFF
--- a/pkg/ingress/model_build_target_group.go
+++ b/pkg/ingress/model_build_target_group.go
@@ -138,7 +138,7 @@ func (t *defaultModelBuildTask) buildTargetGroupSpec(ctx context.Context,
 
 var invalidTargetGroupNamePattern = regexp.MustCompile("[[:^alnum:]]")
 
-// buildTargetGroupName will calculate the targetGroup's
+// buildTargetGroupName will calculate the targetGroup's name.
 func (t *defaultModelBuildTask) buildTargetGroupName(_ context.Context,
 	ingKey types.NamespacedName, svc *corev1.Service, port intstr.IntOrString, tgPort int64,
 	targetType elbv2model.TargetType, tgProtocol elbv2model.Protocol) string {
--- a/pkg/ingress/model_build_target_group.go
+++ b/pkg/ingress/model_build_target_group.go
@@ -173,8 +173,8 @@ func (t *defaultModelBuildTask) buildTargetGroupTargetType(_ context.Context, sv
 }
 
 // buildTargetGroupPort constructs the TargetGroup's port.
-// Note: TargetGroup's port setting don't actually matter to this controller since we always register targets with port specified.
-// however, we'll do our best to use the most appropriate port as TargetGroup's port to avoid UX confusing.
+// Note: TargetGroup's port is not in the data path as we always register targets with port specified.
+// so this settings don't really matter to our controller, and we do our best to use the most appropriate port as targetGroup's port to avoid UX confusing.
 func (t *defaultModelBuildTask) buildTargetGroupPort(_ context.Context, targetType elbv2model.TargetType, svcPort corev1.ServicePort) int64 {
 	if targetType == elbv2model.TargetTypeInstance {
 		return int64(svcPort.NodePort)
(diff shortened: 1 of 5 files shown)


kishorj (approved):

pkg/ingress/model_build_target_group.go
```diff
@@ -121,26 +122,25 @@ func (t *defaultModelBuildTask) buildTargetGroupSpec(ctx context.Context,
...
 		HealthCheckConfig:     &healthCheckConfig,
-		TargetGroupAttributes: targetGroupAttributes,
+		TargetGroupAttributes: tgAttributes,
 		Tags:                  tags,
 	}, nil
 }
 
 var invalidTargetGroupNamePattern = regexp.MustCompile("[[:^alnum:]]")
 
+// buildTargetGroupName will calculate the targetGroup's
```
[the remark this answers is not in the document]
M00nF1sh: thanks :D addressed :D

pkg/ingress/model_build_target_group.go
```diff
@@ -171,6 +172,21 @@ func (t *defaultModelBuildTask) buildTargetGroupTargetType(_ context.Context, sv
...
+func (t *defaultModelBuildTask) buildTargetGroupPort(_ context.Context, targetType elbv2model.TargetType, svcPort corev1.ServicePort) int64 {
+	if targetType == elbv2model.TargetTypeInstance {
+		return int64(svcPort.NodePort)
+	}
+	if svcPort.TargetPort.Type == intstr.Int {
+		return int64(svcPort.TargetPort.IntValue())
+	}
+
+	// when a literal targetPort is used, it can actually be different ports for different pods. so we use a fixed 1 here.
+	return 1
```
fawadkhaliq, line 187: Is target group->port optional? How come this dummy port value has no impact?

fawadkhaliq (approved):
Let's pls add a comment why port=1 has no impact on the datapath as we override with a target 

/lgtm otherwise

pkg/ingress/model_build_target_group.go
M00nF1sh: it's not in the data path. we override each target's port on registeration.

fawadkhaliq (commented):
/lgtm

PR #1645, opened 2020-11-12
feat: "nlb-instance" type and filtering listener targets by node selector

jeffherald:
Greetings,
We are submitting this PR with some features we have found useful for our use case and believe they may be of interest to others in the community as well.

- `nlb-instance` type allows targets for the nlb to be registered by instance rather than IP much like the native k8s provider. This is useful in managed environments like EKS to use features gated features such as `ServiceNodeExclusion` that cannot be enabled in the native aws provider due to lack of access to control plane configuration.
- Listener target filtering by node selector allows filtering of listener targets by a node selector defined in a configurable annotation on the service. This is useful in larger clusters and clusters with a lot of node volatility to limit the number of targets registered, ensure stability and decrease the likelihood of hitting cloud provider limits.


controllers/service/eventhandlers/service_events.go
```diff
@@ -14,6 +14,7 @@ import (
 )
 
 const loadBalancerTypeNLBIP = "nlb-ip"
+const LoadBalancerTypeNLBInstance = "nlb-instance"
```
kishorj, line 17: This will not work very well with the in-tree service controller for service type LoadBalancer. The in-tree controller will end up provisioning a CLB for the service.

The path we want to follow is to use the load balancer type 'external' so that the in-tree code ignores the service object, and have an additional annotation to determine the instance mode.

kishorj (commented):
Also need to support annotation service.beta.kubernetes.io/aws-load-balancer-target-node-labels.

controllers/service/eventhandlers/service_events.go
jeffherald: Thank you for the feedback. We are on 1.18 so we are using type `NodePort`. To clarify the path forward, you are suggestion the `service.beta.kubernetes.io/aws-load-balancer-type` annotation be set to `external` and another annotation to define instance mode, either `nlb-ip` or `nlb-instance`?

kishorj: We are also working on the instance mode support and targeting it post reInvent 2020.

PR #1666, opened 2020-11-17
Add basic GitHub Action for running unit tests  + code coverage

### LINKED ISSUE #1665: Unit tests not run on PRs/CI

backjo:
Closes #1665

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,32 @@
+
+name: Unit Test
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+
+  build:
+    name: Build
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -0,0 +1,32 @@
...
+name: Unit Test
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+
```
kishorj, line 11: why so many empty lines in the spec? Is this a norm for github config?

backjo: Came with the auto-generated template for Go. I removed a few, but I think they're still useful to have in-between steps.

M00nF1sh:
thanks dude, we'll replace this with prow, but this will help for now :D

PR #1669, opened 2020-11-18
Add support for NLB ALPN policy

### LINKED ISSUE #1546: Feature Request: Please enable management for Alpn Policy
To be able to host gRPC services on EKS we need to set the AlpnPolicy for the NLB. Now that NLB support is support in version 2 of this ingress controller, it would be great if we could manage the AlpnPolicy with a NLB annotation (https://github.com/kubernetes-sigs/aws-alb-ingress-controller/blob/main/docs/guide/service/annotations.md)

kishorj:
Fixes #1546 Allow configuration of ALPN policies for NLB

Support additional annotation `service.beta.kubernetes.io/aws-load-balancer-alpn-policy` on the service object which allows users to configure ALPN policy on the NLB. ALPN can be configured only if both the listener and the target group protocol is TLS.

Tests run
Service requires at least the following annotations
```
    service.beta.kubernetes.io/aws-load-balancer-alpn-policy: <ALPN>
    service.beta.kubernetes.io/aws-load-balancer-backend-protocol: ssl
    service.beta.kubernetes.io/aws-load-balancer-ssl-cert: <certs>
    service.beta.kubernetes.io/aws-load-balancer-ssl-cert: nlb-ip
```
- When the alpn policy annotation is not specified, the default gets used
- ALPN policy can be specified via the annotation.
- When previously specified alpn policy annotation gets deleted, the existing configuration remains in place
- Invalid ALPN policy configuration results in error

### DIFF
--- a/pkg/model/elbv2/listener.go
+++ b/pkg/model/elbv2/listener.go
@@ -292,6 +292,18 @@ type Certificate struct {
 	CertificateARN *string `json:"certificateARN,omitempty"`
 }
 
+// ALPNPolicy ALPN policy configuration for TLS listeners forwarding to TLS target groups
+type ALPNPolicy string
+
+// Supported ALPN policies
+const (
+	ALPNPolicyNone           ALPNPolicy = "None"
+	ALPNPolicyHTTP1Only      ALPNPolicy = "HTTP1Only"
+	ALPNPolicyHTTP2Only      ALPNPolicy = "HTTP2Only"
+	ALPNPolicyHTTP2Optional  ALPNPolicy = "HTTP2Optional"
+	ALPNPolicyHTTP2Preferred ALPNPolicy = "HTTP2Preferred"
+)
+
 // ListenerSpec defines the desired state of Listener
 type ListenerSpec struct {
 	// The Amazon Resource Name (ARN) of the load balancer.
--- a/docs/guide/service/annotations.md
+++ b/docs/guide/service/annotations.md
@@ -33,6 +33,7 @@
 | service.beta.kubernetes.io/aws-load-balancer-eip-allocations                   | stringList |                           |                        |
 | [service.beta.kubernetes.io/aws-load-balancer-target-group-attributes](#target-group-attributes)  | stringMap  |        |                        |
 | [service.beta.kubernetes.io/aws-load-balancer-subnets](#subnets)              | stringList  |                           |                        |
+| [service.beta.kubernetes.io/aws-load-balancer-alpn-policy](#alpn-policy)      | stringList  |                           |                        |
 
 
 ## Traffic Routing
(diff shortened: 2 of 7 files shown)


pkg/model/elbv2/listener.go
```diff
@@ -292,6 +292,18 @@ type Certificate struct {
 	CertificateARN *string `json:"certificateARN,omitempty"`
 }
 
+//ALPNPolicy ALPN policy configuration for TLS listeners forwarding to TLS target groups
```
M00nF1sh, line 295: nit. let's have a space after '//'

M00nF1sh:
1. If nil alpn policy array is specified we when create the listener, what's the value of alpn policy array from AWS API
2. If [`None`] alpn policy array is specified when we create the listener, what's the value of alpn policy array from AWS API?
3. did this AWS API supports alpn policy array  with multiple values or just one

pkg/model/elbv2/listener.go
kishorj: updated

kishorj:
Here is the summary of AWS API behavior

>     1. If nil alpn policy array is specified we when create the listener, what's the value of alpn policy array from AWS API
> 
If nil is specified, the previous ALPN Policy configuration takes effect. Used the `--alpn-policy '[]'` argument to aws cli  . The output shows the previous policy configuration.

>     2. If [`None`] alpn policy array is specified when we create the listener, what's the value of alpn policy array from AWS API?
> 
If `["None"]`  is specified, ALPN is disabled. `--alpn-policy '["None"]'`. The output shows ALPN policy `None`.


>     3. did this AWS API supports alpn policy array  with multiple values or just one

If multiple values are specified, for example `--alpn-policy '["None", "Http"]'`, we get the following error from AWS
```
      An error occurred (ValidationError) when calling the ModifyListener operation: Up to '1' ALPN policy can be specified, but '2' were specified
```

M00nF1sh (approved):
awesome 👍 
/lgtm

julianxhokaxhiu:
Sorry to necropost on this PR, but I'm currently not understanding which is the benefit to allow ALPN policies only when the `backend-protocol` is set to `ssl`.

Look at this example of mine:
```yaml
service.beta.kubernetes.io/aws-load-balancer-backend-protocol: http
service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: '60'
service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: 'true'
service.beta.kubernetes.io/aws-load-balancer-type: external
service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
service.beta.kubernetes.io/aws-load-balancer-scheme: ${load_balancer_scheme}
service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: "*"
service.beta.kubernetes.io/aws-load-balancer-ssl-cert: ${aws_ssl_cert_arn}
service.beta.kubernetes.io/aws-load-balancer-ssl-ports: https
service.beta.kubernetes.io/aws-load-balancer-alpn-policy: HTTP2Preferred
```

Expected behavior:
1) I connect to port 80, I move the traffic to whatever is listening as ingress ( nginx on my own side ), I get redirected to https
2) I connect to port 443 using TLS, I would like to prefer HTTP/2 connections but I can't because ALPN is set to None, but I continue using HTTP/1.1

How can I therefore allow this redirection flow to continue AND set up the ALPN for the TLS part on the Load Balancer?
As you can see from the screenshot it is possible to configure it via WebGUI, I don't see why it would not be via annotations.
image


Thank you in advance for any help given!

PR #1670, opened 2020-11-19
add ingress class support

M00nF1sh:
## Add ingress class support
This change will add Ingress class support, the only supported field in IngressClass is the `controller` field. In the future, we'll leverage the `parameters` field for more customizations.

## Behavior
* If only `kubernete.io/ingress.class` annotations presents, we consider IngressClass matches this controller if 
   * when `--ingress-class` is specified with non-empty value, and it matches annotation value (when install via YAML/Helm, by default it's specified as `alb`):
   * when `--ingress-class` is specified with empty value, and annotation value is empty or `alb`.

* If only `spec.ingressClassName` non-nil, we consider IngressClass matches this controller if
   * the corresponding IngressClass exists and the `spec.controller` of IngressClass have value `ingress.k8s.aws/alb`

* If both `kubernete.io/ingress.class` annotation presents and `spec.ingressClassName` non-nil, we will favor `kubernete.io/ingress.class` and issue a warning event to Ingress if match result conflicts. 
    we do this to align with IngressSpec's recommendation:`For backwards compatibility, when that annotation is set, it must be given precedence over this field. The controller may emit a warning if the field and annotation have different values.`

* If neither `kubernete.io/ingress.class` annotation presents and `spec.ingressClassName` non-nil, we consider IngressClass matches this controller if `--ingress-class` is specified with empty value

## Future plans
1. we'll offer an option to disable `kubernete.io/ingress.class` annotation support in future versions with default value `false` in future versions. (this will help us enforce permission control via IngressClass)
2. we'll change the option above to have default value `true` post X versions.
3. we'll remove the code that support `kubernete.io/ingress.class` post Y versions.
4. IngressClass's parameters will only be respected if IngressClass is decided from `spec.ingressClassName`. (i.e. `kubernete.io/ingress.class` annotation absents)

## Note to reviewer:
1. I'll have follow up PR to add RBAC permissions
2. I'll have follow up PR to adjust the behavior of decide whether Ingress should be groupMember if there are errors like ingressClassName not found.

### DIFF
--- a/pkg/ingress/group_loader.go
+++ b/pkg/ingress/group_loader.go
@@ -3,8 +3,11 @@ package ingress
 import (
 	"context"
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	networking "k8s.io/api/networking/v1beta1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/record"
 	"regexp"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/annotations"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/k8s"
@@ -13,11 +16,13 @@ import (
 )
 
 const (
-	defaultGroupOrder   int64 = 0
-	minGroupOrder       int64 = 1
-	maxGroupOder        int64 = 1000
-	maxGroupNameLength  int   = 63
-	defaultIngressClass       = "alb"
+	defaultGroupOrder  int64 = 0
+	minGroupOrder      int64 = 1
+	maxGroupOder       int64 = 1000
+	maxGroupNameLength int   = 63
+	ingressClassALB          = "alb"
+	// the controller name used in IngressClass for ALB.
+	ingressClassControllerALB = "ingress.k8s.aws/alb"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


pkg/k8s/events.go
```diff
@@ -2,6 +2,7 @@ package k8s
 
 const (
 	// Ingress events
+	IngressEventReasonConflictIngressClass   = "ConflictIngressClass"
```
kishorj, line 5: nit: ConflictingIngressClass instead?

pkg/ingress/group_loader.go
```diff
@@ -104,13 +115,52 @@ func (m *defaultGroupLoader) Load(ctx context.Context, groupID GroupID) (Group,
...
+		matchesIngressClass, err := m.matchesIngressClassName(ctx, *ing.Spec.IngressClassName)
+		if err != nil {
+			return false, err
+		}
+		matchesIngressClassResults = append(matchesIngressClassResults, matchesIngressClass)
+	}
+
+	if len(matchesIngressClassResults) == 2 {
+		if matchesIngressClassResults[0] != matchesIngressClassResults[1] {
+			m.eventRecorder.Event(ing, corev1.EventTypeWarning, k8s.IngressEventReasonConflictIngressClass, "conflict IngressClass by `spec.IngressClass` and `kubernetes.io/ingress.class` annotation")
```
kishorj, line 136: nit: lets reword the error message
conflicting values for IngressClass ..

M00nF1sh: thanks,  addressed it :D

pkg/k8s/events.go
M00nF1sh: thanks,  addressed it :D

kishorj (approved):

PR #1674, opened 2020-11-19
Allow stringMap annotation values containing equal sign

### LINKED ISSUE #1673: Failed to parse stringMap annotation with equal sign
We use AWS tags to store some base64 encoded JSON metadata. Looks like this no longer works in version 2 of the LB controller where as it used to work in version 1. Is there a different way to escape the equal signs?

```
{"level":"error","ts":1605798917.9130397,"logger":"controller","msg":"Reconciler error","controller":"ingress","name":"vproxy","namespace":"default","error":"failed to parse stringMap annotation, alb.ingress.kubernetes.io/tags: foo=Zm9vCg==,another=tag"}

# With escaping.
{"level":"error","ts":1605798989.6451142,"logger":"controller","msg":"Reconciler error","controller":"ingress","name":"vproxy","namespace":"default","error":"failed to parse stringMap annotation, alb.ingress.kubernetes.io/tags: foo=Zm9vCg\\=\\=,another=tag"}
```

kishorj:
Fixes #1673
Fix stringMap annotation type to accept values with the equal sign. For example, the following annotation will now be accepted - 
```
alb.ingress.kubernetes.io/tags: foo=ZXF1YWxzCg==, another=tag"
```
and would be parsed to 
```
map[string]string{
	"foo":     "ZXF1YWxzCg==,",
	"another": "tag",
}
```

### DIFF
--- a/pkg/annotations/parser.go
+++ b/pkg/annotations/parser.go
@@ -136,7 +136,7 @@ func (p *suffixAnnotationParser) ParseStringMapAnnotation(annotation string, val
 	rawKVPairs := splitCommaSeparatedString(raw)
 	keyValues := make(map[string]string)
 	for _, kvPair := range rawKVPairs {
-		parts := strings.Split(kvPair, "=")
+		parts := strings.SplitN(kvPair, "=", 2)
 		if len(parts) != 2 {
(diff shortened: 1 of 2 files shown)


M00nF1sh (approved):
/lgtm
nice 👍

Issue #1655, opened 2020-11-14
NLB IPv6 support


bootc:
Now that this is a thing: https://aws.amazon.com/about-aws/whats-new/2020/11/network-load-balancer-supports-ipv6/

Please add support for configuring NLB services with IPv6, including support for static IPv6 addresses.

keperry:
Hi, I was looking at implementing this.  Any objections?  Is anyone else already working on this that you know of?

kishorj:
Awesome, thanks.

M00nF1sh:
closing this since the PR is merged and will be release in v2.1.0 soon :D

PR #1677, opened 2020-11-20
Support for IPv6 and NLB  

csnitker:
- Added optional `service.beta.kubernetes.io/aws-load-balancer-ip-address-type: (ipv4|dualstack)` annotation for NLB

Implements #1655

### DIFF
--- a/docs/guide/service/annotations.md
+++ b/docs/guide/service/annotations.md
@@ -14,7 +14,6 @@
 | service.beta.kubernetes.io/aws-load-balancer-type                              | string     |                           |                        |
 | service.beta.kubernetes.io/aws-load-balancer-internal                          | boolean    | false                     |                        |
 | [service.beta.kubernetes.io/aws-load-balancer-proxy-protocol](#proxy-protocol-v2)                 | string     |        | Set to `"*"` to enable |
-| service.beta.kubernetes.io/aws-load-balancer-type                              | string     |                           |                        |
 | service.beta.kubernetes.io/aws-load-balancer-ip-address-type                   | string     | ipv4                      | ipv4 \| dualstack      |
 | service.beta.kubernetes.io/aws-load-balancer-access-log-enabled                | boolean    | false                     |                        |
 | service.beta.kubernetes.io/aws-load-balancer-access-log-s3-bucket-name         | string     |                           |                        |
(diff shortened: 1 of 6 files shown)


kishorj:
I appreciate your efforts, thanks. Changes look good, I'm requesting for additional changes 
- Lets document the annotation in the file docs/guide/service/annotations.md
- Include the tests results the PR. For example, when the annotation gets edited/removed
- Traffic tests (if any). I will do it post merge as well.

kishorj:
Fixes #1655

csnitker:
Just updated the documentation. We've test these changes in one of our AWS accounts and verified that adding, changing, or removing the annotation reflects on the NLB. Do we need a unit/integration test for that? If so, is there a test already setup that I can use as an example?

docs/guide/service/annotations.md
```diff
@@ -14,7 +14,8 @@
 | service.beta.kubernetes.io/aws-load-balancer-type                              | string     |                           |                        |
 | service.beta.kubernetes.io/aws-load-balancer-internal                          | boolean    | false                     |                        |
 | [service.beta.kubernetes.io/aws-load-balancer-proxy-protocol](#proxy-protocol-v2)                 | string     |        | Set to `"*"` to enable |
-| service.beta.kubernetes.io/aws-load-balancer-access-log-enabled                | boolean    | false                     |                        |
```
kishorj, line 17: keep the aws-load-balancer-access-log-enabled annotation as is, and add a row for aws-load-balancer-ip-address-type instead of aws-load-balancer-type

kishorj:
> Just updated the documentation. We've test these changes in one of our AWS accounts and verified that adding, changing, or removing the annotation reflects on the NLB. Do we need a unit/integration test for that? If so, is there a test already setup that I can use as an example?

This is good enough for now.

docs/guide/service/annotations.md
csnitker: Yeah sorry, it's been a long day 😆 

It should be correct now.

kishorj (approved):

M00nF1sh (approved):
/lgtm
thanks for contributing.

We'll need follow up changes to make this work for perserveSourceIP feature.
However, this is good for now since perserveSourceIP is not going to release before reinvent.

integrii:
Bump.  I am a little confused on the state of this PR.  It says 'merged' but is not in the docs and appears to not be passing tests here.  @M00nF1sh @csnitker 

Can this feature still be a thing?  This looks great.

integrii:
Ah, awesome!  We were using [these docs](https://kubernetes-sigs.github.io/aws-load-balancer-controller/guide/service/annotations/) instead of the one you linked.  We will try the v2.1.0 release.  Thank you!

PR #1679, opened 2020-11-21
group.order is string type

tanmayakp:
The type of alb.ingress.kubernetes.io/group.order is string.


M00nF1sh:
@tanmayakp 
Hi, `group.order` is an integer type instead of string type 😄 .

You have to quote integer like "100". it's a requirement by YAML/kubernetes.

Quoting requirement is documented at top of this page: https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/ingress/annotations/.

PR #1680, opened 2020-11-22
docs: make external-dns example to 0.7.2+ working in AWS China

zxkane:
external-dns supports AWS China since 0.7.2, it also requires additional **endpoints** permission


M00nF1sh (commented):
/lgtm
thanks for contributing 👍

PR #1681, opened 2020-11-23
promote install instructions to top level menu item, enable mkdocs gh action

geoffcline:


geoffcline:
draft new page visible here

https://geoffcline.github.io/aws-load-balancer-controller/install/installation/

geoffcline:
ready for review @M00nF1sh :)

.github/workflows/mkdocs-deploy.yml
```diff
@@ -0,0 +1,32 @@
+# GCLINE@ 23NOV2020
```
M00nF1sh, line 1: guess this should be amazon's copy right or leave empty :D

.github/workflows/mkdocs-deploy.yml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  build:
+    name: Deploy docs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v2
+
+      - name: Deploy docs
```
M00nF1sh, line 28: we are using [mike](https://github.com/jimporter/mike) to deploy our multi-version docs.
I think we need to have our own actions or update `mhausenblas/mkdocs-deploy-gh-pages` to support mike.

M00nF1sh:
btw, how about the following layout
```
Welcome
Deployment
    Installation Guide
    Configurations
    Subnet Discovery
    Pod Readiness Gate
    Upgrade
        Migrate v1 to v2
User Guide
    Ingress
        ...
    Service
        ...
    TargetGroupBinding
        ...
    Tasks
        ...
Examples
    EchoServer
```

.github/workflows/mkdocs-deploy.yml
geoffcline: I meant that more as a "who wrote this" but I can see the confusion, I'll fix

.github/workflows/mkdocs-deploy.yml
geoffcline: I'll take a look at how difficult that would be ... hopefully we can still do CD

geoffcline:
ok, I can work using that layout tomorrow 👍

geoffcline:
Ok @M00nF1sh -  I refactored based on your suggested nav.

I also removed the github action, until I can get an action for mike working. 

Thanks!
Geoffrey

PR #1689, opened 2020-11-25
update service.beta.kubernetes.io/aws-load-balancer-healthcheck-port doc

M00nF1sh:
`service.beta.kubernetes.io/aws-load-balancer-healthcheck-port` should be interger or `traffic-port`


astrived (approved):
Looks good!

PR #1710, opened 2020-12-04
doc: fix wrong itemization format

shibataka000:
Itemization are broken at https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/how-it-works/ because of wrong itemization format. So I fix it.

image


kishorj (approved):
Thanks for the changes

PR #1716, opened 2020-12-08
Fix list in how-it-works doc

calvinbui:
Doesn't appear correctly

image


M00nF1sh:
/lgtm
thanks for fixing this :D

PR #1722, opened 2020-12-10
Fix annotation documentation about ALB attributes

giannisbetas:
The documentation mentions that the `MergeBehavior` for the load-balancer-attributes annotation is `Merge`.  
However, if the same attributes are defined across the ingresses of the same ingressGroup, they need to have the same values in order to be merged.


M00nF1sh:
thanks for making this change.
/lgtm
/approve

just FYI, we'll add a new crd as IngressClass's parameter, where we can force customers using IngressGroup to set LB attributes there instead of on individual Ingresses in the future.

PR #1726, opened 2020-12-12
enable e2e tests for service

kishorj:
# Description
Enable Prow e2e tests for service

# Changes
The service e2e tests take additional arguments `--s3-bucket-name` and `--certificate-arns`. The S3 bucket is used for access logging feature, and the certificates are used for creating TLS listeners. If the arguments are not specified, those tests get skipped.

# Tests
- verified the e2e script works locally with/without the arguments
- access logging and TLS listener tests get skipped when the options are not specified
- make e2e-tests locally

### DIFF
--- a/scripts/ci_e2e_test.sh
+++ b/scripts/ci_e2e_test.sh
@@ -214,12 +214,22 @@ test_controller_image() {
     return 1
   fi
 
-  ginkgo -v -r test/e2e/ingress/ -- \
+  AWS_ACCOUNT_ID=$(aws sts get-caller-identity --region ${AWS_REGION} --query Account --output text)
+  S3_BUCKET=${S3_BUCKET:-"lb-controller-e2e-${AWS_ACCOUNT_ID}"}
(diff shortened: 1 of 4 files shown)


scripts/ci_e2e_test.sh
```diff
@@ -220,6 +220,23 @@ test_controller_image() {
     --aws-region=${AWS_REGION} \
     --aws-vpc-id=${cluster_vpc_id} \
     --controller-image=${CONTROLLER_IMAGE_NAME}
+
+  AWS_ACCOUNT_ID=$(aws sts get-caller-identity --region ${AWS_REGION} --query Account --output text)
```
M00nF1sh, line 224: we can put these at top of this file, and use
`ginko -v -r test/e2e/` (which will use invoke both test).

kishorj: sounds good

M00nF1sh (approved):
/lgtm

PR #1727, opened 2020-12-12
ignore target group health check timeout for NLB

### LINKED ISSUE #1714: Unable to Support HTTP health checks with nlb-ip type load balancer
When using the following annotations to configure an NLB:

service.beta.kubernetes.io/aws-load-balancer-type: "nlb-ip"
service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: "true"
service.beta.kubernetes.io/aws-load-balancer-target-group-attributes: stickiness.enabled=true, stickiness.type=source_ip
service.beta.kubernetes.io/aws-load-balancer-healthcheck-path: ‘/test’
service.beta.kubernetes.io/aws-load-balancer-healthcheck-protocol: ‘HTTP’
service.beta.kubernetes.io/aws-load-balancer-healthcheck-port: "8080"

you get the following error from the AWS Load Balancer Controller

{"level":"error","ts":1607377231.8265452,"logger":"controller","msg":"Reconciler error","controller":"service","name":"echo-server","namespace":"default","error":"Unsupported health check protocol ‘HTTP’"}

kishorj:
Fixes #1714
AWS ELBv2 does not support specifying timeout for target group health check config as of now. The default is 10s for TCP/HTTPS protocols, and 6s for HTTP protocol and it cannot be modified. This change will ignore the healthcheck timeout annotation for now.

Tests
- Verified timeout is 10s for health check protocol HTTPS/TCP, 6s for HTTP
- Veriried annotation `service.beta.kubernetes.io/aws-load-balancer-healthcheck-timeout` gets ignored


M00nF1sh (commented):
/ok-to-test
/lgtm

M00nF1sh (approved):
/lgtm
/approve

PR #1773, opened 2021-01-23
Relax cluster tag requirement on subnets during auto-discovery

kishorj:
Relax the kubernetes cluster tag requirement on subnets during auto-discovery. The prior behavior was to discover only those subnets with the tag `kubernetes.io/cluster/<clustername>` and values `{owned, shared}`. If there are subnets with the cluster tag, those still get choosen. In case there are multiple subnets per AZ, auto-discovery will break the tie in the following order 
 1. prefer subnets with the cluster tag kubernetes.io/cluster/<Cluster Name>
 1. prefer the subnet that is first in lexicographic order

If the subnet has tag for a different cluster, auto-discovery will ignore such subnets. 

## Tests done
-  subnets with cluster tag got choosen during auto discovery
-  subnets got sorted in lexicographic order if there were ties
-  subnets without cluster tags were also considered by auto-discovery
-  subnets with tag for different cluster did not get auto-discovered

### DIFF
--- a/pkg/networking/subnet_resolver.go
+++ b/pkg/networking/subnet_resolver.go
@@ -112,12 +112,7 @@ func (r *defaultSubnetsResolver) ResolveViaDiscovery(ctx context.Context, opts .
 	case elbv2model.LoadBalancerSchemeInternetFacing:
 		subnetRoleTagKey = TagKeySubnetPublicELB
 	}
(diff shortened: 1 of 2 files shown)


M00nF1sh (approved):
/lgtm

Issue #1719, opened 2020-12-10
Getting 502/504 with Pod Readiness Gates during rolling updates


calvinbui:
I'm making use of the [Pod Readiness Gate](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/deploy/pod_readiness_gate/) on Kubernetes Deployments running Golang-based APIs. The goal is to achieve full zero downtime deployments.

During a rolling update of the Kubernetes Deployment, I'm getting 502/504 responses from these APIs. This did not happen when setting `target-type: instance`. 

**I believe the problem is that AWS does not drain the pod from the LB before Kubernetes terminates it**

Timeline of events:

1. Perform a rolling update on the deployment (1 replica)
2. A second pod is created in the deployment
3. AWS registers a second target in the Load Balancing Target Group
4. Both pods begin receiving traffic
5. I'm not sure what happens first at this point: 
   a. AWS begins de-registered/drained the target
   b. Kubernetes begins terminating the pod
6. Traffic sent to the deployment begins receiving 502 and 504 errors
7. The old pod is deleted
8. Traffic returns to normal (200)
9. The target is de-registered/drained (depending on delay)

This is tested with a looping curl command:

```bash
while true; do
  curl --write-out '%{url_effective} - %{http_code} -' --silent --output /dev/null -L https://example.com | pv -N "$(date +"%T")" -t
  sleep 1
done
```

Results:

```
https://example.com - 200 - 13:04:16: 0:00:00
https://example.com - 502 - 13:04:17: 0:00:01
https://example.com - 200 - 13:04:20: 0:00:00
https://example.com - 504 - 13:04:31: 0:00:10
https://example.com - 200 - 13:04:32: 0:00:00
https://example.com - 200 - 13:04:33: 0:00:00
https://example.com - 200 - 13:04:34: 0:00:00
https://example.com - 200 - 13:04:35: 0:00:00
https://example.com - 200 - 13:04:36: 0:00:00
```

AirbornePorcine:
We've been having the same issue. We confirmed with AWS that there is some propagation time between when some target is marked draining in a target group, and when that target actually stops receiving new connections. So, at the suggestion of other issues I've seen in the old project for this, we added a 20s sleep in a `preStop` script. This hasn't entirely eliminated them though, they still happen on deployment, just not with as much volume. Following this to see if anyone else has any good ideas, as troubleshooting these 502s has been infuriatingly difficult.

[...]

Issue #1065, opened 2019-11-06
400/502/504 errors while doing rollout restart or rolling update


bmbferreira:
Hi, I'm getting errors while doing a rolling update and I can reproduce this problem consistently doing a rollout restart. I already tried the recommendations pointed in other issues related with this problem (#814), such as adding a preStop hook to sleep for some seconds so the pods can finish the on going requests but it doesn't solve the problem. 

I have also changed the configuration on the load balancer to have the health check interval and threshold-count lower than what I'm setting for the pod's readiness probe so the load balancer could stop sending requests to the pods before receiving the SIGTERM, but without success. 

Currently this is the configuration for the ingress, service and deployment:

## Ingress
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    alb.ingress.kubernetes.io/actions.ssl-redirect: '{"Type": "redirect", "RedirectConfig":
      { "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}'
    alb.ingress.kubernetes.io/certificate-arn: ...
    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS": 443}]'
    alb.ingress.kubernetes.io/scheme: internet-facing
    external-dns.alpha.kubernetes.io/hostname: app.dev.codacy.org, api.dev.codacy.org
    external-dns.alpha.kubernetes.io/scope: public
    kubernetes.io/ingress.class: alb
  labels:
    app.kubernetes.io/instance: codacy-api
    app.kubernetes.io/managed-by: Tiller
    app.kubernetes.io/name: codacy-api
    app.kubernetes.io/version: 4.93.0-SNAPSHOT.d94f47083
    helm.sh/chart: codacy-api-4.93.0-SNAPSHOT.d94f47083
  name: codacy-api
  namespace: codacy
spec:
  rules:
  - host: app.dev.codacy.org
    http:
      paths:
      - backend:
          serviceName: ssl-redirect
          servicePort: use-annotation
        path: /*
      - backend:
          serviceName: codacy-api
          servicePort: http
        path: /*
  - host: api.dev.codacy.org
    http:
      paths:
      - backend:
          serviceName: ssl-redirect
          servicePort: use-annotation
        path: /*
      - backend:
          serviceName: codacy-api
          servicePort: http
        path: /*
```

## Service
```yaml
apiVersion: v1
kind: Service
metadata:
[...]

Issue #1403, opened 2020-09-10
Feature Request: Annotation to delay the start of deregistration from target group


omnibs:
This feature request might be a stretch, but Kubernetes not having a concept of create-before-destroy when killing Pods makes life difficult. We could get around it with long sleeps in `preStop` hooks, but the ALB Ingress Controller acting as fast as it does to drain Pods in Terminating state blocks working around this issue:

### Context

When manually draining instances, or when the cluster autoscaler drains an instance to remove it because we have excess capacity, or when killing misbehaving Pods.

### What happens

In my understanding, the following happens (probably heavily simplified and somewhat inaccurate)

1) Concurrently:
  * Pod state is set to `Terminating`
  * `preStop` hook is fired
  * Pod's IP is removed from the service
2) Concurrently:
  * Another Pod is scheduled to satisfy the current replica count in the deployment
  * AWS ALB Ingress Controller starts to drain the Pod's target in its the ALB
3) New pod is healthy and starts receiving requests from the ALB

### Problem

In a situation where we have few pods, between ALB starting the drain in `2` and the new pod starting to receive requests in `3`, the application will run with up to 50% less capacity than it should (50% when there's 2 Pods, 33% when there's 3 Pods, etc).

In our environments, going healthy takes anywhere from 15s to 150s, depending on the application and whether new nodes have to be brought up to schedule the new Pod.

This requires us to overprovision to reduce the damage, and still causes temporary ripples of slowness and queueing throughout our service topology.

### Solution

Introduce an annotation that allows the ALB to wait X amount of seconds before starting to drain a Pod, after spotting it in `Terminating` state, so we can:
* Sleep for X seconds on `preStop`, still processing requests while in `Terminating` state
* Give the new Pod enough time to go healthy

PR #1775, opened 2021-01-23
Delay pod deletion to handle deregistraton delay

foriequal0:
Pods are deleted during Deployment rollout and then removed from
Endpoints. When deleted, the controller deregisters from TargetGroups
and then the ELB drains the traffic. The delay between Pod deletion and
the propagation of draining is known to be the cause of ELB 5xx error.
ELB tries to send traffic to Pods that are tereminated without knowing
that they are terminated until Pods are completely deregistered.

We solve this problem by keeping Pods alive for a delay time. There is a
well-known solution that attaches a "sleep" command to preStop lifecycle
hook to keep them alive for the delay. However, this is an ugly
mitigation like duct tape, and it might be difficult to hook up a "sleep
9999" to some Deployments depending on their configurations.

In this commit, we use ValidatingAdmissionWebhook to be notified about
Pods that need be deleted, and block their immediate deletions. By doing
so, we can do whatever we want with them while keeping them alive. We
remove all labels from them, so they are removed from the Endpoints, and
are deregistered from TargetGroups. We also remove ownerReferences so it
doesn't kick GC in. Most importantly, they are still alive so they can
continue to serve incoming new traffics received during the delay. After
letting them alive for the delay, we can safely delete them as there
shouldn't be any new traffic coming in.

You can test with the prebuilt docker image: https://github.com/users/foriequal0/packages/container/aws-load-balancer-controller/901501
I've also forked eks-charts/aws-load-balancer-controller https://github.com/foriequal0/eks-charts/tree/drain/stable/aws-load-balancer-controller
You can use this helm chart value:
```yaml
podGracefulDrainDelay: 90s
image:
  repository: ghcr.io/foriequal0/aws-load-balancer-controller
  tag: v2.1.1-2-geb716265
```

closes: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1719
closes: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1065
related?: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1403

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -45,6 +45,8 @@ rules:
   - get
   - list
   - watch
+  - patch
+  - delete
 - apiGroups:
   - ""
   resources:
--- a/config/webhook/manifests.v1beta1.yaml
+++ b/config/webhook/manifests.v1beta1.yaml
@@ -53,6 +53,24 @@ webhooks:
       service:
         name: webhook-service
         namespace: system
+        path: /validate-v1-pod
+    failurePolicy: Ignore
+    name: vpod.elbv2.k8s.aws
+    rules:
+      - apiGroups:
+        - ""
+        apiVersions:
+          - v1
+        operations:
+          - DELETE
+        resources:
+          - pods
+    sideEffects: NoneOnDryRun
+  - clientConfig:
+      caBundle: Cg==
+      service:
+        name: webhook-service
+        namespace: system
         path: /validate-elbv2-k8s-aws-v1beta1-targetgroupbinding
     failurePolicy: Fail
     name: vtargetgroupbinding.elbv2.k8s.aws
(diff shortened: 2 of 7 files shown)


foriequal0:
I think we shouldn't remove all labels. Should I remove only labels that are defined by the service's selector? Also shoud add support an option to delay them until the draining is fully finished (deregistration_delay. timeout_seconds?) if the app is not support connection draining?

foriequal0:
aws-node-termination-handler, which depends on cluster-api uses eviction api, and it seems that it doesn't rely on admissions. :/

mihasya:
Came here after several days of trying to configure the this thing JUST SO and still getting handfulls of 500s on every deploy. Thank you for this work, I really hope a clean way to make this work can be figured out. Of course, `drain node` is an extremely important use-case. If you come up with a solution and need help testing, I would gladly run this PR on our staging environment. Thanks for digging into this!

foriequal0:
@mihasha I've made this a seperated package here https://github.com/foriequal0/pod-graceful-drain
Can you test that? If you have any question, please leave an issue to there.

Issue #1575, opened 2020-10-28
[Feature Request] expose nodeSelector on targetGroupBindings, so users can optionally specify different nodes for different tgbs.


wlin-jin:
Do we have plans to develop this features?
  expose nodeSelector on targetGroupBindings, so users can optionally specify different nodes for different tgbs.
  https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/backend/endpoint_utils.go

M00nF1sh:
@wlin-jin 
Yes, it's planed as it's a TODO 😄 
BTW, why would you prefer instance mode over IP mode?

PR #1785, opened 2021-01-29
Add support for NodeSelector in TargetGroupBindings

Smirl:
# Summary

- fixes https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1575
- add optional`NodeSelector` to `TargetGroupBindingSpec `
- does not add support for ingress or service annotation at this time, can be added later.
- Update validating webhook to ensure that `NodeSelector` is only set when `TargetType: instance`
- Update tests for validating webhook
  - update tests for ValidateCreate
  - update tests for ValidateUpdate
  - add tests for checkNodeSelector
- Update `GetTrafficProxyNodeSelector`
  -  to use `NodeSelector` is set, fallback to default
  - return an error
  - update all usages of this function to check for error
- Add tests for `GetTrafficProxyNodeSelector` 
- Update docs in `docs/guide/targetgroupbinding/targetgroupbinding.md`

### DIFF
--- a/config/crd/bases/elbv2.k8s.aws_targetgroupbindings.yaml
+++ b/config/crd/bases/elbv2.k8s.aws_targetgroupbindings.yaml
@@ -40,143 +40,262 @@ spec:
   scope: Namespaced
   subresources:
     status: {}
-  validation:
-    openAPIV3Schema:
-      description: TargetGroupBinding is the Schema for the TargetGroupBinding API
-      properties:
-        apiVersion:
-          description: 'APIVersion defines the versioned schema of this representation
-            of an object. Servers should convert recognized schemas to the latest
-            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
-          type: string
-        kind:
-          description: 'Kind is a string value representing the REST resource this
-            object represents. Servers may infer this from the endpoint the client
-            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
-          type: string
-        metadata:
-          type: object
-        spec:
-          description: TargetGroupBindingSpec defines the desired state of TargetGroupBinding
-          properties:
-            networking:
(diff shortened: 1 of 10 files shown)


config/crd/bases/elbv2.k8s.aws_targetgroupbindings.yaml
```diff
@@ -40,143 +40,262 @@ spec:
   scope: Namespaced
   subresources:
     status: {}
-  validation:
```
Smirl, line 43: This looks to the old format. Do I have a different version of `controller-gen`? I use 0.4.0 so I think this is fine to check in.

M00nF1sh:
/ok-to-test
thanks for contributing this feature 👍

config/crd/bases/elbv2.k8s.aws_targetgroupbindings.yaml
M00nF1sh: the controller-gen seems to have upgraded, i haven't updated the CRDs since launch, need to figure out what's changed

Smirl: Did you get a chance to take a look. Let me know if you want me to add any changes.

M00nF1sh (commented):
overall lgtm.
However, for the NodeSelector, i'm wondering what should be the semantic.
whether it should override the default selector or be in addition to the default selectors.

Smirl:
I can make the changes to the code and docs to make them additional if you prefer?

It might make sense as the defaults are quite sensible. Let me push the commit when I get time this weekend.

config/crd/bases/elbv2.k8s.aws_targetgroupbindings.yaml
M00nF1sh: it's fine to include these CRD changes. we'll adjust it if needed when cutting a release

M00nF1sh:
@Smirl 
Thanks, i discussed with our PM, adding the labelSelector on top of default labelSelector makes more sense. 
The changes lgtm if we can change to that.

PR #1793, opened 2021-02-03
docs: update helm install command

hi-rustin:
Update helm install command.
We need to specify the name or use the generated name, and in addition we need to set the cluster name.

### DIFF
--- a/docs/deploy/installation.md
+++ b/docs/deploy/installation.md
@@ -66,7 +66,7 @@ curl -o iam-policy.json https://raw.githubusercontent.com/kubernetes-sigs/aws-lo
     ```
     1. Install the helm chart
     ```
-    helm install eks/aws-load-balancer-controller -n kube-system --generate-name --set clusterName=<cluster-name>
+    helm install aws-load-balancer-controller eks/aws-load-balancer-controller -n kube-system --set clusterName=<cluster-name>
     ```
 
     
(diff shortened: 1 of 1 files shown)


docs/deploy/installation.md
```diff
@@ -66,7 +66,7 @@ curl -o iam-policy.json https://raw.githubusercontent.com/kubernetes-sigs/aws-lo
     ```
     1. Install the helm chart
     ```
-    helm install eks/aws-load-balancer-controller
+    helm install eks/aws-load-balancer-controller --generate-name --set clusterName=<cluster-name>
```
kishorj, line 69: - default name instead of the `--generate-name` looks better to me
- lets suggest a namespace, for example -n kube-system

hi-rustin: >     * default name instead of the `--generate-name` looks better to me
> 

It does not use the default name, but prompts you that you must specify a name or use --generate-name.

My env:
➜  ~ helm version
version.BuildInfo{Version:"v3.3.2", GitCommit:"e5077257b6ca106d1f65652b4ca994736d221ab1", GitTreeState:"dirty", GoVersion:"go1.15.2"}

hi-rustin: >     * lets suggest a namespace, for example -n kube-system

done

kishorj: by default name, I meant to use aws-load-balancer-controller. For example - 

```
helm install aws-load-balancer-controller  eks/aws-load-balancer-controller -n kube-system --set clusterName=<cluster-name>
```

hi-rustin: Got it.

kishorj: Thanks for your help.

kishorj (approved):

PR #1795, opened 2021-02-04
docs: update iam service account creation command

weiwarren:
Keep the sig doc consistent with documentation from AWS public documentations
https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html
--override-existing-serviceaccounts flag is required if service account is already in the cluster


M00nF1sh:
Thanks so much for fixing this doc 👍 
/lgtm
/approve

PR #1800, opened 2021-02-05
add mod=readonly

M00nF1sh:
add mod=readonly to prevent go command from updating go.sum and go.mod during build


kishorj (approved):
Thanks for fixing this.

Issue #1814, opened 2021-02-10
Issue with pod readiness gates not getting injected


itskingori:
I've been trying to understand why pod readiness gates are not being injected into my pods. After double/triple checking everything I'm starting to think that the issue might be because of this code ...

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/9c719f75e7c3e8252343e5b1eaf2192d87dfb374/pkg/inject/pod_readiness_gate.go#L90-L93

Correct me if I'm wrong but that piece of code means that the `Service` _'selector'_ labels and the pod labels **must be an exact match** with the `Pod` labels. So with this configuration that condition would evaluate to `false` and therefore not set up the readiness gates, right?

```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: example-app-nodeport
  namespace: default
  labels:
    app: example
    role: app
    team: none
spec:
  type: NodePort
  selector:
    app: example
    role: app
  ports:
    - name: http
      port: 80
      targetPort: http
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: example-app
  namespace: default
  labels:
    app: example
    role: app
    team: none
spec:
  selector:
    matchLabels:
      app: example
      role: app
      team: none
  template:
    metadata:
      namespace: default
      labels:
        app: example
        role: app
        team: none
    spec:
      containers:
        - name: caddy
          image: caddy:2.1.1-alpine
          ports:
            - containerPort: 80
              name: http
```

If that holds true, then isn't it better if we check if at least each `Service` selector label is in the `Pod` labels? So we're a subset would match even if they aren't exact?

itskingori:
I'm realising now that the issue is likely because my target type is `instance` and not `ip` ... 🤦🏽‍♂️ ... [and I quote from here](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.1/deploy/pod_readiness_gate/):

> Once labelled, the controller will add the pod readiness gates config to all the pods created subsequently that meet all the following conditions
>
> * There exists a service matching the pod labels in the same namespace
> * There exists at least one target group binding that refers to the matching service
> * The target type is IP

There's also https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1348#issuecomment-675783036:

> * whether it's headless service doesn't matter.
[...]

PR #1815, opened 2021-02-10
State supported target-type more clearly in pod readiness gate documentation

itskingori:
Because I spent too much time trying to figure out why pod readiness gates weren't being injected when I had `target-type: instance` all along. 🤦🏽‍♂️ 😅 

Closes https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1814.


M00nF1sh:
/lgtm
/approve
Thanks so much for helping improve the doc 😃

Issue #1744, opened 2020-12-22
nlb instance support


jurgenweber:
As per this; https://github.com/kubernetes/kubernetes/issues/57250#issuecomment-723390608

Apparently this project is taking over the NLB instance support, for those that need proxy-protocol v2 on + NLB and are not on k8s 1.20 (which is a [prerequisite for nlb-ip](https://kubernetes-sigs.github.io/aws-load-balancer-controller/guide/service/nlb_ip_mode/)). We are on eks, so 1.18.

Any status update on this? Are you aware? :)

kishorj:
@jurgenweber, It is in our roadmap, we are working on it.

jurgenweber:
since you answered this question; https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1745 meaning I don't specifically need this but there are people out there going from those threads that are crying out for it. It was hard to find good information so hopefully this helps those whom are searching.

indrekj:
Any idea when v2.2 will be released?

hanjh0814:
Any updates with regards to this issue?

kishorj:
@hanjh0814, v2.2 is already released. Please refer to the release notes at - https://github.com/kubernetes-sigs/aws-load-balancer-controller/releases/tag/v2.2.0.

PR #1832, opened 2021-02-18
Add support for NLB instance mode

kishorj:
Fixes: #1744 

## Overview
The support for NLB instance mode is available from the k8s service controller, and any changes or improvements are tied to the upstream release schedule. In addition, there are also limitations on using CRDs in the in-tree controller.

This PR is for adding support for NLB instance mode in the AWS Load balancer controller. The controller already supports instance mode for ALB, and we will build up on  the `TargetGroupBinding` feature for NLB instance mode as well. 

Moving forward, the k8s AWS service controller will have support for NLB instance mode and CLB with basic features only. For more NLB features like the TargetGroupBinding, proxy v2 support, improved health checks, IP target mode we recommend running the AWS load balancer controller. Each service of type LoadBalancer on AWS k8s is reconciled either by the in-tree controller or the external controller based on the annotations provided the version requirements are met.

## k8s version requirement
The instance mode support for service type LoadBalancer requires k8s 1.20 (non-EKS) or later where we've modified the in-tree controller to ignore service with aws-load-balancer-type annotation of `external` or `nlb-ip`. For EKS, we've backported the changes to EKS 1.16 and later releases.

## How to configure load balancer target type?
If the service annotation `service.beta.kubernetes.io/aws-load-balancer-type` is external, this controller will look for an additional annotation `service.beta.kubernetes.io/aws-load-balancer-target-type` to determine the target type. The current supported values for the target types annotation are `nlb-instance` and `nlb-ip`.

### Instance target
Example service spec for instance target
```
kind: Service
apiVersion: v1
metadata:
  name: nlb-svc
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-type: "external"
    service.beta.kubernetes.io/aws-load-balancer-target-type: "nlb-instance"
spec:
  ports:
    - name: port
      port: 80
      targetPort: 80
      protocol: TCP
  type: LoadBalancer
  selector:
    app: instance-mode
```

### IP target
For consistency with instance target mode, IP mode will also be determined based on the following annotations - 
```
service.beta.kubernetes.io/aws-load-balancer-type: external
service.beta.kubernetes.io/aws-load-balancer-target-type: nlb-ip
```
For backwards compatibility, the `service.beta.kubernetes.io/aws-load-balancer-type` annotation value of `nlb-ip` will still be interpreted as IP mode NLB.

### Alternative considered
Modify the in-tree controller to support additional value for the `service.beta.kubernetes.io/aws-load-balancer-type` annotation, for example 
```
service.beta.kubernetes.io/aws-load-balancer-type: nlb-instance
```
Pros:
- Single annotation, and is similar to `nlb-ip`

Cons:
- We have to backport the change to all supported EKS versions, and will be dependent on the patch rollout. This results in different behavior across multiple platform versions and will be more confusing to the end users
- One more patch to the in-tree controller we need to maintain.

## Testing
- Create service of type Load Balancer with instance targets, External Traffic Policy set to Cluster
- Create service of type Load Balancer with instance targets, External Traffic Policy set to Local
- health check annotations get honored
- proxy protocol v2
- preserve client IP can be disabled for instance mode
- Security group rules get added as expected
- The labels specified in`service.beta.kubernetes.io/aws-load-balancer-target-node-labels` annotation works as expected
   - Without this annotation, the default filter gets applied
   - When service is created/edited with this annotaton, only matching nodes get registered
   - On edit, any nodes not matching the labels get de-registered from the target group
   - Node label edits are correctly reconciled

### DIFF
--- a/pkg/model/elbv2/target_group_binding.go
+++ b/pkg/model/elbv2/target_group_binding.go
@@ -97,7 +97,7 @@ type TargetGroupBindingSpec struct {
 	Networking *TargetGroupBindingNetworking `json:"networking,omitempty"`
 
 	// node selector for instance type target groups to only register certain nodes
-	//+optional
+	// +optional
 	NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
 }
 
--- a/pkg/model/elbv2/target_group_binding.go
+++ b/pkg/model/elbv2/target_group_binding.go
@@ -95,6 +95,10 @@ type TargetGroupBindingSpec struct {
 	// networking provides the networking setup for ELBV2 LoadBalancer to access targets in TargetGroup.
 	// +optional
 	Networking *TargetGroupBindingNetworking `json:"networking,omitempty"`
+
+	// node selector for instance type target groups to only register certain nodes
+	//+optional
+	NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
 }
 
 // Template for TargetGroupBinding Custom Resource.
--- a/pkg/service/model_build_target_group.go
+++ b/pkg/service/model_build_target_group.go
@@ -53,8 +53,11 @@ func (t *defaultModelBuildTask) buildTargetGroup(ctx context.Context, port corev
 		return nil, err
 	}
 	targetGroup := elbv2model.NewTargetGroup(t.stack, tgResourceID, tgSpec)
+	_, err = t.buildTargetGroupBinding(ctx, targetGroup, preserveClientIP, port, healthCheckConfig)
+	if err != nil {
+		return nil, err
+	}
 	t.tgByResID[tgResourceID] = targetGroup
-	_ = t.buildTargetGroupBinding(ctx, targetGroup, preserveClientIP, port, healthCheckConfig)
 	return targetGroup, nil
 }
 
@@ -196,12 +199,12 @@ func (t *defaultModelBuildTask) buildPreserveClientIPFlag(_ context.Context, tar
 func (t *defaultModelBuildTask) buildTargetGroupHealthCheckPort(_ context.Context) (intstr.IntOrString, error) {
 	rawHealthCheckPort := t.defaultHealthCheckPort
 	t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixHCPort, &rawHealthCheckPort, t.service.Annotations)
-	if rawHealthCheckPort == t.defaultHealthCheckPort {
+	if rawHealthCheckPort == healthCheckPortTrafficPort {
 		return intstr.FromString(rawHealthCheckPort), nil
 	}
-	var portVal int64
-	if _, err := t.annotationParser.ParseInt64Annotation(annotations.SvcLBSuffixHCPort, &portVal, t.service.Annotations); err != nil {
-		return intstr.IntOrString{}, err
+	portVal, err := strconv.ParseInt(rawHealthCheckPort, 10, 64)
+	if err != nil {
+		return intstr.IntOrString{}, errors.Errorf("health check port \"%v\" not supported", rawHealthCheckPort)
 	}
 	return intstr.FromInt(int(portVal)), nil
 }
@@ -276,7 +279,17 @@ func (t *defaultModelBuildTask) buildTargetGroupHealthCheckUnhealthyThresholdCou
 }
 
 func (t *defaultModelBuildTask) buildTargetType(_ context.Context) (elbv2model.TargetType, error) {
-	return elbv2model.TargetTypeIP, nil
+	var lbType string
+	_ = t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixLoadBalancerType, &lbType, t.service.Annotations)
+	var lbTargetType string
+	_ = t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixTargetType, &lbTargetType, t.service.Annotations)
+	if lbType == LoadBalancerTargetTypeNLBIP || (lbType == LoadBalancerTypeExternal && lbTargetType == LoadBalancerTargetTypeNLBIP) {
+		return elbv2model.TargetTypeIP, nil
+	}
+	if lbType == LoadBalancerTypeExternal && lbTargetType == LoadBalancerTargetTypeNLBInstance {
+		return elbv2model.TargetTypeInstance, nil
+	}
+	return "", errors.New("unsupported target type")
 }
 
 func (t *defaultModelBuildTask) buildTargetGroupResourceID(svcKey types.NamespacedName, port intstr.IntOrString) string {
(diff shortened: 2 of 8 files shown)


TBBle:
Does this implementation also suffer from https://github.com/kubernetes/cloud-provider-aws/issues/87 with `externalTrafficPolicy: Local`? And if so, would it make sense to (later) have `service.beta.kubernetes.io/aws-load-balancer-target-type: nlb-instance` and `externalTrafficPolicy: Local` together trigger a different registration mode which registers only instances with local `Endpoints` for the given `Service`? I don't believe this was possible via the Cloud Provider APIs (there was no notification I could see for `Endpoint` changes), but with the distinct Load Balancer Controller, I think this could be feasible. It perhaps presents scaling concerns due to having to react to `Endpoint` create/delete instead of just `Node` create/delete.

Alternatively, just documenting that if you want `externalTrafficPolicy: Local`, you're better off with `service.beta.kubernetes.io/aws-load-balancer-target-type: nlb-ip` to avoid this "known issue" would be fine.

Asking due to #1842. It's not _directly_ related due to as it doesn't actually mention `externalTrafficPolicy: Local`, but it prompted me to wonder this.

M00nF1sh:
@TBBle Yes, the initial implementation will still suffer from this NLB issue as well. (NLB is working on changes on their end to avoid initial targets receive traffic).

We might follow up PRs to optimize the issue in the way as you described. (However, it will still suffer from the long NLB registration delay you saw in NLB-IP mode, e.g. pod relocates to another Node).

pkg/model/elbv2/target_group_binding.go
```diff
@@ -95,6 +95,10 @@ type TargetGroupBindingSpec struct {
 	// networking provides the networking setup for ELBV2 LoadBalancer to access targets in TargetGroup.
 	// +optional
 	Networking *TargetGroupBindingNetworking `json:"networking,omitempty"`
+
+	// node selector for instance type target groups to only register certain nodes
+	//+optional
```
M00nF1sh, line 100: nit, it need to be `// +optional`

pkg/service/model_build_target_group.go
```diff
@@ -276,7 +279,17 @@ func (t *defaultModelBuildTask) buildTargetGroupHealthCheckUnhealthyThresholdCou
...
+	_ = t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixLoadBalancerType, &lbType, t.service.Annotations)
+	var lbTargetType string
+	_ = t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixTargetType, &lbTargetType, t.service.Annotations)
+	if lbType == LoadBalancerTargetTypeNLBIP || (lbType == LoadBalancerTypeExternal && lbTargetType == LoadBalancerTargetTypeNLBIP) {
+		return elbv2model.TargetTypeIP, nil
+	}
+	if lbType == LoadBalancerTypeExternal && lbTargetType == LoadBalancerTargetTypeNLBInstance {
+		return elbv2model.TargetTypeInstance, nil
+	}
+	return "", errors.New("unsupported target type")
```
M00nF1sh, line 292: nit: better to have lbType and lbTargetType in err message

pkg/service/model_builder.go
```diff
@@ -63,6 +71,20 @@ func (b *defaultModelBuilder) Build(ctx context.Context, service *corev1.Service
 		defaultHealthCheckHealthyThreshold:   3,
 		defaultHealthCheckUnhealthyThreshold: 3,
 	}
+	targetType, err := task.buildTargetType(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+	if targetType == elbv2model.TargetTypeInstance &&
```
M00nF1sh, line 78: tbh, i think this is a little bit hacky here. model_builder's build should be simply and delegate to task run directly(instead of have different default values here).

instead, i think it's better to split these default values into different variables.
e.g. `defaultHealthCheckintervalForIPMode` and `defaultHealthCheckIntervalForLocalInstanceMode`.

pkg/model/elbv2/target_group_binding.go
kishorj: ok

pkg/service/model_build_target_group.go
kishorj: I'll change it.

pkg/service/model_builder.go
kishorj: I was trying to avoid more code. I will decouple the default values and update the PR.

pkg/service/model_build_target_group.go
```diff
@@ -221,62 +282,56 @@ func (t *defaultModelBuildTask) buildTargetGroupHealthCheckProtocol(_ context.Co
...
-	}
-
-	// when a literal targetPort is used, we just use a fixed 1 here as this setting is not in the data path.
-	// also, under extreme edge case, it can actually be different ports for different pods.
-	return 1
-}
-
-func (t *defaultModelBuildTask) buildTargetGroupHealthCheckPath(_ context.Context) *string {
-	healthCheckPath := t.defaultHealthCheckPath
+func (t *defaultModelBuildTask) buildTargetGroupHealthCheckPath(_ context.Context, defaultHealthCheckPath string) *string {
```
M00nF1sh, line 285: personally i think we should return string instead. since it always going to be a string instead of nil.

pkg/service/model_build_target_group.go
```diff
@@ -221,62 +282,56 @@ func (t *defaultModelBuildTask) buildTargetGroupHealthCheckProtocol(_ context.Co
...
+	_ = t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixLoadBalancerType, &lbType, t.service.Annotations)
+	var lbTargetType string
+	_ = t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixTargetType, &lbTargetType, t.service.Annotations)
+	if lbType == LoadBalancerTargetTypeNLBIP || (lbType == LoadBalancerTypeExternal && lbTargetType == LoadBalancerTargetTypeNLBIP) {
+		return elbv2model.TargetTypeIP, nil
+	}
+	if lbType == LoadBalancerTypeExternal && lbTargetType == LoadBalancerTargetTypeNLBInstance {
+		return elbv2model.TargetTypeInstance, nil
+	}
+	return "", errors.Errorf("unsupported target type \"%v\" for load balancer type \"%v\"", lbTargetType, lbType)
```
M00nF1sh, line 334: not a real code comment here but a node. this error should never trigger because we filtered out it in eventHandlers.

M00nF1sh (approved):
/lgtm

pkg/service/model_build_target_group.go
kishorj: I will change it in a follow up commit.

pkg/service/model_build_target_group.go
kishorj: correct. I added since model builder is independent from the event handler module. I will consider removing this error in subsequent commits.

PR #1837, opened 2021-02-19
add support for ssl-redirect annotation

M00nF1sh:
# add a new annotation "ssl-redirect" to simply the configuration for HTTP-HTTPS redirect

## What is the problem
1. The current solution (https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/tasks/ssl_redirect/) to configure HTTP→HTTPS redirect via AWSLoadBalancer is complicated and intrusive to customer’s Ingress rules.
2. The current solution (https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/tasks/ssl_redirect/) to configure HTTP→HTTPS redirect via AWSLoadBalancer is not AWS-config compatible.
    1. AWS Config is a AWS service that validates best practices on AWS resource configurations.
    2. with current solution, we’ll have a catch all rule(*path:/**) to do SSL-redirect instead of using the default rule per listener (which AWS config expects).
3. The current solution (https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/tasks/ssl_redirect/) to configure HTTP→HTTPS redirect is tricky, that relies on an implementation detail that the controller will ignore infinite redirect rules.
    1. when generate rules for port 80, the controller inserts the redirect-to-ssl rule as expected.
    2. when generate rules for port 443, the controller ignores the redirect-to-ssl rule as it will be a infinite redirect.

## What will be changed
We’ll introduce a new annotation called “alb.ingress.kubernetes.io/ssl-redirect” to do SSL redirection.
The value of alb.ingress.kubernetes.io/ssl-redirect is simply the SSL port to redirect to. e.g. alb.ingress.kubernetes.io/ssl-redirect: `443`

1. The port must be appears in the listen-port for Ingresses, and must be a HTTPS port.
2. All HTTP port for Ingresses will be redirect to this SSL port under *HTTP_301* redirect.

Note: 

1. currently ALB only supports *HTTP_301* or *HTTP_302* for ** redirect actions. And HTTP_301 is the preferred status_code for HTTP to HTTPS redirection. 
2. There are also a HTTP_308 status code can be used be achieve HTTP to HTTPS redirection. However, ALB currently don’t support it. If ALB supports it in the future, we can add a separate annotation like alb.ingress.kubernetes.io/ssl-redirect-status-code: HTTP_308 to further customize the behavior for advanced users.

Alternative designs considered:

1. we can allow specify both port and status code for the SSL redirect, like alb.ingress.kubernetes.io/ssl-redirect: '{"port":443, "status_code": "HTTP_301"}'
    1. pros: more flexible since we can have more settings like status_code.
    2. cons: more complicated to configure.

### DIFF
--- a/pkg/ingress/model_build_actions_test.go
+++ b/pkg/ingress/model_build_actions_test.go
@@ -248,3 +248,39 @@ func Test_defaultModelBuildTask_buildAuthenticateOIDCAction(t *testing.T) {
 		})
 	}
 }
+
+func Test_defaultModelBuildTask_buildSSLRedirectAction(t *testing.T) {
+	type args struct {
+		sslRedirectConfig SSLRedirectConfig
+	}
+	tests := []struct {
+		name string
+		args args
+		want elbv2model.Action
+	}{
+		{
+			name: "SSLRedirect to 443 with 301",
+			args: args{
+				sslRedirectConfig: SSLRedirectConfig{
+					SSLPort:    443,
+					StatusCode: "HTTP_301",
+				},
+			},
+			want: elbv2model.Action{
+				Type: elbv2model.ActionTypeRedirect,
+				RedirectConfig: &elbv2model.RedirectActionConfig{
+					Port:       awssdk.String("443"),
+					Protocol:   awssdk.String("HTTPS"),
+					StatusCode: "HTTP_301",
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t1 *testing.T) {
+			task := &defaultModelBuildTask{}
+			got := task.buildSSLRedirectAction(context.Background(), tt.args.sslRedirectConfig)
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
--- a/pkg/ingress/model_builder.go
+++ b/pkg/ingress/model_builder.go
@@ -3,6 +3,7 @@ package ingress
 import (
 	"context"
 	awssdk "github.com/aws/aws-sdk-go/aws"
+	elbv2sdk "github.com/aws/aws-sdk-go/service/elbv2"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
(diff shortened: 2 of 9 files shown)


pkg/ingress/model_builder.go
```diff
@@ -276,3 +283,36 @@ func (t *defaultModelBuildTask) mergeListenPortConfigs(_ context.Context, listen
...
+
+	if len(explicitSSLRedirectPorts) == 0 {
+		return nil, nil
+	}
+	if len(explicitSSLRedirectPorts) > 1 {
+		return nil, errors.Errorf("conflicting sslRedirect port: %v", explicitSSLRedirectPorts.List())
+	}
+	rawSSLRedirectPort, _ := explicitSSLRedirectPorts.PopAny()
+	if listenPortConfig, ok := listenPortConfigByPort[rawSSLRedirectPort]; !ok {
+		return nil, errors.Errorf("listener non-exists for SSLRedirect port: %v", rawSSLRedirectPort)
```
kishorj, line 309: "listener non-existent ... " or "listener does not exist ... " looks better to me.

kishorj (commented):
I have some minor comments, changes look good otherwise.

pkg/ingress/model_build_actions_test.go
```diff
@@ -248,3 +248,39 @@ func Test_defaultModelBuildTask_buildAuthenticateOIDCAction(t *testing.T) {
...
+	type args struct {
+		sslRedirectConfig SSLRedirectConfig
+	}
+	tests := []struct {
+		name string
+		args args
+		want elbv2model.Action
+	}{
+		{
+			name: "SSLRedirect to 332 with 301",
```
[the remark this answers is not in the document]
M00nF1sh: thanks :D

pkg/ingress/model_builder.go
M00nF1sh: fixed :D make sense

PR #1845, opened 2021-02-24
added all to targetgroup update

### LINKED ISSUE #1569: [Feature Request] TargetGroupBinding can't bind pod ip to another vpc's  targetgroup
**Short Version**
I create one  TargetGroupBinding resource , want to binding pod ip to a targetGroup (type is ip) which belong to another vpc, but aws-load-balancer-controller report error with **"ValidationError: The Availability Zone for IP address '192.168.11.189' must be 'all' for Application Load Balancer target groups, when not within the VPC"**.

Full error message: "{"level":"error","ts":1603805204.1610796,"logger":"controller","msg":"Reconciler error","reconcilerGroup":"elbv2.k8s.aws","reconcilerKind":"TargetGroupBinding","controller":"targetGroupBinding","name":"php-apache-1","namespace":"default","error":"ValidationError: The Availability Zone for IP address '192.168.11.189' must be 'all' for Application Load Balancer target groups, when not within the VPC\n\tstatus code: 400, request id"

**Long Version**

1. Create a targetgroup which belong to VPC A , Target type is ip.
1. Create a TargetGroupBinding, spec.targetType is also ip, spec.targetGroupARN is the targetgroup created in step one.
1. Get the error.

There have another similar bug reported but had been fixed :https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/783. So i think this error maybe triggered by a bug.

acjohnson:
Not finished but I'd like to keep going with this, if you have any questions or comments let me know.

Fixes #1569

### DIFF
--- a/pkg/targetgroupbinding/resource_manager.go
+++ b/pkg/targetgroupbinding/resource_manager.go
@@ -296,8 +296,9 @@ func (m *defaultResourceManager) registerPodEndpoints(ctx context.Context, tgARN
 	sdkTargets := make([]elbv2sdk.TargetDescription, 0, len(endpoints))
 	for _, endpoint := range endpoints {
 		sdkTargets = append(sdkTargets, elbv2sdk.TargetDescription{
-			Id:   awssdk.String(endpoint.IP),
-			Port: awssdk.Int64(endpoint.Port),
+			Id:               awssdk.String(endpoint.IP),
+			Port:             awssdk.Int64(endpoint.Port),
+			AvailabilityZone: awssdk.String("all"),
 		})
 	}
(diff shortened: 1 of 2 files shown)


M00nF1sh:
@acjohnson 
Thanks for this PR. however, it cannot be simply "ALL". We need to check whether the IP address is within VPC.

acjohnson:
@M00nF1sh the value needs to be set to string `"all"` for cross VPC access to work with target groups. No checking whether the IP is within the VPC should be necessary.

That said I don't intend to just hardcode it. It should be a boolean annotation in my opinion.

If I add the annotation code and go tests will that work?

Thanks!

acjohnson:
Closed in favor of #1862

PR #1850, opened 2021-03-01
remove depericated apiVersoin

ngoyal16:
Depericated apiVersion removed in Ingress resource

### DIFF
--- a/docs/examples/2048/2048_full.yaml
+++ b/docs/examples/2048/2048_full.yaml
@@ -40,7 +40,7 @@ spec:
   selector:
     app.kubernetes.io/name: app-2048
 ---
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   namespace: game-2048
--- a/docs/examples/2048/2048_full.yaml
+++ b/docs/examples/2048/2048_full.yaml
@@ -40,7 +40,7 @@ spec:
   selector:
     app.kubernetes.io/name: app-2048
 ---
-apiVersion: networking.k8s.io/v1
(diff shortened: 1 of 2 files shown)


M00nF1sh:
@ngoyal16 
Thanks for making this PR.
would mind change it to networking.k8s.io/v1beta1 (the backend syntax differs from v1 slightly as well).
`networking.k8s.io/v1` is not available for old supported EKS versions(e.g. 1.17) currently, and this file is used in EKS docs.

Another optional alternative is create another file like `docs/examples/2048/2048_full_latest.yaml` for latest apis.

ngoyal16:
@M00nF1sh  I have chanced the api version for now.

tomislater:
Hmm, is this example valid?

We use `networking.k8s.io/v1beta1`, but the section

```
service:
  name: service-2048
  port:
    number: 80
```
is valid only in `networking.k8s.io/v1` as far as I know.

In `1.18` we still must use `networking.k8s.io/v1beta1`: https://v1-18.docs.kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource

Since `1.19` we can use `networking.k8s.io/v1` (it will be removed in `1.22`), and in this version we must use `service.name`, `service.port.number`, etc.: https://v1-19.docs.kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource

BTW. Can I spawn on my EKS Cluster (`1.19`) a new Ingress object (`networking.k8s.io/v1`) and `aws-load-balancer-controller` will create (or update existing one) ALB on AWS?

ngoyal16:
@tomislater issue fixed.

M00nF1sh (approved):
/lgtm
thanks for updating this :D

Issue #1836, opened 2021-02-19
alb.ingress.kubernetes.io/certificate-arn ordering: first certificate arn should be load bal default


bbacker:
Thank you for all your work on this project.

We are running aws-load-balancer-controller:v2.1.1 and have a couple load balancers with multiple certificates.
It was our understanding that the first listed certificate should be the default certificate.

However after applying an alb definition starting like this:

    apiVersion: extensions/v1beta1
    kind: Ingress
    metadata:
      name: "ingress-vert1"
      namespace: prod
      annotations:
        kubernetes.io/ingress.class: alb
        alb.ingress.kubernetes.io/subnets: 'subnet-1, subnet-2, subnet-3'
        alb.ingress.kubernetes.io/certificate-arn: 'arn1, arn2, arn3'
        alb.ingress.kubernetes.io/listen-ports: '[{"HTTP":80,"HTTPS": 443}]'

However after applying, I find my load balancer with arn2 as the default.  (full arns shortened for brevity)

Is this expected behavior? If ordering is not yet important, could this be an enhancement request that the first listed
arn1 be made the load balancer default?

We have a very old client that is sensitive to which certificate is the default.

M00nF1sh:
/kind bug
@bbacker 
It is the extract behavior. This is a bug we introduced during v1->v2 rewrite. we'll fix this in next release.

bbacker:
@M00nF1sh - thank you for the quick reply, and glad to hear it's on the list to fix. thanks

BEvgeniyS:
Can we at least update the documentation so it would warn users of potential issues? Docs still claim that order is maintained:


```The first certificate in the list will be added as default certificate. And remaining certificate will be added to the optional certificate list.```

kishorj:
Fixed in v2.2.0 release - https://github.com/kubernetes-sigs/aws-load-balancer-controller/releases/tag/v2.2.0

PR #1863, opened 2021-03-08
Fixed an issue where mergedTLSCerts would be sorted.

yutachaos:
# Description
- Added model builder using multiple certificate cases.
- I changed to use Unsorted List because the order of certificate list changes even though I want to default certificate in the order of annotation.
    - https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/9776d298fddc1bb7fe9245510467d2507580ec96/docs/guide/ingress/annotation.md#ssl 

## Issue
- https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1836

### DIFF
--- a/pkg/ingress/model_builder.go
+++ b/pkg/ingress/model_builder.go
@@ -232,7 +232,7 @@ func (t *defaultModelBuildTask) mergeListenPortConfigs(_ context.Context, listen
 	var mergedSSLPolicyProvider *types.NamespacedName
 	var mergedSSLPolicy *string
 
-	mergedTLSCerts := sets.NewString()
+	var mergedTLSCerts []string
 
 	for ingKey, cfg := range listenPortConfigByIngress {
 		if mergedProtocolProvider == nil {
@@ -267,7 +267,7 @@ func (t *defaultModelBuildTask) mergeListenPortConfigs(_ context.Context, listen
 					*mergedSSLPolicyProvider, awssdk.StringValue(mergedSSLPolicy), ingKey, awssdk.StringValue(cfg.sslPolicy))
 			}
 		}
-		mergedTLSCerts.Insert(cfg.tlsCerts...)
+		mergedTLSCerts = append(mergedTLSCerts, cfg.tlsCerts...)
 	}
 
 	if len(mergedInboundCIDRv4s) == 0 && len(mergedInboundCIDRv6s) == 0 {
@@ -283,7 +283,7 @@ func (t *defaultModelBuildTask) mergeListenPortConfigs(_ context.Context, listen
 		inboundCIDRv4s: mergedInboundCIDRv4s.List(),
 		inboundCIDRv6s: mergedInboundCIDRv6s.List(),
 		sslPolicy:      mergedSSLPolicy,
-		tlsCerts:       mergedTLSCerts.UnsortedList(),
+		tlsCerts:       mergedTLSCerts,
 	}, nil
 }
 
--- a/pkg/ingress/model_builder.go
+++ b/pkg/ingress/model_builder.go
@@ -171,26 +171,27 @@ func (t *defaultModelBuildTask) run(ctx context.Context) error {
 	}
 
 	ingListByPort := make(map[int64][]*networking.Ingress)
-	listenPortConfigsByPort := make(map[int64]map[types.NamespacedName]listenPortConfig)
+	listenPortConfigsByPort := make(map[int64][]listenPortConfigWithIngress)
 	for _, member := range t.ingGroup.Members {
+		ingKey := k8s.NamespacedName(member.Ing)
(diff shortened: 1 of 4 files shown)


pkg/ingress/model_builder.go
```diff
@@ -280,7 +280,7 @@ func (t *defaultModelBuildTask) mergeListenPortConfigs(_ context.Context, listen
 		inboundCIDRv4s: mergedInboundCIDRv4s.List(),
 		inboundCIDRv6s: mergedInboundCIDRv6s.List(),
 		sslPolicy:      mergedSSLPolicy,
-		tlsCerts:       mergedTLSCerts.List(),
+		tlsCerts:       mergedTLSCerts.UnsortedList(),
```
kishorj, line 283: This will list out the certs in random order and the outcome is not deterministic. We'd need to consider the following issues
- maintain the order specified in the annotation
- handle multiple ingresses with the certificate annotation (provide some reasonable ordering)

yutachaos: > maintain the order specified in the annotation

I have two ideas.

1 
    Create a field called defaultCertificate in the request struct for the v1 implementation.
    https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/9776d298fddc1bb7fe9245510467d2507580ec96/internal/alb/ls/listener.go#L241-L246
2. 
    Stop using sets in mergedTLSCerts and change to slice.
    
> handle multiple ingresses with the certificate annotation (provide some reasonable ordering)

I don't have an idea, but I think it's going to be a big change because I can't go without changing the specifications below.

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/9776d298fddc1bb7fe9245510467d2507580ec96/docs/guide/ingress/annotation.md#certificate-arn

However, I have the same problem as the issue below, so I would like a solution.
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1836

yutachaos: Alternatively, evaluate in the order of group order, and make group order the largest one first.
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/48b4a7dd3177cab816dd678cdec460550e0cd528/docs/guide/ingress/annotations.md#group.order

M00nF1sh:
@yutachaos 
Thanks for making this change.
For certificate specified in multiple Ingresses within IngressGroup, i think the semantic should be use the first certificate of **lowest** group.order.    (since we used the first certificate as default when deal with a list of certificates, even the default certificate is evaluated last by alb).

yutachaos:
@M00nF1sh 
Thank you for your comment.

> @yutachaos
> BTW, `t.ingGroup.Members` is already ordered Ingresses by its group.order, so you'll need to first get all certificate into a list, dedup from them while maintain order.

https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/1863/files#diff-b467d2353d5bd2732f344b8bf4127d6318126b5dd85dec3c1726a970a19138a4R283

Certainly, there is a duplication in certs here.

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/deploy/elbv2/listener_manager.go#L128-L132

At this point, entering sets will remove the duplicates, so it looks okay.

yutachaos:
> BTW, `t.ingGroup.Members` is already ordered Ingresses by its group.order, so you'll need to first get all certificate into a list, dedup from them while maintain order.

I tried to remove duplicates with buildSDKCertificates function. How about that?

https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/1863/commits/a705fd359ead866dd19cd5bc9fc18c605aa8c5f9

M00nF1sh:
> > BTW, `t.ingGroup.Members` is already ordered Ingresses by its group.order, so you'll need to first get all certificate into a list, dedup from them while maintain order.
> 
> I tried to remove duplicates with buildSDKCertificates function. How about that?
> 
> [a705fd3](https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/a705fd359ead866dd19cd5bc9fc18c605aa8c5f9)

this works, but could we do this deduplication in  model builder? (the model deployer should be treated like a cloudFormation stack, it accepts **already sanitized/normalized data**).

also, the current implementation didn't handle ordering among multiple Ingresses within ingressGroup. (we are looping through `map[types.NamespacedName]listenPortConfig`, where the iteration order is undefined)

yutachaos:
> this works, but could we do this deduplication in model builder? (the model deployer should be treated like a cloudFormation stack, it accepts already sanitized/normalized data).

Understood.
It may take some time, but let's think about how to do it

M00nF1sh:
Hi @yutachaos 
I pushed an additional commit on top of your changes, which should cover the IngressGroup ordering issue.

kishorj (approved):

PR #1865, opened 2021-03-08
feat: Add WatchIPBlocks and WatchInstanceFilters to TargetGroupBinding

nicolasgarfinkiel:
### Introduction

This PR adds `WatchIPBlocks` and `WatchInstanceFilters` to `TargetGroupBinding`.

In our use case our product is deployed in multiple EKS clusters that are load balanced through a common ALB. When sharing a single target group among many `aws-load-balancer-controller` instances registration/deregistration of one controller affects other controllers. This PR makes a `TargetGroupBinding` to only care about a specific instance group or address space and leave the rest unmanaged.

### WatchIPBlocks

When using `TargetType: ip`, you can specify IP blocks in CIDR notation
to only list (from AWS) ip targets that fall within their ranges. This is 
useful when sharing a single target group among many `aws-load-balancer-controller`
instances (for example across different clusters), so registration/deregistration
of one controller does not affect other controllers.

```yaml
kind: TargetGroupBinding
metadata:
  name: my-tgb
spec:
  watchIPBlocks:
    - cidr: 10.1.1.0/24
    - cidr: 10.1.2.0/24
  ...
```

### WatchInstanceFilters

When using `TargetType: instance`, you can specify filters to only list
(from AWS) instance targets that match the specified filters. This is 
useful when sharing a single target group among many `aws-load-balancer-controller`
instances (for example across different clusters), so registration/deregistration
of one controller does not affect other controllers.

Please read [this reference](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html)
for more information about these filters.

```yaml
kind: TargetGroupBinding
metadata:
  name: my-tgb
spec:
  watchInstanceFilters:
    - name: "tag:someTag"
      values:
        - value1
        - value2
  ...
```

### DIFF
--- a/config/crd/bases/elbv2.k8s.aws_targetgroupbindings.yaml
+++ b/config/crd/bases/elbv2.k8s.aws_targetgroupbindings.yaml
@@ -48,10 +48,14 @@ spec:
         description: TargetGroupBinding is the Schema for the TargetGroupBinding API
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
             type: string
           kind:
-            description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
             type: string
           metadata:
             type: object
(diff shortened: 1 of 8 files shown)


M00nF1sh:
@nicolasgarfinkiel 
Hi, i think this works but it's too hacky tbh..
ideally, we should only support this if ELBv2 API supports tags on targets registered.

nicolasgarfinkiel:
@M00nF1sh
I think it opens the door to some configuration corner cases but I don't see why you consider it hacky. Can you please elaborate?

We are narrowing the controller area of responsibility using current API constructs: in the case `type: instances` by AWS API filters and the case of `type: ip` by IPBlock type already present on the controller.

I actually added these as arguments to the controller's executable but since there was `nodeSelector` in `TargetGroupBinding` I thought it was best to keep filtering config in the same place. Do you see that less hacky and more viable?

We need to have this feature because of our use case, so in case you guys don't accept this, what is your recommendation to support this feature, other than maintaining our own fork? I would like to avoid maintaining a fork as much as possible.

M00nF1sh:
@nicolasgarfinkiel 

I think it's hacky in the way how it sub-divides a TargetGroup and how it treats IP/Instance Targets.

If we sub-divides a targetGroup, ideally it should work like `a controller only manages targets in TargetGroup added by it`. However, in this case, we are identify the `targets added a controller` using IPFilter and InstanceTags filter, which are limited to this specific use case.  (an sample case cannot be satisfied by this is like share same targetGroup across multiple services within a single cluster).
Ideally, it should work like some tags/attributes associated with each target(like how cloudmap API did: https://docs.aws.amazon.com/cloud-map/latest/api/API_RegisterInstance.html). But ELB team doesn't have plans to support this.

another approach is instead of sub-divides a TargetGroup, we use weighted routing, so each set of Targets get it's own TargetGroup. However, only ALB supports this currently, and NLB support for this is planed for 2022.


so, personally i'm opposite to add this as fields into TGB's API as it's not generic and adds unnecessary burden for generic users.
but i'm open to add it as a flag to the controller binary for this specific use case/setup. 

wondering what's the opinion of other folks? @kishorj

kishorj:
My concerns
- Having multiple controllers across clusters modify same target group will make it difficult to maintain consistency. For example, if the watch IP block gets misconfigured on a cluster, then we might end up with a hard situation to debug.
- With the proposed changes we also assume that all of the clusters are at the same security level, and this enforcing this is upto the end user. If clusters are at different security levels, we open up for privilege escalation where controller on lower security level can interfere with the targets registered by other clusters.
- Even if we are to go the flag route, the CRDs will still have the fields. I'm assuming we will disable this flag by default. In that case, there is a potential for confusion.

With the current controller, we aren't ready to support cross-cluster modification of resources in a reliable and secure way. While this use case is compelling, we should look at whether alternative solutions are possible.

nicolasgarfinkiel:
@M00nF1sh

I did consider what you mention as _"a TGB only manages targets in TargetGroup added by it"_, and concluded that by filtering we can achieve that without actually managing state, which would require substancial work and also would open a can of worms.

I like the idea of adding this feature as a flag to the controller binary (as originally intended) as this makes more sense on a cluster wide level. I will update the PR accordingly.

@kishorj

Thanks for your comments.

- AFAICT if you miss configure IP blocks are either ip targets registered that never get removed, easily identifiable when they become unhealthy, or two controllers fighting each other for register/deregister, which is what would happen if you currently point many TGBs to the same Target Group.
- It is a very valid point that we can specify as a caveat in the docs, IMO.
- I see no need to keep the fields in the CRDs IMO, do you think we should?

Thanks.

nicolasgarfinkiel:
@M00nF1sh @kishorj Hello. Can you please advice if this PR will be admitted? If so I'll rebase it to leave it ready to merge. Otherwise I'll close it.

Thanks.

PR #1874, opened 2021-03-10
Support target node labels for ingress

kishorj:
Add support for target node labels for ingress resources. This label will allow the user to register the matching nodes in the target group.  The labels are specified via the following StringMap annotation - 
```
alb.ingress.kubernetes.io/target-node-labels
```
This annotation can be specified in the ingress or the service resources. The annotation on the service resource has preference over ingress. This annotation is significant in case of `instance` targets, and is ignored for `ip` targets.

## Testing done
- Without this annotation, the default filter gets applied
- When ingress/service has this annotation, only the matching nodes get registered
- The annotation on service has preference over the one on the ingress
- On ingress/service edit, any nodes not matching the labels get de-registered from the target group 
- When a node resource gets modified, the target group gets reconciled as expected


M00nF1sh (approved):
/lgtm
/approve

Issue #1483, opened 2020-10-10
[Feature Request] Add the option to specify a load balancer name


moosahmed:
Similar to Issue #360 

What is the annotation to specify the alb name?

kishorj:
We do not support this feature yet. 
kind /feature

M00nF1sh:
we'll add support for this after our first v2.0.0 release. Should be fairly easy to support it.

moosahmed:
Thanks,

@M00nF1sh, no rush, just wondering when you are targeting v2.0.0 release?

lexore:
Hello! I just want to say that version 2.0.0. is out and we can continue discuss this feature :-)

M00nF1sh:
Hi, we'll add support to use existing LB and also this override LB name.

vdahmane:
Hello, do you have ETA for this feature?

LevonBecker:
How about now?  We're on v2.1.3... This feature is important to us for chargeback of stacks etc. Plus it makes easy to filter ALBs for a specific stack etc. 

Thanks in advance!

kishorj:
/reopen
need to support for ingress as well

pcolazurdo:
Will this feature allow to reuse existing NLBs? This is important to avoid having to recreate multiple Private Link configuration if for any reason the LoadBalancer has to be removed. Allowing the Controller to manage an existing NLB allows (which is connected to a Private Link) allow us to replace the service without having to ask the receiving end to recreate all the Private Endpoints setups

kishorj:
@pcolazurdo, this feature is for specifying a custom load balancer name via annotation.

#1573 is the one for reusing existing LB.

PR #1880, opened 2021-03-12
Add support for specifying the load balancer's name via annotation

rifelpet:
This adds a new "aws-load-balancer-name" annotation that can be used to specify the associated load balancer's name, overriding the default name format.

Some implementation notes:
* This doesn't perform any input validation on the annotation value to ensure it matches the allowed naming format. I'm happy to add this but didnt see validation on any other annotation value strings.
* Adding or updating the annotation on existing ingresses has no effect once a load balancer already exists.
* This doesn't modify target group names. I was wondering if another annotation would be accepted for specifying target group names. Given the one-to-many relationship of k8s Services to target groups, the annotation could be a stringMap of service ports to target group names, something like:
```yaml
service.beta.kubernetes.io/aws-target-group-name: "{\"80\": \"target-group-1\", \"443\": \"target-group-2\"}"
```

fixes https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1483

### DIFF
--- a/test/e2e/service/service.go
+++ b/test/e2e/service/service.go
@@ -48,6 +48,7 @@ type TargetGroupHC struct {
 }
 
 type LoadBalancerExpectation struct {
+	Name          *string
 	Type          string
 	Scheme        string
 	TargetType    string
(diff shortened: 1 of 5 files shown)


test/e2e/service/service.go
```diff
@@ -48,6 +48,7 @@ type TargetGroupHC struct {
 }
 
 type LoadBalancerExpectation struct {
+	Name          *string
```
kishorj, line 51: nit: for consistency, string instead of a pointer looks better to me

M00nF1sh:
thanks for contributing this feature,  really valuable 👍 

for target-group-name, i wonder the use case to explicitly control it since there will be multiple ones.(it's also complicated when IngressGroup is involved). i think we can pending that until there is a real use case for control over targetGroup names. (For loadbalancer, the use case exists since the name impacts DNS name)

kishorj (commented):
@rifelpet, thanks for the contribution. I appreciate your efforts on the e2e tests.

kishorj (approved):

PR #1881, opened 2021-03-12
Add a flag to set the default SSL Policy

laurovenancio:
The default SSL Policy is used when the SSL policy is not set by
annotations.


M00nF1sh:
/ok-to-test
Thanks for contributing this feature, it think it's a valuable addition.

would you help fix the commit to include your user info according to https://docs.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user

laurovenancio:
> would you help fix the commit to include your user info according to https://docs.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user

@M00nF1sh fixed. Thanks.

M00nF1sh (approved):
/lgtm
/approve

PR #1896, opened 2021-03-18
Update installation instructions

### LINKED ISSUE #1885: EKS IAM role as service does not work as expected, the node is also requiring permission
I have followed all the steps mentioned in the link below
[https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/deploy/installation/ ](url)

**Environment details**
1. EKS
2. Kubernetes v1.18
3. eksctl to create EKS cluster (v 0.40.0)
4. Used helm chart to deploy the aws load balancer controller, helm client version: "v3.5.1"

**Expected behavior** - No IAM permission should be needed to be applied to the EKS worker node role as outlined in the step 4 of the link [https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/deploy/installation/](url) 

**Issue description:**
When I created an Ingress resource, I saw error logs in my aws load balancer controller pod stating the following:

_"Failed deploy model due to UnauthorizedOperation: You are not authorized to perform this operation. Encoded authorization failure message: ........"_

**What I did to make it work:**
I looked at AWS CloudTrail and decoded the encoded log message from my aws load balancer controller pod and found out that the worker nodes of EKS needed some access rights, I did not check to see what exact rights were needed but instead applied the IAM policy in the step 2 of the link [https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/deploy/installation/](url)  to my worker nodes IAM role. As soon as I did that it started working and a ALB was provisioned. But this defeats the whole purpose of removing access from worker nodes and assigning them to the pods via IAM role as a service approach.
Is this not a bug?  Am I missing something

kishorj:
Fixes  #1885
Provide separate helm install commands with and without IRSA enabled.

### DIFF
--- a/docs/deploy/installation.md
+++ b/docs/deploy/installation.md
@@ -65,11 +65,19 @@ curl -o iam-policy.json https://raw.githubusercontent.com/kubernetes-sigs/aws-lo
     ```
     helm repo add eks https://aws.github.io/eks-charts
     ```
-    1. Install the TargetGroupBinding CRDs
+    1. Install the TargetGroupBinding CRDs if upgrading the chart via `helm upgrade`.
     ```
     kubectl apply -k "github.com/aws/eks-charts/stable/aws-load-balancer-controller//crds?ref=master"
     ```
-    1. Install the helm chart
+
+        !!!tip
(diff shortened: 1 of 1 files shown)


M00nF1sh (approved):
/approve
/lgtm

PR #1901, opened 2021-03-22
gRPC docs

### LINKED ISSUE #1706: add gRPC ingress example
Are there any available examples of how to enable a GRPC-only endpoint?

I'm trying to use v2.1.0 of the controller on EKS and the controller reports the following error:

```
{"level":"error","ts":1606949022.5301569,"logger":"controller","msg":"Reconciler error","controller":"ingress","name":"prod-buildgrid-frontend2","namespace":"prod","error":"InvalidParameter: 1 validation error(s) found.\n- minimum field value of 1, CreateTargetGroupInput.Port.\n"}
```

I can supply the annotations I'm using and the Ingress resource, but would rather first try and use a known-good example of how to use gRPC, and then see if the issue still persists.

(As for the error at issue, looking at the resource model output to the controller's log, the `Port` attribute for the `AWS::ElasticLoadBalancingV2::TargetGroup` resource is set to `0` instead of being set to the port on which the service is listening.)

brianannis:
Add documentation and example manifests for secure gRPC services via ALB. Fixes #1706.

### DIFF
--- a/docs/examples/grpc/grpcserver-deployment.yaml
+++ b/docs/examples/grpc/grpcserver-deployment.yaml
@@ -1,0 +1,28 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: grpcserver
+  namespace: grpcserver
+spec:
+  selector:
+    matchLabels:
+      app: grpcserver
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: grpcserver
+    spec:
+      containers:
+      - name: grpc-demo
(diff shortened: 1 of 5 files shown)


brianannis:
CLA bump

brianannis:
Apologies for the spam, needed to triage CLA requirement

brianannis:
@kishorj @M00nF1sh this is ready for review

kishorj:
@brianannis, thanks for the contribution. We will merge the changes shortly.

PR #1904, opened 2021-03-26
Highlight the need to tag subnets

isqrd:
While this info is part of the AWS eks userguid documentation, I think it is worth including the need to tag subnets with kubernetes.io/role/ tag here as well.

### DIFF
--- a/docs/guide/service/nlb_ip_mode.md
+++ b/docs/guide/service/nlb_ip_mode.md
@@ -34,20 +34,7 @@ service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: "*"
 ```
 
 ## Subnet tagging requirements
-You must tag the subnets you wish to attach your ELB to with the following tags. 
-  - For any subnet used by node group. 
-    - **Key:** *kubernetes.io/cluster/$CLUSTER_NAME*
-    - **Value:** (shared|owned)
-      - Shared -  Allow more than one cluster to use the subnet
-      - Owned -  Only allow one cluster to use the subnet. 
-
-  - For Private internal ELB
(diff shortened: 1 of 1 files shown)


docs/guide/service/nlb_ip_mode.md
```diff
@@ -33,5 +33,22 @@ to enable proxy protocol v2, apply the following annotation to your service:
 service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: "*"
 ```
 
+## Subnet tagging requirements
+You must tag the subnets you wish to attach your ELB to with the following tags.
```
kishorj, line 37: Subnets tags are useful only for auto-discovery.

docs/guide/service/nlb_ip_mode.md
```diff
@@ -33,5 +33,22 @@ to enable proxy protocol v2, apply the following annotation to your service:
 service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: "*"
 ```
 
+## Subnet tagging requirements
+You must tag the subnets you wish to attach your ELB to with the following tags. 
+  - For any subnet used by node group. 
+    - **Key:** *kubernetes.io/cluster/$CLUSTER_NAME*
```
kishorj, line 39: The cluster tag is optional. In case of ties (for example there are multiple subnets in the same AZ), auto-discovery will prefer the one with the cluster tag. This has been the behavior since EKS 1.19 for in-tree and v2.1.1 release of the lb controller (ref https://docs.aws.amazon.com/eks/latest/userguide/load-balancing.html)

kishorj:
Lets reference the file deploy/subnet_discovery.md instead.

PR #1912, opened 2021-04-02
Add IAM policy for US GovCloud

### LINKED ISSUE #1910: [v2] iam_policy for AWS GovCloud regions
Similar to #1557 when creating the IAM policy for the AWS load balancer controller with AWS GovCloud regions:

```
curl -o iam_policy.json  https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/main/docs/install/iam_policy.json
aws iam create-policy --policy-name ALBIngressControllerIAMPolicy --policy-document file://iam_policy.json
```
Results in the partition error:
```
An error occurred (MalformedPolicyDocument) when calling the CreatePolicy operation: Partition "aws" is not valid for resource "arn:aws:ec2:*:*:security-group/*".
```
The fix is to replace as with aws-us-gov in the arn stuff in this file. Would be nice to have this file available much like iam_policy_ch.json

[Reference for the arn names in GovCloud (US) regions](https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-arns.html)

1riggs:
Fixes #1910
Adds a new IAM policy file called iam_policy_usgov.json which has the arn region to work in the AWS US GovCloud region. 

[Reference for the arn names in GovCloud (US) regions](https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-arns.html)


1riggs:
I signed it

kishorj:
Thanks for the PR.
/lgtm

kishorj (approved):

M00nF1sh (changes requested):
lgtm.
would you mind change the filename to be `iam_policy_us-gov.json` to stay same naming format as well files :D

1riggs:
> lgtm.
> would you mind change the filename to be `iam_policy_us-gov.json` to stay same naming format as well files :D

Totally! Could've sworn I name it `iam_policy_usgov.json` initially but obviously, that isn't the case.

PR #1919, opened 2021-04-06
Adding a note about the preservation of ALB

nataizya-s:
It would be beneficial to have this also explicitly state that the ALB will not be preserved if using a version lower than v.1.1.3. In this event, based on my tests, a new load balancer is created and the old one remains. The ingress resource is updated to use the new ALB.


nataizya-s:
assign @kishorj

M00nF1sh:
/lgtm
/approve
thanks for making this update, appreciate it 👍

PR #1920, opened 2021-04-07
add support for local zone

M00nF1sh:
## This PR add support for local zone
1.  The controller will do an additional EC2:DescribeAvailabilityZone API call for each zone to discover zone type. (the result is cached indefinitely, so at most one API call will be called for each zone in entire life of this controller)

Additional Notes:
1. waveLength zone is currently not supported by ELB team, however, this change accounted for waveLengthZone as well, but ELBv2 API will reject the LB creation.


kishorj:
upgrade would require modifying existing IAM permissions. Lets make sure we capture this in our documentation.

/lgtm

PR #1926, opened 2021-04-14
Update annotations.md

aws-sidhartha:
service.beta.kubernetes.io/aws-load-balancer-subnets is supported only from EKS 1.19 but not on EKS 1.18


kishorj:
@aws-sidhartha, thanks for your contribution.

This controller supports service annotation `service.beta.kubernetes.io/aws-load-balancer-subnets` for NLB with IP targets starting controller version `v2.1.2`. Since this controller is external, it is independent of the k8s version as long as the prerequisites are met.

The kubernetes in-tree controller supports service annotation `service.beta.kubernetes.io/aws-load-balancer-subnets` for NLB with instance targets starting EKS 1.19 release.

Since this controller is not tied to a specific k8s version, the proposed doc changes would create more confusion.

aws-sidhartha:
@kishorj  as discussed, If customers are reading through the prerequisites section then it might not be clear for them to understand that AWS load balancer is needed only for IP targets hence we may need to update the documentation. Please consider making following changes.

Thank you!

PR #1933, opened 2021-04-17
Use YAML separator in docs

bnutt:
Added yaml separator to all examples and install manifests


kishorj:
@bnutt, thanks for the changes. Any reason to explicitly add the separators?

PR #1887, opened 2021-03-17
Resource tagging for Listener and Rules

kishorj:
Fixes: #1660
Add support for tagging load balancer listener and listener rules.

### Tests
- upgraded controller adds tags to the listener and rules for existing ingress/services resources. This requires additional iam permissions
- for new service or ingress resources, controller adds the default tags to the listener and rules as well
- For ALB, additional tags can be added via `alb.ingress.kubernetes.io/tags` annotation
- For NLB, additional tags can be added via `service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags` annotation
- Additional tags can be successfully removed as well
- Describe tag on ALB listener resource
```
aws elbv2 describe-tags --resource-arn arn:aws:elasticloadbalancing:us-west-2:<AWS_ACCOUNT_ID>:listener/app/k8s-grouper-xxxxxxxxxxxx/yyyyyyyyyyyyy/zzzzzzzzzzzz
{
    "TagDescriptions": [
        {
            "ResourceArn": "arn:aws:elasticloadbalancing:us-west-2:<AWS_ACCOUNT_ID>:listener/app/k8s-grouper-xxxxxxxxxxxx/yyyyyyyyyyyyy/zzzzzzzzzzzz",
            "Tags": [
                {
                    "Key": "ingress.k8s.aws/resource",
                    "Value": "80"
                },
                {
                    "Key": "ingress.k8s.aws/stack",
                    "Value": "grouper"
                },
                {
                    "Key": "elbv2.k8s.aws/cluster",
                    "Value": "eighteen"
                }
            ]
        }
    ]
}

```
- Describe tag on ALB listener rule resource
```
aws elbv2 describe-tags --resource-arns arn:aws:elasticloadbalancing:us-west-2:<AWS_ACCOUNT_ID>:listener-rule/app/k8s-grouper-xxxxxxxxxxxx/yyyyyyyyyyyyy/wwwwwwwwwww
{
    "TagDescriptions": [
        {
            "ResourceArn": "arn:aws:elasticloadbalancing:us-west-2:019453415603:listener-rule/app/k8s-grouper-xxxxxxxxxxxx/yyyyyyyyyyyyy/wwwwwwwwwww",
            "Tags": [
                {
                    "Key": "ingress.k8s.aws/resource",
                    "Value": "80:1"
                },
                {
                    "Key": "ingress.k8s.aws/stack",
                    "Value": "grouper"
                },
                {
                    "Key": "elbv2.k8s.aws/cluster",
[...]

PR #1944, opened 2021-04-21
Fix new tagging ARNs in aws-cn IAM policy

rifelpet:
Followup to https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/1887

This policy JSON is for AWS China which requires using the aws-cn partition (as evidenced by the other ARNs in the policy document)


kishorj:
Thanks for the changes.

PR #1947, opened 2021-04-22
add aws-region / aws-vpc-id args in deployment

alphaxo:
when nodegroup IMDS disabled, args `--aws-region` and `--aws-vpc-id` is required

### DIFF
--- a/docs/install/v2_1_3_full.yaml
+++ b/docs/install/v2_1_3_full.yaml
@@ -485,6 +485,8 @@ spec:
         - args:
             - --cluster-name=your-cluster-name
             - --ingress-class=alb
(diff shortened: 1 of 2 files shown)


docs/install/v2_1_3_full.yaml
```diff
@@ -485,6 +485,8 @@ spec:
         - args:
             - --cluster-name=your-cluster-name
             - --ingress-class=alb
+            - --aws-region=eks-vpc-region
```
kishorj, line 488: These are auto-discovered if IMDS is enabled. We didn't put in the sample manifest so as to avoid confusion. These should be added manually during installation.

IMO it is better if we mention this in the intallation guide.

Issue #1859, opened 2021-03-05
[Feature Request] PrivateLink Support


rifelpet:
We have a need to create PrivateLink Endpoint Services with our NLBs. A Service resource would correspond to one NLB as well as one Endpoint Service configured to use that NLB.

I imagine the annotation design would be something like:

```yaml
service.beta.kubernetes.io/aws-load-balancer-endpoint-service-acceptance-required: 'true'
service.beta.kubernetes.io/aws-load-balancer-endpoint-service-allowed-principals: '["arn:aws:iam::123456789012:root"]'
service.beta.kubernetes.io/aws-load-balancer-endpoint-service-private-dns-name: 'foobar.example.com'
```

I don't have a good idea for how to conditionally add the DNS verification record... typically external-dns would handle that but given that both the name and value are computed, it doesn't align as well with the external-dns pattern of reading the name from a spec or annotation and the value from a resource's status or annotation.

A stretch goal would be to support creating Endpoints in a specified set of VPCs but given that (in our case) many of the VPCs would be in different AWS accounts and aws-load-balancer-controller doesn't support using multiple credentials for multiple accounts, I understand if we don't want to incorporate that complexity directly into aws-load-balancer-controller.

Would a PR for supporting Endpoint Services be accepted in the load balancer controller? Or are there any existing plans to add something similar?

fandujar:
@rifelpet I need something similar. I got some issues when deleting a load balancer associated with a VPC Endpoint Service.

Maybe it would worth adding an annotation to your request to reuse a VPC Endpoint Service:
`service.beta.kubernetes.io/aws-load-balancer-endpoint-service-id: 'vpce-svc-xxxxxx'`

That way you can manage your VPC Endpoint Service without add complexity to aws-load-balancer-controller and still manage this association lifecycle of a load balancer with private link.

M00nF1sh:
@fandujar @rifelpet 

We don't have plans internally for this. PRs are always welcome, however given this is a little complicated, i think we need a placeholder PR finalize the interfaces first before any code implementations, which we can review internally along with community review.

Personally i'd favor the `aws-load-balancer-endpoint-service-id` options if we can create a endpointService without NLB. (however seems an NLB is required when create a endpointService. I'm not familiar with privateLink so i might be wrong).

[...]

PR #1948, opened 2021-04-22
Initial VPC Endpoint Service (Private Link) Interface

rifelpet:
This adds a stubbed interface for VPC Endpoint Service support. This was mostly copied from the security group implementation since I view them as similar auxiliary resources to the main loadbalancer/listener/targetgroup resource types.

A few discussion topics and notes for the future implementation:
* Should the controller detach other NLBs or GWLBs besides the Service's NLB?
* I haven't checked whether Endpoint Services are supported in all regions and partitions, but I anticipate the new addon CLI flag should be sufficient if not.
* EndpointServices can't be deleted if they're currently used by an interface Endpoint in another VPC, should this block the finalizer on k8s Service deletion?
* We'll want to decide on how to expose the DNS verification information (record name, type, and value). If we choose to use annotations, I believe this would be the first case of the controller writing annotations back to the resource so we would need to think through the implications of doing so. If we can define the records in a way compatible with external-dns that would be ideal.
* I plan on including the same tags as on the NLB itself and using them in the tagging filter of `DescribeVpcEndpointServiceConfigurations` to find existing endpoint services.
* The [AWS docs](https://docs.aws.amazon.com/vpc/latest/privatelink/endpoint-service-overview.html) recommend exposing Endpoint Services in all AZs of the region which requires the NLB be configured for a subnet in every AZ, this is probably worth mentioning in the docs.

ref: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1859

### DIFF
--- a/pkg/annotations/constants.go
+++ b/pkg/annotations/constants.go
@@ -81,4 +81,8 @@ const (
 	SvcLBSuffixTargetNodeLabels              = "aws-load-balancer-target-node-labels"
 	SvcLBSuffixLoadBalancerAttributes        = "aws-load-balancer-attributes"
 	SvcLBSuffixManageSGRules                 = "aws-load-balancer-manage-backend-security-group-rules"
+
+	SvcLBSuffixEndpointServiceAcceptanceRequired = "aws-load-balancer-endpoint-service-acceptance-required"
+	SvcLBSuffixEndpointServiceAllowedPrincipals  = "aws-load-balancer-endpoint-service-allowed-principals"
+	SvcLBSuffixEndpointServicePrivateDNSName     = "aws-load-balancer-endpoint-service-private-dns-name"
 )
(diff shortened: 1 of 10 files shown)


hintofbasil:
Hi @rifelpet,

My work is very interested in getting this functionality added in.  Is there anything we can do to help you push this feature over the line?

rifelpet:
I'm likely not going to continue working on this but as it stands the maintainers and EKS team were going to review the interface proposal and then someone can begin the implementation work. It looks like it has been bumped out to 2.5.0 so I'm not sure how soon the review will occur.

hintofbasil:
Thanks for the context.  We will bring this up with AWS ourselves.
If we decide that we need this feature before AWS can provide it do you mind if we build on top of your existing work?

PR #1958, opened 2021-04-29
Use objectSelector for pod mutator webhook

### LINKED ISSUE #1789: Disable pod readiness gate for a specific ingress?
Hi,

Is it possible to have the webhook skip injecting a pod with the readiness gate configuration, according to an annotation perhaps?

Thanks

kishorj:
Fixes #1789

Use objectSelector in the pod mutator configuration to not match aws-load-balancer-controller pods. This removes the circular dependency between the controller pods and the webhook configuration.

The `objectSelector` in the webhook configuration can also be used to select/unselect specific pods for the readiness gate injection. Provide some examples in the pod readiness gate documentation.

### Tests
- Added elbv2.k8s.aws/pod-readiness-gate-inject=enabled label to the kube-system namespace, scaled aws load balancer controller deployment to 0 replicas, and when the controller pod was no longer running, scaled the deployment back to 1 replicas. The controller pod was came up right away with the modified webhook configuration.
- Verified that readiness gate config got injected as expected for pods running in a labelled namespace.

### DIFF
--- a/config/webhook/pod_mutator_patch.yaml
+++ b/config/webhook/pod_mutator_patch.yaml
@@ -10,3 +10,9 @@ webhooks:
           operator: In
           values:
             - enabled
+    objectSelector:
+      matchExpressions:
+        - key: app.kubernetes.io/name
+          operator: NotIn
+          values:
+            - aws-load-balancer-controller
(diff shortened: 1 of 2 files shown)


M00nF1sh (approved):
/lgtm
/approve

PR #1964, opened 2021-04-30
Provision internal NLB by default

kishorj:
In case the load balancer scheme annotation is not specified explicitly, provision internal NLB. This is to follow secure by default guidelines to limit the access to EC2 nodes unless the user wants to allow it explicitly. The default value in the previous release was internet-facing, so for backwards compatibility the scheme of existing load balancers will not be modified. To summarize 

- If load balancer scheme is not explicitly specified
   - on service create, provision internal NLB
   - on service update, continue to use the existing scheme specified in the NLB
- If scheme is explicitly mentioned, follow the annotation

As a side effect, if a previously specified scheme annotation on a service resource gets deleted, controller will preserve the previous scheme.

This change also adds new annotation to specify the load balancer scheme
```
service.beta.kubernetes.io/aws-load-balancer-scheme
```  
Values can be `internal` or `internet-facing`.  Default is `internal`.

The prior annotation for this purpose will be deprecated as of this release, and support will be removed in future.
```
service.beta.kubernetes.io/aws-load-balancer-internal
```
Note: In case of conflicts, `aws-load-balancer-scheme` will get higher precedence.

### Testing
- Internal NLB gets provisioned by default, VPC CIDRs get added to SG rule
- Existing NLB scheme does not get modified on controller update
- If scheme is explicitly mentioned, controller honors the annotation
- When scheme annotation gets deleted, NLB scheme does not get modified
- When both  annotations`service.beta.kubernetes.io/aws-load-balancer-scheme` and `service.beta.kubernetes.io/aws-load-balancer-internal` are specified, `service.beta.kubernetes.io/aws-load-balancer-scheme` gets precedence.


pkg/service/model_build_target_group.go
```diff
@@ -362,7 +362,14 @@ func (t *defaultModelBuildTask) buildTargetGroupBindingSpec(ctx context.Context,
 	if targetType == elbv2api.TargetTypeInstance {
 		targetPort = intstr.FromInt(int(port.NodePort))
 	}
-	tgbNetworking := t.buildTargetGroupBindingNetworking(ctx, targetPort, preserveClientIP, *hc.Port, port.Protocol)
+	defaultSourceRanges := []string{"0.0.0.0/0"}
```
M00nF1sh, line 365: nit: since defaultSourceRanges is only used in buildTargetGroupBindingNetworking, 
i'd suggest pass `scheme` into buildTargetGroupBindingNetworking instead, and construct defaultSourceRanges within there. either inline or have a function for build default source ranges.

M00nF1sh (approved):
/approve
/lgtm

Issue #1963, opened 2021-04-29
[feature request] IP from Pod Labels


santosh-kodali:
There are certain scenarios in which the IP we forward traffic to will not be the IP of the pod. 

Users might have additional ENIs configured for the pod and would need the load balancer to pick those IPs rather than the default Pod IP. If these IPs are stored on the pod's labels, could we have an annotation that would allow picking those IPs from the pod label instead of the default Pod IP?

PR #1968, opened 2021-05-03
support for annotation aws-load-balancer-ip-target-label

santosh-kodali:
#1963 

This PR will allow for specification of Target IP from pod label with label specified from `service.beta.kubernetes.io/aws-load-balancer-ip-target-label`

Please advise on any modifications or additions.

### DIFF
--- a/pkg/annotations/constants.go
+++ b/pkg/annotations/constants.go
@@ -77,4 +77,5 @@ const (
 	SvcLBSuffixSubnets                       = "aws-load-balancer-subnets"
 	SvcLBSuffixALPNPolicy                    = "aws-load-balancer-alpn-policy"
 	SvcLBSuffixTargetNodeLabels              = "aws-load-balancer-target-node-labels"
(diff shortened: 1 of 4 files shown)


M00nF1sh:
@santosh-kodali 
Thanks for making this contribution.
Would you explain the background for this PR? why the pod labels is needed. since ideally all these pods behind service should be registered

santosh-kodali:
The reason behind this PR is that default service registration is based on podIP, however, we might want certain LBs to point towards a different IP attached to the node. I suppose we could have a custom endpoint reconciler to update the endpoints according to custom needs. I will close this PR if this functionality is not wanted.

M00nF1sh:
@santosh-kodali 
Yes,  i think we don't want to introduce such feature(to further select pods behind service). 

For such use case, it's better to do what you suggested(have a separate controller to generate endpoints)

santosh-kodali:
feature not required

PR #1970, opened 2021-05-03
add support for external-managed-tags & prefer defaultTags

M00nF1sh:
This PR contains two changes:
1. add support for external-managed-tags. Tag keys specified via "--external-managed-tags" will be ignored during tag reconciliation.
    1. If a tag key specified in --external-managed-tags controller-level flag collisions with tags below(used for track the mapping between AWS resources and K8s resources), the controller should error out during start up and exit.
        * elbv2.k8s.aws/cluster
        * ingress.k8s.aws/stack
        * ingress.k8s.aws/resource
        * service.k8s.aws/stack
        * service.k8s.aws/resource
    2. If a tag key specified in both --external-managed-tags controller-level flag and --default-tags controller-level flag, the controller should error out during start up and exit.
    3. If a tag key specified in --external-managed-tags controller-level flag collisions with tags specified via annotation on K8s resources, the controller should refuse to reconcile for that K8s resource and emit error events/logs.

1. [**Behavior change**] prefer tag value from defaultTags if there are conflicts of tags specified via "--default-tags" and via annotation on Ingress/Service. (the default Tags is specified by administrator, which should be enforced)


pkg/config/controller_config.go
```diff
@@ -76,5 +83,57 @@ func (cfg *ControllerConfig) Validate() error {
...
+	for tagKey := range cfg.DefaultTags {
+		if trackingTagKeys.Has(tagKey) {
+			return errors.Errorf("tag key %v cannot be specified in %v flag", tagKey, flagDefaultTags)
+		}
+	}
+	return nil
+}
+
+func (cfg *ControllerConfig) validateExternalManagedTagsCollisionWithTrackingTags() error {
+	trackingTagKeys := sets.NewString(
```
kishorj, line 116: better to have a single source of truth for this.

pkg/service/model_build_load_balancer.go
```diff
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"regexp"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/algorithm"
```
kishorj, line 10: nit: lets group this with the local imports

pkg/config/controller_config.go
M00nF1sh: make sense, let me do the change

pkg/service/model_build_load_balancer.go
M00nF1sh: 👍

PR #1982, opened 2021-05-06
update site_name

mikutas:
trivial change of site_name
because I feel it's a bit confusing


kishorj:
@mikutas thanks for the contribution.
/approve

PR #1983, opened 2021-05-06
Add flags to set server key and certificate

### LINKED ISSUE #1942: Expose flag to configure certs for webhook
Problem:
AWS load balancer controller use [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) to build the webhook server. The ALB controller will crash, if it failed to open ca.crt, tls.key and tls.crt from /tmp/k8s-webhook-server/serving-certs.

Solution:
We should expose additional flags in [Controller command line flags
](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/deploy/configurations.md#controller-command-line-flags) to configure CertDir, CertName, KeyName and ClientCAName for [ServerOptions](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/webhook#ServerOptions).

lxlxok:
This PR resolve #1942

### DIFF
--- a/pkg/config/runtime_config.go
+++ b/pkg/config/runtime_config.go
@@ -20,6 +20,9 @@ const (
 	flagWatchNamespace          = "watch-namespace"
 	flagSyncPeriod              = "sync-period"
 	flagKubeconfig              = "kubeconfig"
+	flagWebhookCertDir          = "webhook-cert-dir"
+	flagWebhookCertName         = "webhook-cert-file"
+	flagWebhookKeyName          = "webhook-key-file"
 
 	defaultKubeconfig              = ""
 	defaultLeaderElectionID        = "aws-load-balancer-controller-leader"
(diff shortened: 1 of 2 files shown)


pkg/config/runtime_config.go
```diff
@@ -20,6 +20,9 @@ const (
 	flagWatchNamespace          = "watch-namespace"
 	flagSyncPeriod              = "sync-period"
 	flagKubeconfig              = "kubeconfig"
+	flagCertDir                 = "cert-dir"
```
kishorj, line 23: lets use webhook-cert-dir, webhook-cert-file, webhook-key-file instead, that will make it clear that these configurations are intended for the webhook.

pkg/config/runtime_config.go
```diff
@@ -49,6 +55,9 @@ type RuntimeConfig struct {
 	LeaderElectionNamespace string
 	WatchNamespace          string
 	SyncPeriod              time.Duration
+	CertDir                 string
```
kishorj, line 58: WebhookCertDir ... for consistency

lxlxok:
Hi @kishorj, thanks for your review, all the comments have been addressed.  PTAL

Issue #1978, opened 2021-05-05
Work in non-EKS clusters?


kolorful:
I'm unable to find any resource on setting up the AWS Load Balancer Controller in non-EKS Kubernetes clusters. Has anyone tried it with success, or does it only work in EKS?

kishorj:
@kolorful, please refer to the linked PR. This will be part of the live docs once merged.

kolorful:
@kishorj I saw the [annotations for Service](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.1/guide/service/annotations/) in the aws-lb-controller is exactly the same as the ones in [legacy-cloud-providers](https://github.com/kubernetes/kubernetes/blob/v1.18.16/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L105). I didn't find any additional configs to toggle between the two, would you shed some light on how the controller won't conflict with the legacy-aws-provider? Thanks!

kishorj:
@kolorful, some of the newer annotations aren't supported by the legacy-cloud-provider. For example those related to proxy protocol v2, private ipv4 addresses, target node labels, custom lb name.

The annotation `service.beta.kubernetes.io/aws-load-balancer-type` is used to determine which controller reconciles the service. If annotation value is `nlb-ip` or `external`, legacy cloud provider ignores the service resource so that the AWS LB controller can take over. For all other values of the annotation, the legacy cloud provider will handle the service. Note that this annotation should be specified during service creation and not edited later.

kishorj:
k8s 1.20 and later have the cloud provider fix. 
I've backported the fix to 1.18 and 1.19 versions, here are the requirements 
v1.18.18 and later for 1.18 release
v1.19.10 and later for 1.19 release

kolorful:
Thank you for the insight, that's very interesting. I'm running Kubernetes v1.18 and I would like to use `service.beta.kubernetes.io/aws-load-balancer-subnets`, but it's not available until v1.21. I was hoping to install the aws-load-balancer-controller which [supports this feautre](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.1/guide/service/annotations/), so I don't have to upgrade the cluster.

Base on my understanding, this should work? I'd imagine the legacy-cloud provider will ignore and let the controller take over?

kishorj:
The `aws-load-balancer-subnets` support will not be backported to 1.18. 

[...]

PR #1988, opened 2021-05-06
Add docs around how does the AWS LB controller work with legacy cloud provider

kolorful:
Add docs around how does the AWS LB controller work with legacy cloud provider, which was mentioned in https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1978#issuecomment-833666471


kishorj:
@kolorful, thanks for the contribution.

/lgtm
/approve

PR #2006, opened 2021-05-13
filter redundant health check SG rules

kishorj:
Filter out redundant health check security group rules when the following conditions are met
- health check and traffic ports are the same
- preserve client IP is enabled
- custom source ranges is not configured on the service resource, or 0.0.0.0/0 is allowed

In case of preserve client IP disabled, separate set of health check rules do not get added if traffic and health check ports are the same.

### DIFF
--- a/pkg/service/model_build_target_group.go
+++ b/pkg/service/model_build_target_group.go
@@ -5,20 +5,21 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	"regexp"
 	elbv2api "sigs.k8s.io/aws-load-balancer-controller/apis/elbv2/v1beta1"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/annotations"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/k8s"
 	elbv2model "sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2"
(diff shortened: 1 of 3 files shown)


pkg/service/model_build_target_group.go
```diff
@@ -483,3 +475,35 @@ func (t *defaultModelBuildTask) buildTargetGroupBindingNodeSelector(_ context.Co
 		MatchLabels: targetNodeLabels,
 	}, nil
 }
+
+func (t *defaultModelBuildTask) buildHealthCheckNetworkingIngressRules(trafficSource, hcSource []elbv2model.NetworkingPeer, tgPort, hcPort intstr.IntOrString,
+	tgProtocol corev1.Protocol, preserveClientIP, customSoureRanges bool) []elbv2model.NetworkingIngressRule {
+	if tgProtocol != corev1.ProtocolUDP &&
+		(hcPort.String() == healthCheckPortTrafficPort || hcPort.IntValue() == tgPort.IntValue()) {
```
M00nF1sh, line 482: Nit:
`hcPort.IntValue() == tgPort.IntValue` this check is hacky.
If somehow in the future, both hcPort and tgPort is a string but not equal, this will evaluate to true.  

ideally we should check hcPort's type.

pkg/service/model_build_target_group.go
```diff
@@ -483,3 +475,35 @@ func (t *defaultModelBuildTask) buildTargetGroupBindingNodeSelector(_ context.Co
...
+		if !preserveClientIP {
+			return []elbv2model.NetworkingIngressRule{}
+		}
+		if !customSoureRanges {
```
M00nF1sh, line 486: LGTM for this version.
however, this becomes really complicated. we should refactor this logic here by some thing like compare CIDRs in the future.

M00nF1sh (approved):
lgtm

pkg/service/model_build_target_group.go
kishorj: currently health check port is limited to int values. Will revisit this logic when we support named ports as well.

pkg/service/model_build_target_group.go
kishorj: once we have SG support for NLB, this wouldn't be a matter of concern.

PR #2008, opened 2021-05-14
resolve VPC CIDRs for UDP protocol

kishorj:
Restrict SG rules for UDP protocol in case of an internal load balancer

Tests:
- Created UDP service with default internal scheme, verified the SG rules were restricted to the VPC CIDR ranges
- Created UDP service with internet-facing scheme, verified SG rules were as per the loadBalancerSourceRanges configuration
- Verified TCP health check rules were added for UDP service


M00nF1sh (approved):
/lgtm
/approve

PR #2012, opened 2021-05-16
Don't consider nodes with exclude-from-external-load-balancer eligible worker nodes in e2e

olemarkus:
The NLB instance target e2e tests in particular considers all nodes eligible targets. This on kOps, which labels control plane nodes with the exclude-from-external-load-balancer label.


kishorj (approved):
Thanks for the contribution.

PR #2029, opened 2021-05-25
Add max backoff flag for TargetGroupBinding reconciler

### LINKED ISSUE #1974: Missing reconcile on targetgroupbinding causing delayed registration
We are seeing missed reconciles on targetgroupbinding when endpoint object is updated sometimes. This happened on 2.1.2 but I don't think there are changes related to this on 2.1.3
What we observed is that sometimes:
1) Pod gets removed from endpoint
2) Reconcile removes the target
3) Within 1 minute of above reconcile, new pod gets added to target to notready subset in the endpoint
4) 4 minutes from previous reconcile pod gets to containers ready
5) 10 minutes from previous reconcile a new reconcile on targetgroupbinding happens (no endpoint updates have happened)
6) Register the pod as target
7) This delays rollout of deployments

I saw the comment 
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/da8951f80521651e0a1ffe1361c011d6baad7706/pkg/backend/endpoint_resolver.go#L18

I was trying to understand whether the above is happening because the endpoint object changes are not observed or its because endpoint doesn't meet your criterea as the comment says?

eytan-avisror:
Allows control over max exponential backoff for reconciler, while keeping existing default values

Signed-off-by: Eytan Avisror <eytan_avisror@intuit.com>

Fixes #1974 

This retains the defaults as in default rate limiter:
https://github.com/kubernetes/client-go/blob/39cb8cd8c11db40f5ec58c16a31b91c535946e7e/util/workqueue/default_rate_limiters.go#L85-L87

- [x] Reproduce/Test

### DIFF
--- a/controllers/elbv2/targetgroupbinding_controller.go
+++ b/controllers/elbv2/targetgroupbinding_controller.go
@@ -164,7 +164,7 @@ func (r *targetGroupBindingReconciler) SetupWithManager(ctx context.Context, mgr
 		Watches(&source.Kind{Type: &corev1.Node{}}, nodeEventsHandler).
 		WithOptions(controller.Options{
 			MaxConcurrentReconciles: r.maxConcurrentReconciles,
-			RateLimiter:             workqueue.NewItemExponentialFailureRateLimiter(time.Millisecond, r.maxExponentialBackoffDelay)}).
+			RateLimiter:             workqueue.NewItemExponentialFailureRateLimiter(5*time.Millisecond, r.maxExponentialBackoffDelay)}).
 		Complete(r)
 }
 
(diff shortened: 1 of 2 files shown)


eytan-avisror:
I verified this fixes the problem and that backoff gets capped correctly according to the flag:

e.g. used 60s max here
```
{"level":"debug","ts":1621969829.9372506,"logger":"controllers.targetGroupBinding","msg":"requeue","reason":"monitor potential ready endpoints"}
{"level":"debug","ts":1621969840.224212,"logger":"controllers.targetGroupBinding","msg":"requeue","reason":"monitor potential ready endpoints"}
{"level":"debug","ts":1621969860.769745,"logger":"controllers.targetGroupBinding","msg":"requeue","reason":"monitor potential ready endpoints"}
{"level":"debug","ts":1621969901.7841525,"logger":"controllers.targetGroupBinding","msg":"requeue","reason":"monitor potential ready endpoints"}
{"level":"debug","ts":1621969961.8460197,"logger":"controllers.targetGroupBinding","msg":"requeue","reason":"monitor potential ready endpoints"}
{"level":"debug","ts":1621970021.8983068,"logger":"controllers.targetGroupBinding","msg":"requeue","reason":"monitor potential ready endpoints"}
```

Target got registered during the 1m window from pod ready.

@M00nF1sh please review

M00nF1sh (approved):
thanks for making this change 😄 
/lgtm
/approve

PR #2033, opened 2021-05-25
Update ssl_redirect.md

csahli:
Syntax to implement ssl redirection changed apparently with v2.2. Current documentation seems to refer to v2.1 syntax for redirect.
Example should include the documented annotation: alb.ingress.kubernetes.io/ssl-redirect: '443'


kishorj:
@csahli, thanks for your contribution. The master branch is no longer active, we've kept it for historical reference only. Please update the main branch instead. Here is the corresponding file in the main branch -  https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/guide/tasks/ssl_redirect.md

Issue #1997, opened 2021-05-10
TargetGroupBinding stuck in FailedCleanup loop (Failed cleanup due to ValidationError) when created without a targetGroupARN


wjrogers:
I accidentally created a TargetGroupBinding without a `targetGroupARN`. I misspelled the key, and I explicitly included `targetType: ip`, so the resource was successfully created missing the ARN:

```yaml
spec:
  serviceRef:
    name: beanstalkd
    port: 11300
  targetGroupARN: ""
  targetType: ip
```

I subsequently tried to delete it, and now the resource seems permanently stuck in this event loop:

> Warning  FailedCleanup  80s (x8 over 6m45s)  targetGroupBinding  (combined from similar events): Failed cleanup due to ValidationError: A target group ARN must be specified
>           status code: 400, request id: 4c885c04-894f-4f40-930b-0fb62e96ede1

Workaround: remove the finalizer from the broken TargetGroupBinding (e.g. `kubectl edit`) so it can be deleted.

M00nF1sh:
we can add the validation in the CRD via kubebuilder mark to verify it's arn(via regex) or simply verify the length.

shuheiktgw:
Let me work on this issue!

/assign @shuheiktgw

PR #2035, opened 2021-05-26
Add MinLength validation to TargetGroupARN

shuheiktgw:
Close https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1997

I added MinLength validation to TargetGroupARN. Thank you for your review! 👍

### DIFF
--- a/apis/elbv2/v1beta1/targetgroupbinding_types.go
+++ b/apis/elbv2/v1beta1/targetgroupbinding_types.go
@@ -115,7 +115,7 @@ type TargetGroupBindingNetworking struct {
 // TargetGroupBindingSpec defines the desired state of TargetGroupBinding
 type TargetGroupBindingSpec struct {
 	// targetGroupARN is the Amazon Resource Name (ARN) for the TargetGroup.
-	// +kubebuilder:validation:MinLength:=1
+	// +kubebuilder:validation:MinLength=1
 	TargetGroupARN string `json:"targetGroupARN"`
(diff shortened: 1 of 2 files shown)


apis/elbv2/v1beta1/targetgroupbinding_types.go
```diff
@@ -115,6 +115,7 @@ type TargetGroupBindingNetworking struct {
 // TargetGroupBindingSpec defines the desired state of TargetGroupBinding
 type TargetGroupBindingSpec struct {
 	// targetGroupARN is the Amazon Resource Name (ARN) for the TargetGroup.
+	// +kubebuilder:validation:MinLength:=1
```
M00nF1sh, line 118: i believe it used to be `kubebuilder:validation:MinLength=1` did `+kubebuilder:validation:MinLength:=1` works as well?

M00nF1sh:
thanks for contributing this :D

apis/elbv2/v1beta1/targetgroupbinding_types.go
shuheiktgw: It seems both `:=` and `=` output the same manifests and https://book.kubebuilder.io/reference/markers/crd-validation.html uses `:=` instead of `=`. I updated it to use `=` instead!

M00nF1sh:
/lgtm
/approve
thanks a lot 😄

PR #2043, opened 2021-05-31
Add missing tests for throttler

shuheiktgw:
I've added some missing tests for throttler. Thank you for your review!

### DIFF
--- a/pkg/aws/throttle/throttler_test.go
+++ b/pkg/aws/throttle/throttler_test.go
@@ -2,15 +2,98 @@ package throttle
 
 import (
 	"context"
+	"github.com/aws/aws-sdk-go/aws/client/metadata"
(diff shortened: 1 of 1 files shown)


pkg/aws/throttle/throttler_test.go
```diff
@@ -2,16 +2,99 @@ package throttle
...
 	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
 )
 
+func Test_NewThrottler(t *testing.T) {
+	config := ServiceOperationsThrottleConfig{
+		value: map[string][]throttleConfig{
+			appmesh.ServiceID: {
```
fawadkhaliq, line 22: Thanks for adding the tests. 

nit: would be nice to use services/service names associated with this project. AppMesh looks a bit odd here given the controller has nothing to do with it.

Issue #2051, opened 2021-06-04
Feature Request: NLB deletion protection


ttony:
Hi,

i would like to ask if the feature can be implemented?

kishorj:
@ttony, thanks for bringing it up. 

We could add an annotation to support this 
```
service.beta.kubernetes.io/aws-load-balancer-attributes: deletion_protection.enabled=true
```

/kind feature

JacekDuszenko:
Hey, I'd like to take a stab at implementing this :) 
/assign @JacekDuszenko

JacekDuszenko:
On the ALB attributes annotation parsing we use a generic approach of parsing the whole config map provided by the `alb.ingress.kubernetes.io/load-balancer-attributes` annotation (https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/ingress/model_build_load_balancer.go#L280) whereas on the NLB we are essentially hardcoding supported lb attributes (https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/service/model_build_load_balancer.go#L234). 

I don't think that going with `service.beta.kubernetes.io/aws-load-balancer-attributes: deletion_protection.enabled=true
` annotation is a good idea here because this PR would have to include parsing all of the attributes in the form of a config map as well. I think this is a bigger change and should be done in a separate PR. This change would also involve changing the code in ingress to allow for code reuse.

What I propose instead and will implement in my PR is to use a different annotation specific to  `deletion_protection.enabled=true` attribute and implement it on a similar basis as the rest in (https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/service/model_build_load_balancer.go#L234).

I suggest that we create a boolean annotation `service.beta.kubernetes.io/aws-load-balancer-deletion-protection-enabled`

What do you think about this?

kishorj:
@JacekDuszenko, I'm inclined more towards specifying load balancer attributes annotation since it will be consistent with the ALB. Here is what I recommend - 
1) Parse `service.beta.kubernetes.io/aws-load-balancer-attributes` into a string map
2) If there are specific annotations defined for access logs, cross zone load balancing, override the configmap from step 1.
3) Do not use a default values like we do currently. If we do this, the controller will ignore the attributes not specified via an annotation. The new behavior will be consistent with ALB.

Also deprecate the existing annotations to configure individual lb attributes, and remove the support in future releases.

[...]

PR #2057, opened 2021-06-08
add NLB deletion protection annotation

JacekDuszenko:
fixes: #2051
Config map annotation `service.beta.kubernetes.io/aws-load-balancer-attributes` was added. It allows to specify all attributes annotations, which will be overridden by annotation-specific attributes for now until we delete annotation-specific attributes. Now, attribute-specific annotations have been deprecated in docs

### DIFF
--- a/docs/guide/service/annotations.md
+++ b/docs/guide/service/annotations.md
@@ -213,7 +213,7 @@ Load balancer access can be controllerd via following annotations:
         ```
         service.beta.kubernetes.io/aws-load-balancer-internal: "true"
         ```
-  
+
 ## Legacy Cloud Provider
 The AWS Load Balancer Controller manages Kubernetes Services in a compatible way with the legacy aws cloud provider. The annotation `service.beta.kubernetes.io/aws-load-balancer-type` is used to determine which controller reconciles the service. If the annotation value is `nlb-ip` or `external`, legacy cloud provider ignores the service resource (provided it has the correct patch) so that the AWS Load Balancer controller can take over. For all other values of the annotation, the legacy cloud provider will handle the service. Note that this annotation should be specified during service creation and not edited later.
 
--- a/docs/guide/service/annotations.md
+++ b/docs/guide/service/annotations.md
@@ -19,10 +19,10 @@
 | [service.beta.kubernetes.io/aws-load-balancer-scheme](#lb-scheme)                                | string                  | internal                  |                                                        |
 | [service.beta.kubernetes.io/aws-load-balancer-proxy-protocol](#proxy-protocol-v2)                | string                  |                           | Set to `"*"` to enable                                 |
 | [service.beta.kubernetes.io/aws-load-balancer-ip-address-type](#ip-address-type)                 | string                  | ipv4                      | ipv4 \| dualstack                                      |
-| service.beta.kubernetes.io/aws-load-balancer-access-log-enabled                                  | boolean                 | false                     |                                                        |
-| service.beta.kubernetes.io/aws-load-balancer-access-log-s3-bucket-name                           | string                  |                           |                                                        |
-| service.beta.kubernetes.io/aws-load-balancer-access-log-s3-bucket-prefix                         | string                  |                           |                                                        |
-| service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled                   | boolean                 | false                     |                                                        |
+| service.beta.kubernetes.io/aws-load-balancer-access-log-enabled                                  | boolean                 | false                     | deprecated, in favor of [aws-load-balancer-attributes](#load-balancer-attributes)|
+| service.beta.kubernetes.io/aws-load-balancer-access-log-s3-bucket-name                           | string                  |                           | deprecated, in favor of [aws-load-balancer-attributes](#load-balancer-attributes)|
+| service.beta.kubernetes.io/aws-load-balancer-access-log-s3-bucket-prefix                         | string                  |                           | deprecated, in favor of [aws-load-balancer-attributes](#load-balancer-attributes)|
+| service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled                   | boolean                 | false                     | deprecated, in favor of [aws-load-balancer-attributes](#load-balancer-attributes)|
 | service.beta.kubernetes.io/aws-load-balancer-ssl-cert                                            | stringList              |                           |                                                        |
 | service.beta.kubernetes.io/aws-load-balancer-ssl-ports                                           | stringList              |                           |                                                        |
 | service.beta.kubernetes.io/aws-load-balancer-ssl-negotiation-policy                              | string                  | ELBSecurityPolicy-2016-08 |                                                        |
@@ -41,8 +41,7 @@
 | [service.beta.kubernetes.io/aws-load-balancer-subnets](#subnets)                                 | stringList              |                           |                                                        |
 | [service.beta.kubernetes.io/aws-load-balancer-alpn-policy](#alpn-policy)                         | stringList              |                           |                                                        |
 | [service.beta.kubernetes.io/aws-load-balancer-target-node-labels](#target-node-labels)           | stringMap               |                           |                                                        |
-| service.beta.kubernetes.io/aws-load-balancer-deletion-protection-enabled                         | boolean                 | false                     |                                                        |
-
+| [service.beta.kubernetes.io/aws-load-balancer-attributes](#load-balancer-attributes)             | stringMap               |                           |                                                        |
 ## Traffic Routing
 Traffic Routing can be controlled with following annotations:
 
@@ -218,3 +217,25 @@ Load balancer access can be controllerd via following annotations:
 The AWS Load Balancer Controller manages Kubernetes Services in a compatible way with the legacy aws cloud provider. The annotation `service.beta.kubernetes.io/aws-load-balancer-type` is used to determine which controller reconciles the service. If the annotation value is `nlb-ip` or `external`, legacy cloud provider ignores the service resource (provided it has the correct patch) so that the AWS Load Balancer controller can take over. For all other values of the annotation, the legacy cloud provider will handle the service. Note that this annotation should be specified during service creation and not edited later.
 
 The legacy cloud provider patch was added in Kubernetes v1.20 and is backported to Kubernetes v1.18.18+, v1.19.10+.
+
+## Custom attributes
+
+- <a name="load-balancer-attributes">`service.beta.kubernetes.io/aws-load-balancer-attributes`</a> specifies [Load Balancer Attributes](http://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_LoadBalancerAttribute.html) that should be applied to the NLB.
+
+  !!!warning ""
+  Only attributes defined in the annotation will be updated. To unset any AWS defaults(e.g. Disabling access logs after having them enabled once), the values need to be explicitly set to the original values(`access_logs.s3.enabled=false`) and omitting them is not sufficient.
+  Custom attributes set in this annotation's config map will be overriden by annotation-specific attributes. This is to ensure backwards compatibility but using annotation-specific attributes is deprecated and will be removed in the future.
+
+  !!!example
+  - enable access log to s3
+  ```
+  service.beta.kubernetes.io/aws-load-balancer-attributes: access_logs.s3.enabled=true,access_logs.s3.bucket=my-access-log-bucket,access_logs.s3.prefix=my-app
+  ```
+  - enable NLB deletion protection
+  ```
+  service.beta.kubernetes.io/aws-load-balancer-attributes: deletion_protection.enabled=true
+  ```
+  - enable cross zone load balancing
+  ```
+  service.beta.kubernetes.io/aws-load-balancer-attributes: load_balancing.cross_zone.enabled=true
+  ```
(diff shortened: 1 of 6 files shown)


kishorj:
@JacekDuszenko, thanks a lot for your contribution. I'm copy pasting my earlier comments from the issue. 

I'm inclined more towards specifying load balancer attributes annotation since it will be consistent with the ALB. Here is what I recommend -

    Parse service.beta.kubernetes.io/aws-load-balancer-attributes into a string map
    If there are specific annotations defined for access logs, cross zone load balancing, override the configmap from step 1.
    Do not use a default values like we do currently. If we do this, the controller will ignore the attributes not specified via an annotation. The new behavior will be consistent with ALB.

Also deprecate the existing annotations to configure individual lb attributes, and remove the support in future releases.

There is no need to modify ingress code.

JacekDuszenko:
Sure @kishorj . I've just posted a PR at the same time as you posted your comment :) I will fix the PR to parse the whole config map according to your advice. Thanks!

pkg/service/model_build_load_balancer.go
```diff
@@ -232,42 +232,94 @@ func (t *defaultModelBuildTask) resolveLoadBalancerSubnets(ctx context.Context,
...
-		},
-		{
-			Key:   lbAttrsAccessLogsS3Prefix,
-			Value: bucketPrefix,
-		},
-		{
-			Key:   lbAttrsLoadBalancingCrossZoneEnabled,
-			Value: strconv.FormatBool(crossZoneEnabled),
-		},
+func (t *defaultModelBuildTask) getDefaultAnnotationLbAttributes() map[string]string {
```
kishorj, line 271: lets not specify default attributes, so the AWS defaults get used, or the controller preserves existing attributes.

docs/guide/service/annotations.md
```diff
@@ -218,3 +217,25 @@ Load balancer access can be controllerd via following annotations:
 The AWS Load Balancer Controller manages Kubernetes Services in a compatible way with the legacy aws cloud provider. The annotation `service.beta.kubernetes.io/aws-load-balancer-type` is used to determine which controller reconciles the service. If the annotation value is `nlb-ip` or `external`, legacy cloud provider ignores the service resource (provided it has the correct patch) so that the AWS Load Balancer controller can take over. For all other values of the annotation, the legacy cloud provider will handle the service. Note that this annotation should be specified during service creation and not edited later.
 
 The legacy cloud provider patch was added in Kubernetes v1.20 and is backported to Kubernetes v1.18.18+, v1.19.10+.
+
```
kishorj, line 220: Did you verify the mkdocs rendering?
You can run `make docs-preview` to serve the pages locally

docs/guide/service/annotations.md
```diff
@@ -218,3 +217,25 @@ Load balancer access can be controllerd via following annotations:
...
+## Custom attributes
+
+- <a name="load-balancer-attributes">`service.beta.kubernetes.io/aws-load-balancer-attributes`</a> specifies [Load Balancer Attributes](http://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_LoadBalancerAttribute.html) that should be applied to the NLB.
+
+  !!!warning ""
+  Only attributes defined in the annotation will be updated. To unset any AWS defaults(e.g. Disabling access logs after having them enabled once), the values need to be explicitly set to the original values(`access_logs.s3.enabled=false`) and omitting them is not sufficient.
+  Custom attributes set in this annotation's config map will be overriden by annotation-specific attributes. This is to ensure backwards compatibility but using annotation-specific attributes is deprecated and will be removed in the future.
```
kishorj, line 227: Lets reword as
For backwards compatibility, existing annotations for the individual load balancer attributes get precedence in case of ties.

pkg/service/model_build_load_balancer.go
```diff
@@ -232,42 +232,94 @@ func (t *defaultModelBuildTask) resolveLoadBalancerSubnets(ctx context.Context,
...
-		t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixAccessLogS3BucketName, &bucketName, t.service.Annotations)
-		t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixAccessLogS3BucketPrefix, &bucketPrefix, t.service.Annotations)
-	}
-	crossZoneEnabled := t.defaultLoadBalancingCrossZoneEnabled
-	if _, err := t.annotationParser.ParseBoolAnnotation(annotations.SvcLBSuffixCrossZoneLoadBalancingEnabled, &crossZoneEnabled, t.service.Annotations); err != nil {
+	specificAttribs, err := t.getAnnotationSpecificLbAttributes()
+	if err != nil {
 		return []elbv2model.LoadBalancerAttribute{}, err
 	}
+	mergedAttribs := mergeAttributeMaps(defaultAttribs, mergeAttributeMaps(configMapAttribs, specificAttribs))
```
kishorj, line 244: lets use the existing algorithm.MergeStringMap which takes in multiple maps in order of precedence from high to low.

pkg/service/model_build_load_balancer.go
```diff
@@ -232,42 +232,94 @@ func (t *defaultModelBuildTask) resolveLoadBalancerSubnets(ctx context.Context,
...
+	return firstMap
+}
+
+func mkAttributesSliceFromConfigMap(configMapAttrs map[string]string) []elbv2model.LoadBalancerAttribute {
+	attributes := make([]elbv2model.LoadBalancerAttribute, 0, len(configMapAttrs))
+	keys := make([]string, 0, len(configMapAttrs))
+	for k := range configMapAttrs {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
kishorj, line 261: no need to sort here. You could sort and compare in the unit tests instead.

pkg/service/model_build_load_balancer.go
```diff
@@ -232,42 +232,94 @@ func (t *defaultModelBuildTask) resolveLoadBalancerSubnets(ctx context.Context,
...
+}
+
+func mergeAttributeMaps(firstMap map[string]string, secondMap map[string]string) map[string]string {
+	for key, value := range secondMap {
+		firstMap[key] = value
+	}
+	return firstMap
+}
+
+func mkAttributesSliceFromConfigMap(configMapAttrs map[string]string) []elbv2model.LoadBalancerAttribute {
```
kishorj, line 255: makeAttributesSliceFromMap instead. I prefer not to use the term configMap, arg could be named loadBalancerAttributesMap

pkg/service/model_build_load_balancer.go
```diff
@@ -232,42 +232,94 @@ func (t *defaultModelBuildTask) resolveLoadBalancerSubnets(ctx context.Context,
 }
 
 func (t *defaultModelBuildTask) buildLoadBalancerAttributes(_ context.Context) ([]elbv2model.LoadBalancerAttribute, error) {
-	var attrs []elbv2model.LoadBalancerAttribute
-	accessLogEnabled := t.defaultAccessLogS3Enabled
-	bucketName := t.defaultAccessLogsS3Bucket
-	bucketPrefix := t.defaultAccessLogsS3Prefix
-	if _, err := t.annotationParser.ParseBoolAnnotation(annotations.SvcLBSuffixAccessLogEnabled, &accessLogEnabled, t.service.Annotations); err != nil {
+	defaultAttribs := t.getDefaultAnnotationLbAttributes()
+	configMapAttribs, err := t.getConfigMapLbAttributes()
```
kishorj, line 236: lets rename to getLoadBalancerAttributes instead. Config map is a bit confusing since there is no configmap resource involved. Also, loadBalancerAttributes instead of configMapAttributes.
nit: loadBalancerAttributes instead of configMapAttribs

pkg/service/model_build_load_balancer.go
```diff
@@ -232,42 +232,94 @@ func (t *defaultModelBuildTask) resolveLoadBalancerSubnets(ctx context.Context,
...
+	if err != nil {
 		return []elbv2model.LoadBalancerAttribute{}, err
 	}
-	if accessLogEnabled {
-		t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixAccessLogS3BucketName, &bucketName, t.service.Annotations)
-		t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixAccessLogS3BucketPrefix, &bucketPrefix, t.service.Annotations)
-	}
-	crossZoneEnabled := t.defaultLoadBalancingCrossZoneEnabled
-	if _, err := t.annotationParser.ParseBoolAnnotation(annotations.SvcLBSuffixCrossZoneLoadBalancingEnabled, &crossZoneEnabled, t.service.Annotations); err != nil {
+	specificAttribs, err := t.getAnnotationSpecificLbAttributes()
```
kishorj, line 240: nit: lets use Attributes instead of Attribs

pkg/service/model_build_load_balancer.go
JacekDuszenko: It would be really hard to sort in unit tests or at least I don't know how to do it.
The structure of tests that were failing is as follows:
1. There's a raw json string passed as a `wanted` value: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/service/model_builder_test.go#L109
2. There's a single validation method that checks if both json strings are equal from the point of json document view (particular keys ordering doesn't matter, list ordering does matter) 
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/service/model_builder_test.go#L2100

These tests implicitly depend on the ordering and I've no idea how to sort in the test itself. Could you please elaborate on how would you sort inside the test here?

kishorj: I was referring to the unit test Test_defaultModelBuilderTask_buildLBAttributes where you can sort the output and the expectation before comparing.

You are right on the Test_defaultModelBuilderTask_Build test. I assumed that with the support for the default removed, the model generated in the Test_defaultModelBuilderTask_Build function would not have the load balancer attributes - but I just realized some tests require it.

could use `sort.Slice` instead of sorting the keys

JacekDuszenko: Alright, thanks! I addressed the rest of your comments in code, was just waiting for the response on that one. I'll improve sorting and push another commit later in the day and let you know.

JacekDuszenko:
@kishorj I made suggested changes and pushed the code. It's ready for a re-review :) Thanks again for fast responses.

JacekDuszenko:
Is this codecov report acceptable for a merge or should I change/add something?

kishorj:
> @kishorj, Is this codecov report acceptable for a merge or should I change/add something?

codecov looks fine to me. I will have to delay merging this PR until we finalize the patch release v2.2.1. Once the review is complete and v2.2.1 is finalized, I will merge the changes for v2.3.0.

JacekDuszenko:
Great to hear that, I'm glad I was able to help :) And thank you for fast feedback

docs/guide/service/annotations.md
```diff
@@ -218,3 +217,25 @@ Load balancer access can be controllerd via following annotations:
 The AWS Load Balancer Controller manages Kubernetes Services in a compatible way with the legacy aws cloud provider. The annotation `service.beta.kubernetes.io/aws-load-balancer-type` is used to determine which controller reconciles the service. If the annotation value is `nlb-ip` or `external`, legacy cloud provider ignores the service resource (provided it has the correct patch) so that the AWS Load Balancer controller can take over. For all other values of the annotation, the legacy cloud provider will handle the service. Note that this annotation should be specified during service creation and not edited later.
 
 The legacy cloud provider patch was added in Kubernetes v1.20 and is backported to Kubernetes v1.18.18+, v1.19.10+.
+
+## Custom attributes
+
+- <a name="load-balancer-attributes">`service.beta.kubernetes.io/aws-load-balancer-attributes`</a> specifies [Load Balancer Attributes](http://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_LoadBalancerAttribute.html) that should be applied to the NLB.
```
kishorj, line 223: Lets move this to the "## Resource attributes" section

pkg/service/model_build_load_balancer.go
```diff
@@ -232,42 +232,75 @@ func (t *defaultModelBuildTask) resolveLoadBalancerSubnets(ctx context.Context,
...
+	var rawAttributes map[string]string
+	if _, err := t.annotationParser.ParseStringMapAnnotation(annotations.SvcLBSuffixLoadBalancerAttributes, &rawAttributes, t.service.Annotations); err != nil {
+		return nil, err
 	}
+	for attrKey, attrValue := range rawAttributes {
+		if existingAttrValue, exists := mergedAttributes[attrKey]; exists && existingAttrValue != attrValue {
+			return nil, errors.Errorf("conflicting loadBalancerAttribute %v: %v | %v", attrKey, existingAttrValue, attrValue)
+		}
+		mergedAttributes[attrKey] = attrValue
+	}
```
kishorj, line 272: this is not necessary

pkg/service/model_build_load_balancer_test.go
```diff
@@ -17,6 +17,8 @@ import (
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2"
 )
 
+const lbAttrsDeletionProtectionEnabled = "deletion_protection.enabled"
```
kishorj, line 20: you could group it together with the rest of the constants in model_build_load_balancer.go

kishorj (commented):
I have some minor comments, changes look good otherwise.

pkg/service/model_build_load_balancer_test.go
JacekDuszenko: I could but its used only in test and I don't think we should put constants used only in test in production source code if that makes sense

JacekDuszenko:
Hi @kishorj, thanks a lot for your comments! I addressed them, let me know what's the final decision on https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/2057#discussion_r651998471

pkg/service/model_build_load_balancer_test.go
kishorj: ok

kishorj:
hold merge until v2.2.1 release is cut

ingshtrom:
Do we need to get this cherry-picked into the release-2.2 branch in order to get it in an upcoming release or do we need to wait for 2.3?

kishorj:
@ingshtrom, this fix will be included in v2.3.0 and later releases.

PR #2062, opened 2021-06-09
Update installation.md

tomahawk-pilot:
cert manager version update .
fix on load balancer apply kubectl command


tomahawk-pilot:
i have multiple email ids linked to my github account and don't know which one was used to do this commit since it was done via the console.

PR #2067, opened 2021-06-10
Update spec for new apiVersion

### LINKED ISSUE #2066: The spec for Ingress resource is incorrect
The manifest for `apiVersion: networking.k8s.io/v1` has the incorrect specs.
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/3a6ba6a7c16e5fe9f63949381b24cebadd7e8c9f/docs/examples/2048/2048_full.yaml#L52-L59

The spec should conform to the spec referenced here:
https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource
```yaml
spec:
  rules:
    - http:
        paths:
        - path: /*
          pathType: Prefix
          backend:
             service:
               name:  service-2048
               port:
                 number: 80
```

tlliddle:
This change updates the spec to support `apiVersion: networking.k8s.io/v1` for Kubernetes version 1.19+

### DIFF
--- a/docs/examples/2048/2048_full.yaml
+++ b/docs/examples/2048/2048_full.yaml
@@ -53,7 +53,7 @@ spec:
   rules:
     - http:
         paths:
-        - path: /*
+        - path: /
           pathType: Prefix
           backend:
              service:
--- a/docs/examples/2048/2048_full.yaml
+++ b/docs/examples/2048/2048_full.yaml
@@ -53,7 +53,11 @@ spec:
   rules:
     - http:
         paths:
-          - path: /*
-            backend:
-              serviceName: service-2048
-              servicePort: 80
+        - path: /*
+          pathType: Prefix
+          backend:
+             service:
+               name:  service-2048
+               port:
+                 number: 80
+


tlliddle:
@kishorj - I tried to retitle but I I don't have permissions.

M00nF1sh:
@tlliddle 
Thanks so much for your effort. 
we decided to rollback https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/175319d9152a0c9fd1b60df6dac53aafdc1c112b instead. 

The reason is EKS is still supporting 1.16 as a managed solution, and networking/v1 is not available there.  This YAML have been used in multiple blogs/docs, and should support minimal supported EKS versions.

tlliddle:
@M00nF1sh - reverting makes sense.  I would like to put a PR for your suggestion to make a latest file to support the newer apiVersion.  

If we agree that is the path forward, I can start that work right away?

M00nF1sh:
Yes, we can do that, thanks so much to help on this 👍 
would you help add some comments at the top of the new file to state the minimal supported k8s version as well 😃

kishorj:
lets also update the guide/ingress/spec.md to include both examples. We can remove the v1beta1 example later.

tlliddle:
@kishorj - Great call out.  I will make that part of the PR also.

tlliddle:
@kishorj , @M00nF1sh - I'm going to close out this PR (mistakes to commit msg and title) and open a new one based on this discussion.  The new PR is #2072.

Issue #2087, opened 2021-06-21
Host aws-load-balancer-controller Helm chart in the controller repo


fawadkhaliq:
**Description**
Helm chart is the preferred way of packaging and installing the AWS Load Balancer controller. Not having the chart source code in the `aws-load-balancer-controller` repo leads to reactive/post-merge/post-release/manual validation of the chart. We would like maintain the charts in the controller repo and sync with the eks-charts on releases (controller release or just a Helm chart release). This will also allow us to 1) test helm charts in integration and end to end tests 2) avoid the need to manually copy over CRD changes etc

PR #2088, opened 2021-06-21
Add aws-load-balancer-controller Helm chart

fawadkhaliq:
**Issue**
#2087 

**Description**
Helm chart is the preferred way of packaging and installing the AWS Load Balancer controller. Not having the chart source code in the aws-load-balancer-controller repo leads to reactive/post-merge/post-release/manual validation of the chart. We would like maintain the charts in the controller repo and sync with the eks-charts on releases (controller release or just a Helm chart release). This will also allow us to 1) test helm charts in integration and end to end tests 2) avoid the need to manually copy over CRD changes 

This PR adds the AWS Load Balancer controller Helm chart. Next, we will add scripts to sync Helm charts with eks-charts on release and migrate the existing integration tests to setup controller using Helm instead

### DIFF
--- a/helm/aws-load-balancer-controller/.helmignore
+++ b/helm/aws-load-balancer-controller/.helmignore
@@ -1,0 +1,24 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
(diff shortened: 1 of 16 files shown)


M00nF1sh (approved):
/lgtm

PR #2089, opened 2021-06-22
Added issue and pull request templates

fawadkhaliq:
**Description**
To speed review process for pull requests and have a consistent structure in managing issues, added issue and pull request templates


dhawal1304:
Hi All ,

Does this controller support multiple protocols e.g TCP and UDP in the same YAML. 
I was not able to create one and getting the below messages.

Error syncing load balancer: failed to ensure load balancer: mixed protocol is not supported for LoadBalancer

Issue #2092, opened 2021-06-23
Prevent creation of TargetGroupBinding if its already referenced by an existing TargetGroup


fawadkhaliq:
**Describe the bug**
TargetGroupBinding has a one-to-one mapping with TargetGroup as per the datamodel, however, the controller does not enforce this behavior. This can lead to two TargetGroupBinding resources point to the same TargetGroup. 

**Steps to reproduce**
Create a TargetGroupBinding-1 that maps to TargetGroup-A. Create a TargetGroupBinding-2 that maps to same TargetGroup-A. The controller will attempt assign the same target group to TargetGroupBinding-2

**Expected outcome**
Based on the example above, TargetGroupBinding-2 creation should result in an error that TargetGroupA is already in use by TargetGroupBinding-1

**Environment**

* AWS Load Balancer controller version: v2.2.0
* Kubernetes version: v1.19
* Using EKS (yes/no), if so version? Yes, EKS v1.19

**Additional Context**:
None

fawadkhaliq:
Fixed it here https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/2098

PR #2098, opened 2021-06-24
Enforce one target group per target group binding

fawadkhaliq:
### Issue
#2092 

### Description
Prevent creation of TargetGroupBinding if its already referenced by an existing TargetGroup

TargetGroupBinding has a one-to-one mapping with TargetGroup as per the data model, however, the controller does not enforce this behavior. This can lead to two TargetGroupBinding resources point to the same TargetGroup.

### Checklist
- [X] Manually tested


M00nF1sh (approved):
/lgtm
/approve
🥂

Issue #2056, opened 2021-06-07
Feature Request: Subnet discovery to be able to select subnet based on available address spaces


anthonysuterag:
We have public subnets that span across 3 AZs, and we use these subnets to host our internet facing LBs. 
Recently, we started to run out of address spaces, so we would like to extend those by adding new subnets, which means we currently have 6 public subnets (two in each availability zone). 

However, I figured out that subnet discovery will select subnet based on lexicographical order by the Subnet IDs (https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/networking/subnet_resolver.go#L148). I think it would be great to adjust this behavior to be able to select subnet based on available address spaces in case of multiple tagged subnets are available.

shinji62:
I would like this to happen too, and happy to make a PR if required.

fabasoad:
Would be great to have it 👍

kishorj:
@anthonysuterag,
Controller already supports configuring custom subnets via annotation. 
`alb.ingress.kubernetes.io/subnets` for ALB
`service.beta.kubernetes.io/aws-load-balancer-subnets` for NLB

Would this not be helpful for your use case?

anthonysuterag:
@kishorj Might be a workaround, but if it is possible, I would like to avoid assigning subnets statically to the Ingress, because in the long run, it would be difficult to maintain and also to keep track as we have a lot of ingresses. So, I would prefer to let subnet discovery to handle the subnets assigning process.

M00nF1sh:
If we add a logic based on available IP addresses, we might need to restrict the subnet auto-discovery for new LB creation only. for existing LBs, we shouldn't invoke subnet auto-discovery as it might change subnets and cause downtime. 

we can do similar things as we currently did for service model builder here: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/service/model_build_load_balancer.go#L118

shinji62:
@M00nF1sh Sounds ok for me.

PR #2129, opened 2021-07-21
Restrict subnet auto-discovery to new LB creation on service side

oliviassss:
### Issue

#2056

### Description

Only use subnet auto-discovery when there is no existing Load Balancer on service side.

- When building the scheme, use specified scheme in annotation if any. If there's no such annotation, try to use the scheme in existing Load Balancer. If there's no existing Load Balancer, use "Internal-Facing" as default scheme.
- When building the subnets, resolve subnets by specified subnet name or ID in annotation if any. if there is no such annotation, try to resolve by the subnet name of ID in existing Load Balancer. If there is no existing Load Balancer, then we use auto discovery to resolve subnets.

#### Tests:

- Created service resource, verified the subnet auto-discovery was restricted to the circumstances with no existing load balancers
- Created service resource, verified the build subnet used the subnet Name or ID from annotation directly
- Created service resource, verified the build subnets used the subnet Name or ID from existing Load Balancer when there was no Name or ID in annotation

### Checklist
- [x] Manually tested


pkg/service/model_build_load_balancer.go
```diff
@@ -217,14 +241,27 @@ func (t *defaultModelBuildTask) getMatchingIPforSubnet(_ context.Context, subnet
...
 		)
 	}
+
+	existingLB, err := t.fetchExistingLoadBalancer(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if existingLB != nil {
+		subnetIDs := make([]string, 0, len(existingLB.LoadBalancer.AvailabilityZones))
+		return t.subnetsResolver.ResolveViaNameOrIDSlice(ctx, subnetIDs,
```
kishorj, line 259: You need to iterate through the subnets in the existingLB.LoadBalancer.AvailabilityZones and push the subnetIDs to the list.

oliviassss: Changed, thanks!

kishorj (approved):

PR #2146, opened 2021-07-29
Discovery subnets by available ip addresses

oliviassss:
### Issue

#2056 

### Description

Add a new step in subnet discovery, which is to select subnets based on the count of available IP addresses.

- ALB：To ensure that the load balancer can scale properly, each Availability Zone subnet has at least 8 free IP addresses per subnet. _(Ref: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#subnets-load-balancer)_

- NLB:  For internet-facing load balancers, the subnets must have at least 8 available IP addresses. For internal load balancers, this is only required if the user let AWS select a private IPv4 address from the subnet. _(Ref: https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html)_

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/service/model_build_load_balancer.go
+++ b/pkg/service/model_build_load_balancer.go
@@ -27,33 +27,35 @@ const (
 	lbAttrsAccessLogsS3Prefix            = "access_logs.s3.prefix"
 	lbAttrsLoadBalancingCrossZoneEnabled = "load_balancing.cross_zone.enabled"
 	resourceIDLoadBalancer               = "LoadBalancer"
+	minimalAvailableIPAddressCount 		 = int64(8)
 )
 
-func (t *defaultModelBuildTask) buildLoadBalancer(ctx context.Context, scheme elbv2model.LoadBalancerScheme) error {
-	spec, err := t.buildLoadBalancerSpec(ctx, scheme)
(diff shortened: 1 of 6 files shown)


pkg/service/model_build_load_balancer.go
```diff
@@ -267,6 +271,16 @@ func (t *defaultModelBuildTask) buildLoadBalancerSubnets(ctx context.Context, sc
 		)
 	}
 
+	// for internet-facing Load Balancers, the subnets mush have at least 8 available IP addresses;
+	// for internal Load Balancers, this is only required if private ip address is not assigned
```
kishorj, line 275: why not check if the service has the  `service.beta.kubernetes.io/aws-load-balancer-private-ipv4-addresses` annotation set in this function? If private annotation is set on an internal load balancer, we can skip the min available check. If we check for the annotation here, then this function doesn't need the ipv4Configured arg to be passed in. This will also help us decouple the subnet discovery logic from the build load balancer function - and possibly less change :).

oliviassss: changes committed, thanks

kishorj (approved):

mikestef9:
Can we add a doc update for this feature? Probably here https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.3/deploy/subnet_discovery/

oliviassss:
@mikestef9 Sure, will update it in the patch release, thanks.

Issue #1900, opened 2021-03-22
delete.protection does not reconcile the resource


mridu23:
We recently migrated from ALB ingress controller (v1) to Load Balancer controller(v2) and noticed the following behaviour:

As per the expected behaviour, when load balancer deletion protection annotation is used with the ingress, the controller does not directly goes ahead and deletes the resources until the load balancer deletion protection is turned off.
 alb.ingress.kubernetes.io/load-balancer-attributes: deletion_protection.enabled=true

With v1 - if ingress with load balancer deletion protection tag set to true was deleted, and later updated to set this tag value to false, the controller reconciled the resource and honoured the changes made

However, with v2 - if ingress with load balancer protection tag set to true is deleted, and later the ingress resource is updated, the controller does not reconcile the resource. It is required to change the protection tag to false before the ingress deletion., otherwise the resource just hangs.

M00nF1sh:
@mridu23 
Hi, there is a behavior change from v1 to v2 regards the `alb.ingress.kubernetes.io/load-balancer-attributes` annotation. There is an PR to update the docs about this: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/1869/files

In v1, we defined default values for these attributes, so remove the attribute from annotation means revert the attribute to the default value defined by the controller. (for deletion protection, the default value is disabled, so remove it clears deletion_protection).

in v2, we didn't define any default values for these attributes. remove the attribute from annotation means "keep the attribute value as what it is in AWS). you need to explicitly set deletion_protection to false in annotation to disable it.

We did this change to gain two benefit:
1. support future new attributes without controller changes. (we don't have default values for these new attributes, and we don't want to remove these future new attributes when they are not specified in annotations).

2. align with AWS API. (the ELBv2 API only allows to modify attributes with a set of attributes to update, and un-specified attributes are left as it is).

kishorj:
@mridu23 
If you delete an ingress with the deletion_protection attribute enabled, controller will not be able to delete the underlying AWS resources. To recover from this situation, you can manually disable the deletion_protection attribute from the AWS console for the affected load balancer. We will call this out in our docs.

/kind documentation

mridu23:
@M00nF1sh @kishorj 
Thanks for addressing my query. I completely understand the reason for change and the expected behavior.

My concern is more around the resource reconciliation.

[...]

PR #2172, opened 2021-08-11
Force delete lb when deletion_protection is disabled

oliviassss:
### Issue

#1900 

### Description
This PR is to handle the deletion_protection.enable for both ALB and NLB.

- If `deletion_protection.enable=true` is in the annotation and the customer deletes the load balancer, the controller will return error right away and refuse to reconcile further
- Once the attribute gets edited to `deletion_protection.enable=false`, the deployer will call ModifyLoadBalancerAttributes() to disable the deletion_protection, so it can continue to delete the resource.

#### Tests:

- Created NLB, with `deletion_protection.enable=true`, verified the service controller will return error and stop reconciling.
- Then edited the attributes using `kubectl edit svc/<my-svc> -n <my-namespace>,` to set  `deletion_protection.enable=false`, verified the deployer will disable the deletion_protection and the resource got deleted.
- Created ALB, with `deletion_protection.enable=true`, verified the group controller will return error and stop reconciling.
- Then edited the attributes using `kubectl edit ingress/<my-ingress> -n <my-namespace>,` to set  `deletion_protection.enable=false`, verified the deployer will disable the deletion_protection and the resource got deleted.

### Checklist
- [x] Manually tested

### DIFF
--- a/controllers/ingress/group_controller.go
+++ b/controllers/ingress/group_controller.go
@@ -29,6 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/source"
+	"strconv"
 )
 
 const (
@@ -36,8 +37,9 @@ const (
 	controllerName   = "ingress"
 
 	// the groupVersion of used Ingress & IngressClass resource.
-	ingressResourcesGroupVersion = "networking.k8s.io/v1beta1"
-	ingressClassKind             = "IngressClass"
+	ingressResourcesGroupVersion 	 = "networking.k8s.io/v1beta1"
+	ingressClassKind             	 = "IngressClass"
+	lbAttrsDeletionProtectionEnabled = "deletion_protection.enabled"
 )
 
 // NewGroupReconciler constructs new GroupReconciler
@@ -74,6 +76,7 @@ func NewGroupReconciler(cloud aws.Cloud, k8sClient client.Client, eventRecorder 
 		modelBuilder:     modelBuilder,
 		stackMarshaller:  stackMarshaller,
 		stackDeployer:    stackDeployer,
+		annotationParser: annotationParser,
 
 		groupLoader:           groupLoader,
 		groupFinalizerManager: groupFinalizerManager,
@@ -91,6 +94,7 @@ type groupReconciler struct {
 	modelBuilder     ingress.ModelBuilder
 	stackMarshaller  deploy.StackMarshaller
 	stackDeployer    deploy.StackDeployer
+	annotationParser annotations.Parser
 
 	groupLoader           ingress.GroupLoader
 	groupFinalizerManager ingress.FinalizerManager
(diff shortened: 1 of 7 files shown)


tanvp112:
@oliviassss, how do you set deletion_protection.enable=true for a nlb? *alb.ingress.kubernetes.io/load-balancer-attributes* is an annotation for ingress (does not support NLB) and not service. Can you shade a light here?

kishorj:
> @oliviassss, how do you set deletion_protection.enable=true for a nlb? _alb.ingress.kubernetes.io/load-balancer-attributes_ is an annotation for ingress (does not support NLB) and not service. Can you shade a light here?

The deletion protection will be supported for NLB in the upcoming v2.3.0 release (PR #2051).

oliviassss:
@tanvp112 You can add this annotation in the yaml file to enable deletion_protection for NLB: `service.beta.kubernetes.io/aws-load-balancer-attributes: deletion_protection.enabled=true`. Check this [link](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/guide/service/annotations.md)

pkg/deploy/elbv2/load_balancer_synthesizer.go
```diff
@@ -55,7 +62,13 @@ func (s *loadBalancerSynthesizer) Synthesize(ctx context.Context) error {
 	// I don't like this, but it's the easiest solution to meet our requirement :D.
 	for _, sdkLB := range unmatchedSDKLBs {
 		if err := s.lbManager.Delete(ctx, sdkLB); err != nil {
-			return err
+			errMessage := err.Error()
+			if strings.Contains(errMessage,"OperationNotPermitted") {
```
kishorj, line 66: Is it possible to check for more specific error?

controllers/service/service_controller.go
```diff
@@ -88,6 +90,13 @@ func (r *serviceReconciler) reconcile(ctx context.Context, req ctrl.Request) err
 		return client.IgnoreNotFound(err)
 	}
 	if !svc.DeletionTimestamp.IsZero() {
+		deletionProtectionEnabled, deletionProtectionSpecified, err := r.getDeletionProtectionViaAnnotation(*svc)
+		if err != nil {
+			return err
+		}
+		if deletionProtectionSpecified && deletionProtectionEnabled {
```
kishorj, line 97: deleteProtectionEnabled true would always imply deleteProtectionSpecified, so no need to check for both.

kishorj (commented):
With the proposed change, I see that if delete protection is not specified via annotation, but set on the NLB out-of-band, controller still force deletes the underlying NLB. This is contrary to the current design where controller doesn't nodify attributes not explicitly specified. This is a corner case, I agree, but lets see if it would make more sense to block the service deletion.

pkg/deploy/elbv2/load_balancer_synthesizer.go
oliviassss: I can check if the error message contains keyword like "deletion protection", but not sure if it's better than check the error code as it is now. Or I can check both.

controllers/service/service_controller.go
oliviassss: Changed.

controllers/ingress/group_controller.go
```diff
@@ -124,7 +128,17 @@ func (r *groupReconciler) reconcile(ctx context.Context, req ctrl.Request) error
 		r.recordIngressGroupEvent(ctx, ingGroup, corev1.EventTypeWarning, k8s.IngressEventReasonFailedAddFinalizer, fmt.Sprintf("Failed add finalizer due to %v", err))
 		return err
 	}
-
+	for _, inactiveMember := range ingGroup.InactiveMembers {
```
kishorj, line 131: we could do this check from the model builder task, any reason not to do so?
If we check this from the model builder task, it will be easier to UT. Same comment for service controller as well.

pkg/deploy/elbv2/load_balancer_synthesizer.go
oliviassss: As far as I can see, the error message is `"OperationNotPermitted: Load balancer 'arn:aws:elasticloadbalancing:xyz123' cannot be deleted because deletion protection is enabled\n\tstatus code: 400, request id: abc123"`
So I changed to check both keywords "OperationNotPermitted" and "deletion protection".

controllers/ingress/group_controller.go
oliviassss: Changed accordingly.

oliviassss:
> With the proposed change, I see that if delete protection is not specified via annotation, but set on the NLB out-of-band, controller still force deletes the underlying NLB. This is contrary to the current design where controller doesn't nodify attributes not explicitly specified. This is a corner case, I agree, but lets see if it would make more sense to block the service deletion.

According to our internal discussion, we will keep the method as it is now, and record this corner case situation in our documentation properly.

kishorj (approved):

PR #2189, opened 2021-08-22
cert-manager apiversion to v1

cw-sakamoto:
Signed-off-by: cw-sakamoto <sakamoto@chatwork.com>

### Description

I recreate PR for https://github.com/aws/eks-charts/pull/579.

### Checklist
- [x] Manually tested


helm/aws-load-balancer-controller/templates/webhook.yaml
```diff
@@ -142,7 +142,7 @@ data:
   tls.crt: {{ $tls.clientCert }}
   tls.key: {{ $tls.clientKey }}
 {{- else }}
-apiVersion: cert-manager.io/v1alpha2
+apiVersion: cert-manager.io/v1
```
kishorj, line 145: Customers with old cert-manager setup will not be able to upgrade to the new load balancer controller chart since the old setup doesn't recognize v1 version. This will confuse the end users.

Is it possible to use `.Capabilities.APIVersions.Has` to use v1 only if support is available?

kishorj (commented):
I have a concern regarding setups with existing cert-manager installation. If we can address the issue I've commented below, I feel comfortable to include the changes as part of next release.

helm/aws-load-balancer-controller/templates/webhook.yaml
cw-sakamoto: It can be done as follows, but `--validate` is required if we want to use CRD apiversion for `.Capabilities.APIVersions.Has`.

```
helm template aws-load-balancer-controller ./aws-load-balancer-controller -f settings.yaml -n kube-system --validate
```


https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/2189/commits/fe800008ac1599b881b22c2640402aaf6f66dfcb

kishorj (approved):

PR #2190, opened 2021-08-23
add load balancer attributes support for IngressClassParams

oliviassss:
### Description
AWS Load balancer controller v2.2.0 and later support IngressClassParams which allows the admins to specify and enforce common configurations for the load balancers and provide fine grained access-control for the ingress group membership.

This PR adds an optional setting **loadBalancerAttributes** for IngressClassParams, to specify the ALB attributes to apply for all Ingresses of the IngressClass.

- If  **loadBalancerAttributes** are defined, controller will configure all of the load balancers with the specified attributes. These attributes cannot be overridden via ingress annotations.
- Currently supported Load balancer attributes are listed in the [AWS documentation](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#load-balancer-attributes).
- In case of invalid attribute keys or values, controller will fail to reconcile the ingresses belonging to the ingress class.
- In case of the conflict attributes values for the Ingresses under the same ingress group, the controller will fail to reconcile.

#### Tests

1.  Added unit tests.
2. Created IngressClass, IngressClassParams and Ingresses, verified the following circumstances:

- If both IngressClassParams and  Ingress have the same attribute keys, but different attribute values, the controller will use the value from IngressClassParams to build the load balancers;
- If the attributes are only specified in IngressClassParams, the controller will configure the attributes to all the ALBs of the IngressClass;
- If there is invalid attributes keys or values in IngressClassParams or Ingresses, the controller will fail to reconcile;
- If there are conflicting attributes values for the Ingresses under the same ingress group, the controller will fail to reconcile.

### Checklist
- [ ] Manually tested


stevehipwell:
Is there a timeframe for when this is likely to be released?

PR #2195, opened 2021-08-26
Set maximum length for custom load-balancer-name

bnutt:
### Description

AWS only allows loadbalancer names with a length up to 32 characters, this change trims the length of the provided value when a custom load balancer name is provided.

### Checklist
- [ ] Manually tested


kishorj (approved):

M00nF1sh:
I think we should change this logic

we shouldn't trim the LB name silently if it exceeds 32.
instead, we should validate that it matches the regex & within 32 character and error out if validation fails.

Issue #2118, opened 2021-07-07
Optimized security group rules


kishorj:
# AWS Load balancer controller - optimized SG rules

## Overview

There are two limitations to the security groups handling currently

* If custom security groups are specified for ALB, i.e. controller doesn’t auto-create one for the ingress group, then we expect the users to manually configure their ENI/Node security groups to permit the ingress traffic from the load balancer. 
* In case of auto-generated SG, controller creates a unique security group for each LB. This will cause the node group SG rules to grow linearly with the number of load balancers. As a result, the SG rule quota effectively limits the total number of LBs.

The goal of this feature is to enable specifying shared security groups for load balancers and the controller automatically add to the ENI/node group security groups to allow traffic from the load balancer. We will provide a controller flag to specify a default SG to use. The shared SG approach will enable users to use a constant number of SG rules for the LB ingress traffic and the SG rule quota will no longer be a bottleneck for the number of LBs.

## Security groups

We will classify security groups into two categories

1. Frontend SG
2. Backend SG

### Frontend SG

Frontend security groups control the clients that can access the load balancer. These SGs contain rules from the `inbound-cidrs` to the `listen-ports`. The frontend security groups can be configured via the exclusive annotation on the ingress resource -

```
`alb.ingress.kubernetes.io/security-groups`
```

If the annotation is not specified, controller will automatically create one security group per load balancer to allow the traffic from `inbound-cidrs` to `listen-ports`.

### Backend SG

Backend security groups control the traffic between the load balancer and the EC2 instances or the ENIs. These SGs are attached to the load balancer to tag the LB traffic and are used as traffic source in the ENI/Instance SG rules. The backend security groups are shared between multiple load balancers, the controller will ensure that the shared rules do not get deleted prematurely. Backend SGs are be configured via the following mechanisms - 

* Default security groups for cluster
    * Specified via controller command line flag or auto-generated based on the flag
* Fallback to the auto-created frontend SG
    * In case the default backend SG is not specified
    * Auto-generated frontend SG gets used as backend SG
    * This configuration provides backwards compatibility with prior releases

### Management of Backend SG rules

When the controller auto-creates the frontend SG for a load balancer, it automatically adds the security group rules to allow egress traffic from the load balancer to the EC2/Fargate instances. In case security group is specified via annotation, the SG rules do not get added by default. The automatic management of instance/ENI security group can be controlled via an additional annotation on the ingress resource

[...]

PR #2205, opened 2021-09-01
Support optimized security group rules for ALB

kishorj:
### Issue
Issue: #2118

### Description

Add support for using shared backend security groups for ALB. This will result in using a constant number of SG rules per cluster for the purpose of allowing access to ALB traffic. Also enable management of security group rules for manually specified security groups. For details refer to issue #2118.

#### Tests
- Backend SG feature get enabled by default, SG is auto-generated if needed, sg name is of the format k8s-<clusterName>-traffic-<15 hex chars of sha256(clusterName + VPCID)>
- backend SG is attached to the ALB
- Ingress rule gets added to the instance security group to allow ALB traffic
- Ingresses in the same cluster share the backend SG
- backend SG gets deleted automatically if no longer needed
- Ingresses with `alb.ingress.kubernetes.io/security-groups` annotation don't use the backend SG by default
- When `alb.ingress.kubernetes.io/manage-security-group-rules: "true"` annotation is applied to ingress with manual SG configuration, the backend SG gets attached to the corresponding ALB, and the node SG gets updated to allow traffic
- With manual security group configuration and manage rules set to true on an ingress, model builder returns error if backend security group feature is not enabled
- When backend SG feature is disabled via the controller flag, the auto-generated SG gets used in the node SG rules as before
- Backend SGs can be configured via command line flag `--backend-security-groups` in lieu of auto-generated ones
 
### Checklist
- [x] Manually tested


controllers/ingress/group_controller.go
```diff
@@ -148,6 +149,12 @@ func (r *groupReconciler) reconcile(ctx context.Context, req ctrl.Request) error
 		}
 	}
 
+	if len(ingGroup.Members) == 0 {
```
M00nF1sh, line 152: should this happen before we release group finalizer?

pkg/annotations/constants.go
```diff
@@ -45,6 +45,7 @@ const (
 	IngressSuffixAuthSessionCookie            = "auth-session-cookie"
 	IngressSuffixAuthSessionTimeout           = "auth-session-timeout"
 	IngressSuffixTargetNodeLabels             = "target-node-labels"
+	IngressSuffixManageSecurityGroupRules     = "manage-security-group-rules"
```
M00nF1sh, line 48: wondering whether it would be better to be specific like `manage-node-security-group-rules` or `manage-backend-security-group-rules`

pkg/config/controller_config.go
```diff
@@ -68,6 +71,13 @@ type ControllerConfig struct {
 	TargetGroupBindingMaxConcurrentReconciles int
 	// Max exponential backoff delay for reconcile failures of TargetGroupBinding
 	TargetGroupBindingMaxExponentialBackoffDelay time.Duration
+
+	// EnableBackendSecurityGroup specifies whether to use optimized security group rules
+	EnableBackendSecurityGroup bool
+
+	// BackendSecurityGroups specifies the configured backend security groups to use
+	// for optimized security group rules
+	BackendSecurityGroups []string
```
M00nF1sh, line 80: what's the use case of multiple backSG vs a single one?

pkg/ingress/model_build_load_balancer.go
```diff
@@ -255,7 +255,18 @@ func (t *defaultModelBuildTask) buildLoadBalancerSecurityGroups(ctx context.Cont
 		if err != nil {
 			return nil, err
 		}
-		return []core.StringToken{sg.GroupID()}, nil
+		sgs := []core.StringToken{sg.GroupID()}
+
```
M00nF1sh, line 259: 1. i'm wondering whether we should make IngressSuffixManageSecurityGroupRules generic for auto-generated SG case as well. e.g. If it's explicitly specified as false, then we still generate the frontend SG but don't generate backendSG and don't manage the worker node SG rules. 
2. it would be good to make the logic of handling backendSG for both `auto-generate frontend SG` and `manually specified frontend SG` in a common generic place here.
3. it would be good to handle the knob of whether enable backendSG feature here instead of inside backendSGProvider. i.e. we store the backendSG into the modelBuildTask, for 'auto-generated frontend SG` case, it will be either the managed SG or the backendSG depends on whether backendSG feature is enabled or not. Instead of handling it in the TGB model build task. (this is more flexible by leaking less implementation detail to TGB model build tasks)

pkg/ingress/model_build_target_group.go
```diff
@@ -74,21 +75,31 @@ func (t *defaultModelBuildTask) buildTargetGroupBindingSpec(ctx context.Context,
 	}
 }
 
-func (t *defaultModelBuildTask) buildTargetGroupBindingNetworking(_ context.Context) *elbv2model.TargetGroupBindingNetworking {
-	if t.managedSG == nil {
+func (t *defaultModelBuildTask) buildTargetGroupBindingNetworking(ctx context.Context) *elbv2model.TargetGroupBindingNetworking {
+	if !t.manageBackendSecurityGroupRules {
```
M00nF1sh, line 79: see comments above, i think the logic can be simplified if we only have a single 
t.backendSGIDTokens variable captured in the modelBuilder

pkg/networking/backend_sg_provider.go
```diff
@@ -0,0 +1,245 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
+	"sync"
+	"time"
+)
+
+const (
+	resourceTypeSecurityGroup = "security-group"
+	tagKeyK8sCluster          = "elbv2.k8s.aws/cluster"
+	tagKeySGType              = "elbv2.k8s.aws/type"
```
M00nF1sh, line 25: shall we use the `elbv2.k8s.aws/resource` (came from our resourceID of `ingress.k8s.aws/resource` and `service.k8s.aws/resource`

so maybe in the future, when we support multiple controllers in a cluster, the SG can be tagged like
```
elbv2.k8s.aws/cluster: cluster-name
elbv2.k8s.aws/stack: controller-deployment-specific-id
elbv2.k8s.aws/resource: controller-specific-resource-id(in our case , backend-sg)
```

pkg/networking/backend_sg_provider.go
```diff
@@ -0,0 +1,245 @@
...
+	tagValueBackend           = "backend"
```
M00nF1sh, line 26: maybe backend-sg?

pkg/networking/backend_sg_provider.go
```diff
@@ -0,0 +1,245 @@
...
+	Get(ctx context.Context) ([]string, error)
+	// Release cleans up the auto-generated backend SG if necessary
+	Release(ctx context.Context) error
+}
+
+// NewBackendSGProvider constructs a new  defaultBackendSGProvider
+func NewBackendSGProvider(clusterName string, backendSGEnabled bool, backendSGs []string, vpcID string,
+	ec2Client services.EC2, k8sClient client.Client, logger logr.Logger) *defaultBackendSGProvider {
+	var sanitizedBackendSGs []string
+	for _, sg := range backendSGs {
```
M00nF1sh, line 46: i prefer to validate the format strictly and early in https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/config/controller_config.go#L100
so we don't do the trimp and validate that if any of the specified sgIDs don't match the "sg-xxxxx" regex format, we fail.

pkg/networking/backend_sg_provider.go
```diff
@@ -0,0 +1,245 @@
...
+// BackendSGProvider is responsible for providing backend security groups
+type BackendSGProvider interface {
+	// Get returns the backend security groups to use
+	Get(ctx context.Context) ([]string, error)
+	// Release cleans up the auto-generated backend SG if necessary
+	Release(ctx context.Context) error
+}
+
+// NewBackendSGProvider constructs a new  defaultBackendSGProvider
+func NewBackendSGProvider(clusterName string, backendSGEnabled bool, backendSGs []string, vpcID string,
```
M00nF1sh, line 43: i prefer not to pass backendSGEnabled in, 
this component should solely generates a backendSG if required. 
Whether to call this component(decided by backendSGEnabled) should be handled in caller-side.

pkg/networking/backend_sg_provider.go
```diff
@@ -0,0 +1,245 @@
...
+
+func (p *defaultBackendSGProvider) allocateBackendSG(ctx context.Context) error {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+
+	if len(p.autoGeneratedSG) > 0 {
+		return nil
+	}
+
+	sgName := p.getBackendSGName()
```
M00nF1sh, line 109: let's use resource tags instead of name to identity existing resources in cluster. 

Name pattern might need to change given our past experience. (e.g. deprecate a unsafe hash algorithm)

pkg/networking/backend_sg_provider.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	p.logger.Info("deleted securityGroup", "ID", p.autoGeneratedSG)
+
+	p.autoGeneratedSG = ""
+	return nil
+}
+
+func (p *defaultBackendSGProvider) getBackendSGName() string {
+	sgNameHash := sha256.New()
+	_, _ = sgNameHash.Write([]byte(p.vpcID))
```
M00nF1sh, line 225: since we have can same sgName across multiple vpcs, seems use vpcID as hash component is unnecessary here.

pkg/networking/backend_sg_provider.go
```diff
@@ -0,0 +1,245 @@
...
+		return "", err
+	}
+	if len(sgs) > 0 {
+		return awssdk.StringValue(sgs[0].GroupId), nil
+	}
+	return "", nil
+}
+
+func (p *defaultBackendSGProvider) isBackendSGRequired(ctx context.Context) (bool, error) {
+	ingList := &networking.IngressList{}
```
M00nF1sh, line 179: personally i feel this depends on ingress implementation details too much.

also, seems the backendSG is not cleaned if there are some ingress in cluster even the left Ingresses don't need the backendSG.

shall we have a dedicated finalizer so that only ingresses/services needs backSG is added these finalizers?
if we have that, we can have a dedicated function to test whether ingress/service needs backendSG, and then add this dedicated finalizer and invoke backendSGProvider.Get inside controller, and pass obtained backendSG into modelBuilder.
and then call backendSGProvider.Release if any group member have this finalizer, and remove the finalizer if it succeeds.

pkg/networking/backend_sg_provider.go
kishorj: just the regex match sg-xxx, or actually query the EC2 configuration? I'm thinking about a less strict validation where we accept sg name as well, and validate only when the actual resource gets created. This gives the opportunity to create the sg resources later. Validation would make sense for the sg-xxx format since it is less likely the exact value to be created later.

pkg/networking/backend_sg_provider.go
kishorj: I was trying to avoid getting the tracking provider in this module - didn't remember if I ran into dependency issues earlier.

pkg/networking/backend_sg_provider.go
```diff
@@ -0,0 +1,245 @@
...
+	p.autoGeneratedSG = ""
+	return nil
+}
+
+func (p *defaultBackendSGProvider) getBackendSGName() string {
+	sgNameHash := sha256.New()
+	_, _ = sgNameHash.Write([]byte(p.vpcID))
+	_, _ = sgNameHash.Write([]byte(p.clusterName))
+	sgHash := hex.EncodeToString(sgNameHash.Sum(nil))
+	return fmt.Sprintf("k8s-%s-traffic-%.15s", p.clusterName, sgHash)
```
M00nF1sh, line 228: cluster name should be capped as well

pkg/config/controller_config.go
kishorj: updated in the new commit.

pkg/config/controller_config.go
```diff
@@ -141,3 +159,13 @@ func (cfg *ControllerConfig) validateExternalManagedTagsCollisionWithDefaultTags
 	}
 	return nil
 }
+
+func (cfg *ControllerConfig) validateBackendSecurityGroupConfiguration() error {
+	if len(cfg.BackendSecurityGroup) == 0 {
+		return nil
+	}
+	if strings.HasPrefix(cfg.BackendSecurityGroup, "sg-") {
```
M00nF1sh, line 167: should this be !strings.HasPrefix(cfg.BackendSecurityGroup, "sg-")

pkg/config/controller_config.go
```diff
@@ -19,10 +20,13 @@ const (
 	flagTargetGroupBindingMaxConcurrentReconciles    = "targetgroupbinding-max-concurrent-reconciles"
 	flagTargetGroupBindingMaxExponentialBackoffDelay = "targetgroupbinding-max-exponential-backoff-delay"
 	flagDefaultSSLPolicy                             = "default-ssl-policy"
+	flagEnableBackendSG                              = "enable-backend-security-group"
+	flagBackendSecurityGroups                        = "backend-security-groups"
```
M00nF1sh, line 24: should this be "backend-security-group"

pkg/networking/backend_sg_provider.go
kishorj: updated the ingress model builder to rely on the enabledBackendSG configuration.

Since there will be multiple modules relying on this configuration, I feel it is cleaner to have the backend SG provider abstract out the configuration as well. In case feature is disabled, the return value would be empty, and callers can take further action based on it. For now, the backend SG provider doesn't take in the configuration - it is up to the caller to decide.

pkg/networking/backend_sg_provider.go
kishorj: done

pkg/config/controller_config.go
kishorj: Yes. Thanks.

pkg/ingress/model_build_load_balancer.go
```diff
@@ -242,6 +242,58 @@ func (t *defaultModelBuildTask) buildLoadBalancerSubnetMappings(ctx context.Cont
...
+				return nil, errors.New("backendSG feature is required to manage worker node SG rules when frontendSG manually specified")
+			}
+			backendSGID, err := t.backendSGProvider.Get(ctx)
+			if err != nil {
+				return nil, err
+			}
+			t.backendSGIDToken = core.LiteralStringToken(backendSGID)
+			lbSGTokens = append(lbSGTokens, t.backendSGIDToken)
+		}
+		t.logger.Info("SG configured via annotation", "LB SGs", lbSGTokens, "backend SG", t.backendSGIDToken)
```
M00nF1sh, line 291: wondering what will be the value of lbSGTokens in printed format if a token is been used..

pkg/networking/backend_sg_provider.go
kishorj: using the following tags for now

```
elbv2.k8s.aws/cluster: cluster-name
elbv2.k8s.aws/resource: controller-specific-resource-id(in our case , backend-sg)
```

Will add the stack tag once we add support for multi controller setups.

pkg/config/controller_config.go
kishorj: oops. updated in the new rev.

pkg/ingress/model_build_load_balancer.go
kishorj: Similar log for auto created SG
```
{"level":"info","ts":1631823355.1664798,"logger":"controllers.ingress","msg":"Auto create SG","SGs":[{"$ref":"#/resources/AWS::EC2::SecurityGroup/ManagedLBSecurityGroup/status/groupID"},"sg-00xxx"],"backend SGs":["sg-00xxx"]}
```

Issue #2139, opened 2021-07-26
Acquire Region from Env


nairb774:
**Is your feature request related to a problem?**

I'm trying to reduce the IMDS hop limit from 2 to 1, and this resulted in the aws-load-balancer-controller erroring out with `failed to introspect region from EC2Metadata, specify --aws-region instead if EC2Metadata is unavailable: EC2MetadataRequestError: failed to get EC2 instance identity document\ncaused by: EC2MetadataError: failed to make EC2Metadata request\n\tstatus code: 401, request id: \ncaused by:`. There are other ways to determine the region which don't require manual configuration.

**Describe the solution you'd like**
A description of what you want to happen.

It would seem that the controller should try picking this information up from the environment. The `AWS_REGION` env var is automatically set by the [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) tooling, and this would be a natural replacement for using the IMDS endpoints.

**Describe alternatives you've considered**
A description of any alternative solutions or features you've considered.

I could set the flag, but it seems redundant when the controller already needs a role for manipulating various AWS objects, and thus the AWS_REGION env var is being set correctly.

M00nF1sh:
/kind good-first-issue
/kind feature
@nairb774 I think this is a good feature request to add. BTW, after we implemented above, you'll still need the `--aws-vpc-id` to specify the vpc ID.

Shreya027:
The issue  has been addressed in #2217, so closing the issue

PR #2217, opened 2021-09-07
Add code for acquiring region from env; Refactor code to reduce EC2 c…

Shreya027:
### Issue

[#2139](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2139)

### Description
When deploying the  ALB controller to EC2 nodes, we end up getting Region and Vpc Id from EC2  metadata service (IMDS). In case of Fargate however, we need to specify --set region=region-code and --set vpcId=vpc-xxxxxxxx flags during the deployment. The controller code currently makes an IMDS call irrespective of the scenario  without checking the cloud configuration set through flags. The code has two main changes listed below:

1. The code is refactored to only make an IMDS call if required, thus reducing the number of EC2 calls per instance. It makes sure to check flags set in cloud configurations or environment variables before making the call.

2. The region value is acquired from the environment variable "AWS_DEFAULT_REGION" or "AWS_REGION" if it is not obtained through the configuration flags. 

### Research
My initial attempt involved trying to eliminate IMDS calls completely and also retrieve cluster name through node data instead of in the deployment parameters. Currently for non Fargate case, both Vpc Id and Region are acquired from  configuration flags and if not specified in flags, it is got through  EC2 metadata. For Fargate case, Vpc Id and Region are acquired only from the configuration flags. Cluster name is retrieved from the parameters specified during deployment of the controller for both cases. I was able to eliminate the need for IMDS call for region by retrieving it through the environment variables set by  IAM roles for service accounts tooling. However, I wasn't able to find a solution to eliminate the need for IMDS call for Vpc Id in case it is not specified in the flags. Also, I wasn't able to find a way to eliminate the need for specifying cluster name in the deployment parameters. 

In case of Vpc Id, it can be found as a field in the Instance struct, whereas Cluster Name can be found as a field in the Node struct. The following line of code in main.go 

`podInfoRepo := k8s.NewDefaultPodInfoRepo(clientSet.CoreV1().RESTClient(), rtOpts.Namespace, ctrl.Log)`

returns podInfoRepo which in turn can be used to get podInfo using Get() and ListKeys() handlers. The podInfo retrieved can be used to get NodeName which is present as a field in the struct. The NodeName fetched can be used to retrieve the node using the k8s client. The cluster name then can be got straight forward from the node struct retrieved. As far as fetching the Vpc Id is concerned, it too can be retrieved using the FetchNodeInstances() function which in turn consequently retrieves instances that have Vpc Id present in the struct. 

However, for a lot of the above functions, a context need to be passed which is initialized as follows 

`ctx := ctrl.SetupSignalHandler()`

and is only done once ingGroupReconciler, svcReconciler and tgbReconciler are initialized using Vpc Id, cluster name and region bringing us back to our original problem. This seems to me like a case of chicken and egg problem. 

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/aws/cloud.go
+++ b/pkg/aws/cloud.go
@@ -43,10 +43,9 @@ type Cloud interface {
 
 // NewCloud constructs new Cloud implementation.
 func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud, error) {
-	metadata := (services.EC2Metadata)(nil)
+	metadataSess := session.Must(session.NewSession(aws.NewConfig()))
+	metadata := services.NewEC2Metadata(metadataSess)
 	if len(cfg.VpcID) == 0 {
-		metadataSess := session.Must(session.NewSession(aws.NewConfig()))
-		metadata = services.NewEC2Metadata(metadataSess)
 		vpcId, err := metadata.VpcID()
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to introspect vpcID from EC2Metadata, specify --aws-vpc-id instead if EC2Metadata is unavailable")
@@ -61,10 +60,6 @@ func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud, 
 		}
 
 		if region == ""{
-			if metadata  == nil {
-				metadataSess := session.Must(session.NewSession(aws.NewConfig()))
-				metadata = services.NewEC2Metadata(metadataSess)
-			}
 			err := (error)(nil)
 			region, err = metadata.Region()
 			if err != nil {
--- a/pkg/aws/cloud.go
+++ b/pkg/aws/cloud.go
@@ -43,10 +43,10 @@ type Cloud interface {
 
 // NewCloud constructs new Cloud implementation.
 func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud, error) {
-
+	metadata := (services.EC2Metadata)(nil)
 	if len(cfg.VpcID) == 0 {
 		metadataSess := session.Must(session.NewSession(aws.NewConfig()))
-		metadata := services.NewEC2Metadata(metadataSess)
+		metadata = services.NewEC2Metadata(metadataSess)
 		vpcId, err := metadata.VpcID()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/aws/cloud.go
```diff
@@ -42,24 +43,25 @@ type Cloud interface {
...
+		metadata := services.NewEC2Metadata(metadataSess)
 		vpcId, err := metadata.VpcID()
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to introspect vpcID from EC2Metadata, specify --aws-vpc-id instead if EC2Metadata is unavailable")
 		}
 		cfg.VpcID = vpcId
 	}
 
+	if len(cfg.Region) == 0 {
+		region := os.Getenv("AWS_DEFAULT_REGION")
```
kishorj, line 58: These env variables are set only if the IRSA is configured. We still need metadata introspection for the region configuration as a fallback.

pkg/aws/cloud.go
```diff
@@ -42,24 +43,37 @@ type Cloud interface {
...
-		region, err := metadata.Region()
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to introspect region from EC2Metadata, specify --aws-region instead if EC2Metadata is unavailable")
-		}
-		cfg.Region = region
-	}
-
+	metadata := (services.EC2Metadata)(nil)
 	if len(cfg.VpcID) == 0 {
+		metadataSess := session.Must(session.NewSession(aws.NewConfig()))
```
kishorj, line 48: This code path is very likely, so  I don't see any benefit of moving these two variables inside the if condition.

pkg/aws/cloud.go
```diff
@@ -42,24 +43,37 @@ type Cloud interface {
 
 // NewCloud constructs new Cloud implementation.
 func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud, error) {
-	metadataSess := session.Must(session.NewSession(aws.NewConfig()))
-	metadata := services.NewEC2Metadata(metadataSess)
-	if len(cfg.Region) == 0 {
```
kishorj, line 47: why not modify this block instead?

Shreya027: I actually added the IMDS call inside the if to reduce the number of IMDS calls per instance. So I am making the call only if it is required.

PR #2235, opened 2021-09-20
Helm chart: Add app.kubernetes.io/component label

imgx64:
### Issue

Fixes aws/eks-charts#518

### Description

The [non-helm install][1] and the [documentation][2] add the
`app.kubernetes.io/component: controller` label to the Deployment,
Service, and ServiceAccount.
The helm chart does not, which makes the ServiceAccount non-functional.
This adds the label and fixes it.

[1]: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/v2.2.1/docs/install/v2_2_1_full.yaml
[2]: https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html

### Checklist
- [ ] Manually tested

### DIFF
--- a/helm/aws-load-balancer-controller/Chart.yaml
+++ b/helm/aws-load-balancer-controller/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: aws-load-balancer-controller
 description: AWS Load Balancer Controller Helm chart for Kubernetes
-version: 1.2.8
+version: 1.2.9
 appVersion: v2.2.4
 home: https://github.com/aws/eks-charts
 icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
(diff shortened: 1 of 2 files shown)


imgx64:
CLA signed.

kishorj:
@imgx64, it is not clear why the label mismatch would cause ServiceAccount non-functional.
- if serviceAccount.create is true, helm will use matching labels on the sa resources
- if you create SA externally, since you'd want to use IRSA, deployment would still be functional, are there any specific errors?

imgx64:
When I tried it before, I had permission errors from the `aws-load-balancer-controller` deployment and it was unable to create the ELB. Manually creating the service account with this label fixed it. I also found the aws/eks-charts#518 issue which confirmed that other people were having the same problem.

However, when I tried it again now on a fresh EKS cluster to grab the exact error message, it worked correctly with the helm-created service account. I'm not sure what changed, or if it was my own mistake.

Anyway it looks like this is not needed after all. Sorry to waste your time.

Issue #1993, opened 2021-05-08
Feature Request: restrict ingress ports for node SG


flyguybravo:
I expose several nodeport services from a single ALB. Currently on deployment of the ingress object, when the Security Group of my nodes is modified by the controller it adds a new ingress rule for port range 0 - 65535 to permit traffic from the SG just created for the ALB. 

It would be very helpful to be able to specify via annotation a min/max value for this rule, or have it default to only those ports in Kubernetes default nodeport range (30000 - 32767).

kishorj:
v2.3.1 release has the feature to restrict SG rules to the the port ranges required for ingress access. Closing the issue.

PR #2236, opened 2021-09-20
Add port range restriction for SG

oliviassss:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1993

### Description

Dynamically adjust the port range for SG of ALB based on the minimum and maximum values for target and health check ports seen by the controller. This will make the feature work without getting input from the end user. And this PR will not affect the port range for NLB side.

#### TGB model changes

The target group model builder generates networking rules with specific ports instead of the current nil ports. If the health check ports are different from traffic ports, additional rules get added to the TGB spec. For IP targets, the targetPort is used, for instance targets, nodeport gets used. For example, if there are two TGBs targetting node ports 31223 and 32331, backend SG sg-backend the consolidated networking rules are as follows

```
from sg-backend, TCP, port 31223
from sg-backend, TCP, port 32331
```

#### Networking manager

Networking manager will take the consolidated rules as input and calculate the optimized list of rules using the port ranges. The rules are grouped by the protocol and the source. Then for each matching source and protocol, calculate the min and max ports and use a port range based rule. In the above example, the optimized rule is as follows

`from sg-backend, TCP, port range 31223 - 32331`

In case a TGB rule doesn’t have the port specified, use 0 as the min and 65535 as the max for the affected (source, protocol).

#### Command line flags

Provide a command line flag `--disable-restricted-sg-rules` if set to true, revert to the existing behavior of using unrestricted SG rules. The default value for this flag is false.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ingress/model_build_target_group.go
+++ b/pkg/ingress/model_build_target_group.go
@@ -79,6 +79,27 @@ func (t *defaultModelBuildTask) buildTargetGroupBindingNetworking(ctx context.Co
 		return nil
 	}
 	protocolTCP := elbv2api.NetworkingProtocolTCP
+	if t.disableRestrictedSGRules {
+		return &elbv2model.TargetGroupBindingNetworking{
+			Ingress: []elbv2model.NetworkingIngressRule{
+				{
+					From: []elbv2model.NetworkingPeer{
+						{
+							SecurityGroup: &elbv2model.SecurityGroup{
+								GroupID: t.backendSGIDToken,
+							},
+						},
+					},
+					Ports: []elbv2api.NetworkingPort{
+						{
+							Protocol: &protocolTCP,
+							Port:     nil,
+						},
+					},
+				},
+			},
+		}
+	}
 	var networkingPorts []elbv2api.NetworkingPort
 	var networkingRules []elbv2model.NetworkingIngressRule
 	tgPort := intstr.FromInt(int(targetGroupPort))
--- a/pkg/ingress/model_build_target_group.go
+++ b/pkg/ingress/model_build_target_group.go
@@ -53,7 +53,7 @@ func (t *defaultModelBuildTask) buildTargetGroupBinding(ctx context.Context, tg 
 
 func (t *defaultModelBuildTask) buildTargetGroupBindingSpec(ctx context.Context, tg *elbv2model.TargetGroup, svc *corev1.Service, port intstr.IntOrString, nodeSelector *metav1.LabelSelector) elbv2model.TargetGroupBindingResourceSpec {
 	targetType := elbv2api.TargetType(tg.Spec.TargetType)
-	tgbNetworking := t.buildTargetGroupBindingNetworking(ctx)
+	tgbNetworking := t.buildTargetGroupBindingNetworking(ctx, tg.Spec.Port, *tg.Spec.HealthCheckConfig.Port)
 	return elbv2model.TargetGroupBindingResourceSpec{
 		Template: elbv2model.TargetGroupBindingTemplate{
 			ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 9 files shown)


pkg/ingress/model_build_target_group.go
```diff
@@ -74,29 +74,38 @@ func (t *defaultModelBuildTask) buildTargetGroupBindingSpec(ctx context.Context,
 	}
 }
 
-func (t *defaultModelBuildTask) buildTargetGroupBindingNetworking(ctx context.Context) *elbv2model.TargetGroupBindingNetworking {
+func (t *defaultModelBuildTask) buildTargetGroupBindingNetworking(ctx context.Context, targetGroupPort int64, healthCheckPort intstr.IntOrString) *elbv2model.TargetGroupBindingNetworking {
```
kishorj, line 77: Lets make the model builder revert to prior behavior if `disable-restricted-sg-rules` is set to true so that users can get existing behavior in case.

pkg/targetgroupbinding/networking_manager_test.go
```diff
@@ -854,7 +854,154 @@ func Test_defaultNetworkingManager_computeAggregatedIngressPermissionsPerSG(t *t
...
+func Test_defaultNetworkingManager_computeRestrictedIngressPermissionsPerSG(t *testing.T) {
+	type fields struct {
+		ingressPermissionsPerSGByTGB map[types.NamespacedName]map[string][]networking.IPPermissionInfo
+	}
+	var tests = []struct {
+		name   string
+		fields fields
+		want   map[string][]networking.IPPermissionInfo
+	}{
+		{
```
kishorj, line 872: Also add tests for CIDRs

pkg/targetgroupbinding/networking_manager_test.go
```diff
@@ -995,6 +995,214 @@ func Test_defaultNetworkingManager_computeRestrictedIngressPermissionsPerSG(t *t
...
+			name: "test for CIDRs",
+			fields: fields{
+				ingressPermissionsPerSGByTGB: map[types.NamespacedName]map[string][]networking.IPPermissionInfo{
+					types.NamespacedName{Namespace: "ns-1", Name: "tgb-1"}: {
+						"sg-a": {
+							{
+								Permission: ec2sdk.IpPermission{
+									IpProtocol: awssdk.String("tcp"),
+									FromPort:   awssdk.Int64(80),
```
kishorj, line 1007: for CIDR based rules, the exact port values get used, could you modify the UTs as well so we can verify that the signle ports don't get merged into ranges.

pkg/ingress/model_build_target_group.go
oliviassss: Changed.

pkg/targetgroupbinding/networking_manager_test.go
oliviassss: Added.

pkg/targetgroupbinding/networking_manager_test.go
oliviassss: Modified. I also manually tested the CIDR case on EKS cluster, the single ports used did not get merged into ranges.

pkg/targetgroupbinding/networking_manager.go
```diff
@@ -225,7 +230,103 @@ func (m *defaultNetworkingManager) consolidateIngressPermissionsPerSGByTGB(_ con
...
+						minPort = *perm.Permission.FromPort
+					}
+					if awssdk.Int64Value(perm.Permission.ToPort) > maxPort {
+						maxPort = *perm.Permission.ToPort
+					}
+				}
+				if minPort > maxPort {
+					minPort, maxPort = defaultTgbMinPort, defaultTgbMaxPort
+				}
+				*optimizedPermByProtocolAndSourcePerSG[sgID][protocol][groupID].Permission.FromPort = *awssdk.Int64(minPort)
```
kishorj, line 306: imo, is is simpler to track the rules in a map[string][]networking.IPPermissionInfo? key is the sgID and the value is the list of permissions once we calculate the min and max ports.

pkg/targetgroupbinding/networking_manager.go
```diff
@@ -225,7 +230,103 @@ func (m *defaultNetworkingManager) consolidateIngressPermissionsPerSGByTGB(_ con
...
-func (m *defaultNetworkingManager) computeAggregatedIngressPermissionsPerSG(_ context.Context) map[string][]networking.IPPermissionInfo {
+func (m *defaultNetworkingManager) computeAggregatedIngressPermissionsPerSG(ctx context.Context) map[string][]networking.IPPermissionInfo {
+	if m.disableRestrictedSGRules {
+		return m.computeUnrestrictedIngressPermissionsPerSG(ctx)
+	}
+	return m.computeRestrictedIngressPermissionsPerSG(ctx)
+}
+
+func (m *defaultNetworkingManager) groupIngressPermsBySourceAndProtocolPerSG(_ context.Context) (map[string][]networking.IPPermissionInfo, map[string]map[string]map[string][]networking.IPPermissionInfo) {
+	permsFromNLBPerSG := make(map[string][]networking.IPPermissionInfo)
```
kishorj, line 241: lets rename to permsFromIPRangeRulesPerSG or something similar

oliviassss: Done.

pkg/targetgroupbinding/networking_manager.go
oliviassss: Done.

pkg/targetgroupbinding/networking_manager.go
```diff
@@ -225,7 +230,93 @@ func (m *defaultNetworkingManager) consolidateIngressPermissionsPerSGByTGB(_ con
...
+
+	restrictedPermByProtocolPerSG := make(map[string][]networking.IPPermissionInfo)
+	for sgID, permsByProtocolAndSource := range permsByProtocolAndSourcePerSG {
+		if _, ok := restrictedPermByProtocolPerSG[sgID]; !ok {
+			restrictedPermByProtocolPerSG[sgID] = []networking.IPPermissionInfo{}
+		}
+		for _, permsBySource := range permsByProtocolAndSource {
+			for _, perms := range permsBySource {
+				minPort, maxPort := defaultTgbMaxPort, defaultTgbMinPort
+				if len(perms) > 0 {
```
kishorj, line 286: When is this false? If this is false, then we will end up modifying the previous rule in line 300!
Lets do this instead  -
if len(perms) == 0 {
  continue
}

pkg/targetgroupbinding/networking_manager.go
```diff
@@ -225,7 +230,93 @@ func (m *defaultNetworkingManager) consolidateIngressPermissionsPerSGByTGB(_ con
...
+					restrictedPermByProtocolPerSG[sgID] = append(restrictedPermByProtocolPerSG[sgID], perms[0])
```
kishorj, line 287: We can construct a new Permission object based on perms[0], minPort, maxPort and append to the list at the end of this loop.

kishorj (approved):

peterbosalliandercom:
@oliviassss @kishorj - Is it possible to come in contact about the usecase of this change and why this is approved? We have a problem with the port range coming on our ALB when setting two ports for example 8444 and 9555 (healthcheck) Which becomes a port range in the security group itself. We think this is a major security risk on our loadbalancer which is public available. Can you explain the ratio behind this solution? There is an open issue about that also: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3460

PR #2245, opened 2021-09-24
Allows SG port range restriction for NLB services

dickeyf:
### Issue

I did not open an issue in this repository.

This PR is for solving Security Rules quota limits which can easily get exceeded because each NodePort gets its own rule added.  Our project uses lots of instance-based NLBs and this means that have many NodePorts open, each requiring one rule.

### Description

I based my PR on PR #2236

PR 2236 solved the exact problem we had by merging the rules into a single rule for a port range going from the smallest to the biggest.  As a result PR-2236 allows a single rule cover a port range which dynamically adjusts as needed.

However PR 2236 was limited to only the Security Rules whose source are a Security group (Rules for the Ingresses ALBs).  
The NLBs creates rules whose source are a IPv4 or IPV6 CIDR and rules with such sources will be ignore by PR 2236 changes.
Hence 2236 helps reduce the number of rules needed for Ingresses but not for NLBs.  PR-2236 intentionally avoids changing the behaviour for NLBs, so I made my change not take effect by default.

When my flag "enable-nlb-restricted-sg-rules" is enabled, I made that code also coalesce the NLB Security Rules into ranges.  This way, when enabled, the NLBs node ports will be covered by a single rule for a port range.

A new Flag is added to the controller "--enable-nlb-restricted-sg-rules".  So adding "--enable-nlb-restricted-sg-rules=true" the to controller's command will enable this functionality.  I also added it to the Helm chart's parameter: "enableNLBRestrictedSGRules".  The controller will not have this functionality enabled by default.

### Checklist
- [X] Manually tested


kishorj:
@dickeyf, thank you for the contribution. When the controller adds the IP/CIDR based rules to the worker node SG for NLB, direct traffic from those IP ranges also get allowed as a side effect. With optimization, we risk opening a wide range of ports on the compute instances. For example, if there is a targetPort 80 required for IP target, and 31223 required for NodePort, and the source range is 0.0.0.0/0, then with the optimization, controller would end up opening port range 80-31223 for 0.0.0.0/0. Since there are potential security consequences, we deliberately choose not to optimize the IP CIDR based rules.

dickeyf:
Would it be acceptable if I limit this optimization for NLB only when the source CIDR range is not 0.0.0.0/0, and that ports are between the node port range ?

dickeyf:
Closing this PR.  This approach would effectively make loadBalancerSourceRanges useless, as I could override another service's `loadBalancerSourceRanges` by creating a IP NLB with ports 80 to 65535, with `loadBalancerSourceRanges` of 0.0.0.0/0 as mentioned by @kishorj.

I will open an issue to track the problem I want to tackle, which is to increase the number of instance-type NLBs ports a EKS cluster can have, which is currently limited by the SG rule limit.

PR #2246, opened 2021-09-24
e2e test helper script

cgchinmay:
### Issue
N/A

### Description
Added helper script to deploy load balancer controller and trigger e2e tests against a given cluster.
Needed script to run tests on prow 

### Checklist
Manually tested


test/e2e/run.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
+
+# This script runs e2e tests on the AWS Load Balancer Controller
+
+set -e
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+echo "Running AppMeshController integration test with the following variables
```
kishorj, line 8: "Load Balancer Controller" :)

test/e2e/run.sh
```diff
@@ -0,0 +1,92 @@
...
+KUBE CONFIG: $KUBE_CONFIG_PATH
```
kishorj, line 9: any reason to not use KUBECONFIG?

cgchinmay: We do specify $KUBE_CONFIG_PATH when triggering ginkgo tests . This KUBE CONFIG is only for logging information.

test/e2e/run-canary-test.sh
```diff
@@ -0,0 +1,77 @@
...
+else
+  CLUSTER_INFO=$($GET_CLUSTER_INFO_CMD --endpoint $ENDPOINT)
+fi
+
+VPC_ID=$(echo $CLUSTER_INFO | jq -r '.cluster.resourcesVpcConfig.vpcId')
+ACCOUNT_ID=$(aws sts get-caller-identity | jq -r '.Account')
+
+echo "VPC ID: $VPC_ID"
+
+eksctl utils associate-iam-oidc-provider \
```
M00nF1sh, line 31: can this be done in the centralized image by default?

test/e2e/run-canary-test.sh
```diff
@@ -0,0 +1,77 @@
...
+
+echo "Update helm repo eks"
+helm repo add eks https://aws.github.io/eks-charts
+
+helm repo update
+
+echo "Install TargetGroupBinding CRDs"
+kubectl apply -k "github.com/aws/eks-charts/stable/aws-load-balancer-controller//crds?ref=master"
+
+echo "Install aws-load-balacner-controller"
```
M00nF1sh, line 60: s/balacner/balancer/

M00nF1sh (approved):
/lgtm

test/e2e/run-canary-test.sh
cgchinmay: We have similar script for other components as well and those scripts are reused to run tests, we can keep this in default image but then other component scripts will have to be modified as well. So kept this code here to have consistent pattern.

Issue #2303, opened 2021-10-19
Internal NLB - EKS worker node security group does not allow traffic from 0.0.0.0/0


dseravalli:
**Describe the bug**
When creating an internal NLB, the security group for the EKS worker nodes is modified to allow ingress to the NLB traffic port from the VPC CIDR, but not from `0.0.0.0/0`, as [the documentation](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/annotations/#access-control) says this is the default when source CIDRs are not specified.

**Steps to reproduce**
Create a `Service` of type `LoadBalancer` with at least these annotations:

```
service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: instance
service.beta.kubernetes.io/aws-load-balancer-scheme: internal
service.beta.kubernetes.io/aws-load-balancer-type: external
 ```

**Expected outcome**

One of:
- The security group of the EKS worker nodes would be modified to include an ingress rule to the NLB's traffic port from source `0.0.0.0/0`
- Or, the documentation explains that when using an internal NLB, it is expected that the worker node SG will be modified to only allow traffic from the VPC CIDR by default.

**Environment**

* aws-load-balancer-controller v2.2.3
* EKS v1.21

kishorj:
Thanks. The targetgroupbinding resource has rule for CIDR "10.0.0.0/16", not 0.0.0.0/0. Could you confirm whether you have the `service.beta.kubernetes.io/load-balancer-source-ranges` annotation or `Spec.LoadBalancerSourceRanges` configured on your service.

dseravalli:
@kishorj I can confirm that we have neither of those set.

PR #2306, opened 2021-10-19
Update internal nlb service access control docs

dseravalli:
Ref. #2303 
Updates docs to explain that VPC CIDR is used for `load-balancer-source-ranges` by default when NLB is internal

/sig docs
/kind documentation
/assign @kishorj


dseravalli:
@kishorj CLA signed

PR #2324, opened 2021-10-27
 implement dnsPolicy in aws-load-balancer-controller chart

notjames:
Signed-off-by: Jim Conner <snafu.x@gmail.com>

### Issue

N/A

### Description

Implement `dnsPolicy` in `aws-load-balancer-controller`

### Testing

* instantiated `kind` cluster (vanilla 1.20.2)
* ran: `helm template aws-load-balancer-controller aws-load-balancer-controller --set  clusterName=jimcluster`
  * verified deployment showed `dnsPolicy` as `Default` as `values.yaml` file is set as `Default`
* ran: `helm template aws-load-balancer-controller aws-load-balancer-controller --set  clusterName=jimcluster --set dnsPolicy=aslakjsdfasd`
  * Verified deployment showed `dnsPolicy: aslakjsdfasd`
* removed `dnsPolicy` from `values.yaml` then ran: `helm template aws-load-balancer-controller aws-load-balancer-controller --set  clusterName=jimcluster`
  * Verified `dnsPolicy` is not in deployment (though the template does use `default` but that's just redundant goodness; the output should still show no `dnsPolicy` as the pod will use whatever the cluster dictates should be used by default).

### Checklist
- [X] Manually tested


notjames:
💢 when I try to rebase this I get conflicts which have nothing to do with my change. This kind of git crap is so frustrating.

helm/aws-load-balancer-controller/values.yaml
```diff
@@ -184,6 +184,13 @@ env:
 # recommended if using the Amazon VPC CNI plugin.
 hostNetwork: false
 
+# Specifies the dnsPolicy that should be used for pods in the deployment
+#
+# This may need to be used to be changed given certain conditions. For instance, if one uses the cilium CNI
+# with certain settings, one may need to set `hostNetwork: true` and webhooks won't work unless `dnsPolicy`
+# is set to `ClusterFirstWithHostNet`. See https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy
+dnsPolicy: Default
```
kishorj, line 192: why not leave the default value as blank and let k8s use the appropriate default?

notjames: Sounds fair. Let's try that. Otherwise, I was mistaken and we probably shouldn't use "Default" but instead It would actually be better to set it to what the API will use by default which is `ClusterFirst`. We'll go with blank for now and see how that goes.

notjames:
ping

notjames:
anyone know how long it might be before this gets reviewed/merged?

helm/aws-load-balancer-controller/README.md
```diff
@@ -202,6 +202,7 @@ The default values set by the application itself can be confirmed [here](https:/
 | `livenessProbe`                             | Liveness probe settings for the controller                                                               | (see `values.yaml`)                                                                |
 | `env`                                       | Environment variables to set for aws-load-balancer-controller pod                                        | None                                                                               |
 | `hostNetwork`                               | If `true`, use hostNetwork                                                                               | `false`                                                                            |
+| `dnsPolicy`                                 | Set dnsPolicy if required                                                                                | `Default`                                                                          |
```
M00nF1sh, line 205: "Default" is not the Default DNS policy, instead `ClusterFirst` is

helm/aws-load-balancer-controller/templates/deployment.yaml
```diff
@@ -46,6 +46,9 @@ spec:
       {{- if .Values.hostNetwork }}
       hostNetwork: true
       {{- end }}
+      {{- with .Values.dnsPolicy }}
```
M00nF1sh, line 49: i think it's better to use if to avoid renderDNSPolicy at all if dnsPolicy is not set
```
{{- if .Values. dnsPolicy }}
dnsPolicy: {{ .Values.dnsPolicy }}
{{- end }}
```

helm/aws-load-balancer-controller/README.md
notjames: Indeed. I'll fix this.

helm/aws-load-balancer-controller/templates/deployment.yaml
notjames: fixed.

helm/aws-load-balancer-controller/test.yaml
```diff
@@ -154,6 +154,13 @@ env:
 # recommended if using the Amazon VPC CNI plugin.
 hostNetwork: false
 
+# Specifies the dnsPolicy that should be used for pods in the deployment
+#
+# This may need to be used to be changed given certain conditions. For instance, if one uses the cilium CNI
+# with certain settings, one may need to set `hostNetwork: true` and webhooks won't work unless `dnsPolicy`
+# is set to `ClusterFirstWithHostNet`. See https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy
+dnsPolicy: Default
```
kishorj, line 162:
```suggestion
dnsPolicy:
```

kishorj (approved):

stevo-f3:
@notjames would it make sense, besides `dnsPolicy`, to also support configuring `dnsConfig` https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-dns-config, for completeness, when `dnsPolicy: None` is used?

notjames:
> @notjames would it make sense, besides `dnsPolicy`, to also support configuring `dnsConfig` https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-dns-config, for completeness, when `dnsPolicy: None` is used?

tbh, such a use-case wasn't in the scope of work I was performing for my own company at the time I submitted this PR. However, your suggestion is a good one I think, though I'm not sure how often people need to manually configure their containers' /etc/hosts IE dnsConfig. I've had to do it like once in my life, so there are times where it's desired. I can't prioritize the work now, but I don't mind doing the work in the near future.

stevo-f3:
Ticket for `dnsConfig` support https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4154

PR #2342, opened 2021-11-05
add feature gate to enable or disable listener rules tagging

oliviassss:
### Description

Add a feature gate as a command line flag to enable / disable features. It is a set of key=value pairs with default value true. Right now there is only 1 feature controlled by this feature gate, that is to enable / disable the tagging of listener and rules.

#### Test

- Tested with `--feature-gate=enable-listener-rules-tagging=true`, verified the the tags were created for listeners and rules
- Tested with `--feature-gate=enable-listener-rules-tagging=false`, verified when creating listeners and rules, the requests did not have tags related field; and the listener and rules did not call the describeResourceTags API.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/config/feature_gate.go
+++ b/pkg/config/feature_gate.go
@@ -10,7 +10,7 @@ import (
 type Feature string
 
 const (
-	EnableListenerRulesTagging Feature = "enable-listenerRules-tagging"
+	EnableListenerRulesTagging Feature = "enable-listener-rules-tagging"
 )
 
 type FeatureGate interface {
--- a/pkg/config/feature_gate.go
+++ b/pkg/config/feature_gate.go
@@ -1,0 +1,108 @@
+package config
+
+import (
+	"fmt"
+	"github.com/spf13/pflag"
+	"strconv"
+	"strings"
+)
+
+type Feature string
+
+const (
+	EnableListenerRulesTagging Feature = "enable-listenerRules-tagging"
+)
+
+type FeatureGate interface {
+	// Enabled returns whether a feature is enabled
+	Enabled(feature Feature) bool
+
+	// Enable will enable a feature
+	Enable(feature Feature)
+
+	// Disable will disable a feature
+	Disable(feature Feature)
+
+	// BindFlags bind featureGate flags
+	BindFlags(fs *pflag.FlagSet)
+}
+
+var _ FeatureGate = (*defaultFeatureGate)(nil)
+var _ pflag.Value = (*defaultFeatureGate)(nil)
+
+type defaultFeatureGate struct {
+	featureState map[Feature]bool
+}
+
+// NewFeatureGate constructs new featureGate
+func NewFeatureGate() FeatureGate {
+	return &defaultFeatureGate{
+		featureState: map[Feature]bool{
+			EnableListenerRulesTagging: true,
+		},
+	}
+}
+
+func (f *defaultFeatureGate) BindFlags(fs *pflag.FlagSet) {
+	fs.Var(f, "feature-gate", "A set of key=bool pairs enable/disable features")
+}
+
+func (f *defaultFeatureGate) Enabled(feature Feature) bool {
+	return f.featureState[feature]
+}
+
+func (f *defaultFeatureGate) Enable(feature Feature) {
+	f.featureState[feature] = true
+}
+
+func (f *defaultFeatureGate) Disable(feature Feature) {
+	f.featureState[feature] = false
+}
+
+func (f *defaultFeatureGate) String() string {
+	var featureSettings []string
+	for feature, enabled := range f.featureState {
+		featureSettings = append(featureSettings, fmt.Sprintf("%v=%v", feature, enabled))
+	}
+	return strings.Join(featureSettings, ",")
+}
+
+// SplitMapStringBool parse comma-separated string of key1=value1,key2=value2. value is either true or false
+func (f *defaultFeatureGate) SplitMapStringBool(str string) (map[string]bool, error) {
+	result := make(map[string]bool)
+	for _, s := range strings.Split(str, ",") {
+		if len(s) == 0 {
+			continue
+		}
+		parts := strings.SplitN(s, "=", 2)
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("invalid mapStringBool: %v", s)
+		}
+		k := strings.TrimSpace(parts[0])
+		v, err := strconv.ParseBool(strings.TrimSpace(parts[1]))
+		if err != nil {
+			return nil, fmt.Errorf("invalid mapStringBool: %v", s)
+		}
+		result[k] = v
+	}
+	return result, nil
+}
+
+func (f *defaultFeatureGate) Set(value string) error {
+	settings, err := f.SplitMapStringBool(value)
+	if err != nil {
+		return fmt.Errorf("failed to parse feature-gate settings due to %v", err)
+	}
+	for k, v := range settings {
+		_, ok := f.featureState[Feature(k)]
+		if !ok {
+			return fmt.Errorf("unknown feature: %v", k)
+		}
+		f.featureState[Feature(k)] = v
+	}
+	return nil
+}
+
+func (f *defaultFeatureGate) Type() string {
+	return "mapStringBool"
+}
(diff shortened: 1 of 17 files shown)


pkg/config/feature_gate.go
```diff
@@ -0,0 +1,108 @@
...
+	"fmt"
+	"github.com/spf13/pflag"
+	"strconv"
+	"strings"
+)
+
+type Feature string
+
+const (
+	EnableListenerRulesTagging Feature = "enable-listenerRules-tagging"
```
M00nF1sh, line 13: let's use either `-` or case to cases to separate words.
maybe `enable-listener-rule-tagging`

pkg/deploy/elbv2/listener_manager.go
```diff
@@ -10,13 +10,18 @@ import (
...
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/config"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/deploy/tracking"
 	elbv2equality "sigs.k8s.io/aws-load-balancer-controller/pkg/equality/elbv2"
 	elbv2model "sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/runtime"
 	"time"
 )
 
+const (
+	EnableListenerRulesTagging = "enable-listenerRules-tagging"
```
M00nF1sh, line 22: we should use the constant defined in config package instead of redefine here

pkg/deploy/elbv2/listener_manager.go
```diff
@@ -58,7 +65,10 @@ func (m *defaultListenerManager) Create(ctx context.Context, resLS *elbv2model.L
 	if err != nil {
 		return elbv2model.ListenerStatus{}, err
 	}
-	lsTags := m.trackingProvider.ResourceTags(resLS.Stack(), resLS, resLS.Spec.Tags)
+	lsTags := map[string]string{}
```
M00nF1sh, line 68: maybe `var lsTags  map[string]string` instead, which avoids the unnecessary map allocation

pkg/deploy/elbv2/listener_rule_manager.go
```diff
@@ -55,7 +58,10 @@ func (m *defaultListenerRuleManager) Create(ctx context.Context, resLR *elbv2mod
 	if err != nil {
 		return elbv2model.ListenerRuleStatus{}, err
 	}
-	ruleTags := m.trackingProvider.ResourceTags(resLR.Stack(), resLR, resLR.Spec.Tags)
+	ruleTags := map[string]string{}
```
M00nF1sh, line 61: maybe `var ruleTags map[string]string` instead to avoid unnecessary map allocation

pkg/deploy/elbv2/tagging_manager.go
```diff
@@ -183,9 +186,12 @@ func (m *defaultTaggingManager) ListListeners(ctx context.Context, lbARN string)
 		lsARNs = append(lsARNs, lsARN)
 		lsByARN[lsARN] = listener
 	}
-	tagsByARN, err := m.describeResourceTags(ctx, lsARNs)
-	if err != nil {
-		return nil, err
+	tagsByARN := map[string]map[string]string{}
```
M00nF1sh, line 189: maybe `var tagsByARN map[string]map[string]string` instead.

pkg/deploy/elbv2/tagging_manager.go
```diff
@@ -213,9 +219,12 @@ func (m *defaultTaggingManager) ListListenerRules(ctx context.Context, lsARN str
 		lrARNs = append(lrARNs, lrARN)
 		lrByARN[lrARN] = rule
 	}
-	tagsByARN, err := m.describeResourceTags(ctx, lrARNs)
-	if err != nil {
-		return nil, err
+	tagsByARN := map[string]map[string]string{}
```
M00nF1sh, line 222: maybe var tagsByARN map[string]map[string]string instead.

pkg/ingress/model_build_listener.go
```diff
@@ -19,6 +19,10 @@ import (
 	elbv2model "sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2"
 )
 
+const (
```
M00nF1sh, line 22: let's use the constant in config package directly

pkg/ingress/model_build_listener.go
```diff
@@ -34,9 +38,12 @@ func (t *defaultModelBuildTask) buildListenerSpec(ctx context.Context, lbARN cor
 	if err != nil {
 		return elbv2model.ListenerSpec{}, err
 	}
-	tags, err := t.buildListenerTags(ctx, ingList)
-	if err != nil {
-		return elbv2model.ListenerSpec{}, err
+	tags := map[string]string{}
```
M00nF1sh, line 41: maybe `var tags map[string]string` instead

pkg/ingress/model_build_listener.go
```diff
@@ -34,9 +38,12 @@ func (t *defaultModelBuildTask) buildListenerSpec(ctx context.Context, lbARN cor
...
+	if t.featureGate.Enabled(EnableListenerRulesTagging) {
+		tags, err = t.buildListenerTags(ctx, ingList)
```
M00nF1sh, line 43: actually, let's minimize the changes, we don't need to change the model_builder sections, and let it continue build tags as normal.

These tags built here won't be used in model_deployer anyway when EnableListenerRulesTagging is not enabled.

pkg/ingress/model_build_listener_rules.go
```diff
@@ -39,9 +39,12 @@ func (t *defaultModelBuildTask) buildListenerRules(ctx context.Context, lsARN co
 				if err != nil {
 					return errors.Wrapf(err, "ingress: %v", k8s.NamespacedName(ing.Ing))
 				}
-				tags, err := t.buildListenerRuleTags(ctx, ing)
-				if err != nil {
-					return errors.Wrapf(err, "ingress: %v", k8s.NamespacedName(ing.Ing))
+				tags := map[string]string{}
+				if t.featureGate.Enabled(EnableListenerRulesTagging) {
```
M00nF1sh, line 43: actually, let's minimize the changes, we don't need to change the model_builder sections, and let it continue build tags as normal.

These tags built here won't be used in model_deployer anyway when EnableListenerRulesTagging is not enabled.

pkg/ingress/model_builder.go
```diff
@@ -40,7 +41,7 @@ func NewDefaultModelBuilder(k8sClient client.Client, eventRecorder record.EventR
 	authConfigBuilder AuthConfigBuilder, enhancedBackendBuilder EnhancedBackendBuilder,
 	trackingProvider tracking.Provider, elbv2TaggingManager elbv2deploy.TaggingManager,
 	vpcID string, clusterName string, defaultTags map[string]string, externalManagedTags []string, defaultSSLPolicy string,
-	backendSGProvider networkingpkg.BackendSGProvider, enableBackendSG bool, disableRestrictedSGRules bool, logger logr.Logger) *defaultModelBuilder {
+	backendSGProvider networkingpkg.BackendSGProvider, enableBackendSG bool, disableRestrictedSGRules bool, featureGate config.FeatureGate, logger logr.Logger) *defaultModelBuilder {
```
M00nF1sh, line 44: we don't need to change model builder

pkg/service/model_build_listener.go
```diff
@@ -47,15 +51,19 @@ func (t *defaultModelBuildTask) buildListenerSpec(ctx context.Context, port core
...
-		return elbv2model.ListenerSpec{}, err
-	}
 	targetGroup, err := t.buildTargetGroup(ctx, port, tgProtocol, scheme)
 	if err != nil {
 		return elbv2model.ListenerSpec{}, err
 	}
 
+	tags := map[string]string{}
+	if t.featureGate.Enabled(EnableListenerRulesTagging) {
+		tags, err = t.buildListenerTags(ctx)
```
M00nF1sh, line 61: we don't need to change model builder

pkg/config/feature_gate.go
oliviassss: done.

pkg/deploy/elbv2/listener_manager.go
oliviassss: done.

pkg/deploy/elbv2/listener_manager.go
oliviassss: done.

pkg/ingress/model_build_listener.go
oliviassss: Done. Revoked the changes in model builder side. Thanks for pointing out!

M00nF1sh (approved):
/lgtm

PR #2360, opened 2021-11-15
change ingress version in docs under docs/examples

michael12312:
change ingress version in docs under docs/examples folder

### DIFF
--- a/docs/examples/cognito-ingress-template.yaml
+++ b/docs/examples/cognito-ingress-template.yaml
@@ -17,7 +17,7 @@
 #    - Related Kubernetes service/application
 #
 
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: <ingress-name>  # app.example.com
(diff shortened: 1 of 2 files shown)


M00nF1sh:
@michael12312
Thanks for making contributing.
We currently cannot merge this due to this controller only supports networking.k8s.io/v1beta1 instead of networking.k8s.io/v1 due to compatibility with older k8s versions.

working is in progress to support the v1 API.

metacyclic:
@M00nF1sh something is wrong here.  Trying to upgrade our ingresses after updating the controller from v1.1.3 -> 2.3.0, and following along with what is [here](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.3/guide/ingress/spec/), when I modify the ingress to be apiVersion extensions/v1, I get the following error
```
error: unable to recognize "ingress.yaml": no matches for kind "Ingress" in version "extensions/v1"
```
we're using Server Version: v1.21.2-eks-0389ca3.

Where can I find a valid example of a networking.k8s.io/v1beta1 apiVersion to use with the latest controller?

johngmyers:
Redundant with #2433 
/close

PR #2364, opened 2021-11-16
Update-the-ingress-version-to-networking.k8s.io/v1

michael12312:
Update-the-ingress-version-to-networking.k8s.io/v1

### DIFF
--- a/docs/deploy/configurations.md
+++ b/docs/deploy/configurations.md
@@ -32,7 +32,7 @@ spec:
 Now, only ingress resources with the appropriate annotation are picked up, as seen below.
 
 ```yaml
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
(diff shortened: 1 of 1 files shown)


M00nF1sh:
@michael12312 
Thanks for making contributing.
We currently cannot merge this due to this controller only supports `networking.k8s.io/v1beta1` instead of `networking.k8s.io/v1` due to compatibility with older k8s versions.

working is in progress to support the v1 API.

Kristjanf-droid:
> @michael12312 Thanks for making contributing. We currently cannot merge this due to this controller only supports `networking.k8s.io/v1beta1` instead of `networking.k8s.io/v1` due to compatibility with older k8s versions.
> 
> working is in progress to support the v1 API.

Any news on this? Was hoping this would be fixed in 2.3.1

johngmyers:
Redundant with #2433 
/close

Issue #2297, opened 2021-10-15
Support for NLBs targeting ALBs


gshutler:
A [few weeks ago](https://aws.amazon.com/blogs/networking-and-content-delivery/application-load-balancer-type-target-group-for-network-load-balancer/) it was announced that NLBs could now use and Application Load Balancer-type Target Group ([docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/application-load-balancer-target.html)).

We are currently using the [previously documented workaround involving Lambda](https://aws.amazon.com/blogs/networking-and-content-delivery/using-aws-lambda-to-enable-static-ip-addresses-for-application-load-balancers/) which this feature is intended to replace and would love to switch to the new method.

kurtbomya:
Oh this is critical for an architecture I'm working on!

wallhided:
bump

45cali:
This too is a feature I could use. create a nlb and define the target for a port to a alb. Is this still on the roadmap?

limoges:
Also have interest in this feature. It is a surprise when looking to configure the NLB that I can use other target types but not ALB.

philstevenson:
This is would be ideal for architectures using private link and API gateway VPC link, such as: https://aws.amazon.com/blogs/networking-and-content-delivery/application-load-balancer-type-target-group-for-network-load-balancer/

Currently looks like we will have create a shared NLB and target group + binding in terraform to achieve this.

NLB/TargetGroup CRDs or a dummy service to chain NLB -> ALB could be an option

oprinmarius:
We would also like to use this feature. Any updates ? Anyone actively working on this ? Issue is open for 900 days

eugenejen:
+1 for this feature to create NLB and set target group to an ALB.

scila1996:
+1 I must configure NLB either using Terraform or manually. It is not straightforward with Kubernetes.

PR #2379, opened 2021-11-22
Support for NLBs targeting ALBs

jon-fearer:
### Issue

[Support for NLBs targeting ALBs](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2297)

### Description

The goal of this pull request is to add support for [targeting ALBs from NLBs](https://aws.amazon.com/blogs/networking-and-content-delivery/application-load-balancer-type-target-group-for-network-load-balancer/) by adding the "alb" value to the existing nlb target type annotation. **_It is currently a work in progress._** If possible, it would be great to hear from the maintainers whether this is a feature that you are willing to support in this project. And if so, do you have any thoughts on what the service spec should look like for targeting an ALB? (see below). Thank you in advance!

Example:
```
apiVersion: v1
kind: Service
metadata:
  name: echoserver-nlb
  namespace: echoserver
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
    service.beta.kubernetes.io/aws-load-balancer-type: "external"
    service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: "alb" <-- new alb option
spec:
  # Not sure how everything below should look
  ports:
    - port: 80
      targetPort: 8080
      protocol: TCP
  type: LoadBalancer
  selector:
    (?)app: ???
```

### Checklist
- [x] Added tests that cover your change (if possible) **IN PROGRESS**
- [ ] Added/modified documentation as required (such as the `README.md`, or the `docs` directory) **TODO**
- [x] Manually tested **IN PROGRESS**

### DIFF
--- a/pkg/service/model_build_target_group.go
+++ b/pkg/service/model_build_target_group.go
@@ -95,6 +95,9 @@ func (t *defaultModelBuildTask) buildTargetGroupHealthCheckConfig(ctx context.Co
 		t.service.Spec.Type == corev1.ServiceTypeLoadBalancer {
 		return t.buildTargetGroupHealthCheckConfigForInstanceModeLocal(ctx)
 	}
+	if targetType == elbv2model.TargetTypeALB {
+		return t.buildTargetGroupHealthCheckConfigForALBTargetType(ctx)
+	}
 	return t.buildTargetGroupHealthCheckConfigDefault(ctx)
 }
 
@@ -155,6 +158,41 @@ func (t *defaultModelBuildTask) buildTargetGroupHealthCheckConfigForInstanceMode
 		return nil, err
 	}
 	unhealthyThresholdCount, err := t.buildTargetGroupHealthCheckUnhealthyThresholdCount(ctx, t.defaultHealthCheckUnhealthyThresholdForInstanceModeLocal)
+	if err != nil {
+		return nil, err
+	}
+	return &elbv2model.TargetGroupHealthCheckConfig{
+		Port:                    &healthCheckPort,
+		Protocol:                &healthCheckProtocol,
+		Path:                    healthCheckPathPtr,
+		IntervalSeconds:         &intervalSeconds,
+		HealthyThresholdCount:   &healthyThresholdCount,
+		UnhealthyThresholdCount: &unhealthyThresholdCount,
+	}, nil
+}
+
+func (t *defaultModelBuildTask) buildTargetGroupHealthCheckConfigForALBTargetType(ctx context.Context) (*elbv2model.TargetGroupHealthCheckConfig, error) {
+	healthCheckProtocol, err := t.buildTargetGroupHealthCheckProtocol(ctx, t.defaultHealthCheckProtocolForALBTargetType)
+	if err != nil {
+		return nil, err
+	}
+	var healthCheckPathPtr *string
+	if healthCheckProtocol != elbv2model.ProtocolTCP {
+		healthCheckPathPtr = t.buildTargetGroupHealthCheckPath(ctx, t.defaultHealthCheckPath)
+	}
+	healthCheckPort, err := t.buildTargetGroupHealthCheckPort(ctx, t.defaultHealthCheckPort)
+	if err != nil {
+		return nil, err
+	}
+	intervalSeconds, err := t.buildTargetGroupHealthCheckIntervalSeconds(ctx, t.defaultHealthCheckInterval)
+	if err != nil {
+		return nil, err
+	}
+	healthyThresholdCount, err := t.buildTargetGroupHealthCheckHealthyThresholdCount(ctx, t.defaultHealthCheckHealthyThreshold)
+	if err != nil {
+		return nil, err
+	}
+	unhealthyThresholdCount, err := t.buildTargetGroupHealthCheckUnhealthyThresholdCount(ctx, t.defaultHealthCheckUnhealthyThreshold)
 	if err != nil {
 		return nil, err
 	}
@@ -348,6 +386,9 @@ func (t *defaultModelBuildTask) buildTargetType(_ context.Context) (elbv2model.T
 			return "", errors.Errorf("unsupported service type \"%v\" for load balancer target type \"%v\"", svcType, lbTargetType)
 		}
 		return elbv2model.TargetTypeInstance, nil
+	}
+	if lbType == LoadBalancerTypeExternal && lbTargetType == LoadBalancerTargetTypeALB {
+		return elbv2model.TargetTypeALB, nil
 	}
 	return "", errors.Errorf("unsupported target type \"%v\" for load balancer type \"%v\"", lbTargetType, lbType)
 }
--- a/pkg/targetgroupbinding/resource_manager.go
+++ b/pkg/targetgroupbinding/resource_manager.go
@@ -89,7 +89,10 @@ func (m *defaultResourceManager) Reconcile(ctx context.Context, tgb *elbv2api.Ta
 	if *tgb.Spec.TargetType == elbv2api.TargetTypeIP {
 		return m.reconcileWithIPTargetType(ctx, tgb)
 	}
-	return m.reconcileWithInstanceTargetType(ctx, tgb)
+	if *tgb.Spec.TargetType == elbv2api.TargetTypeInstance {
+		return m.reconcileWithInstanceTargetType(ctx, tgb)
+	}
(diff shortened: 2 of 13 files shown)


pjaudiomv:
This would be amazing, def willing to test.

jon-fearer:
I've implemented the spec changes outlined above and tested manually. I'll hold off on any other changes for now

sftim:
@jon-fearer is this still WIP? If so, consider marking it as a draft (via GitHub's UI / API)

pkg/service/model_build_target_group.go
```diff
@@ -538,6 +585,36 @@ func (t *defaultModelBuildTask) buildTargetGroupBindingNodeSelector(_ context.Co
...
+			Name: t.service.Name,
+			Port: intstr.FromInt(int(port.Port)),
+		}
+	}
+	return serviceRef, nil
+}
+
+func (t *defaultModelBuildTask) buildTargetGroupBindingIngressRef(_ context.Context, targetType elbv2api.TargetType,
+	port corev1.ServicePort) (elbv2api.IngressReference, error) {
+	var ingressRef elbv2api.IngressReference
```
sftim, line 602: Do we check that the `ingressClassName` of that Ingress refers to an Ingress backed by this controller?

pkg/targetgroupbinding/resource_manager.go
```diff
@@ -209,15 +215,75 @@ func (m *defaultResourceManager) reconcileWithInstanceTargetType(ctx context.Con
...
+			Id:   awssdk.String(loadBalancerARN),
+			Port: awssdk.Int64(int64(tgb.Spec.IngressRef.Port.IntVal)),
+		},
+	})
+}
+
+func (m *defaultResourceManager) buildTargetLoadBalancerArn(ctx context.Context, tgb *elbv2api.TargetGroupBinding) (string, error) {
+	ingressName := tgb.Spec.IngressRef.Name
+	ingressKey := types.NamespacedName{Namespace: tgb.Namespace, Name: ingressName}
+	ingress := &v1beta1.Ingress{}
```
sftim, line 243: I think this should use the v1 Ingress API. Earlier APIs are no longer supported in Kubernetes.

sftim (changes requested):

pkg/targetgroupbinding/resource_manager.go
jon-fearer: Updated from networking v1beta1 to networking v1

pkg/service/model_build_target_group.go
jon-fearer: There is nothing doing that currently. The `IngressRef` is only validated in the `buildTargetLoadBalancerArn` function in `resource_manager.go`, when we attempt to register the ALB as a target. If a matching ingress isn't found, then that function returns an error and no target is registered. Do you think we should also be checking for an `IngressClass` resource (with `controller: ingress.k8s.aws/alb`)? Is there a scenario that this would protect against? (Still wrapping my head around this project and related k8s topics)

lado-golijashvili-rft:
Any update about it? We need this functionality for our solution.

jon-fearer:
@lado-golijashvili-rft I am looking for some directional feedback from maintainers (see proposed schemas above). I guess that would be @kishorj or @M00nF1sh ? Not 100% sure though

pjaudiomv:
> @jon-fearer any active work on thir pr ? This will be amazing.

The lack of any maintainer feedback has stalled his work.

erickfaustino:
Hi, folks!

This feature would be a life saver for us!

Can anyone unblock this, please?

Thank you very much!

liad5h:
Would definitely use this.

smcavallo:
@kishorj @M00nF1sh - any update?  We have been waiting for this functionality.  It is a common use case.  
Another common use case is creating and managing both the NLB and associated backend ALB and targetgroup with this controller

M00nF1sh:
@jon-fearer 
@smcavallo 
For your use case, would there be multiple ALBs behind NLB or just a single ALB behind a NLB?

Personally i don't favor current approach as the Kubernetes Service object is designed to target pods only. Using Service objects to model AWS NLB in this scenario is not appropriate to me. 

Possible alternatives in my mind:
1. use Gateway API(https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1alpha2.TCPRouteRule) and backendRefs to reference ALBs. We have plans to support gateway APIs in upcoming major releases. (This feels better to me)
2. use annotation on Ingress or field on IngressClass to specify a unique identifier representing a NLB(similar how ingressGroup's name works), and for each such identifier create a NLB and attach these NLBs to it.

any thoughts? @kishorj

smcavallo:
@M00nF1sh - this is our use case:
We accept both tcp and http traffic from customers.
We have a public facing NLB at the perimeter managed by `aws-load-balancer-controller` which accepts both tcp and http traffic
* tcp traffic is routed to kubernetes ingress/pods just like any other use case of `aws-load-balancer-controller`
* http traffic is routed to ALB target group (below)

we have an internal facing ALB managed by `aws-load-balancer-controller` with WAF and TLS/SNI for http traffic
* http traffic is routed to kubernetes ingress/pods just like any other use case of `aws-load-balancer-controller`
(we can't offload tls/sni to something inside the cluster - we need WAF, aws certificates, and other aws specific integrations and ALB functionality.)

Everything is out of the box `aws-load-balancer-controller` and working as it's designed to be used.
The only thing we can't do is wire up the ALB (managed by `aws-load-balancer-controller`) to sit behind the NLB (managed by the `aws-load-balancer-controller`)
Since `aws-load-balancer-controller` is managing BOTH the NLB and ALB it seems like it should be able to connect the 2.

The alternative workaround is for us to manage both the NLB and ALB via infrastructure-as-code (crossplane!) and then use the `TargetGroup` functionality of `aws-load-balancer-controller` to make sure the nodes/pods exist as part of the target groups. 

It feels like overkill to use to manage the entire config some other way when `aws-load-balancer-controller` can create and maintain everything, it just can't connect the infrastructure it created.

sftim:
We could make a very narrow implementation of the Gateway API that deploys an NLB and ALB (and, if we want, integrates with PrivateLink).

In Kubernetes, the Service API doesn't express a desired state that looks like that diagram @jon-fearer. That is why we shouldn't use Service to manage it.

If someone wants to add complexity to the AWS Load Balancer controller to add Gateway, they can - it's open source. However, that is a *big* ask.

matthewdaltonfanduel:
@M00nF1sh 

I am wondering if you can expand upon why you find it inappropriate to model the nlb with the service object? 

I am encountering the use case needing to inject an ALB for http to https redirection at the nlb. Load Balancers managed by aws-load-balancer-controller, this becomes quite the difficult ask. As @smcavallo pointed out earlier we could move this logic into IaC and manage with target groups but I find this to be a less desirable option as we prefer to manage the LB's with the controller rather than with IaC. 

I think this functionality also opens additional feature functionality that does align with this controller and how it functions. 
This feature would also allow implementations of a service defining an nlb targetting an ingress defining an alb. This functionality would be able to model any complex networking use-cases that load-balancers support.

I think this change can only be positive. I think that the aws-load-balancer-controller should be able to model and implement any of the features and capacities of load balancers in aws. Allowing the controller to be agnostic of implementation details allows it to remain in feature parity and be in lock step with additional aws features released down the road.

M00nF1sh:
@matthewdaltonfanduel 
The main reason is Kubernetes Service object is designed to target pods instead of targeting other objects like ALB. We are just abusing Kubernetes Service spec if we implement this way.
Personally i still prefer we implement this by the Gateway API, https://gateway-api.sigs.k8s.io/, which is more flexible. 

There is a on-going Gateway API implementation from community folks: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3146, we can consider implement this use case with that.

scila1996:
> @matthewdaltonfanduel The main reason is Kubernetes Service object is designed to target pods instead of targeting other objects like ALB. We are just abusing Kubernetes Service spec if we implement this way. Personally i still prefer we implement this by the Gateway API, https://gateway-api.sigs.k8s.io/, which is more flexible.
> 
> There is a on-going Gateway API implementation from community folks: #3146, we can consider implement this use case with that.

Just create a normal ingress with the class: alb. However, we should add a feature to support an annotation called 'static IP.' The ALB will automatically create an additional NLB and target the ALB. We should not create an additional service of type LoadBalancer.

PR #2385, opened 2021-11-29
Minimalist IAM permissions for TargetGroupBinding [documentation]

marcosdiez:
Small update on the documentation for those who want to use aws-load-balancer-controller only to add and remove targets into the target group and manage everything else elsewhere (like terraform or cloud formation)

### DIFF
--- a/docs/deploy/installation.md
+++ b/docs/deploy/installation.md
@@ -22,7 +22,7 @@
 If you are using the IMDSv2 you must set the hop limit to 2 or higher in order to allow the AWS Load Balancer Controller to perform the metadata introspection. Otherwise you have to manually specify the AWS region and the VPC via the controller flags `--aws-region` and `--aws-vpc-id`.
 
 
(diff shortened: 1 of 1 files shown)


docs/deploy/installation.md
```diff
@@ -102,14 +102,38 @@ If not setting up IAM for ServiceAccount, apply the IAM policies from the follow
 ```
 curl -o iam-policy.json https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/v2.3.0/docs/install/iam_policy.json
 ```
+
+##### IAM permission subsets for those who use *TargetGroupBinding* only and don't plan to use the AWS Load Balancer Controller to manage security group rules:
```
kishorj, line 106:
```suggestion
##### IAM permission subset for those who use *TargetGroupBinding* only and don't plan to use the AWS Load Balancer Controller to manage security group rules:
```

PR #2391, opened 2021-12-02
Set a namespace on the helm PodDisruptionBudget template

Dutchy-:
The poddisruptionbudget was missing a namespace - this adds it


Dutchy-:
I've just signed the CLA

kishorj (approved):

Dutchy-:
/retest

The failure does not look like something I caused

Issue #2203, opened 2021-09-01
ingress rules gets random order / priority in load balancer listener rules


ThomasSteinbach:
/kind bug

**Describe the bug**

The aws-load-balancer-controller creates Load Balancer Listener Rules in the same order as K8s ingress rules are defined in the ingress object. This does NOT correctly implement the K8s ingress specification, as the ingress documentation says:

> In some cases, multiple paths within an Ingress will match a request. In those cases precedence will be given first to the longest matching path.
( https://kubernetes.io/docs/concepts/services-networking/ingress/#multiple-matches )

More worse - the order of rules within an ingress specification is not ensured, as those rules are just a YAML/JSON list with no defined order. When processing the specification, any tool could possibly "sort" the items or alter the order. This is for instance the case if you process your k8s yaml specifications with official kubernetes kustomize.

That means that the priority of load balancer listener is random. At least NOT ordered by the longest matching path.

**Steps to reproduce**

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
  name: my-service
spec:
  rules:
  - host: my-service.hello.world
    http:
      paths:
      - backend:
          service:
            name: my-service
            port:
              name: application
        path: /
        pathType: Prefix
      - backend:
          service:
            name: my-service
            port:
              name: monitoring
        path: /actuator/
        pathType: Prefix
```

The AWS Load Balancer created will forward requests always to the `application` port. Even in the case of the request path `/actuator/` he will not forward to the `monitoring` port, as the rule with the path `/` get a higher priority in the Load Balancer Listener Rules.

**Expected outcome**

* The listener rule with the more specific path `/actuator` should get a higher priority in the Load Balancer Listener Rules.
* For the upper example requests to `my-service.hello.world/actuator` should be forwarded to the `monitoring` port.

**Environment**

* AWS Load Balancer controller version: helm chart versiono 1.2.3

kishorj:
@ThomasSteinbach, do the listener rules follow the order you specify in the manifest?

ThomasSteinbach:
@kishorj yes, they do.

kishorj:
/kind bug
we will fix it in the next release

sergeyshevch:
Also same behavior if we have multiple kind:Ingress with the same domain and different paths:
Example:

Ingress a contains host example.com and path /
Ingress b contains host example.com and path /blog/

sergeyshevch:
[...]

PR #2409, opened 2021-12-10
Sort ingress rules by path length and pathType

oliviassss:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2203

### Description

Sort Ingress rules based on the strategy:
- all exact match paths come first, no need to sort since exact match has to be unique
- followed by prefix paths, sort by lengths - longer paths get precedence
- followed by ImplementationSpecific paths or paths with no pathType specified, keep the original order
 
ref: Kubernetes Ingress (https://kubernetes.io/docs/concepts/services-networking/ingress/#multiple-matches).

#### Tests

- Added unit test
- Manual tested on Ingress creation with multiple ingress rules, verified that the rule with exact pathType has precedence, followed by paths with prefix pathType (order by length desc.), then followed by paths with implementationSpecific pathType.
- Created Ingresses with controller on v2.4.1 image, then upgraded to the controller with this change. Verified that the ingresses got update, and the order of paths with ImplementationSpecific PathType will remain the same.

### Checklist
- [x] Manually tested


pkg/ingress/rule_optimizer.go
```diff
@@ -83,6 +89,33 @@ func (o *defaultRuleOptimizer) omitOvershadowedRulesAfterRedirectRules(rules []R
...
+	for _, rule := range rules {
+		if _, ok := path2RulesMap[rule.Ingresspath]; ok {
+			path2RulesMap[rule.Ingresspath] = append(path2RulesMap[rule.Ingresspath], rule)
+		} else {
+			path2RulesMap[rule.Ingresspath] = []Rule{rule}
+			paths = append(paths, rule.Ingresspath)
+		}
+	}
+	// sort paths by length, if lengths equal, the path with Exact pathType has precedence
+	sort.Slice(paths, func(i, j int) bool {
```
kishorj, line 105: could we not sort the rules directly?

pkg/ingress/rule_optimizer_test.go
```diff
@@ -229,6 +230,300 @@ func Test_defaultRuleOptimizer_Optimize(t *testing.T) {
...
+					},
+					Ingresspath: networking.HTTPIngressPath{
+						Path:     "/",
+						PathType: (*networking.PathType)(awssdk.String("Exact")),
+					},
+				},
+			},
+		},
+		{
+			name: "Rules sorted by path length descending, when tie, rules with Exact path type precedes",
```
kishorj, line 381: also add a test with the same values for prefix and exact path, and verify the exact match takes precedence

pkg/ingress/model_build_listener_rules.go
```diff
@@ -72,6 +74,19 @@ func (t *defaultModelBuildTask) buildListenerRules(ctx context.Context, lsARN co
 	return nil
 }
 
+// sort paths by length, if lengths equal, the path with Exact pathType has precedence
+func (t *defaultModelBuildTask) sortIngressPaths(paths []networking.HTTPIngressPath) []networking.HTTPIngressPath {
+	sort.Slice(paths, func(i, j int) bool {
+		if len(paths[i].Path) == len(paths[j].Path) {
+			if paths[i].PathType != nil {
+				return *paths[i].PathType == networking.PathTypeExact
```
kishorj, line 82: does it work for all cases?

pkg/ingress/model_build_listener_rules_test.go
```diff
@@ -1,12 +1,123 @@
...
+				{
+					Path:     "/abc",
+					PathType: (*networking.PathType)(awssdk.String("Prefix")),
+				},
+			},
+		},
+		{
+			name: "Paths are the same, the path with Exact path type precedes",
+			args: args{
+				paths: []networking.HTTPIngressPath{
```
kishorj, line 89: what happens when exact comes before prefix in the input?

oliviassss: Just checked the exact path will still have the precedence.

pkg/ingress/model_build_listener_rules.go
oliviassss: I thought about these cases, when 2 paths have the same length, the current algorithm will return below results:
1. path[i] is exact type, path[j] is exact type, other type, or just nil - > path[i] precedes 
2. path[i] is other type, path[j] is exact type, other type or just nil -> path[j] precedes
3. path[i] PathType is nil, path[j] is exact type, other type or just nil -> path[j] precedes

So I think that would be sufficient. Since upstream only indicates the exact type should have higher priority over  others.

oliviassss: I thought about these cases, when 2 paths have the same length, the current algorithm will return below results:

1. path[i] is exact type, path[j] is exact type, other type, or just nil - > path[i] precedes
2. path[i] is other type, path[j] is exact type, other type or just nil -> path[j] precedes
3. path[i] PathType is nil, path[j] is exact type, other type or just nil -> path[j] precedes

So I think that would be sufficient. Since upstream only indicates the exact type should have higher priority over others.

pkg/ingress/model_build_listener_rules_test.go
oliviassss: Just checked the exact path will still have the precedence.

kishorj: thanks

kishorj (approved):

sergeyshevch:
@oliviassss Is there any plans to get this merged?

oliviassss:
Sort Ingress rules based on the strategy:

- all exact match paths come first, no need to sort since exact match has to be unique
- followed by prefix paths, sort by lengths - longer paths get precedence
- followed by ImplementationSpecific paths or paths with no pathType specified, keep the original order

pkg/ingress/model_build_listener_rules.go
```diff
@@ -72,6 +77,47 @@ func (t *defaultModelBuildTask) buildListenerRules(ctx context.Context, lsARN co
...
+}
+
+// classifyIngressPathsByType will classify the paths by type Exact, Prefix and ImplementationSpecific
+func (t *defaultModelBuildTask) classifyIngressPathsByType(paths []networking.HTTPIngressPath) ([]networking.HTTPIngressPath, []networking.HTTPIngressPath, []networking.HTTPIngressPath, error) {
+	var exactPaths []networking.HTTPIngressPath
+	var prefixPaths []networking.HTTPIngressPath
+	var implementationSpecificPaths []networking.HTTPIngressPath
+	for _, path := range paths {
+		if path.PathType != nil {
+			if *path.PathType == networking.PathTypeExact {
```
kishorj, line 105: use a switch statement instead of if-else

oliviassss: Updated

pkg/ingress/model_build_listener_rules.go
```diff
@@ -74,17 +77,46 @@ func (t *defaultModelBuildTask) buildListenerRules(ctx context.Context, lsARN co
...
+// all exact match paths come first, no need to sort since exact match has to be unique
+// followed by prefix paths, sort by lengths - longer paths get precedence
+// followed by ImplementationSpecific paths or paths with no pathType specified, keep the original order
+func (t *defaultModelBuildTask) sortIngressPaths(paths []networking.HTTPIngressPath) ([]networking.HTTPIngressPath, error) {
+	exactPaths, prefixPaths, implementationSpecificPaths, err := t.classifyIngressPathsByType(paths)
+	if err != nil {
+		return nil, err
+	}
+	sortedPaths := exactPaths
+	sort.Slice(prefixPaths, func(i, j int) bool {
```
kishorj, line 90: sort.Slice is not stable and there is a potential for reversing the original order for equal length paths. Lets use sort.SliceStable instead

oliviassss: Updated. Thanks

PR #2431, opened 2021-12-26
WIP: Add --service-default-target-type flag

johngmyers:
### Description

Adds a `--service-default-target-type` flag to specify the target type for `Service` objects that don't have a `service.beta.kubernetes.io/aws-load-balancer-nlb-target-type` annotation.

This is useful when kube-controller-manager's service controller is disabled. It is particularly useful for IPv6-only clusters with nodes in IPv6-only subnets, since KCM's service controller defaults to CLBs and CLBs don't support IPv6 backends.

### Checklist
- [x] Manually tested

### DIFF
--- a/controllers/service/eventhandlers/service_events.go
+++ b/controllers/service/eventhandlers/service_events.go
@@ -60,7 +60,7 @@ func (h *enqueueRequestsForServiceEvent) Generic(e event.GenericEvent, queue wor
 }
 
 func (h *enqueueRequestsForServiceEvent) isServiceSupported(service *corev1.Service) bool {
-	if h.handleByDefault {
+	if h.handleByDefault && service.Spec.Type == corev1.ServiceTypeLoadBalancer {
 		return true
 	}
 	lbType := ""
--- a/controllers/service/eventhandlers/service_events.go
+++ b/controllers/service/eventhandlers/service_events.go
@@ -15,11 +15,12 @@ import (
 )
 
 // NewEnqueueRequestForServiceEvent constructs new enqueueRequestsForServiceEvent.
-func NewEnqueueRequestForServiceEvent(eventRecorder record.EventRecorder, annotationParser annotations.Parser, logger logr.Logger) *enqueueRequestsForServiceEvent {
+func NewEnqueueRequestForServiceEvent(eventRecorder record.EventRecorder, annotationParser annotations.Parser, logger logr.Logger, handleByDefault bool) *enqueueRequestsForServiceEvent {
 	return &enqueueRequestsForServiceEvent{
 		eventRecorder:    eventRecorder,
 		annotationParser: annotationParser,
 		logger:           logger,
+		handleByDefault:  handleByDefault,
 	}
 }
 
@@ -29,6 +30,7 @@ type enqueueRequestsForServiceEvent struct {
 	eventRecorder    record.EventRecorder
 	annotationParser annotations.Parser
 	logger           logr.Logger
+	handleByDefault  bool
 }
 
 func (h *enqueueRequestsForServiceEvent) Create(e event.CreateEvent, queue workqueue.RateLimitingInterface) {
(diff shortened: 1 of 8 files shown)


johngmyers:
This appears to be trying to create loadbalancers for Services not of type LoadBalancer.

controllers/service/service_controller.go
```diff
@@ -184,7 +187,7 @@ func (r *serviceReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manag
 
 func (r *serviceReconciler) setupWatches(_ context.Context, c controller.Controller) error {
 	svcEventHandler := eventhandlers.NewEnqueueRequestForServiceEvent(r.eventRecorder, r.annotationParser,
-		r.logger.WithName("eventHandlers").WithName("service"))
+		r.logger.WithName("eventHandlers").WithName("service"), r.defaultTargetType != "")
```
kishorj, line 190: this will create a confusion to the end users. If they happen to enable this setting without realizing the kcm needs to be disabled, then both kcm and this controller would end up provisioning lb for the same service.

controllers/service/eventhandlers/service_events.go
```diff
@@ -58,6 +60,9 @@ func (h *enqueueRequestsForServiceEvent) Generic(e event.GenericEvent, queue wor
 }
 
 func (h *enqueueRequestsForServiceEvent) isServiceSupported(service *corev1.Service) bool {
+	if h.handleByDefault {
```
kishorj, line 63: this will apply for all service type without considering the load balancer type annotation values.

kishorj (commented):
I have some concerns with the proposed changes
1) We are effectively making NLB the default load balancer - we do want it eventually, but need some more time to ensure there are no issues with the changes. Some prior concerns related to NLB was the support for security groups, and traffic hairpinning
2) How do we disable the just the service controller from the kcm? Is there a feature gate, or a flag? If so, then it might not be practical for the EKS users.

johngmyers:
In response to comments:

1. CLBs are not useful in clusters with IPv6-only nodes, as CLBs do not support IPv6 targets. Thus, NLBs are the only practical option for the default load balancer on such clusters.
2. To disable just the service controller from either KCM or CCM, pass it the flags `--controllers=*` and `--controllers=-service`. (KCM cannot support IPv6-only nodes as it lacks kubernetes/cloud-provider-aws#286. One has to use CCM 1.23 or higher on such clusters.)
3. This feature's not being usable with EKS is not a concern. There is nothing stopping EKS from supporting it if they so choose.

johngmyers:
I'll additionally note that since CCM's service controller only supports NLBs with instance targets and NLBs don't support IPv6-only instance targets, the CCM's service controller is not at all useful for clusters with IPv6-only nodes.

docs/deploy/configurations.md
```diff
@@ -91,6 +91,7 @@ Currently, you can set only 1 namespace to watch in this flag. See [this Kuberne
 |leader-election-namespace              | string                          |                 | Name of the leader election ID to use for this controller |
 |log-level                              | string                          | info            | Set the controller log level - info, debug |
 |metrics-bind-addr                      | string                          | :8080           | The address the metric endpoint binds to |
+|service-default-target-type            | string                          |                 | Default target type for Services - ip, instance |
```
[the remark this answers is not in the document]
johngmyers: This flag only affects Services; it does not affect NLBs for Ingresses at all. So the name `default-nlb-target-type` would not be appropriate.

controllers/service/service_controller.go
johngmyers: I'm failing to understand the relevance of the comment to the context that it was made in.

Assuming it's a comment about the feature as a whole, it is inherent to the feature of being able to act as the default service controller that no other thing in the cluster can also be the default service controller at the same time.

johngmyers:
In today's Provider-AWS meeting I made the suggestion of having LBC add support for CLBs. Then CCM could deprecate and eventually remove its service controller. That and the ability for LBC to use CLBs as the default for Services of type LoadBalancer would be a followup to this PR.

PR #2435, opened 2021-12-27
e2e: Use Helm chart from the commit

johngmyers:
### Description

Change the e2e tests to use the Helm chart from the commit, not from the `main` branch. The Helm chart from the commit should be what is under the test.

### Checklist
- [x] Added tests that cover your change (if possible) (the change is the test)
- [ ] Added/modified documentation as required (such as the `README.md`, or the `docs` directory) (N/A)
- [ ] Manually tested (N/A)

### DIFF
--- a/test/framework/options.go
+++ b/test/framework/options.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 
 	"github.com/pkg/errors"
+	"sigs.k8s.io/aws-load-balancer-controller/test/framework/controller"
 )
 
 var globalOptions Options
@@ -17,7 +18,7 @@ type Options struct {
 	ClusterName string
 	AWSRegion   string
 	AWSVPCID    string
-	HelmDir     string
+	HelmChart   string
 	KubeConfig  string
 
 	// AWS Load Balancer Controller image. leave empty to use default one from helm chart.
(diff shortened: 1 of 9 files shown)


test/framework/options.go
```diff
@@ -51,6 +53,9 @@ func (options *Options) Validate() error {
 	if len(options.AWSVPCID) == 0 {
 		return errors.Errorf("%s must be set!", "aws-vpc-id")
 	}
+	if len(options.HelmDir) == 0 {
```
kishorj, line 56: Comment from PR 2433, 

@kishorj  2 hours ago

why not make it optional? If set, test uses the specified directory, otherwise default to the aws eks-charts repo.


@johngmyers  1 hour ago

Making it optional adds complexity for no obvious reason and admits the possibility of missing an incompatibility between the repo's chart and the repo's image.

Could you give more information on the use case for this? Why would one not want to put the repo's Helm chart under test?

@kishorj  now

One example is the canary script which we run against the release version. In this case we want the last published image and the chart, not the recent one from the repo.

johngmyers: One possibility would be to download the published chart in the script, outside of ginkgo. But it might work to rely on the scripts' current `helm repo add` and pass the name of the chart.

johngmyers:
I can squash if requested.

kishorj:
> I can squash if requested.

not necessary, I set to squash on merge :)

kishorj (approved):

Issue #1598, opened 2020-11-01
[v2] Error when there are multiple certificates for a host


postmaxin:
It is perfectly fine for there two be two certificates available for a particular host. I have this right now because I am in mid-migration and have both a wildcard, and host-specific, cert. In these cases, the controller should either add both certificates, or add (in my opinion) the least specific certificate.

```
{"level":"error","ts":1604259893.3354607,"logger":"controller","msg":"Reconciler error","controller":"ingress","name":"dex","namespace":"kube-system","error":"multiple certificate found for host: dex.admin.rex.sh, certARNs: [arn:aws:acm:us-west-2:355508092300:certificate/c6d7d07b-0aec-49ae-aff6-3616afb4b0a7 arn:aws:acm:us-west-2:355508092300:certificate/eeba7334-1671-4ad9-bc2f-d458ef972fdb]"}
```

M00nF1sh:
@postmaxin 
That's why we allow to specify the `certificate-arn` annotation to override the cert-discovery feature. The controller cannot make a choice which certificate to use when there are multiple matches. 😄

postmaxin:
It can, it just does not. It has no criteria. Example criteria would be: Length of domain name, expiry date, etc.
If it makes a choice based on some criteria, it may not be the exact criteria everyone who  uses AWSv2 would choose. However, it would be a choice, that is predictable, that breaks things less than having AWSv2 give up.

postmaxin:
I would suggest that in this situation, the controller should pick a certificate in a deterministic way, but log a warning saying that there were other choices. That's better than it just throwing it's hands up and walking away.

M00nF1sh:
@postmaxin 
I agree we can do it if we have a deterministic way. (e.g. pure sort based on lexical order of certificate arn). will it suits your need?

jorrit-wehelp:
It is behaviour that changed between v1 and v2 I think, so I'd say it should do whatever v1 did to keep the 'backwards-compatible' promise :)

Scenario that I have ingress rules for "example.com", "\*.example.com", "sub.example.com" and "\*.sub.example.com".
With two certificates, "\*.example.com" with additional domains "example.com" and "\*.sub.example.com" with additional domain "sub.example.com". So there is some overlap in the domain names.
But this worked in v1, 
v2 complains that the sub.example.com ingress now matches two certificates, which is correct. But v1 seemed to pick one (the 'exact match')...?

[...]

PR #2448, opened 2022-01-07
Enable having multiple matching certs in autodiscover mode

LechG:
### Issue
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1598

### Description

When you don't provide the annotation for certificate arn, the controller tries to find a matching certificate in ACM by itself. However when multiple matching certs are found - it fails with 
"multiple certificate found for host: host, certARNs: [arn1, arn2]"

Which I believe is wronge because:
1. YOU CAN add multiple certificates via annotation
2. AWS has it's own algorithm of determining the best certificate to be used for the requests

To adjust this behavior, I removed the if clause that threw an error during setup of ALB in case there was more than one domain matching certificate in ACM.

### Checklist
- [X] Manually tested


allupaku:
Hello team, any specific version , this will get merged into ? We are also looking forward for this

blakepettersson:
@kishorj @M00nF1sh apart from the conflict with master and the CLA, what would be the showstopper to get this in?

kishorj:
@LechG, thanks for your contribution. We will relax the single cert per host restriction in the v2.5.0 release, PR #3028 will address this. Closing this PR in favor of 3028.

PR #2453, opened 2022-01-10
Query IMDS over IPv6 if no IPv4 interface address

johngmyers:
### Description

Queries IMDS over IPv6 if there is no non-loopback IPv4 interface address.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/aws/cloud.go
+++ b/pkg/aws/cloud.go
@@ -1,12 +1,15 @@
 package aws
 
 import (
+	"net"
+	"os"
+	"strings"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
-	"os"
 	epresolver "sigs.k8s.io/aws-load-balancer-controller/pkg/aws/endpoints"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/aws/metrics"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/aws/services"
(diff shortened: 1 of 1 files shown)


pkg/aws/cloud.go
```diff
@@ -44,9 +47,28 @@ type Cloud interface {
 
 // NewCloud constructs new Cloud implementation.
 func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud, error) {
+	hasIPv4 := true
+	addrs, err := net.InterfaceAddrs()
+	if err == nil {
```
kishorj, line 52: in case of error, why not return right away?
Are there situations where the net.InterfaceAddrs() return error, but we still want to continue processing?

johngmyers: The idea was to avoid any negative effects in existing cases. So if the process doesn't have permission to enumerate the interfaces, it would assume IPv4.

kishorj (approved):

pkg/aws/cloud.go
```diff
@@ -44,9 +47,28 @@ type Cloud interface {
...
+		hasIPv4 = false
+		for _, addr := range addrs {
+			str := addr.String()
+			if !strings.HasPrefix(str, "127.") && !strings.Contains(str, ":") {
```
M00nF1sh, line 56: i think this check is a bit hacky.
maybe just upgrade the sdk and let users specify `AWS_EC2_METADATA_SERVICE_ENDPOINT_MODE` env variable via helm chart?

pkg/deploy/elbv2/load_balancer_synthesizer.go
```diff
@@ -99,7 +98,7 @@ func (s *loadBalancerSynthesizer) disableDeletionProtection(lb *elbv2sdk.LoadBal
 		},
 		LoadBalancerArn: lb.LoadBalancerArn,
 	}
-	_, err := svc.ModifyLoadBalancerAttributes(input)
+	_, err := s.elbv2Client.ModifyLoadBalancerAttributes(input)
```
M00nF1sh, line 101: this is a good catch :D

pkg/aws/cloud.go
johngmyers: Why should users have to specify AWS_EC2_METADATA_SERVICE_ENDPOINT_MODE at all?

olemarkus (approved):

M00nF1sh:
to be honest, i still don't think we should do this from the controller. It should be AWSSDK-GO to automatically detect IMDS if necessary.
And this detection feels hacky to me as well, e.g. you might have IMDS only available in ipv6 but the container have some other ipv4 addresses as well(e.g. cni's dummy ipv4 addresses for ipv6 only mode). 
Would be better to follow AWSSDK's practice to have the explicitly configure knob before AWSSDK have automatic detection mechanism.

johngmyers:
I would also prefer this be done in AWSSDK-GO, but aws-sdk has stated, without any explanation or discussion, that [they are not going to do that](https://github.com/aws/aws-sdk/issues/227). So this would need to be done client-by-client. The PR aws/aws-sdk-go#4234 has gotten no attention in months.

The container having non-loopback IPv4 addresses is an obscure corner case and is not negatively impacted by this change. Such unusual situations can still be handled by setting the environment variable.

PR #2470, opened 2022-01-21
added correct alb annotation

Shauncoss:
### Checklist
- [X] Manually tested

### DIFF
--- a/docs/examples/2048/2048_full.yaml
+++ b/docs/examples/2048/2048_full.yaml
@@ -48,8 +48,8 @@ metadata:
   annotations:
     alb.ingress.kubernetes.io/scheme: internet-facing
     alb.ingress.kubernetes.io/target-type: ip
+    kubernetes.io/ingress.class: alb
 spec:
-  ingressClassName: alb
   rules:
     - http:
         paths:


kishorj:
@Shauncoss, the kubernetes.io/ingress.class annotation is deprecated, and we recommend specifying ingress class via the ingress field `spec.ingressClassName`. 

If you use helm, you can specify `--set createIngressClassResource=true` during chart installation. We will make it default in the next chart release.

You can still use the old manifest from prior release, https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/v2.3.1/docs/examples/2048/2048_full.yaml which uses the annotation.

docs/examples/2048/2048_full.yaml
```diff
@@ -48,8 +48,8 @@ metadata:
   annotations:
     alb.ingress.kubernetes.io/scheme: internet-facing
     alb.ingress.kubernetes.io/target-type: ip
+    kubernetes.io/ingress.class: alb
 spec:
-  ingressClassName: alb
```
kishorj, line 52: This change was intentional

Shauncoss: @kishorj Thanks for the review. I understand it may be intentional, however when running that manifest as is, the error I posted originally is what is produced. I am running latest EKS cluster version and ALB controller version.

Shauncoss: Adding additional comment: I recreated my eks cluster and load balancer controller using the additional helm switch you recommended. I still receive this error:
Error from server (InternalError): error when creating "2048_full.yaml": Internal error occurred: failed calling webhook "vingress.elbv2.k8s.aws": Post "https://aws-load-balancer-webhook-service.kube-system.svc:443/validate-networking-v1beta1-ingress?timeout=10s": no endpoints available for service "aws-load-balancer-webhook-service"

kishorj: Could you check if the aws-load-balancer-controller pods are up and running? If you still have the problem, you can open a new issue. I will close this PR.

Issue #2411, opened 2021-12-10
[Feature request] Merge tags from alb.ingress.kubernetes.io/tags between service and ingress


michelzanini:
**Is your feature request related to a problem?**

When monitoring healthy/unhealthy targets on a monitoring system it is really hard to identify what K8S service it came from since the target name only contains 8 first chars, and we have services with long names and the same prefixes.

This are two examples of targets that look very much the same and we can't distinguish:

```
k8s-sandbox-systemte-7ba41435fc
k8s-sandbox-systemte-be0d2a4b2e
```

**Describe the solution you'd like**

Will be good if we could count on an AWS tag in the Target Group that only contains the service name, such as:

```
ingress.k8s.aws/service: "my-long-k8s-service-name"
```

or

```
ingress.k8s.aws/target: "my-long-k8s-service-name"
```

**Describe alternatives you've considered**

I know we can add a tag by using:

```
alb.ingress.kubernetes.io/tags: Service=my-long-k8s-service-name
```

But I can't use this because my Ingress already has 5 tags added and when you use `alb.ingress.kubernetes.io/tags` on the service and on the Ingress together, the tags **don't** merge, instead, the service tags take priority. This way I can't add the tag myself without losing other important tags.

Another solution that would work great would be to merge tags from service and ingress for `alb.ingress.kubernetes.io/tags`. That would also help solving the issue.

**Note**:
The issue has been renamed to track the implementation that will merge tag values from `alb.ingress.kubernetes.io/tags` at service and ingress levels.

For example, adding this annotations:

service:
```
alb.ingress.kubernetes.io/tags: ServiceTag=my-tag-service
```

ingress:
```
alb.ingress.kubernetes.io/tags: IngressTag=my-ingress-service
```

Will tag the AWS target group with:

```
ServiceTag=my-tag-service
IngressTag=my-ingress-service
```

Current behaviour only tags `ServiceTag`.
I recommend if the same tag is found on both places that the `service` level overrides the value of the `ingress` (but only if it's the same tag).

kishorj:
@michelzanini, the targets group for ALB will have the following tag:
ingress.k8s.aws/resource: <namespace>/<ingress-name>-<service-name>:<port>
check if parsing the tag works for your case

We will look into supporting merging of the tags specified via annotation between ingress and service.

michelzanini:
[...]

PR #2480, opened 2022-01-27
Merge tags from alb.ingress.kubernetes.io/tags

oliviassss:
### Issue

Fixes https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2411

### Description

This PR will merge the tags specified by `alb.ingress.kubernetes.io/tags` from ingress and service and add them to the corresponding target groups. In case of conflicting values for the tag key, controller gives precedence to the ones specified on the service.

#### Test

- Added unit test
- Created ingress and backend service, verified that the target group has merged tags specified from both ingress and service, and the tags of ingress will not be affected
- Created ingress and beckend service with conflicted tag keys, verified that the target group has the tag values from service for the conflicted tag keys
- Created IngressClass, ingress and backend service, verified that the tags specified via IngressClass still have higher priority than tags specified via annotation on ingress or service

### Checklist
- [x] Manually tested

### DIFF
--- a/docs/guide/ingress/annotations.md
+++ b/docs/guide/ingress/annotations.md
@@ -741,6 +741,8 @@ The AWS Load Balancer Controller automatically applies following tags to the AWS
 In addition, you can use annotations to specify additional tags
 
 - <a name="tags">`alb.ingress.kubernetes.io/tags`</a> specifies additional tags that will be applied to AWS resources created.
+  In case of target group, the controller will merge the tags from the ingress and the backend service giving precedence 
+  to the values specified on the service when there is conflict.
 
     !!!example
         ```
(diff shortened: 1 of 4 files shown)


docs/guide/ingress/annotations.md
```diff
@@ -740,7 +740,9 @@ The AWS Load Balancer Controller automatically applies following tags to the AWS
 
 In addition, you can use annotations to specify additional tags
 
-- <a name="tags">`alb.ingress.kubernetes.io/tags`</a> specifies additional tags that will be applied to AWS resources created.
+- <a name="tags">`alb.ingress.kubernetes.io/tags`</a> specifies additional tags that will be applied to AWS resources created.
```
kishorj, line 743: Lets reword it as follows - 
In case of target group, the controller will merge the tags from the ingress and the backend service giving precedence to the values specified on the service in case of conflict.

kishorj (approved):

Issue #2466, opened 2022-01-20
`loadBalancerClass` support?


TBBle:
**Is your feature request related to a problem?**

Not a problem per-se. I just came across [`Service.spec.loadBalancerClass`](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class) and thought it might be a good alternative or replacement for the `service.beta.kubernetes.io/aws-load-balancer-nlb-target-type` annotation. 

By it's very presence, it implies `service.beta.kubernetes.io/aws-load-balancer-type: "external"` behaviour, so the [legacy AWS Cloud Controller will ignore it](https://github.com/kubernetes/kubernetes/blob/v1.22.0/staging/src/k8s.io/cloud-provider/controllers/service/controller.go#L822-L824).

I'm not sure if the AWS Cloud Controller v1 will also inherit that behaviour. I _expect_ that it already works, since the relevant feature has moved to beta and the KEP didn't mention any work needed by non-legacy Cloud Controller implementations. Otherwise it will need to separately implement the same check (similar to the [current `isLBExternal` annotation check](https://github.com/kubernetes/cloud-provider-aws/blob/v1.22.0-alpha.1/pkg/providers/v1/aws_loadbalancer.go#L83-L88), I guess).

The AWS Cloud Controller v2 doesn't appear to support load balancers at all yet, I recall a discussion that the plan for v2 is to never implement it in favour of this controller (which presents a different challenge), but I may be misremembering anyway. (Edit: I appear to be wrong about this, it looks like there [was intention for LB support in the V2 Cloud Controller](https://github.com/kubernetes/cloud-provider-aws/issues/125), but [work stalled and has staled away](https://github.com/kubernetes/cloud-provider-aws/pull/148)...)

Per [KEP-1959](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cloud-provider/1959-service-lb-class-field/README.md), one of the goals of this feature is:
> prevent every cloud provider from implementing a custom "opt-out" annotation for their load balancer.

which we already have done in this controller and the various AWS Cloud Controllers.

**Describe the solution you'd like**

The simple-obvious solution is that we recognise the following loadBalancerClass values as equivalent to the given `aws-load-balancer-nlb-target-type` annotation. (I'm making up the domain name. Looking around, all the AWS annotations for services seem to be in kubernetes.io domains, I'm not sure if that's legacy or not.)
[...]

PR #2489, opened 2022-02-02
add support for loadBalancerClass

kishorj:
### Issue
Fixes: #2466
Fixes: #2492

### Description
Add support for service `spec.loadBalancerClass`. The load balancer class provides a similar mechanism to the current service annotation `service.beta.kubernetes.io/aws-load-balancer-type: "external"` to offload reconciling of load balancer from the kubernetes in-tree controller, but it is agnostic to the cloudprovider - [KEP-1959](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cloud-provider/1959-service-lb-class-field/README.md) for details. The load balancer class is beta starting k8s 1.22.

The AWS Load Balancer controller will support the `spec.loadBalancerClass` starting v2.4.0 on k8s 1.22 and later. Here is the summary of the changes
- default LoadBalancerClass is `service.k8s.aws/nlb`, users can modify via controller command line flag `--load-balancer-class`
- with `spec.loadBalancerClass`, users no longer need to specify the annotation `service.beta.kubernetes.io/aws-load-balancer-type` on k8s 1.22 and later
- Target type defaults to `instance` for service with loadBalancerClass specified on supported k8s
- If loadBalancerClass is configured, when the user modifies the service type from LoadBalancer to something else, controller will deallocate the provisioned load balancer
- Provide a feature gate `ServiceTypeLoadBalancerOnly` which limits the controller to service of type LoadBalancer for NLB
- error out during reconcile for instance target if NodePort is not allocated 

### Tests:
- controller honors the `spec.loadBalancerClass` configuration on k8s 1.22
- controller works as expected on earlier k8s versions

### Checklist
- [x] Manually tested


M00nF1sh:
i think we can use service.k8s.aws/nlb to align with our IngressClass's ingress.k8s.aws/alb name. also it denotes it's a NLB :D

pkg/config/service_config.go
```diff
@@ -0,0 +1,20 @@
+package config
+
+import "github.com/spf13/pflag"
+
+const (
+	flagLoadBalancerClass    = "load-balancer-class"
+	defaultLoadBalancerClass = "elbv2.k8s.aws"
```
TBBle, line 7: This definitely needs to change to be a prefixed name, since ["Unprefixed names are reserved for end-users."](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class).

kishorj: updated

controllers/service/service_controller.go
```diff
@@ -137,12 +149,16 @@ func (r *serviceReconciler) reconcileLoadBalancerResources(ctx context.Context,
 	return nil
 }
 
-func (r *serviceReconciler) cleanupLoadBalancerResources(ctx context.Context, svc *corev1.Service) error {
+func (r *serviceReconciler) cleanupLoadBalancerResources(ctx context.Context, svc *corev1.Service, stack core.Stack, lb *elbv2model.LoadBalancer) error {
```
M00nF1sh, line 152: seems `lb` is not been used

pkg/service/model_builder.go
```diff
@@ -214,3 +221,19 @@ func (t *defaultModelBuildTask) getDeletionProtectionViaAnnotation(svc corev1.Se
...
+func (t *defaultModelBuildTask) needsCleanup() bool {
+	if !t.service.DeletionTimestamp.IsZero() {
+		return true
+	}
+	if t.service.Spec.Type != corev1.ServiceTypeLoadBalancer {
+		if t.featureGates.Enabled(config.ServiceTypeLoadBalancerOnly) {
+			return true
+		}
+		var lbType string
+		if !t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixLoadBalancerType, &lbType, t.service.Annotations) {
```
M00nF1sh, line 234: how about user changed it to something else like "nlb-ddd", seems we won't delete the NLB here.

controllers/service/service_controller.go
```diff
@@ -87,13 +89,20 @@ func (r *serviceReconciler) reconcile(ctx context.Context, req ctrl.Request) err
 	if err := r.k8sClient.Get(ctx, req.NamespacedName, svc); err != nil {
 		return client.IgnoreNotFound(err)
 	}
-	if !svc.DeletionTimestamp.IsZero() {
-		return r.cleanupLoadBalancerResources(ctx, svc)
+	stack, lb, err := r.buildModel(ctx, svc)
+	if err != nil {
+		return err
+	}
+	var resLBs []*elbv2model.LoadBalancer
```
M00nF1sh, line 96: i check we can simply check whether lb is nil or not here.

pkg/service/model_build_target_group.go
```diff
@@ -332,19 +332,25 @@ func (t *defaultModelBuildTask) buildTargetGroupHealthCheckUnhealthyThresholdCou
 	return unhealthyThresholdCount, nil
 }
 
-func (t *defaultModelBuildTask) buildTargetType(_ context.Context) (elbv2model.TargetType, error) {
+func (t *defaultModelBuildTask) buildTargetType(_ context.Context, port corev1.ServicePort) (elbv2model.TargetType, error) {
 	svcType := t.service.Spec.Type
 	var lbType string
 	_ = t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixLoadBalancerType, &lbType, t.service.Annotations)
 	var lbTargetType string
+	if t.service.Spec.LoadBalancerClass != nil {
```
M00nF1sh, line 340: maybe just remove the check if t.service.Spec.LoadBalancerClass != nil, since the service will only be called when it's have one of the `SvcLBSuffixTargetType` annotation any way if it's with external LB type.

pkg/service/model_builder.go
M00nF1sh: also, how about other cases like when the nlb-target-type annotation is removed while lb-type is still external (when type changed back to node port).

I feel such logic should be in a centralized place(shared by both event handler and here)

kishorj: our docs mention the aws-load-balancer-type annotation should be treated immutable. We can definitely cleanup the NLB if we no longer recognize the annotation.

I agree, we should centralize the logic to determine whether the service is supported or not.

pkg/service/model_build_target_group.go
kishorj: ack

pkg/service/model_builder.go
```diff
@@ -214,3 +226,20 @@ func (t *defaultModelBuildTask) getDeletionProtectionViaAnnotation(svc corev1.Se
 	}
 	return false, nil
 }
+
+func (t *defaultModelBuildTask) needsCleanup() bool {
```
M00nF1sh, line 230: this is still a little hard to understand this part tbh.

I think maybe like have below method in serviceUtils?
* IsServicePendingFinalization
     -> only checks whether the Service contains our finalizer 
* IsServiceSupported/IsServiceAlive(or other better word than Alive/Supported)
     -> only checks whether the Service is supported, and if so, we'll always provision a NLB.
     -> if !t.service.DeletionTimestamp.IsZero, return false
     -> if config.ServiceTypeLoadBalancerOnly and type not LB, return false
     -> this checks ((spec.IngressClass == service.k8s.aws/nlb) || anno.LBType & anno.NLBTargetType meets requirement)


So that, the event handler will check **whether IsServicePendingFinalization or IsServiceSupported**

And modelBuilder's logic will be like
If !IsServiceSupported{
  if IsServicePendingFinalization {
     checkForDeletionProtection
  }
  return nil
}
return non-nil model

And modelBuilder only need to check whether IsServiceSupported.

kishorj: updated

M00nF1sh (approved):
/lgtm
/approve

Issue #2496, opened 2022-02-08
Desired AvailabilityZoneIDs support


shaoxt:
**Is your feature request related to a problem?**

When anything wrong in one of the AvailabilityZone of AWS, how can we turn off the traffic quickly to all ingresses in the cluster?
So that the traffic won't be impacted by the bad AZ.

There is some discussions about the approach to apply target node labels on each Ingress #2354.
Think about the case that couple hundreds ingresses in one cluster, it is tough to update each Ingress.

**Describe the solution you'd like**
The proposal is to have an argument "--desired-availability-zone-ids" on aws-load-balancer-controller, so that the controller can pickup the targets within the desired AZs. This can switch the entire cluster as quick as possible.

Sample:
```
--desired-availability-zone-ids=usw2-az2,usw2-az3
```
This argument will limit the subnets specified on the annotation of Ingress 
```
alb.ingress.kubernetes.io/subnets: IngressSubnetAz1, IngressSubnetAz2, IngressSubnetAz3
```
Even the ingress specified 3 subnets, but the targets NOT in the desired availability zones won't be used in the targetGroup.
The feature provides a lever to control the traffic out of one AZ quickly.

**Describe alternatives you've considered**
Since this feature only enabled once the argument is specified, so it won't impact any major feature.

kishorj:
@shaoxt, I see the following issues with the suggested approach
- In case of NLB, we will not be able to delete existing AZ after creation
- The controller subnet autodiscovery performs discovery only for new load balancers, existing ones won't be affected

shaoxt:
@kishorj 

- Yes, NLB doesn't reconcile when it is created.  
    -   Is there any reason to treat NBL differently with ALB ? 
- The subnet autodiscovery does apply on existing new load balancer after the aws-load-balancer got restarted.
It does limit the subnets from given AZs.

```
Availability Zones
[subnet-07a4c8b03446f6e2b - us-west-2a ](https://console.aws.amazon.com/vpc/home?region=us-west-2#subnets:filter=subnet-07a4c8b03446f6e2b)
IPv4 address: Assigned by AWS

[subnet-0cd52c2c9e095a466 - us-west-2b ](https://console.aws.amazon.com/vpc/home?region=us-west-2#subnets:filter=subnet-0cd52c2c9e095a466)
IPv4 address: Assigned by AWS
```

Issue #2354, opened 2021-11-11
Add support labels selector to `target-node-labels` and `aws-load-balancer-target-node-labels` annotations


alanprot:
**Is your feature request related to a problem?**

We may need to remove from a target group all nodes from a given AZ in case of a AZ outage. Currently 
`service.beta.kubernetes.io/aws-load-balancer-target-node-labels` and `alb.ingress.kubernetes.io/target-node-labels` only supports the `=` operator which makes this not feasible.

https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.3/guide/ingress/annotations/#target-node-labels
https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/annotations/#target-node-labels

**Describe the solution you'd like**
`service.beta.kubernetes.io/aws-load-balancer-target-node-labels` and `alb.ingress.kubernetes.io/target-node-labels` values should be decoded as [labels selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) instead `stringMap` and we so should be able now to do something like:

```
 service.beta.kubernetes.io/aws-load-balancer-target-node-labels: "install=gateway, topology.kubernetes.io/zone!=us-west-2a"
```
**Describe alternatives you've considered**

I tried to work around  issue adding add the healthy AZs but it does not work as the `stringMap` type will return only one value for a given label (as expected once it is a Map):

Ex:

```
 service.beta.kubernetes.io/aws-load-balancer-target-node-labels: "install=gateway, topology.kubernetes.io/zone=us-west-2a, topology.kubernetes.io/zone=us-west-2c"
```
Will be decoded to
```
{
  "install": "gateway",
  "topology.kubernetes.io/zone": "us-west-2-c"
}
```

M00nF1sh:
/kind feature
Thanks for raising this issue. I believe use labelSelectors is the way to go and it should be backwards compatible as well 👍

shaoxt:
This feature is critical for AZ evacuation, when one AWS AZ went back, we should have a quick way to remove the nodes in the BAD AZ from the target group quickly.

alanprot:
I think this is still relevant!

PR #2497, opened 2022-02-08
Desired available zone ids

shaoxt:
### Issue

PR for #2496 

### Description

The proposal is to have an argument "--desired-availability-zone-ids" on aws-load-balancer-controller, so that the controller can pickup the targets within the desired AZs. This can switch the entire cluster as quick as possible.

Sample:

--desired-availability-zone-ids=usw2-az2,usw2-az3
This argument will limit the subnets specified on the annotation of Ingress. The argument won't limit any subnets if it doesn't specified anything. It won't impact anything by default.

alb.ingress.kubernetes.io/subnets: IngressSubnetAz1, IngressSubnetAz2, IngressSubnetAz3
Even the ingress specified 3 subnets, but the targets NOT in the desired availability zones won't be used in the targetGroup.
The feature provides a lever to control the traffic out of one AZ quickly.

In SubnetResolver, it consumes the desired availability zone ids, so that it limit the subnets be selected in this given list.

### Checklist
- [x] Manually tested


shaoxt:
Going through the checklist

M00nF1sh:
Thanks for making this PR.

I believe we won't re-invoke the subnet discovery for existing ALBs. So this change will only impact new ALBs only, which wont suit your requirement per my understanding.

i feel a preferred approach is combine https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2354 with a controller-level flag to set default nodeSelector. (so you can change the controller-level default nodeSelector if you don't set the nodeSelector per Ingress)

shaoxt:
@M00nF1sh  Thanks for the comments.  
Couple questions, about the #2354 
1. How are you going to sync the bad AZ information from one place to all ingresses ? if you have couple hundreds of Ingresses in one cluster.
2. Since the "topology.kubernetes.io/zone" points the bad AZ name in each aws account.  How are you going to deals with 200 aws accounts when the physical AZ ID went bad, but it has different AZ names in different aws accounts ?

M00nF1sh:
@shaoxt 
How are you going to sync the bad AZ information from one place to all ingresses ? if you have couple hundreds of Ingresses in one cluster.
1. i think the solution is to expose a controller-level flag(can be IngressClass level as well), which sets the nodeSelector, thus once the controller is restarted, all Ingresses will be reconciled.

2. Since the "topology.kubernetes.io/zone" points the bad AZ name in each aws account. How are you going to deals with 200 aws accounts when the physical AZ ID went bad, but it has different AZ names in different aws accounts ?
1. i prefer a external source (e.g. gitops tool that manages the controller deployment) to do the az mapping, which translates into the nodeSelector argument into the controller. (this way will make the controller itself agnostic to the detailed nodeSelector).

shaoxt:
@M00nF1sh Like the idea of "-backlisted-az-ids" 
 
```

> @shaoxt
> 
> > How are you going to sync the bad AZ information from one place to all ingresses ? if you have couple hundreds of Ingresses in one cluster.
> 
> 1. i think the solution is to expose a controller-level flag(can be IngressClass level as well), which sets the nodeSelector, thus once the controller is restarted, all Ingresses will be reconciled.
> 
> > Since the "topology.kubernetes.io/zone" points the bad AZ name in each aws account. How are you going to deals with 200 aws accounts when the physical AZ ID went bad, but it has different AZ names in different aws accounts ?
> 
> 1. i prefer a external source (e.g. gitops tool that manages the controller deployment) to do the az mapping, which translates into the nodeSelector argument into the controller. (this way will make the controller itself agnostic to the detailed nodeSelector).
> 2. alternatively, there can be a special flag such as "--backlisted-az-ids", and the controller reads it and translates it into the corresponding `topology.kubernetes.io/zone` nodeSelector per account internally. but this couples the controller with the specific `topology.kubernetes.io/zone` key.

PR #2516, opened 2022-02-18
Make e2e LB name test work with clusters that have . in the name

olemarkus:
### Description

ELB names cannot have `.` in the name, but cluster names can. So when the `load-balancer-name` annotation e2e test directly copies cluster name into the ELB name, the test fails.

This PR replaces `.` with `-` to make tests pass.

### Checklist
- [ ] Manually tested


kishorj (approved):
Thanks

Issue #2393, opened 2021-12-02
New deployment pods never being added to new target group


michael-careplanner:
**Describe the bug**
Changing an ingress group and doing a deployment rollout at the same time can cause new pods to never be registered on the new target group. This causes the readiness gate on the pod to never pass, stalling the deployment rollout. The only fix is to delete the unready pod, at which point a new one is created that gets added to the new target group as expected.

**Steps to reproduce**
[nginx-chart.tar.gz](https://github.com/kubernetes-sigs/aws-load-balancer-controller/files/7642770/nginx-chart.tar.gz)

Above is a simple Helm chart that can be used to replicate this bug. It runs an nginx deployment with a service and ingress. To replicate, install this chart into a namespace with a `elbv2.k8s.aws/pod-readiness-gate-inject: enabled` annotation:

`helm install nginx-chart nginx-chart/ --set group=internal --set tag=stable`

Then, trigger an update of the chart, changing both the nginx image tag (to trigger a deployment rollout) and the ingress group at the same time:

`helm upgrade nginx-chart nginx-chart/ --set group=external --set tag=latest`

Sometimes it takes a few toggles between the group / tag to trigger the buggy behaviour, but after a few tries you should see that there is a single pod with the readiness gate at `0/1`, and if you check the target group that pod's IP will not have been registered with the load balancer. This unready pod will remain indefinitely and never be registered to the target group.

**Expected outcome**
The newly deployed pod is registered on the new target group, becomes healthy, and passes the readiness gate, allowing the deployment to complete successfully.

**Environment**

* AWS Load Balancer controller version:
  * 2.2.3
* Kubernetes version:
  * 1.21
* Using EKS (yes/no), if so version?
  * Yes. 1.21

**Additional Context**:
I think this could be a similar race condition issue to https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1764

Looking at the controller logs as this is happening it:
 - Register the newly created pod against the _old_ target group
 - Deregister this pod against the _old_ target group
 - Delete the _old_ target group
 - Create a new target group
 - Never register the new pod into the new target group

M00nF1sh:
This is a similar race condition for https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1764.
It's due to the TGBs are not created when the pods are created.

we can address this issue by mark pods with corresponding readinessGates to be healthy when deleting a TGB

oliviassss:
/ assign

Issue #1764, opened 2021-01-18
Race condition between pod readiness gate injection and ingress model update?


foriequal0:
I've once tried to update both ingress and deployment at once.
But then new pod's readiness gate stuck at this state for few minutes.
```
Readiness Gates:
  Type                                                           Status
  target-health.elbv2.k8s.aws/old-target-group-id                <none>
```

This is a brief timeline of logs of the `aws-load-balancer-controller`. This happened in < 3 seconds.
```
< left are logs with old-target-group-id
          > right are logs with new-target-group-id
successfully built model (new pod has created around this log's timestamp)
successfully deployed model
          successfully built model
          creating targetGroup
          created targetGroup
          modifying listener rule
          modified listener rule
          creating targetGroupBinding
          created targetGroupBinding
deleting targetGroupBinding
deRegistering targets (old pod)
deRegistered targets (old pod)
          registering targets (old pod)
Reconciler error with "TargetGroupBinding.elbv2.k8s.aws \"old-target-group-id\" not found"
deleted targetGroupBinding
          registered targets (old pod)
deleting targetGroup
deleted targetGroup
          successfully deployed model
```

I also wonder why the events are in this order for the old pod:
```
creating targetGroup
modifying listener rule
creating targetGroupBinding
deRegistering targets
registering targets
```
not in this order:
```
creating targetGroup
creating targetGroupBinding
registering targets
modifying listener rule
deRegistering targets
```
Wouldn't modifying the listener to forward traffic to the target group that targets are registered later introduce downtime?

M00nF1sh:
@foriequal0 
The controller currently doesn't support zero downtime when u modify the Ingress/Service in such way that result in TargetGroup changes(e.g. changed the backend service).  The Targets won't be health-checked before added to LoadBalancer. However, we indeed can do optimizations like modify the rule once new targets are registered(but not health-checked) to reduce the downtime in such case.

For the pod readinessGate issue, i think it's indeed a race condition under such case. the webhook is based on current available TargetGroupBinding in cluster.  I think we can solve it by:
1. delete old TargetGroupBinding first.
2. mark podCondition with associated TargetGroupBinding's readinessGate as healthy when delete a TargetGroupBinding.

/kind bug

Project0:
/remove-lifecycle stale

[...]

PR #2524, opened 2022-02-23
Update pods with readinessGate as healthy for deleted TGB

oliviassss:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2393
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1764

### Description

To avoid race condition between pods readiness gate, this PR adds a step to mark podCondition with associated TargetGroupBinding's readinessGate as healthy when delete a TargetGroupBinding.

#### Test

- Created ingress and deployment in the namespace with pod readinessGate enabled, triggered the ingress update and deployment rollout at the same time, verified that the pod's readinessGate status is `1/1`, and the new target group is registered to the TargetGroupBinding.
- Deleted the resources in the namespace with pod readinessGate enabled, verified these resources can be cleaned up successfully
- Created and deleted ingress and deployment in the namespace with pod readinessGate diabled, verified these resources can be created and deleted successfully, no behavior change in this use case

### Checklist
- [x] Manually tested


kishorj (approved):

pkg/targetgroupbinding/resource_manager.go
```diff
@@ -325,6 +330,34 @@ func (m *defaultResourceManager) updateTargetHealthPodConditionForPod(ctx contex
...
+	targetHealthCondType := BuildTargetHealthPodConditionType(tgb)
+
+	allPodKeys := m.podInfoRepo.ListKeys(ctx)
+	for _, podKey := range allPodKeys {
+		pod, exists, err := m.podInfoRepo.Get(ctx, podKey)
+		if err != nil {
+			return err
+		}
+		if !exists {
+			return errors.New("couldn't find podInfo for ready endpoint")
```
M00nF1sh, line 345: we should just `continue` if pod not found. 
we can fix it in future versions as it's rare to trigger and won't cause much defect.

pkg/targetgroupbinding/resource_manager.go
```diff
@@ -325,6 +330,34 @@ func (m *defaultResourceManager) updateTargetHealthPodConditionForPod(ctx contex
...
 }
 
+// updatePodAsHealthyForDeletedTGB updates pod's targetHealth condition as healthy when deleting a TGB
+// if the pod has readiness Gate.
+func (m *defaultResourceManager) updatePodAsHealthyForDeletedTGB(ctx context.Context, tgb *elbv2api.TargetGroupBinding) error {
+	targetHealthCondType := BuildTargetHealthPodConditionType(tgb)
+
+	allPodKeys := m.podInfoRepo.ListKeys(ctx)
+	for _, podKey := range allPodKeys {
+		pod, exists, err := m.podInfoRepo.Get(ctx, podKey)
```
M00nF1sh, line 340: we should check for pod's namespace here as well. 
Since in theory you can have two targetGroupBinding in two namespace.

In practice, with our default Ingress/Service implementation, the TGB's name is same as TargetGroup's name and will be unique across namespaces.
but this will impact user created targetGroupBinding.

we can fix it in future versions as the impact is minimal

PR #2530, opened 2022-03-01
Add new IPv4 Pool annotation for automated Elastic IP management of NLBs

adammw:
### Issue

#2523

### Description

Add a new annotation `service.beta.kubernetes.io/aws-load-balancer-eip-ipv4-pool` that will direct the aws-load-balancer-controller to create new Elastic IP allocations on-demand from a given public ipv4 pool. This removes the need to manually create a discrete number of Elastic IP allocations as is required by the `service.beta.kubernetes.io/aws-load-balancer-eip-allocations` annotation.

Much of the code is cloned from the Security Group management, and converts the pre-existing AllocationID into a resource token that is resolved when the allocations are created by the ElasticIPAddress Manager/Synthesizer.

I could not see any example tests that tested this sort of resource resolution so unsure how to write tests that directly exercise the main codepaths, but manually tested against our staging environment successfully creating and deleting Elastic IP allocations.

### Checklist
- [ ] Added tests that cover your change (if possible) - partially 
- [x] Manually tested


olemarkus:
This one looks quite interesting.

Would it be possible to also set a default pool through an argument?

/ok-to-test

adammw:
@kishorj @olemarkus @M00nF1sh rebased PR. would be great if we could get this upstreamed so we don't need to rely on a fork.

Issue #2159, opened 2021-08-04
Reduce resource consumption by filtering secrets or namespaces


zack-johnson5455:
**Is your feature request related to a problem?**
We run the aws-load-balancer-controller as a deployment in an EKS cluster. We recently had an incident where the load balancer controller started failing to update load balancer target groups. The root cause here was that the list secrets API call was taking longer than 60 seconds. The exact error we started receiving was "1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.6/tools/cache/reflector.go:125: Failed to list *v1.Secret: stream error when reading response body, may be caused by closed connection. Please retry. Original error: stream error: stream ID 13845; INTERNAL_ERROR".

We resolved the incident by increasing the CPU requests/limits on the controller and reducing the number of secrets in the cluster. We had already set the memory requests/limits fairly high. However, as our clusters see increasing usage, we are concerned that we are going to run into this issue again. We'd like to find a solution that doesn't require just increasing the resource requests more and more.

We did additional testing where we took a test cluster and created a large number of secrets in it. After creating about 10,000 secrets, we saw memory usage of up to 5Gi. We also saw CPU throttling up to 75% at a CPU limit of 500mi. We saw the following log message (and many other similar ones) as well, which we interpreted to mean that the list secrets call was taking up to 37 seconds.

_"I0726 16:50:58.010017       1 trace.go:201] Trace[436340495]: "Reflector ListAndWatch" name:pkg/mod/k8s.io/client-go@v0.18.6/tools/cache/reflector.go:125 (26-Jul-2021 16:50:00.914) (total time: 37095ms):
Trace[436340495]: ---"Objects listed" 37035ms (16:50:00.950)
Trace[436340495]: [37.095120496s] [37.095120496s] END"_

**Describe the solution you'd like**
We see a couple of options here that would be helpful:

A) **Allow the watch-namespace flag to accept multiple values.** We have several deployments of ingress-nginx in our clusters. The namespaces for these deployments are the only namespaces that the aws-load-balancer-controller should need to watch. Because the existing flag only accepts one value, to use it we would have to deploy the aws-load-balancer-controller once per namespace. We'd like to avoid doing this. 

[...]

PR #2550, opened 2022-03-14
monitor secret resources ony if necessary

kishorj:
### Issue
Fixes: #2159

### Description

The controller until this fix required access to all secret resources in the cluster. The secret resources were used for OIDC feature for the clientID and clientSecret stored in the secret resource. The controller setup a watch during initialization in order to reconcile the affected ingress group and and depending on the number of secrets in the cluster, there is a potential for slower initialization and increased resource usage.

The secret access is not required when OIDC feature is not in use. Furthermore, we can restrict the controller to specific secrets required for the OIDC feature. This PR contains the following changes:
* removes the default watch for all secret resources in the cluster
* create watch for specific secret resource if required for OIDC feature
* do not configure rbac permissions for secrets by default
* provide helm chart configuration options for specific secret resources

### Tests
- helm install does not create any permissions for secrets resource by default
- controller works without secret permissions; is able to reconcile ingress/service resources
- if oidc is configured, reconcile fails without rbac configuration
- reconcile works eventually if rbac gets configured
- any change to the secret resource causes affected ingress to be reconciled
- if rbac permission is revoked, controller is not able to reconcile the affected ingresses until the permission is restored

### Checklist
- [x] Manually tested


helm/aws-load-balancer-controller/README.md
```diff
@@ -234,3 +234,5 @@ The default values set by the application itself can be confirmed [here](https:/
...
 | `serviceMonitor.additionalLabels`              | Labels to add to the service account                                                                     | `{}`                                                                               |
 | `serviceMonitor.interval`                      | Prometheus scrape interval                                                                               | `1m`                                                                               |
+| `clusterSecretsPermissions.allowAllSecrets`    | If `true`, controller has access to all secrets in the cluster.                                          | `false`                                                                            |
+| `clusterSecretsPermissions.secretNames`        | Names of the specific secret resources controller should have access to for OIDC feature                 | `[]`                                                                               |
```
M00nF1sh, line 238: how will `secretNames` work given it can be multiple namespaces

helm/aws-load-balancer-controller/templates/rbac.yaml
```diff
@@ -57,8 +57,21 @@ rules:
...
   verbs: [get, list, watch]
+{{- if .Values.clusterSecretsPermissions.allowAllSecrets }}
+- apiGroups: [""]
+  resources: [secrets]
+  verbs: [get, list, watch]
+{{- else if .Values.clusterSecretsPermissions.secretNames }}
+- apiGroups: [""]
+  resources: [secrets]
+{{- with .Values.clusterSecretsPermissions.secretNames }}
+  resourceNames:
```
M00nF1sh, line 70: will resourceNames supports namespace as well?

pkg/k8s/secrets_manager.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+var _ SecretsManager = &defaultSecretsManager{}
+
+type defaultSecretsManager struct {
+	mutex        sync.Mutex
+	secretMap    map[types.NamespacedName]*secretItem
+	eventHandler handler.EventHandler
+	clientSet    kubernetes.Interface
+	queue        workqueue.RateLimitingInterface
```
M00nF1sh, line 46: seems this queue is not initialized.

pkg/k8s/secrets_store.go
```diff
@@ -0,0 +1,90 @@
...
+	queue        workqueue.RateLimitingInterface
+	eventHandler handler.EventHandler
+}
+
+// Add adds the given object to the accumulator associated with the given object's key
+func (s *SecretsStore) Add(obj interface{}) error {
+	if err := s.store.Add(obj); err != nil {
+		return err
+	}
+	s.eventHandler.Create(event.CreateEvent{Object: obj.(*corev1.Secret)}, s.queue)
```
M00nF1sh, line 35: i prefer to use a channel, and make the controller watch events from the channel instead.

helm/aws-load-balancer-controller/templates/rbac.yaml
kishorj: This is for convenience. We should require users to create a namespaced role and role binding for better security.

pkg/k8s/secrets_manager.go
kishorj: this is not used by the secrets eventhandler

helm/aws-load-balancer-controller/templates/rbac.yaml
kishorj: removed the secretNames option

pkg/k8s/secrets_manager.go
kishorj: used channel instead

pkg/k8s/secrets_store.go
kishorj: updated

Issue #2570, opened 2022-03-21
ensure empty affinity chart value results in empty affinity pod policies


BryanStenson-okta:
**Is your feature request related to a problem?**
I'm frustrated that when i set `affinity` to `""`, the chart installs an antiAffinity stanza....this makes setting an "empty" affinity nearly impossible.

**Describe the solution you'd like**
I'd like the helm chart to be updated so that an empty `affinity` value in values.yaml results in no pod affinity -- for the sake of feature parity, the affinity value set in the values.yaml could match the "default" antiAffinity rule now set if omitted.

**Describe alternatives you've considered**
A description of any alternative solutions or features you've considered.

kishorj:
@BryanStenson-okta, the current pod anti affinity depends on the label key `app.kubernetes.io/name`, and the chart name as the value. I'm not currently aware of an easier way to move it to the values.yaml. If we can make the template work via values.yaml, it would be great.

Would an alternate parameter work help with your use case? For example, `noAffinity` if set to `true`  result in empty affinity configuration for the deployment?

aRobinson-R7:
@BryanStenson-okta I saw your issue / PR for topologySpreadConstraints https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/2571
do these work together? does pod-affinity need to be disabled for this to work with Karpenter?

BryanStenson-okta:
Karpenter ignores pods which have `antiAffinity` set...I was thinking of "hacking" the current chart to set an "affinity" rule to match something global for my cluster (arch, or os-type, etc).

BryanStenson-okta:
> Would an alternate parameter work help with your use case? For example, `noAffinity` if set to `true` result in empty affinity configuration for the deployment?

I recognize the difficulty of setting `antiAffinity` based on dynamic chart name, but an additional `noAffinity` key feels like a confusing option (bad design) next to an `affinity` key.

I don't have a great suggestion/alternative here.  But I will say, when `affinity: {}`, it's confusing/frustrating an antiAffinity rule is put in place.

Maybe we change the API of `affinity` to be multi-typed?  If value of `affinity` is a string and that string is "default", setup the current antiAffinity rules.  This would be the default in `values.yaml`.

But, if `affinity` is a map (including an empty one), set the node affinity details directly from the map.

In this way, we avoid having to maintain both `affinity` and `noAffinity` keys....

kishorj:
closing, since we merged the fix.

PR #2576, opened 2022-03-22
helm chart update ; adding support to set affinity in podSpec to the empty map

stensonb:
### Issue

#2570 

### Description

This PR extends the use of the `affinity` key.

`affinity` can now be either a string or map.  if a string, and if that string is "default", then set an antiAffinity policy on the podSpec matching the current helm chart's behavior.  this is the default, as set in `values.yaml`.

if `affinity` is a map, use this map exclusively to set the affinity policy on the podSpec.

This logic will allow for an easy-to-understand way of disabling any/all affinity rules set by the chart.

### Checklist
- [x] Manually tested


kishorj (approved):

kishorj:
Please rebase after PR 2577 gets merged.

Issue #2601, opened 2022-04-06
[chart] Don't set prometheus.io pod annotations when ServiceMonitor is enabled


z0rc:
**Describe the bug**
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/v2.4.1/helm/aws-load-balancer-controller/templates/deployment.yaml#L25-L26 are set always. But with ServiceMonitor enabled, these labels still present. This results in double scraping, as prometheus discovers both pod annotations and ServiceMonitor as independent scraping targets.

**Steps to reproduce**
Enable ServiceMonitor via helm values, observe, that prometheus.io annotations still set on pods.

**Expected outcome**
When ServiceMonitor is enabled, prometheus.io pod annotations should be omitted.

**Environment**

* AWS Load Balancer controller version: 2.4.1
* Kubernetes version: 1.22
* Using EKS (yes/no), if so version? Yes

**Additional Context**: None

tamboliasir1:
Hi @z0rc.. Hi I would like to work on this. Just wanted to confirm so I wanted to check serviceMonitor is enabled or not from values if it is enabled then omit Prometheus annotations right?

z0rc:
@tamboliasir1 you are correct, wrap prometheus.io annotations in if block and enable annotations when `.Values.serviceMonitor.enabled` is false.

tamboliasir1:
Got it.. Thank you @z0rc

tamboliasir1:
Hi @z0rc .. Just wanted to confirm condition will look something like this right?
`{{- if not .Values.serviceMonitor.enabled }}
prometheus.io/scrape: "true"
prometheus.io/port: "{{ (split ":" .Values.metricsBindAddr)._1 | default 8080 }}"
{{- end}}`

z0rc:
@tamboliasir1 aside from github formatting, this looks okay. But it's better to create a PR and do a code review there.

tamboliasir1:
Hi @z0rc I have created a PR please review. Thanks

PR #2606, opened 2022-04-11
Updated deployment for prometheus.io annotations when sericeMonitor i…

tamboliasir1:
Fixes: #2601

Made changes in deployment for #2601 

Thank you for this opportunity.


kishorj:
@tamboliasir1, thanks for your contribution.
Could you also document the tests you ran?
You could run helm template with the serviceMonitor.enabled configuration and verify the template generated is as per the expectation.

kishorj (approved):

PR #2634, opened 2022-05-01
Allow negative and duplicate group.orders

### LINKED ISSUE #1875: Specifying group order smaller than the default (i.e. 0)
## Problem
I have an ingress with a rule that needs to be evaluated first in the ALB. 
In order to ensure that, I would have to assign group orders to all other ingresses, so that this particular ingress has the smallest order which is "0". 
It would be convenient if we could set the order such that one ingress has the smallest order instead of specifying the order for all other present and future ingresses.

I faced this issue while creating a common ingress for redirecting HTTP to HTTPS for all routes, and some common annotations instead of adding boilerplate config to each ingress. ( This request/issue is already tracked in #1768 ) 

## Proposed Solution
If a new order is introduced which is smaller than the default, it would solve the problem. 
e.g. order `-1`, the default would still be `0` so all rules will be added after that. 

## Current Solution / Workaround
By default the rules are ordered based on the lexical order of Ingress’s namespace/name.
I created a dummy namespace (e.g. `1-ns`) such that this ingress rule is always evaluated first and created this ingress in that namespace. 

## Example Ingress
```
kind: Ingress
metadata:
  name: dev-loadbalancer-default
  namespace: 1-ns
  annotations:
    alb.ingress.kubernetes.io/group.name: "dev"
    alb.ingress.kubernetes.io/tags: Environment=dev,Name=dev
    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS":443}]'
    alb.ingress.kubernetes.io/actions.ssl-redirect: '{"Type": "redirect", "RedirectConfig": { "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}'
spec:
  rules:
    - http:
        paths:
          - path: /*
            pathType: Prefix
            backend:
              service:
                name: ssl-redirect
                port:
                  name: use-annotation
```

### LINKED ISSUE #2629: Allow non-unique `group.order` to support more options around rule evaluation order
**Problem**

I would like to have more control over rule evaluation order.

My goal is define a common Ingress (and set of rules) that get evaluated first. 

In this instance, it would be to define global deny policies for urls such as `/healthz` and `/readiness` (that return http forbidden)

I would prefer to only define this once (not per `Ingress`) to avoid getting closer to rule limits on AWS.

---

Is there a reason that duplicate `group.order` values should be an error, and why it can't instead fall back to lexical ordering?

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/46668f11bdc548a3df60d1a921b0e70cbf0f91f1/pkg/ingress/group_loader.go#L254

I would like to define have the option to define duplicate `group.order` values.

Example:
- Global Deny Monitoring Rules would get `group.order: 0`
- All other micro-services we run run would get `group.order: 10`

My initial impl. followed https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1875 although I didn't spot that it was based on the namespace too. 

For my use-case, creating a namespace to deal with this isn't something I would like todo (it's quite a drastic workaround).

Since we run lots of applications and use the `ip` mode, making use of `group.order` would be troublesome as well, since they to be unique.

jtdoepke:
### Issue

- Fixes #2629
- Fixes #1875

### Description

This commit relaxes the restrictions around the `group.order` annotation, allowing both negative and duplicate orders. Duplicate orders are sorted lexicographically.

### Checklist
- [ x] Added tests that cover your change (if possible)
- [ x] Added/modified documentation as required (such as the `README.md`, or the `docs` directory)
- [ ] Manually tested


test/e2e/ingress/multi_path_backend.go
```diff
@@ -32,7 +32,7 @@ type BackendConfig struct {
 
 type MultiPathIngressConfig struct {
 	GroupName  string
-	GroupOrder int32
+	GroupOrder int64
```
kishorj, line 35: any reason for this change?

jtdoepke: Because it's an int64 [here](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/9935d4847f11e2ac07d50e734e4c5bb770f896b5/pkg/ingress/group_loader.go#L263). I figured they should be the same type, even if changing this line doesn't effect the test passage either way right now.

kishorj: cool, thanks

pkg/ingress/group_loader.go
```diff
@@ -3,22 +3,20 @@ package ingress
...
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/annotations"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/k8s"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sort"
-	"strings"
 )
 
 const (
 	defaultGroupOrder  int64 = 0
-	minGroupOrder      int64 = 1
```
M00nF1sh, line 20: can we make it within range -1000~1000. I think this should sufficient for all use cases i can imaging.
We can extend this if there is any valid use case that need access beyond this range.

enforce limits on fields like this can avoid us introduce any backwards incompatible changes. (e.g. some might start set a order with max int64 and we later accidentally set it to int32 in some future release and break them)

M00nF1sh: all rest code lgtm

jtdoepke: Done :+1:

pkg/ingress/group_loader.go
```diff
@@ -260,22 +260,17 @@ func (m *defaultGroupLoader) sortGroupMembers(members []ClassifiedIngress) ([]Cl
...
-	explicitOrders := sets.NewInt64()
 	for _, member := range members {
 		var order = defaultGroupOrder
 		exists, err := m.annotationParser.ParseInt64Annotation(annotations.IngressSuffixGroupOrder, &order, member.Ing.Annotations)
 		if err != nil {
 			return nil, errors.Wrapf(err, "failed to load Ingress group order for ingress: %v", k8s.NamespacedName(member.Ing))
 		}
 		if exists {
-			if order < minGroupOrder || order > maxGroupOder {
+			if order <= minGroupOrder || order >= maxGroupOder {
```
kishorj, line 270: lets make the range inclusive i.e: [1000, -1000]
Otherwise, the change will be backwards incompatible

PR #2645, opened 2022-05-09
helm: Add network policy to aws-load-balancer-controller

jdheyburn:
### Issue

N/A

### Description

Deploys NetworkPolicy resource if the configuration is enabled

I am happy to change the default rules that get created - I found that when I deployed this it was a minimum that I needed when there was a default deny all network policy enabled for the cluster.

### Checklist
~- [ ] Added tests that cover your change (if possible)~
- [x] Manually tested

~- [ ] Backfilled missing tests for code in same general area :tada:~
~- [ ] Refactored something and made the world a better place :star2:~


jdheyburn:
@kishorj Is it possible to get an ok-to-test on this? Thanks!

jdheyburn:
@kishorj @M00nF1sh , is it possible for a review on this when you are available? Thanks!

helm/aws-load-balancer-controller/README.md
```diff
@@ -250,3 +250,7 @@ The default values set by the application itself can be confirmed [here](https:/
...
+| `networkPolicy.enabled`                        | If `true`, creates a NetworkPolicy with minimal connectivity opened                                      | `false`                                                                            |
+| `networkPolicy.annotations`                    | Annotations to add to NetworkPolicy                                                                      | `{}`                                                                               |
+| `networkPolicy.egressRules`                    | Define additional egress network policy rules                                                            | `[]`                                                                               |
+| `networkPolicy.ingressRules`                   | Define additional ingress network policy rules                                                           | `[]`                                                                               |
```
johngmyers, line 256:
```suggestion
| `networkPolicy.egressRules`                    | Defines additional egress network policy rules.                                                            | `[]`                                                                               |
| `networkPolicy.ingressRules`                   | Defines additional ingress network policy rules.                                                           | `[]`                                                                               |
```

johngmyers (commented):
Some minor nits.

Doesn't it need egress to the AWS API? How is that traffic permitted?

jdheyburn:
> Doesn't it need egress to the AWS API? How is that traffic permitted?

@johngmyers Usually on clusters where you restrict network access (via deny all rule) you would permit external traffic on port 443 everywhere before that rule - or you might not, in which case you would add a rule here for the same.

I don't want to open up 443 by default, I'll leave that up to the user who has context on their setup.

Hope that answers the question :)

johngmyers:
The description of `networkPolicy.enabled` states "If `true`, creates a NetworkPolicy with minimal connectivity opened". But it doesn't open minimal connectivity; it opens less than minimal connectivity.

At the very least, the description needs to match the implementation.

Now I consider a network policy that allows everything unrestricted egress to any port on the internet to be a strange thing.

I suppose a NetworkPolicy could be tighter than allow-all-internet-egress if it knew the IPs of the relevant Endpoints. One could perhaps default `egressRules` to an allow-all-internet one, so it would not take effect if `egressRules` were explicitly specified.

jdheyburn:
Thanks for the feedback!

In the PR I have the following which explains what is considered minimal default connectivity. I can update the description to include that connectivity to AWS is required and that it is up to the user to add this.

>   By default the following connectivity is opened:
>   - ingress/egress other aws-load-balancer-controller pods with matching selector labels
>   - ingress to pods on .Values.webhookBindPort

I am not advocating for unrestricted egress to any port. I mention opening egress on port 443 as an example, but it would be up to the user to decide how they want to open connectivity egress. An opinionated and rigid network policy should not make that decision for them.

Given that IP addresses for endpoints can change, I would not want them baked into a helm-chart.

jdheyburn:
@kishorj @M00nF1sh Please could you review? Thank you in advance!

Issue #1747, opened 2020-12-29
NoCredentialProviders: no valid providers in chain


v1nc3nt27:
Hey there,

I'm running a self managed (kubeadm) cluster on EC2 instances and running into some problems with installing the load balancer controller. I've created the [IAM policy manually](https://github.com/aws/eks-charts/tree/master/stable/aws-load-balancer-controller#setup-iam-manually) and added it to my worker node IAM role. The LB controller was installed with the helm chart. I first encountered the problem that the EC2Metadata couldn't be read so I added `vpcId` and `region` parameters and the controller came up fine. However, when deploying the game-2048 example from [here](https://docs.aws.amazon.com/eks/latest/userguide/alb-ingress.html), the controller gives me the following status:

`{"level":"error","ts":1609246673.333103,"logger":"controller","msg":"Reconciler error","controller":"ingress","name":"ingress-2048","namespace":"game-2048","error":"couldn't auto-discover subnets: NoCredentialProviders: no valid providers in chain. Deprecated.\n\tFor verbose messaging see aws.Config.CredentialsChainVerboseErrors"}`

Hence, I've added the annotation `alb.ingress.kubernetes.io/subnets:` with my public subnets to the ingress controller. The error changes to:

`{"level":"error","ts":1609250899.5429265,"logger":"controller","msg":"Reconciler error","controller":"ingress","name":"ingress-2048","namespace":"game-2048","error":"NoCredentialProviders: no valid providers in chain. Deprecated.\n\tFor verbose messaging see aws.Config.CredentialsChainVerboseErrors"}`

It's obvious that the auto discovery of meta data is not working, but I don't see why and also I don't really understand what is missing. Is there anything I've overlooked?

Here is also the initial output of the controller logs:

> {"level":"info","ts":1609243280.836659,"msg":"version","GitVersion":"v2.1.0","GitCommit":"f95827224a66cae20a1af999d8ef1d46ec462547","BuildDate":"2020-12-01T19:27:50+0000"}
> {"level":"info","ts":1609243280.8867862,"logger":"controller-runtime.metrics","msg":"metrics server is starting to listen","addr":":8080"}
> {"level":"info","ts":1609243280.8893495,"logger":"setup","msg":"adding health check for controller"}
> {"level":"info","ts":1609243280.889599,"logger":"controller-runtime.webhook","msg":"registering webhook","path":"/mutate-v1-pod"}
[...]

PR #2664, opened 2022-05-25
Updated run-e2e-test script to remove eksctl iamserviceaccount creation

cgchinmay:
### Issue
NA

### Description
Removed eksctl usage to create iamserviceaccount from run-e2e-test.sh, as in EKS Beta environment, it's not available.
Update Instance metadata and set PutHtttpResponseHopLimit as 2, now that we are adding required IAM policies directly to the NodeInstanceRole. 
If we don't set response hop limit to 2, then we see following [error](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1747)

### Checklist
- [ ] Manually tested


M00nF1sh (approved):
lgtm

scripts/run-e2e-test.sh
```diff
@@ -18,29 +18,78 @@ function toggle_windows_scheduling(){
   done
 }
 
-echo "Cordon off windows nodes"
+ROLE_NAME="AWS-LB-CONTROLLER-ROLE"
```
kishorj, line 21: I recommend using a role name prefix, and attach the e2e test id as well. This will helps us run multiple tests in parallel.

cgchinmay: yes will update it

cgchinmay: Updated

PR #2675, opened 2022-06-03
added additional tags filter

### LINKED ISSUE #2674: Add additional subnet tags filter
EKS workload is running in availability zones and local zones. Local and availability zone subnets use `kubernetes.io/role/internal-elb = 1` tags and alb ingress fails to spin up load balancer with `Failed build model due to couldn't auto-discover subnets: subnets in multiple locales: [availability-zone local-zone]` error.

I would like to add an ability to provide extra tags in alb ingress annotation for subnets filtering to limit alb to either local or availability zone subnets (or any other subnets type)

Now I have to hardcode subnet names for each environment in [alb.ingress.kubernetes.io/subnets](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/ingress/annotations/#subnets), but their names and amount differ from env to env, from project to project

AndrewChubatiuk:
### Issue
Fixes #2674

### Description
Added `subnets-tags` annotation to be able to limit subnets, where to spin up a load balancer

### DIFF
--- a/pkg/annotations/constants.go
+++ b/pkg/annotations/constants.go
@@ -13,6 +13,7 @@ const (
 	IngressSuffixIPAddressType                = "ip-address-type"
 	IngressSuffixScheme                       = "scheme"
 	IngressSuffixSubnets                      = "subnets"
+	IngressSuffixSubnetsTags                  = "subnets-tags"
 	IngressSuffixCustomerOwnedIPv4Pool        = "customer-owned-ipv4-pool"
 	IngressSuffixLoadBalancerAttributes       = "load-balancer-attributes"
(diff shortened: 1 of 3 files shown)


kishorj:
@AndrewChubatiuk, while we will be able to address the pain point with auto-discovery with this change, the burden is still on the end user to tag all of their subnets appropriately. My recommendation is to restrict the auto-discovery to `ZoneType` `availability-zone` by default. That way, subnets can still be specified for the other zones manually via annotation.

We can instead add an annotation `az-type` which defaults to `availability-zone`, with the options for `wavelength-zone` and `local-zone`. The auto-discovery can then be limited to the subnets from the configured zone.

PR #2723, opened 2022-07-08
e2e test changes to support china regions

cgchinmay:
### Issue
NA

### Description
Cannot run e2e tests in china region. The color teller application image is region specific so running e2e tests in china fails to pull this image. So moved this image to public docker registry so that it can be pulled in china region. 

### Checklist
- [ ] Manually tested

### DIFF
--- a/test/framework/manifest/fixed_response_service_builder.go
+++ b/test/framework/manifest/fixed_response_service_builder.go
@@ -2,11 +2,13 @@ package manifest
 
 import (
 	"fmt"
+
 	"github.com/aws/aws-sdk-go/aws"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/aws-load-balancer-controller/test/framework/utils"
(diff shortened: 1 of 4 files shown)


test/framework/utils/constants.go
```diff
@@ -0,0 +1,5 @@
+package utils
+
+const (
+	COLORTELLER_IMAGE = "chinmay5j/colorteller:v1"
```
kishorj, line 4: ControllerName for the const name

kishorj: also why not use the tag latest?

cgchinmay: just to know exact version if anything changes on the image, so usually I don't use latest but in this case I can change it.

test/framework/manifest/fixed_response_service_builder.go
```diff
@@ -93,7 +95,7 @@ func (b *fixedResponseServiceBuilder) buildDeployment(namespace string, name str
 					Containers: []corev1.Container{
 						{
 							Name:  "app",
-							Image: "970805265562.dkr.ecr.us-west-2.amazonaws.com/colorteller:latest",
+							Image: utils.COLORTELLER_IMAGE,
```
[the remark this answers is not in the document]
cgchinmay: I did run all tests under e2e, didn't have any issues

Issue #2042, opened 2021-05-31
controller is stuck if one ingress has empty request condition


aandrushchenko:
Version: 2.1.1

Observed:
one of the ingresses has annotation:
```
alb.ingress.kubernetes.io/conditions.account-service-lumen-100716933: |                                                                                                                                                
  [{"field":"http-header","httpHeaderConfig":{"httpHeaderName": "HEADER","values":[""]}}]
```
the other ingress is not getting reconciled with error:
```
Failed deploy model due to ValidationError: A condition value cannot be empty
```

Expected:
broken annotations on one of the ingresses should not prevent reconciliation of the others

kishorj:
@aandrushchenko, did you group the ingresses together into a single ingress group? if so, the error would impact the entire ingress group until the issue is fixed. 

In recent v2.2.0 release, we have made improvements to tolerate missing services, but validation error is still there. 

If the ingresses are independent or belong to different groups, error on one ingress would not affect other ingresses.

aandrushchenko:
@kishorj you are right about the group, i forgot to mention this. shall we keep this issue open or there's already a known issue reported and prioritised?

M00nF1sh:
we welcome PRs from community given this is a good first issue. 
We'll also evaluate and prioritize this internally if not taken by community members.

r-erema:
I would like to take it
/assign @r-erema

r-erema:
@aandrushchenko 
> one of the ingresses has annotation...
> the other ingress is not getting reconciled...

Could you provide full manifests of both ingresses, please?

PR #2735, opened 2022-07-24
Validate Ingress condition annotations

r-erema:
### Issue

#2042 

### Description

Earlier if at least 1 ingress in the group has an invalid annotation the rest of the ingresses in the group aren't took into account during building listener rules.

I've handled the condition validation error separately to prevent the termination of the process of the building listener rules for valid ingresses if such an error occurs.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ingress/config_types.go
+++ b/pkg/ingress/config_types.go
@@ -322,7 +322,7 @@ type RuleCondition struct {
 	SourceIPConfig *SourceIPConditionConfig `json:"sourceIPConfig"`
 }
 
-func (c *RuleCondition) validate() error {
+func (c *RuleCondition) Validate() error {
 	switch c.Field {
 	case RuleConditionFieldHostHeader:
 		if c.HostHeaderConfig == nil {
--- a/pkg/ingress/enhanced_backend_builder.go
+++ b/pkg/ingress/enhanced_backend_builder.go
@@ -165,7 +165,7 @@ func (b *defaultEnhancedBackendBuilder) buildConditions(_ context.Context, ingAn
 		return nil, err
 	}
 	for _, condition := range conditions {
-		if err := condition.validate(); err != nil {
+		if err := condition.Validate(); err != nil {
 			return nil, err
 		}
 	}
--- a/webhooks/networking/ingress_validator_test.go
+++ b/webhooks/networking/ingress_validator_test.go
@@ -3,6 +3,7 @@ package networking
 import (
 	"context"
 	awssdk "github.com/aws/aws-sdk-go/aws"
+	"github.com/go-logr/logr"
 	"github.com/golang/mock/gomock"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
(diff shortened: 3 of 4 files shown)


r-erema:
@M00nF1sh Hi, kindly reminder about PR. Please, please estimate when it can get to the review. Thanks!

pkg/ingress/config_types.go
```diff
@@ -238,7 +240,7 @@ type HTTPRequestMethodConditionConfig struct {
 
 func (c *HTTPRequestMethodConditionConfig) validate() error {
 	if len(c.Values) == 0 {
-		return errors.New("values cannot be empty")
+		return errors.Wrap(errConditionValidation, "values cannot be empty")
```
johngmyers, line 243: Prefer fmt.Errorf() and `%w` for wrapping errors.

But this isn't the use case for wrapping; one should probably use a custom error type and a constructor.

pkg/ingress/model_build_listener_rules.go
```diff
@@ -33,7 +33,15 @@ func (t *defaultModelBuildTask) buildListenerRules(ctx context.Context, lsARN co
...
 					WithLoadBackendServices(true, t.backendServices),
 					WithLoadAuthConfig(true))
-				if err != nil {
+				if errors.Is(err, errConditionValidation) {
+					t.logger.Error(err, "Ingress has invalid annotation for service, thus will be ignored",
+						"Ingress",
+						ing,
+						"Service name",
+						path.Backend.Service.Name,
+					)
```
johngmyers, line 42: This isn't the clearest error message. And logging an entire `ClassifiedIngress` is noisy.

Perhaps something like "Ignoring Ingress _NAMESPACE_/_NAME_: alb.ingress.kubernetes.io/conditions._SERVICE_ annotation: _ERROR_"

johngmyers (changes requested):
/remove-lifecycle rotten

johngmyers:
Would it be possible to put these checks in the validating admission webhook so such ingresses don't get created in the first place?

pkg/ingress/config_types.go
r-erema: Agree regarding to `fmt.Errorf()`.

> But this isn't the use case for wrapping

What' are cases for wrapping and what for creation custom type? It will be good to understand for me to make it properly in the future. Thanks!

pkg/ingress/model_build_listener_rules.go
r-erema: Sticking the same style of logging error in that project, where I noticed first go a common error message at the beggining and then entities involved in the error in key-values format, e.g.
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/a92e689dfe464f5b24784f398947e0fef31dc470/controllers/elbv2/eventhandlers/node.go#L92
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/a92e689dfe464f5b24784f398947e0fef31dc470/pkg/backend/endpoint_resolver.go#L187
I'll change your suggested error format a little, but anyway make it less noisy
```go
t.logger.Error(err, "Ignoring Ingress invalid annotation",
  "Ingress", fmt.Sprintf("%s/%s", ing.Ing.Namespace, ing.Ing.Name),
  "Service", path.Backend.Service.Name,
)
```

pkg/ingress/config_types.go
johngmyers: The canonical case for wrapping is when you received an error from doing a call and you want to add information about the context in which the error occurred. I tend to start the wrapping message with a gerund, for example "writing file %q".

A custom error type is for when the caller needs to programmatically check if a particular error condition occurred. In this case, the caller wants to know if there was an invalid rule condition.

pkg/ingress/model_build_listener_rules.go
johngmyers: Consider logging the annotation key. It might not be obvious that it's the `actions.` annotation this is referring to.

johngmyers:
If you want to explore the validating admission webhook path, the code is in `webhooks/networking/ingress_validator.go`.

Although the existing code uses generic errors, it would be better to return API validation errors from apimachinery. Examples of this can be found in my PR #2945.

pkg/ingress/config_types.go
r-erema: @johngmyers Thanks a lot!

pkg/ingress/model_build_listener_rules.go
r-erema: > It might not be obvious that it's the actions.

@johngmyers Probably you've meant `conditions.`, I'll add that info to error.

r-erema:
> Would it be possible to put these checks in the validating admission webhook so such ingresses don't get created in the first place?

@johngmyers Do we need **replace** existed validation flow, i.e. take validation from here
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/c94ffd3055892235c77a4024817367207bc6f94c/pkg/ingress/enhanced_backend_builder.go#L167-L171
and put to the validating admission webhook?

Or just **add** the same validation to the validating admission webhook, i.e. we will have same validation in 2 places

johngmyers:
I think it would be fine just being in the webhook.

Putting the checks in both places would cover cases where the webhook wasn't able to act, likely because it wasn't configured when the Ingress was created. I don't think the added complexity would be worth it.

r-erema:
> Would it be possible to put these checks in the validating admission webhook so **such ingresses don't get created** in the first place?

@johngmyers I thought 1 thing - do we actually desire to reject the whole ingress in case of at least 1 bad condition?

Or do we just want to reject the path related to the invalid condition but keep the ingress with the rest of the good conditions? If so, I guess we need involve mutation to exclude wrong paths from the ingress request. Or it will be overhead, and it's not worth making such complexity and just rejecting the whole ingress is ok? :-)

r-erema:
@johngmyers PR has been updated: validation is shifted to the validation webhook.

One question: I have the failed job "pull-aws-load-balancer-controller-unit-test"
Screenshot from 2023-01-28 08-47-28
with error:
`vet: webhooks/networking/ingress_validator_test.go:895:41: NullLogger not declared by package log`
Maybe you're familiar with the problem? Otherwise I will dig in it by my myself

webhooks/networking/ingress_validator_test.go
```diff
@@ -785,3 +785,119 @@ func Test_ingressValidator_checkIngressClassUsage(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			annotationParser := annotations.NewSuffixAnnotationParser("alb.ingress.kubernetes.io")
+			classAnnotationMatcher := ingress.NewDefaultClassAnnotationMatcher("alb")
+			v := &ingressValidator{
+				annotationParser:              annotationParser,
+				classAnnotationMatcher:        classAnnotationMatcher,
+				disableIngressGroupAnnotation: tt.fields.disableIngressGroupAnnotation,
+				logger:                        &log.NullLogger{},
```
johngmyers, line 893:
```suggestion
				logger:                        logr.New(&log.NullLogSink{}),
```

johngmyers:
You should change the title of the PR. Something like "Validate Ingress condition annotations".

webhooks/networking/ingress_validator_test.go
r-erema: The issue was in depricated `&log.NullLogger{}`, need to be used `logr.Discard()` instead. Thanks anyway!

r-erema:
>  You should change the title of the PR. Something like "Validate Ingress condition annotations".

@johngmyers Done

> I have the failed job "pull-aws-load-balancer-controller-unit-test"

Fixed as well

pkg/ingress/enhanced_backend_builder.go
```diff
@@ -164,11 +164,7 @@ func (b *defaultEnhancedBackendBuilder) buildConditions(_ context.Context, ingAn
 	if err != nil {
 		return nil, err
 	}
-	for _, condition := range conditions {
```
kishorj, line 167: Please keep this validation block in place. We've seen feature requests to optionally disable validation webhook for ingress. There would also be a case where the invalid ingress already exists before the controller got updated. We still want to error out from model builder if validation fails.

kishorj (commented):
Thanks for your contribution. We still need the validation block in buildConditions(). I'm fine with the changes otherwise.

pkg/ingress/enhanced_backend_builder.go
johngmyers: I do not think we should support disabling our validation webhook for ingress. If someone disables it, they can deal with the fallout of configurations the webhook would prevent.

We should not add unnecessary complexity.

r-erema: As for me I'm not a fun of duplication of logic/checkings etc, so I'd prefer to keep validation in 1 place. @kishorj @johngmyers can we meet consolidate decision(maybe by voting or etc.)?

kishorj: We discussed further on this matter in the provider-aws meeting.  Since there is a possibility of having ingress resources from a prior version (before the webhook validation), we still do these validation for 2.5.0. We can clean up the validation in future releases.

r-erema: Validation returned back

kishorj (approved):

Issue #2650, opened 2022-05-13
[Feature] TargetGroupBinding for TargetGroup outside of ALB Controller's VPC


KamikazeZirou:
## Is your feature request related to a problem?

I'd like to bind pods to a TargetGroup in a VPC outside the EKS cluster. A similar feature was added in #1862, but it does not work in my case.

The infrastructure configuration I would like to build is shown below.
![image](https://user-images.githubusercontent.com/33488934/168240993-e11b26c0-1857-4086-8592-fc4a12f487b0.png)

When registering the pod's IP address to the TargetGroup that is in a VPC outside the EKS cluster, it is necessary to set "all" for the availability zone. [In the current implementation](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/4c33010f204fef2868a396adf7fac6d13d027250/pkg/targetgroupbinding/resource_manager.go#L373-L402), "all" is set if the pod's IP address is not in the VPC CIDRs configured in the ALB Controller. In my case, the VPC of the ALB Controller is the same as the VPC of the EKS cluster, so "all" is not set.

## Describe the solution you'd like

Set "all" to the availability zone if the pod's IP address is in the VPC of the TargetGroup.

## Describe alternatives you've considered

The problem can be avoided by deploying two ALB Controllers on one cluster.
However, I do not know if it is possible to deploy two ALB Controllers.

## Note

I think that this issue is the same as aws/eks-charts#745.

kishorj:
@KamikazeZirou, are the VPCs in the same account or different accounts?

KamikazeZirou:
@kishorj 

> @KamikazeZirou, are the VPCs in the same account or different accounts?

They are in the same account.

PR #2736, opened 2022-07-26
Add "AvailabilityZone:all" parameter to IP targets when TargetGroup is outside of Controller's VPC

a2ush:
### Issue

#2650

### Description

Currently, when you deploy targetgroupbinding with ALB/NLB TargetGroup which is outside of Loadbalancer Controller's VPC, the Controller cannot do RegisterTargets API because it doesn't add "AvailabilityZone:all" parameter to IP targets.

[Loadbalancer Controller Log]
```
Ex)
{"level":"error","ts":1658843126.3276293,"logger":"controller-runtime.manager.controller.targetGroupBinding","msg":"Reconciler error","reconciler group":"elbv2.k8s.aws","reconciler kind":"TargetGroupBinding","name":"my-tgb","namespace":"default","error":"ValidationError: The Availability Zone is required for IP address '192.168.28.187' because it is not in the VPC\n\tstatus code: 400, request id: b228b6b8-..."}
```
https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_TargetDescription.html
 > If the target type is ip and the IP address is in a subnet of the VPC for the target group, the Availability Zone is automatically detected and this parameter is optional. If the IP address is outside the VPC, this parameter is required.

This PR helps Loadbalancer Controller to add "AvailabilityZone:all" parameter to IP targets in this situation, by comparing TargetGroup's VPC with Loadbalancer Controller's VPC.
That is, you will be able to register pod's IP address to other VPC's TargetGroups.

New Loadbalancer Controller's log
```
Ex)
{"level":"info","ts":1658812216.264621,"msg":"registering targets","arn":"arn:aws:elasticloadbalancing:<Region>:<AccountID>:targetgroup/my-tgb/1234567890","targets":[{"AvailabilityZone":"all","Id":"192.168.36.77","Port":80},{"AvailabilityZone":"all","Id":"192.168.43.133","Port":80},{"AvailabilityZone":"all","Id":"192.168.53.216","Port":80}]}
{"level":"info","ts":1658812216.5066864,"msg":"registered targets","arn":"arn:aws:elasticloadbalancing:<Region>:<AccountID>:targetgroup/my-tgb/1234567890"}
```

### Checklist
- [X] Manually tested

### DIFF
--- a/pkg/targetgroupbinding/resource_manager.go
+++ b/pkg/targetgroupbinding/resource_manager.go
@@ -382,6 +382,10 @@ func (m *defaultResourceManager) registerPodEndpoints(ctx context.Context, tgARN
 	if err != nil {
 		return err
 	}
+	tgInfo, err := m.targetsManager.DescribeTargetGroup(ctx, tgARN)
+	if err != nil {
+		return err
+	}
 
 	sdkTargets := make([]elbv2sdk.TargetDescription, 0, len(endpoints))
 	for _, endpoint := range endpoints {
@@ -395,6 +399,9 @@ func (m *defaultResourceManager) registerPodEndpoints(ctx context.Context, tgARN
 		}
 		if !networking.IsIPWithinCIDRs(podIP, vpcCIDRs) {
 			target.AvailabilityZone = awssdk.String("all")
+		}
+		if *tgInfo.VpcId != m.vpcID {
+		        target.AvailabilityZone = awssdk.String("all")
 		}
 		sdkTargets = append(sdkTargets, target)
 	}
--- a/pkg/targetgroupbinding/targets_manager.go
+++ b/pkg/targetgroupbinding/targets_manager.go
@@ -27,6 +27,9 @@ type TargetsManager interface {
 
 	// List Targets from TargetGroup.
 	ListTargets(ctx context.Context, tgARN string) ([]TargetInfo, error)
+
+	// Describe TargetGroup.
+	DescribeTargetGroup(ctx context.Context, tgARN string) (*elbv2sdk.TargetGroup, error)
 }
 
 // NewCachedTargetsManager constructs new cachedTargetsManager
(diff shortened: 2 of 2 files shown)


daghsentinel:
Any chance of getting anyone to look into this? I am having the exact same problems as staed in the issue

tmehlinger:
I'm also running into this. I'd really appreciate someone reviewing and merging it.

pkg/targetgroupbinding/resource_manager.go
```diff
@@ -382,6 +382,10 @@ func (m *defaultResourceManager) registerPodEndpoints(ctx context.Context, tgARN
 	if err != nil {
 		return err
 	}
+	tgInfo, err := m.targetsManager.DescribeTargetGroup(ctx, tgARN)
```
kishorj, line 385: I'd prefer caching the target group vpc Id so we don't incur the DescribeTargetGroup call with every register.

M00nF1sh: +1, we should minimize DescribeTargetGroup calls in this critical path.

we can either add cache or mutate the targetGroupBinding spec during MutationWebhook to capture vpcID in spec/annotation.

I prefer to have a cache for now given its simper and will be a two way door.

tmehlinger:
@kishorj thanks for checking this out, much appreciated!

pkg/targetgroupbinding/resource_manager.go
a2ush: Thank you for your advice!
I agree that. But I'm afraid I'm not sure how to best implement a cache for the controller.

Could you give me some reference information (e.g. similar implement, sample code)?

pescetto:
Thanks for your work on this, any idea when this bug fix might get merged in? We're running into this also and can't use aws-load-balancer-controller until this gets resolved.

PR #2743, opened 2022-07-28
use a dedicated config to disable default pod affinity

kishorj:
## Issue
Alternate fix for issue #2570

### Description
This PR revert changes from #2576 and adds a separate helm value `configureDefaultAffinity` which controls whether to use the default affinity configuration for the controller pods. This value is true by default, so there is no changes in the behavior from prior version. In case of custom affinity configuration, this value gets ignored.

### Checklist
- [x] Manually tested

### DIFF
--- a/helm/aws-load-balancer-controller/README.md
+++ b/helm/aws-load-balancer-controller/README.md
@@ -157,7 +157,7 @@ eksctl delete iamserviceaccount --cluster <cluster-name> --namespace kube-system
 Chart release v1.2.0 and later enables high availability configuration by default.
 - The default number of replicas is 2. You can pass`--set replicaCount=1` flag during chart installation to disable this. Due to leader election, only one controller will actively reconcile resources.
 - The default priority class for the controller pods is `system-cluster-critical`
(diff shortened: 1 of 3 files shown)


helm/aws-load-balancer-controller/README.md
```diff
@@ -157,7 +157,7 @@ eksctl delete iamserviceaccount --cluster <cluster-name> --namespace kube-system
 Chart release v1.2.0 and later enables high availability configuration by default.
 - The default number of replicas is 2. You can pass`--set replicaCount=1` flag during chart installation to disable this. Due to leader election, only one controller will actively reconcile resources.
 - The default priority class for the controller pods is `system-cluster-critical`
-- Soft pod anti-affinity is enabled for controller pods with `topologyKey: kubernetes.io/hostname` if custom affinity is not configured
+- Soft pod anti-affinity is enabled for controller pods with `topologyKey: kubernetes.io/hostname` if you don't configure custom affinity or set `configureDefaultAffinity` to `true`
```
M00nF1sh, line 160: i think this should be "and" instead of "or"

M00nF1sh (approved):
/lgtm

PR #2756, opened 2022-08-10
feat: add default ingressclass annotation ✨

meysam81:
### Issue

A replacement for aws/eks-charts#792

### Description

Any ingressclass out there has the option to be the default ingressclass if the right value is passed to the chart. This is the reason behind this PR so that with passing a single value, we can have the ingressclass as the default inside the cluster.

### Checklist
- [x] Manually tested


meysam81:
@M00nF1sh
@kishorj 
@oliviassss 
@johngmyers

helm/aws-load-balancer-controller/values.yaml
```diff
@@ -319,3 +319,8 @@ clusterSecretsPermissions:
   # allowAllSecrets allows the controller to access all secrets in the cluster.
   # This is to get backwards compatible behavior, but *NOT* recommended for security reasons
   allowAllSecrets: false
+
+# whether or not this ingressclass is the default ingressclass inside the cluster
+controller:
```
kishorj, line 324: lets have in this structure
```
# ingressClassConfig contains configurations specific to the ingress class
ingressClassConfig:
    default: true
```
We will group the remaining ingress class configurations under this later.

meysam81:
@kishorj done ✅

kishorj (approved):

PR #2760, opened 2022-08-12
feat: Since replicaCount is 2 by default, `podDisruptionBudget.maxUnavailable: 1` also be set by default.

kahirokunn:
### Description

If podDisruptionBudget is set by default, Scale-in is enabled when using Cluster Autoscaler.
I think cheap by default is good.

### Checklist
- [x] Manually tested


kishorj:
@kahirokunn, thanks for your contribution. However, the pod disruption budget is an optional feature and at the moment we don't want to enforce it by default. Since there are possible consequences, we will let the end user make the appropriate choice for now in regards to this feature.

PR #2770, opened 2022-08-18
Allow configuring the actual cluster DNS domain for the TLS certificate

stephan2012:
### Description

Currently, the cluster DNS domain `cluster.local` is hard-coded in the Helm chart for the webhook TLS certificate. This change provides a config option that allows using the actual cluster domain.

This is a non-breaking change, i.e., the cluster DNS domain still defaults to `cluster.local`.

### Checklist
- [X] Manually tested

### DIFF
--- a/helm/aws-load-balancer-controller/values.yaml
+++ b/helm/aws-load-balancer-controller/values.yaml
@@ -78,7 +78,7 @@ configureDefaultAffinity: true
 # topologySpreadConstraints is a stable feature of k8s v1.19 which provides the ability to
 # control how Pods are spread across your cluster among failure-domains such as regions, zones,
 # nodes, and other user-defined topology domains.
-# 
+#
 # more details here: https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/
 topologySpreadConstraints: {}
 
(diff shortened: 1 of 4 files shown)


helm/aws-load-balancer-controller/values.yaml
```diff
@@ -104,6 +104,9 @@ additionalLabels: {}
 # Enable cert-manager
 enableCertManager: false
 
+# Cluster DNS domain (required for generating the TLS certificates)
+clusterDomain: cluster.local
```
kishorj, line 108: Lets group it together with clusterName in the values.yaml file.
Also, lets have it as follows:
```
# cluster contains configurations specific to the kubernetes cluster
cluster:
    domain: cluster.local
```
In future, we can nest the clusterName inside as well.

kishorj (commented):
Thanks for your contribution.

Please add the new values to the configurations table in the README.md file

helm/aws-load-balancer-controller/values.yaml
stephan2012: Thanks for your feedback, I have refactored it to `cluster.dnsDomain`.

kishorj (approved):

Issue #2738, opened 2022-07-27
ingressClass cannot be set as default when deployed using helm chart


robertpolasek-brt:
**Describe the bug**
In the helm chart there is ingressclass.yaml  file which is responsible for creation of ingressClass resource. Annotations are omitted from this file which prevents creation of default ingressClass as a part of helm chart. 

Default ingress class is specified by the annotation  `"ingressclass.kubernetes.io/is-default-class": "true"`.  

There should be a way to supply annotations for ingressClass from helm values and that way support default ingress class creation. 

* AWS Load Balancer controller version
 2.4.2  

* Kubernetes version
1.21

* Helm chart version
1.4.2

kishorj:
@robertpolasek-brt, thanks for raising the issue. Feel free to raise a PR to address this.

PR #2791, opened 2022-09-03
Annotations passed to IngressClass from Helm Values (allowing creation of a default ingressClass)

asim-reza:
### Issue
ingressClass cannot be set as default when deployed using helm chart [#2738](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2738)

### Description
In the helm chart there is a ingressclass.yaml file which is responsible for creation of ingressClass resource. Annotations were omitted from this file which prevented creation of default ingressClass as a part of helm chart.

Default ingress class is specified by the annotation "ingressclass.kubernetes.io/is-default-class": "true"

For this purpose the Value `ingressClassAnnotations` was created that defaults to `{}`. This is used to populate the IngressClass Annotations when the manifest is created from the helm template.

### Checklist
- [x] Manually tested


asim-reza:
Didn't add user ID to commit

Issue #2430, opened 2021-12-25
Auto-discovery of VPC


olemarkus:
**Is your feature request related to a problem?**
When deploying a cluster with LBC through IaC, it is not that easy to inject the VPC ID into the LBC Deployment resource. That is, however, necessary when LBC runs with IRSA and with metadata is unavailable.

**Describe the solution you'd like**
It would be nice if VPC could be discovered through tags the same way subnets are.

M00nF1sh:
/kind feature
we have been think about new ways to discover vpc, clusterName.
One possible way is to use the labels on nodes where the controller run

olemarkus:
There are no labels for this information today, but I could definitely help out adding at least a VPC label through Cloud Controller Manager.

cluster name is less of a problem since we do know that up front.

M00nF1sh:
we could get the region from the zone label possibly, and then find the vpcID by the instanceID of node(spec.providerID).
and hopefully we can get the clusterName from the tags on instance as well.

on the other hand, using tags on the vpc is also a good idea.

We'll evaluate which one is better.

kishorj:
@olemarkus, we could get the Node resource, extract the providerID and invoke the ec2 `DescribeInstances` API to get the VPC id. The controller already has permissions to access the Node resources from k8s, and IAM permissions for the `DescribeInstances` call and would not depend on the CCM putting the appropriate label.

olemarkus:
That sounds good to me.

olemarkus:
@M00nF1sh @kishorj any update on this?

/remove-lifecycle stale

PR #2824, opened 2022-10-03
Retrieve VPC ID from EC2 instance the controller is running on if possible

olemarkus:
### Issue

#2430

### Description

Right now either metadata has to be available or one has to pass VPC ID as a flag.
This PR tries to fetch the VPC ID from the EC2 API if it can infer the instance ID from the node name.

It uses the "new" node name convention from cloud controller manager, which works well for my use case. A follow up would be to try to fetch the node object and extract the instance ID from the provider ID.

This approach was tested running LBC on instances with IMDS max-hop-limit 1 and IRSA enabled and it ran successfully without setting the vpc id flag.

The LBC Deployment requires the following:
```
spec:
  template:
    spec:
      containers:
      - env:
        - name: NODENAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName
```

### Checklist
- [ ] Manually tested

If this approach looks good, I can do any of the above deemed necessary.

### DIFF
--- a/pkg/aws/cloud.go
+++ b/pkg/aws/cloud.go
@@ -1,6 +1,7 @@
 package aws
 
 import (
+	"fmt"
 	"net"
 	"os"
 	"strings"
@@ -8,8 +9,10 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	amerrors "k8s.io/apimachinery/pkg/util/errors"
 	epresolver "sigs.k8s.io/aws-load-balancer-controller/pkg/aws/endpoints"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/aws/metrics"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/aws/services"
@@ -70,13 +73,6 @@ func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud, 
 
 	metadataSess := session.Must(session.NewSessionWithOptions(opts))
 	metadata := services.NewEC2Metadata(metadataSess)
-	if len(cfg.VpcID) == 0 {
-		vpcId, err := metadata.VpcID()
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to introspect vpcID from EC2Metadata, specify --aws-vpc-id instead if EC2Metadata is unavailable")
-		}
-		cfg.VpcID = vpcId
-	}
 
 	if len(cfg.Region) == 0 {
 		region := os.Getenv("AWS_DEFAULT_REGION")
@@ -114,9 +110,19 @@ func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud, 
 		metricsCollector.InjectHandlers(&sess.Handlers)
 	}
 
+	ec2Service := services.NewEC2(sess)
+
+	if len(cfg.VpcID) == 0 {
+		vpcID, err := inferVPCID(metadata, ec2Service)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to introspect vpcID from EC2Metadata or Node name, specify --aws-vpc-id instead if EC2Metadata is unavailable")
+		}
+		cfg.VpcID = vpcID
+	}
+
 	return &defaultCloud{
 		cfg:         cfg,
-		ec2:         services.NewEC2(sess),
+		ec2:         ec2Service,
 		elbv2:       services.NewELBV2(sess),
 		acm:         services.NewACM(sess),
 		wafv2:       services.NewWAFv2(sess),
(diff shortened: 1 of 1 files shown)


pkg/aws/cloud.go
```diff
@@ -114,9 +108,41 @@ func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud,
...
 
+	ec2Service := services.NewEC2(sess)
+
+	if len(cfg.VpcID) == 0 {
+		nodeName := os.Getenv("NODENAME")
+		if strings.HasPrefix(nodeName, "i-") {
+			output, err := ec2Service.DescribeInstances(&ec2.DescribeInstancesInput{
+				InstanceIds: []*string{&nodeName},
+			})
+			if err != nil {
```
M00nF1sh, line 119: we should change our logic here, if we don't have permission to describe the instance, we still should be able to fallback to metadata path. 
similarly, if we check for metadata first then instance id(see my next comment), we should change the logic to not fall when metadata is not available

pkg/aws/cloud.go
```diff
@@ -114,9 +108,41 @@ func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud,
 		metricsCollector.InjectHandlers(&sess.Handlers)
 	}
 
+	ec2Service := services.NewEC2(sess)
+
+	if len(cfg.VpcID) == 0 {
+		nodeName := os.Getenv("NODENAME")
```
M00nF1sh, line 114: a nit comment, given this function is a bit long, 
maybe created a dedicated file like `pkg/aws/utils/vpc_id.go`, and have functions like
```
inferVPCID(sess, metadata), which invokes inferVPCIDFromNodeName
inferVPCIDFromNodeName(sess, nodeName)
```

M00nF1sh (commented):
overall looks good to me with some nit comments

pkg/aws/cloud.go
```diff
@@ -114,9 +108,41 @@ func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud,
...
+			if len(output.Reservations[0].Instances) != 1 {
+				return nil, errors.Wrap(err, "more than one instance returned")
+			}
+
+			vpcID := output.Reservations[0].Instances[0].VpcId
+			if vpcID != nil {
+				cfg.VpcID = *vpcID
+			}
+
+		} else {
```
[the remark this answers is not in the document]
olemarkus: Agree! Done.

pkg/aws/cloud.go
olemarkus: Also done.
Means less meaningful errors returned, but perhaps not very important anyway.

pkg/aws/cloud.go
olemarkus: Didn't spot this comment, but ended up doing pretty much the same you describe here as it makes the fallback logic easier. I put the function below `NewCloud` though.

pkg/aws/cloud.go
```diff
@@ -114,9 +108,19 @@ func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud,
 		metricsCollector.InjectHandlers(&sess.Handlers)
 	}
 
+	ec2Service := services.NewEC2(sess)
+
+	if len(cfg.VpcID) == 0 {
```
M00nF1sh, line 113: nit: i think we should use same check here, either `len(cfg.VpcID) ==0` or `vpcID == ""`

pkg/aws/cloud.go
```diff
@@ -126,6 +130,37 @@ func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud,
 	}, nil
 }
 
+func getVPC(metadata services.EC2Metadata, ec2Service services.EC2) string {
```
M00nF1sh, line 133: this breaks the existing experience as it hide the error if metadata is not available since it hides the error.
I think change to below will be better:
```
inferVPCID(metadata, ec2Service) (vpcID, error) {
	var errors []err
	vpcID, err := metadata.vpcID()
	if err == nil {
		return vpcID
	}
	errors = append(errors, errors.Wrap(err, "cannot infer VPCID from metadata"))
	vpcID, err := inferVPCIDFromNodeName(...)
	if err == nil {
		return vpcID
	}
	errors = append(errors, err)
	return "", errors.NewAggregate(errors) # https://github.com/kubernetes/apimachinery/blob/478dd6ed1ec9b4cc3459dcd4f1d772013ac0103c/pkg/util/errors/errors.go#L46
}

```

pkg/aws/cloud.go
```diff
@@ -126,6 +130,37 @@ func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud,
...
+
+	vpcId, err := metadata.VpcID()
+	if err == nil {
+		return vpcId
+	}
+
+	nodeName := os.Getenv("NODENAME")
+	if strings.HasPrefix(nodeName, "i-") {
```
M00nF1sh, line 141: i haven't tried the new cloudProvider yet. What's the spec.providerID on the new cloudProvider? 
currently the spec.providerID is like ```aws:///us-west-2a/i-029d795cbe1d5c038```, if the new cloudprovider is having similar pattern, maybe we can support inferFromNode for other environment as well..
e.g. by 1. get the node object from node name, then get the providerID, then check the last component splitted by "/" is starting with `i-`(or use regex).

pkg/aws/cloud.go
olemarkus: Ouh. Apimachinery has it's own error aggregate? Today I learnt. Will fix.

pkg/aws/cloud.go
olemarkus: Overall, I agree this is the better solution, but I consider this PR as a stepping stone towards that. It would require more rearranging of code to do k8s API calls and such as well. Especially if one wants to use the node watcher.

The providerID is the same as now when it comes to instance ID. But even now, it is not given that region is a part of the providerID and should rather be fetched from the topology labels.

pkg/aws/cloud.go
olemarkus: You were referring to this one and the one two lines below? The latter was replaced by an err check now that the infer function returns errors.

pkg/aws/cloud.go
M00nF1sh: make sense, i'm fine to take this as a first step and we can have follow up PRs to further improve it.
Thanks for the info.

olemarkus:
Anything more needed here?

M00nF1sh (approved):
/lgtm
/approve

Thanks :D now it looks good to me, we can include this in next release.

PR #2831, opened 2022-10-12
alpn-policy accept only string and not alpn-policy

BouchaaraAdil:
### Issue

fix typo documentation `service.beta.kubernetes.io/aws-load-balancer-alpn-policy`
### Description
we tended to use the annotation as a list to accept different policies and map them to listeners,
but while applying we got an error that it accepts only one value of 
`[HTTP1Only, HTTP2Only, HTTP2Optional, HTTP2Preferred, None]`

it would be nice if we have an annotation which accepts different alpn policies and not only one
something like 
`service.beta.kubernetes.io/aws-load-balancer-alpn-policies` with type StringMap so then we assign port to each alpn policy


M00nF1sh:
thanks for fixing this 😄 
will update the doc shortly.

PR #2836, opened 2022-10-18
refactor backend SG provider

kishorj:
### Issue

Refactor SG provider modules

### Description
The controller currently cleans up auto-allocated backend SG only if there are no active ingress resources managed by the controller. This change causes the lb controller to delete the backend SG as soon as there are no active ingresses requiring automatic SG rules configuration. Also refactor the SG resolver code from the model builder.

### Checklist
- [x] Manually tested


pkg/networking/backend_sg_provider.go
```diff
@@ -76,37 +89,106 @@ type defaultBackendSGProvider struct {
...
+			requiredForAny = true
+			return false
+		}
+		return true
+	})
+	if requiredForAny {
+		return true, nil
+	}
+	if required, err := p.checkIngressListForUnmapped(ctx); required || err != nil {
+		return required, err
```
johngmyers, line 157: When `required` is true, is it safe to return an `err` of nil? It seems to me in that case it will complete deployment of the model and remove the finalizers from the inactive members. The controller might then stop trying to reconcile the group and leave the SG orphaned.

pkg/networking/backend_sg_provider.go
```diff
@@ -76,37 +89,106 @@ type defaultBackendSGProvider struct {
...
+	if err := p.allocateBackendSG(ctx, activeResources); err != nil {
 		p.logger.Error(err, "Failed to auto-create backend SG")
 		return "", err
 	}
 	return p.autoGeneratedSG, nil
 }
 
-func (p *defaultBackendSGProvider) Release(ctx context.Context) error {
+func (p *defaultBackendSGProvider) Release(ctx context.Context, activeResources []metav1.Object,
+	inactiveResources []metav1.Object, backendSGRequired bool) error {
```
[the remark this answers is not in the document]
kishorj: I want to extend it to service as well.

pkg/networking/backend_sg_provider.go
kishorj: Controller will not remove finalizers in case of error. 

If required is true, then controller skips backend SG cleanup.

pkg/networking/backend_sg_provider.go
```diff
@@ -76,37 +89,106 @@ type defaultBackendSGProvider struct {
 	ec2Client       services.EC2
 	k8sClient       client.Client
 	logger          logr.Logger
+	objectsMap      sync.Map
```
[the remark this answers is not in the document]
kishorj: ack, i will add as comments

controllers/ingress/group_controller.go
```diff
@@ -324,6 +323,18 @@ func (r *groupReconciler) setupWatches(_ context.Context, c controller.Controlle
 	return nil
 }
 
+func (r *groupReconciler) getActiveAndInactiveIngressObjects(ingGroup ingress.Group) ([]metav1.Object, []metav1.Object) {
```
M00nF1sh, line 326: seems have a utility like below is more clean :D 
```
func ToSliceOfMetaObject[T any](s []T) []metav1.Object {
    result := make([]metav1.Object, len(s))
    for i, v := range s {
        result[i] = v
    }
    return result
}
```
then we can simply use ```ToSliceOfMetaObject(ingGroup.Members)``` and ```ToSliceOfMetaObject(ingGroup. InactiveMembers)```

pkg/ingress/model_build_load_balancer.go
```diff
@@ -309,18 +310,27 @@ func (t *defaultModelBuildTask) buildLoadBalancerSecurityGroups(ctx context.Cont
...
+			t.backendSGAllocated = true
 			lbSGTokens = append(lbSGTokens, t.backendSGIDToken)
 		}
 		t.logger.Info("SG configured via annotation", "LB SGs", lbSGTokens, "backend SG", t.backendSGIDToken)
 	}
 	return lbSGTokens, nil
 }
 
+func (t *defaultModelBuildTask) getActiveIngressObjects(_ context.Context) []metav1.Object {
+	var activeIngresses []metav1.Object
```
M00nF1sh, line 327: see my previous comment about getActiveAndInactiveIngressObjects, this can be a utility function

controllers/ingress/group_controller.go
kishorj: generic needs type constraints, the assignment result[i] = v doesn't work with any. I've updated the PR with utils.

pkg/ingress/model_build_load_balancer.go
kishorj: updated

controllers/ingress/group_controller.go
```diff
@@ -180,7 +175,11 @@ func (r *groupReconciler) buildAndDeployModel(ctx context.Context, ingGroup ingr
 	}
 	r.logger.Info("successfully deployed model", "ingressGroup", ingGroup.ID)
 	r.secretsManager.MonitorSecrets(ingGroup.ID.String(), secrets)
-	return stack, lb, err
+	if err := r.backendSGProvider.Release(ctx, k8s.ToSliceOfMetaObject(ingress.ExtractIngresses(ingGroup.Members)),
```
M00nF1sh, line 178: nit:
if we change the type of ToSliceOfMetaObject  from `[T metav1.Object]` to be `[T metav1.ObjectMetaAccessor]`, and add a GetObjectMeta() Object into `ClassifiedIngress`, we don't need this `ingress.ExtractIngresses`.

This might be better for future changes since we can treat ClassifiedIngress same as Ingress in our code.

pkg/networking/backend_sg_provider.go
```diff
@@ -40,9 +43,10 @@ const (
 // BackendSGProvider is responsible for providing backend security groups
 type BackendSGProvider interface {
 	// Get returns the backend security group to use
-	Get(ctx context.Context) (string, error)
+	Get(ctx context.Context, activeResources []metav1.Object) (string, error)
 	// Release cleans up the auto-generated backend SG if necessary
-	Release(ctx context.Context) error
+	Release(ctx context.Context, activeResources []metav1.Object,
```
M00nF1sh, line 48: why we do need this `activeResources` and `backendSGRequired` in release
is it possible that we just have 
```
Allocate(ctx context.Context, resources []metav1.Object)
Release(ctx context.Context, resources []metav1.Object)
```
And during release, if all tracked resource is released, and we have invoked reconcile for all resources at least once. then we release the sg?

pkg/networking/backend_sg_provider.go
```diff
@@ -267,3 +332,7 @@ func isEC2SecurityGroupNotFoundError(err error) bool {
 	}
 	return false
 }
+
+func getObjectKey(obj metav1.Object) string {
+	return reflect.TypeOf(obj).String() + "/" + k8s.NamespacedName(obj).String()
```
M00nF1sh, line 337: maybe pass the gvk of object in `get` and `release` explicitly?

pkg/networking/backend_sg_provider.go
```diff
@@ -76,37 +89,111 @@ type defaultBackendSGProvider struct {
...
+	return false, nil
+}
+
+func (p *defaultBackendSGProvider) checkIngressListForUnmapped(ctx context.Context) (bool, error) {
+	ingList := &networking.IngressList{}
+	if err := p.k8sClient.List(ctx, ingList); err != nil {
+		return true, errors.Wrapf(err, "unable to list ingresses")
+	}
+	for _, ing := range ingList.Items {
+		if !p.checkIngressFinalizersFunc(ing.GetFinalizers()) {
```
M00nF1sh, line 173: wondering whether there could be race conditions.

M00nF1sh (approved):
/lgtm
/approve
discussed with kishor, we will take action items to address the comments as follow-ups

pkg/networking/backend_sg_provider.go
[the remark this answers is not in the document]
kishorj: added comments

pkg/networking/backend_sg_provider.go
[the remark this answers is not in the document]
kishorj: I refactored the module to use types.NamespacedName instead.

controllers/ingress/group_controller.go
kishorj: done

pkg/networking/backend_sg_provider.go
kishorj: looks cleaner, refactored the APIs

pkg/networking/backend_sg_provider.go
kishorj: refactored to pass the type string

pkg/networking/backend_sg_provider.go
```diff
@@ -76,37 +94,102 @@ type defaultBackendSGProvider struct {
...
 
-func (p *defaultBackendSGProvider) Release(ctx context.Context) error {
+func (p *defaultBackendSGProvider) Release(ctx context.Context, resourceType ResourceType,
+	inactiveResources []types.NamespacedName) error {
 	if len(p.backendSG) > 0 {
 		return nil
 	}
+	defer func() {
+		for _, res := range inactiveResources {
+			p.objectsMap.Delete(getObjectKey(resourceType, res))
```
M00nF1sh, line 129: why we delete these keys

M00nF1sh (approved):
/lgtm
/approve

pkg/networking/backend_sg_provider.go
kishorj: for garbage collection

PR #2840, opened 2022-10-20
Add --default-target-type flag

### LINKED ISSUE #2690: Support flag to change the default target-type

The desired target type of an ALB or NLB is almost always determined by the the cluster's CNI. If a cluster's CNI uses ENIs (like AWS VPC CNI does) then the desired target type is `ip`. If a cluster's CNI uses masquerading, then the desired target type is `instance`.

When a cluster's CNI uses ENIs, it is a burden to require each workload's `Ingress` and `Service` resources to include an LBC-specific annotation in order to function.

Add a flag to the controller to set the default target-type. The value of the flag would be stored in the `defaultTargetType` field of  `defaultModelBuildTask`. This would require a new field be added to `defaultModelBuilder` and a new parameter be added to `NewDefaultModelBuilder`.

Add a field to `IngressClassParams`. While this would reduce the size of the burden from one-per-Ingress to one-per-IngressClass, there isn't a likely use case where one would want different IngressClasses in the same cluster to have different default target-types.

johngmyers:
### Issue

Closes #2690

### Description

Adds a `--default-target-type` flag which sets the default target type of both Services and Ingresses.

The desired target type of an ALB or NLB is almost always determined by the the cluster's CNI. If a cluster's CNI uses ENIs (like AWS VPC CNI does) then the desired target type is `ip`. If a cluster's CNI uses masquerading, then the desired target type is `instance`.

When a cluster's CNI uses ENIs, it is a burden to require each workload's `Ingress` and `Service` resources to include an LBC-specific annotation in order to function.

This allows clusters with ENI-based CNIs to set the default target type to `ip`, allowing `Ingress` and `Service` resources to work correctly without having to set a target-type annotation on each one.

### Checklist
- [x] Manually tested

### DIFF
--- a/docs/deploy/configurations.md
+++ b/docs/deploy/configurations.md
@@ -75,6 +75,7 @@ Currently, you can set only 1 namespace to watch in this flag. See [this Kuberne
 |cluster-name                           | string                          |                 | Kubernetes cluster name|
 |default-ssl-policy                     | string                          | ELBSecurityPolicy-2016-08 | Default SSL Policy that will be applied to all Ingresses or Services that do not have the SSL Policy annotation |
 |default-tags                           | stringMap                       |                 | AWS Tags that will be applied to all AWS resources managed by this controller. Specified Tags takes highest priority |
+|default-target-type                    | string                          | instance        | Default target type for Ingresses and Services - ip, instance |
 |[disable-ingress-class-annotation](#disable-ingress-class-annotation)       | boolean                         | false           | Disable new usage of the `kubernetes.io/ingress.class` annotation |
 |[disable-ingress-group-name-annotation](#disable-ingress-group-name-annotation)  | boolean                         | false           | Disallow new use of the `alb.ingress.kubernetes.io/group.name` annotation |
 |disable-restricted-sg-rules            | boolean                         | false            | Disable the usage of restricted security group rules |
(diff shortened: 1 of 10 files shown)


johngmyers:
One question would be how to document the default value of annotations whose default value depend on flags.

docs/deploy/configurations.md
```diff
@@ -75,6 +75,7 @@ Currently, you can set only 1 namespace to watch in this flag. See [this Kuberne
 |cluster-name                           | string                          |                 | Kubernetes cluster name|
 |default-ssl-policy                     | string                          | ELBSecurityPolicy-2016-08 | Default SSL Policy that will be applied to all Ingresses or Services that do not have the SSL Policy annotation |
 |default-tags                           | stringMap                       |                 | AWS Tags that will be applied to all AWS resources managed by this controller. Specified Tags takes highest priority |
+|default-target-type                    | string                          |                 | Default target type for Ingresses and Services - ip, instance |
```
kishorj, line 78: default `instance`

kishorj (commented):
nit: default value in the configurations.md
Changes look good to me.

docs/deploy/configurations.md
kishorj: we can modify doc in a follow up commit.

kishorj:
/lgtm
/approve

Thanks a lot for the contribution

PR #2849, opened 2022-10-25
Stop recommending insecure IMDSv2 configuration

### LINKED ISSUE #2665: IMDSv2 Doc Suggests Insecure Configuration
**Describe the bug**
The [IMDSv2 section of the Installation Guide](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/deploy/installation/#using-metadata-server-version-2-imdsv2) suggests increasing the endpoint hop limit to 2, despite this being directly against the recommendation of [EKS best practices](https://aws.github.io/aws-eks-best-practices/security/docs/iam/#restrict-access-to-the-instance-profile-assigned-to-the-worker-node). This Best Practice is relevant even for non-EKS clusters running in AWS.

This is also without any suggestion that this could seriously impact a cluster's security posture.

**Steps to reproduce**
- read the 2 linked documents

**Expected outcome**
A removal of the recommendation, or at least a warning about the security implications, and a guide on a minimally-scoped NetworkPolicy to allow _only_ the AWS LB Controller minimal access.

**Environment**

* AWS Load Balancer controller version N/A
* Kubernetes version N/A
* Using EKS (yes/no), if so version? N/A

**Additional Context**:

johngmyers:
### Issue

Fixes #2665 

### Description

Update installation documentation to replace the advice to configure IMDSv2 hops insecurely.

### Checklist
- [ ] Manually tested

### DIFF
--- a/docs/deploy/installation.md
+++ b/docs/deploy/installation.md
@@ -20,15 +20,14 @@ The LBC is supported by AWS. Some clusters may using legacy "in-tree" functional
 * Ensure subnets are tagged appropriately for auto-discovery to work
 * For IP targets, pods must have IPs from the VPC subnets. You can configure `amazon-vpc-cni-k8s` plugin for this purpose.
 
-### Using metadata server version 2 (IMDSv2)
-If you are using the [IMDSv2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) you must set the hop limit to 2 or higher in order to allow the AWS Load Balancer Controller to perform the metadata introspection.
+### Determining VPC and Region
 
-You can set the IMDSv2 hop limit as follows:
-```
-aws ec2 modify-instance-metadata-options --http-put-response-hop-limit 2 --region <region> --instance-id <instance-id>
-```
+The controller must be able to determine the AWS Region and VPC ID. There are several options:
 
-Instead of depending on IMDSv2, you alternatively may specify the AWS region and the VPC via the controller flags `--aws-region` and `--aws-vpc-id`.
+1. You may specify either or both using the controller flags `--aws-region` and/or `--aws-vpc-id`.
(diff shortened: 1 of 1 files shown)


olemarkus (approved):

docs/deploy/installation.md
```diff
@@ -20,15 +20,14 @@ The LBC is supported by AWS. Some clusters may using legacy "in-tree" functional
...
+### Determining VPC and Region
 
-You can set the IMDSv2 hop limit as follows:
-```
-aws ec2 modify-instance-metadata-options --http-put-response-hop-limit 2 --region <region> --instance-id <instance-id>
-```
+The controller must be able to determine the AWS Region and VPC ID. There are several options:
 
-Instead of depending on IMDSv2, you alternatively may specify the AWS region and the VPC via the controller flags `--aws-region` and `--aws-vpc-id`.
+1. You may specify either or both using the controller flags `--aws-region` and/or `--aws-vpc-id`.
```
kishorj, line 27: With these instructions in place, the current default installation will not end up with a functional controller (assuming the hop limit is 1). I'm more comfortable with these instructions once we are able to determine VPC from the node resource without the node name limitation.

johngmyers: Perhaps we should make the default installation specify host networking?

M00nF1sh: @johngmyers 
I think host networking provides unnecessary permissions to the controller, which is not secure.

Chatted with @kishorj , the best solution to moving forward is to auto-detect the region & vpc-ID to remove IMDS dependency for default installation.

johngmyers: @M00nF1sh Setting the IMDSv2 hop limit to 2 is even less secure than giving host networking to LBC, by a lot. That provides unnecessary permissions to every pod on the node, allowing them to obtain credentials to the node's instance role.

PR #2881, opened 2022-11-13
Add GitHub Action for tagging releases and creating release branches 

johngmyers:
### Issue

N/A

### Description

Adds automation for creating release tags and branches. This allows the decision to create a release to go through the standard code review process.

This also has code to create a release branch upon the creation of a `.0-beta.1` release. LBC doesn't currently create beta releases, so this code won't have effect unless it starts doing so. A possible change would be to have it also create a release branch with a `.0` stable release (if no such branch exists). That, however, would require a change in process to do all patch releases off of the release branch.

### Checklist
- [x] Manually tested

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -1,0 +1,25 @@
+name: tag-release
+
+on:
+  push:
+    branches:
+      - main
+      - 'release-*'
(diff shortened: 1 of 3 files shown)


kishorj (approved):
Thanks

Issue #2358, opened 2021-11-11
Disable adding security group rules to worker node security group for nlb


RommelLayco:
**Is your feature request related to a problem?**
When creating an NLB. The controller adds inbound rules to the node security group or ENI security group.

<img width="839" alt="Screen Shot 2021-11-12 at 10 26 06 AM" src="https://user-images.githubusercontent.com/8358460/141371635-6505e677-3ee6-4feb-952b-e33615df2e32.png">

https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.3/guide/service/nlb/#security-group

When the service has  a lot of ports in its spec. This will cause us to reach the security group limit real quick, As it add a new inbound rule per port.

**Describe the solution you'd like**

We would like one of the following.

Option 1: An annotation to disable the controller for adding new inbound rules to the security group. Instead the admin would open the ports needed on the security groups.

Option 2: An annotation that allows us provide what ports should be open on the security group. Overriding the current logic that uses the port in the service spec.

**Describe alternatives you've considered**
A description of any alternative solutions or features you've considered.

M00nF1sh:
/kind feature
i think option1 makes sense for now before NLB security group support. we'll make it into the v2.4.0 release if NLB SG is not ready by then

alfredkrohmer:
This would help us a lot. We can't manage NLBs through the aws-load-balancer-controller right now due to this issue (our cloud governance doesn't allow us to modify security groups).

bruno-andrad:
Hello everyone, everything good?
Has this problem already been resolved? I have the same problem. I need to prevent my service from creating new rules in the security group. But it does not work.
Besides installing aws-load-balancer-controller and adding the annotations, what else should I do?

PR #2919, opened 2022-12-09
Add new featuregate to enable/disable SG management

alikhanich:
### Issue
We have a lot of services in k8s with type LoadBalancer, and for each of them controller was creating SG in AWS. We've reached the point when quota increase was denied from AWS support. So we decided to disable SG management for controller and switch to static rule which allows 30000-32767 port range to 0.0.0.0/0.  Currently it can only be achieved by placing annotation on EACH service in k8s cluster, as our cluster is very big and there are a lot of different charts it is very hard for us. Hence here comes this PR.

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2358

### Description

I have added new feature flag `ManageSecurityGroups` which is true by default for backwards compatibility, and in method `buildManageSecurityGroupRulesFlag` I preserved the ability to use annotations to disable SG management.
If you like this feature I will gladly write tests for it and will update the docs accordingly.

### Checklist
- [x] Manually tested


djirik:
@kishorj @M00nF1sh Hi, any thoughts on this?

johngmyers:
I'm planning on adding a `securityGroups` field to `IngressClassParams`. Would this solve your problem?

djirik:
@johngmyers how would it work? and do you have ETA?

johngmyers:
It would look like #2945 except for security groups instead of subnets.

I'm completely dependent on code reviews and approvals, so don't have an ETA. Security groups would be fourth on my list because I don't have a need to specify them myself. You could use my PR as a template.

ngnix:
Hi @johngmyers !
I think it will be the best solution to add a `securityGroups` to the `IngressClassParams`.

ngnix:
Maybe it's worth adding the ability to create a security group for each service with the LoadBalancer type? and add them only to those worker nodes on which the corresponding loads are located? The problem is that now it is very easy to run into two AWS limits:
 - number of rules within one security group 
 - number of security groupы per EC2 instance

This is a hard problem in AWS - as these limits are very small !

ngnix:
The current solution proposed by @djirik  will allow turning off the creation of rules by the operator, in this case, ports of 30000+ services are always allowed in the security group and new rules are not added to it, so we do not rest on the limits.
This is an alternative solution.

Issue #2920, opened 2022-12-09
Add additional settings to IngressClassParams


johngmyers:
**Is your feature request related to a problem?**

As a provider of Kubernetes clusters to internal service teams, I want to configure ALB-level settings at the IngressClass level, preventing service teams from impacting other teams using the same IngressClass by misconfiguring annotations on their Ingress objects

**Describe the solution you'd like**

As a rule of thumb, there should be an `IngressClassParams` field for any `Ingress` annotation that has a `MergeBehavior` of `Exclusive`. This field should override/prevent the `Ingress` level annotations.

There should probably be a way to explicitly configure the default behavior so the `Ingress` annotations have no effect in the `IngressClass`, or to state that no `Ingress` annotations of `Exclusive` `MergeBehavior` have effect.

The fields most important to me are:

* subnets
* ssl-policy
* inbound-cidrs

Other `Exclusive` fields that are missing from `IngressClassParams` are:

* security-groups
* load-balancer-name
* customer-owned-ipv4-pool
* wafv2-acl-arn
* waf-acl-id
* shield-advanced-protection

**Describe alternatives you've considered**

Muddle through by configuring on a single `Ingress` in the `IngressClass`. It is extremely inconvenient.

kishorj:
/kind feature
Some of the fields would imply an ingress group either explicit or implicit. For example: the load-balancer-name.

johngmyers:
So perhaps api-validate that a `group` is specified if any of those fields are set?

johngmyers:
Which fields other than `load-balancer-name` imply a group?

kishorj:
> Which fields other than `load-balancer-name` imply a group?

I verified, this is the only one. With the `load-balancer-name`, controller doesn't group multiple ingresses together, however, all ingresses will refer to the same ALB. This will result in a race condition.

wonko:
I would like to have additional parameters in the IngressClassParams as well. My use case asks for a `subnets` setting. I can't update the tags on the subnets in the VPC, as those are supplied and provisioned by another team, and my team has no rights to update tags on the subnets (so I can't set the `kubernetes.io/role/elb = 1` tag for auto-discovery.

Requiring the subnet ids in the annotations for each deployed ingress is cumbersome, people keep forgetting them, and we need to distribute the correct set of ids for each situation and account.

[...]

PR #2945, opened 2022-12-25
Add subnets field to IngressClassParams

johngmyers:
### Issue

#2920

### Description

Adds a `subnets` field to `IngressClassParams`, to allow overriding the `alb.ingress.kubernetes.io/subnets` annotation for all `Ingresses` in an `IngressClass`.

At this point I'm proposing the API, to get feedback before implementing. I'm trying to take advantage of the ability to use the structure that exists in YAML but not annotation values.

### Checklist
- [x] Manually tested


johngmyers:
I'm a bit stuck, having trouble compiling controller-gen v0.5.0, as specified in Makefile. Go 1.17 and above won't compile it. I tried using Go 1.13.15, but it wouldn't download:

```
can't load package: package sigs.k8s.io/controller-tools/cmd/controller-gen@v0.5.0: can only use path@version syntax with 'go get'
```

Which version of Go do I need to compile controller-gen per the Makefile?

kishorj:
> I'm a bit stuck, having trouble compiling controller-gen v0.5.0, as specified in Makefile. Go 1.17 and above won't compile it. I tried using Go 1.13.15, but it wouldn't download:
> 
> ```
> can't load package: package sigs.k8s.io/controller-tools/cmd/controller-gen@v0.5.0: can only use path@version syntax with 'go get'
> ```
> 
> Which version of Go do I need to compile controller-gen per the Makefile?

I was unable to install v0.5.0 on my dev setup as well, since I had the old version in my path, it didn't complain. PR #2953 updates the controller gen to latest.

apis/elbv2/v1beta1/ingressclassparams_types.go
```diff
@@ -41,6 +41,15 @@ const (
 	LoadBalancerSchemeInternetFacing LoadBalancerScheme = "internet-facing"
 )
 
+// SubnetSelector selects one or more existing subnets.
+type SubnetSelector struct {
+	// ID specifies the resource ID of a subnet. Exactly one of this or `nameContains` must be specified.
+	ID string `json:"id,omitempty"`
+	// NameContains specifies all subnets containing a string in their Name tag. Exactly one of this or `id` must be specified.
+	NameContains string `json:"nameContains,omitempty"`
```
kishorj, line 49: There is a possibility of multiple matches here and we need a strategy to break ties in case of multiple subnets per AZ. In case of subnet name/id in the annotation, controller currently expects all of the entries to be resolved to valid subnets, and no multiples in the same AZ.

apis/elbv2/v1beta1/ingressclassparams_types.go
```diff
@@ -41,6 +41,15 @@ const (
...
+	// TODO VPCID string `json:"vpcID,omitempty"`
```
kishorj, line 50: Is this the same VPC as the ALB?

apis/elbv2/v1beta1/ingressclassparams_types.go
johngmyers: I was thinking we'd have the same expectation; that the union of all `SubnetSelector`s contains at most one subnet per AZ. It is true that the situation could change between the time the `IngressClassParams` is last modified and the LB is created or modified.

apis/elbv2/v1beta1/ingressclassparams_types.go
johngmyers: Good question! A better approach would probably be to add a `vpc` field to `IngressClassParamsSpec` instead of hanging it off of `subnets`.

apis/elbv2/v1beta1/ingressclassparams_types.go
johngmyers: Apparently the annotation expects exactly one match. I was expecting this to handle multiple matches, but still at most one subnet per AZ.

johngmyers:
Ah, DescribeSubnets doesn't allow for filtering a substring. I think it makes more sense to allow filtering on arbitrary tags:

```
type SubnetSelector struct {
	// ID specifies the resource ID of a subnet. Exactly one of this or `nameContains` must be specified.
	ID string `json:"id,omitempty"`

	TagKey string `json:"tagKey,omitempty"`
        TagValue string `json:"tagValue,omitempty"`
}
```
or somesuch.

apis/elbv2/v1beta1/ingressclassparams_types.go
johngmyers: I think devising rules for multiple subnets in an AZ would be for a separate PR.

apis/elbv2/v1beta1/ingressclassparams_types.go
```diff
@@ -41,6 +41,22 @@ const (
...
+type SubnetSelector struct {
+	// ID specifies the resource ID of a subnet. Exactly one of this or `tagKey` must be specified.
+	// +kubebuilder:validation:Pattern=subnet-[0-9a-f]+
+	// +optional
+	ID string `json:"id,omitempty"`
+
+	// TagKey specifies all subnets containing a tag with the given key and the value specified in `tagValue`.
+	// Exactly one of this or `id` must be specified.
+	// +optional
+	TagKey string `json:"tagKey,omitempty"`
```
kishorj, line 54: Since the describe is on a single tag key (on the given vpc), if there are multiple matching subnets per AZ, we make the tag feature unusable. It is more usable if we support multiple tag keys

pkg/networking/subnet_resolver.go
```diff
@@ -263,6 +268,76 @@ func (r *defaultSubnetsResolver) ResolveViaNameOrIDSlice(ctx context.Context, su
...
+		if len(subnets) < len(values) {
+			return nil, errors.Errorf("couldn't find all subnets, tag: %v, found: %v", key, len(subnets))
+		}
+		resolvedSubnets = append(resolvedSubnets, subnets...)
+	}
+
+	if len(resolvedSubnets) == 0 {
+		return nil, errors.New("unable to resolve at least one subnet")
+	}
+	if err := r.validateSubnetsAZExclusivity(resolvedSubnets); err != nil {
```
kishorj, line 327: We need to account for multiple matches per AZ, otherwise we make the tag selector feature unusable in those situations.

apis/elbv2/v1beta1/ingressclassparams_types.go
```diff
@@ -41,6 +41,22 @@ const (
 	LoadBalancerSchemeInternetFacing LoadBalancerScheme = "internet-facing"
 )
 
+// SubnetSelector selects one or more existing subnets.
+type SubnetSelector struct {
+	// ID specifies the resource ID of a subnet. Exactly one of this or `tagKey` must be specified.
+	// +kubebuilder:validation:Pattern=subnet-[0-9a-f]+
+	// +optional
+	ID string `json:"id,omitempty"`
```
kishorj, line 49: Why not a list of IDs or a list of tag key/values? We can limit one set of subnet selectors for the ingress group.

docs/guide/ingress/ingress_class.md
```diff
@@ -135,6 +135,26 @@ Cluster administrators can use the `scheme` field to restrict the scheme for all
...
 2. If `scheme` un-specified, Ingresses with this IngressClass can continue to use `alb.ingress.kubernetes.io/scheme annotation` to specify scheme.
 
+#### spec.subnets
+
+`subnets` is an optional setting.
+
+Cluster administrators can use the `subnets` field to specify the subnets for the load balancers that belong to this IngressClass.
+
+1. If `subnets` specified as a non-empty list, all load balancers that belong to this IngressClass will have the specified subnets.
+2. If `subnets` specified as an empty list, all load balancers that belong to this IngressClass will use [Subnet Discovery](../../deploy/subnet_discovery.md).
```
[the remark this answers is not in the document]
johngmyers: I disagree. The admin should have a way of forcing use of subnet discovery, disallowing a single `Ingress` from changing the subnets for the entire `IngressClass` through an annotation.

apis/elbv2/v1beta1/ingressclassparams_types.go
johngmyers: The admin presumably has control over subnet tagging, so they can avoid putting the tag value on more than one subnet in an AZ.

We could make this:

```go
type TagSelector struct {
    Key string
    Value string
}

type SubnetSelector struct {
    ID string
   Tags []TagSelector
```

to be fully generalized, but that seems overkill.

apis/elbv2/v1beta1/ingressclassparams_types.go
johngmyers: Is there a need to prohibit specifying both IDs and tag/values? I'd like the interface to be able to specify anything that the existing annotation can.

pkg/networking/subnet_resolver.go
johngmyers: How is this different from the existing annotation's tag searching semantics? The admin presumably has control over both the assignment of tags to subnets and use of tags in subnet selection; they can assign and select appropriately.

If we do anything other than error out, we risk painting ourselves into a corner should ALBs start supporting multiple subnets per AZ. We could perhaps add a check for this condition in the validating admission webhook so that it fails earlier.

docs/guide/ingress/ingress_class.md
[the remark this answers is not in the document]
johngmyers: One would force subnet discovery with:

```yaml
spec:
    subnets: []
```

In my use case, a soft-multitenant cluster, I will want to disable all ingress annotations with exclusive mergebehavior. My current plan is to specify a value for each of them in the `IngressClassParams` of each `IngressClass`.

pkg/networking/subnet_resolver.go
johngmyers: With `IngressClassParams`, one can also specify appropriate subnet IDs:

```
spec:
    subnets:
    - id: subnet-xxxx
    - tagKey: Name
      tagValue: mySubnet
 ```

to reproduce the example in the annotations documentation. (I'm thinking a followup PR could replace `SubnetsResolver.ResolveViaNameOrIDSlice()` with a call to `ResolveSubnetSpec()`.)

File: apis/elbv2/v1beta1/ingressclassparams_types.go (hunk 5 above)
Comment (johngmyers) [reply apis/elbv2/v1beta1/ingressclassparams_types.go:49]: The current proposal cannot represent the rules for subnet discovery: It cannot specify that two keys must match. If we go with the `TagSelector` approach I mentioned above, then one could do:

```yaml
spec:
    subnets:
    - tags:
      - key: kubernetes.io/role/elb
        value: "1"
      - key: kubernetes.io/cluster/the-cluster-name
        value: owned
    - tags:
      - key: kubernetes.io/role/elb
        value: "1"
      - key: kubernetes.io/cluster/the-cluster-name
        value: shared
    - tags:
      - key: kubernetes.io/role/elb
        value: ""
      - key: kubernetes.io/cluster/the-cluster-name
        value: owned
    - tags:
      - key: kubernetes.io/role/elb
        value: ""
      - key: kubernetes.io/cluster/the-cluster-name
        value: shared
```

assuming we allow a TagSelector to match zero subnets.

This seems overly complicated to me. It's much simpler to rely on the tags being provisioned appropriately:

```yaml
spec:
    subnets:
    - tagKey: ingressclass/the-cluster-name
      tagValue: the-ingress-class
```

pkg/networking/subnet_resolver.go
johngmyers: If we provide semantics for multiple subnets in an AZ that are other than erroring out, then the semantics that ALB later gives multiple subnets in an AZ might turn out to not be backwards compatible with the semantics we provided.

johngmyers:
I'm working on implementing the following:

```go
// TagSelector specifies a tag search requirement.
type TagSelector struct {
	// Key specifies the tag key that must be present.
	Key string `json:"key"`
	// Values specifies the tag values, one of which must match the value set for the tag key.
	Values []string `json:"values"`
}

// SubnetSelector selects one or more existing subnets.
type SubnetSelector struct {
	// IDs specify the resource IDs of subnets. Exactly one of this or `tags` must be specified.
	// +kubebuilder:validation:Pattern=subnet-[0-9a-f]+
	// +optional
	IDs []string `json:"ids,omitempty"`

	// TagKey specifies subnets matching all of a set of tag search requirements.
	// Exactly one of this or `ids` must be specified.
	// +optional
	Tags []TagSelector `json:"tags,omitempty"`
}
```

Furthermore, `tags` will apply the subnet discovery filtering and conflict resolution semantics.

kishorj (approved):

PR #2950, opened 2022-12-30
docs: make external-dns optional in echoserver walkthrough

prasadkatti:
### Description
This PR updates the `echoserver` walkthrough.

- makes it clear that use of `external-dns` is completely optional.
- adds a verification section that uses the load balancer DNS name in case `external-dns` is not used
- removes the `spec.rules[0].host` in the default manifest. With this in, the example does not work if external-dns is not used. The listener rule does not match because the Host Header does not match the ALB DNS name.
- removes instructions to deploy external-dns and instead links to the right page

### Checklist
- [ ] Manually tested


docs/examples/echo_server.md
```diff
@@ -5,8 +5,9 @@ In this walkthrough, you'll
 - Create a cluster with EKS
 - Deploy an aws-load-balancer-controller
 - Create deployments and ingress resources in the cluster
-- Use [external-dns](https://github.com/kubernetes-incubator/external-dns) to create a DNS record
-    - This assumes you have a route53 hosted zone available. Otherwise you can skip this, but you'll only be able to address the service from the ALB's DNS.
+- Verify access to the service
+- (Optional) Use [external-dns](https://github.com/kubernetes-incubator/external-dns) to create a DNS record pointing to the load balancer created by the aws-load-balancer-controller.
```
johngmyers, line 9:
```suggestion
- (Optional) Use [external-dns](https://github.com/kubernetes-sigs/external-dns) to create a DNS record pointing to the load balancer created by the aws-load-balancer-controller.
```

docs/examples/echo_server.md
```diff
@@ -143,8 +144,7 @@ In this walkthrough, you'll
                 alb.ingress.kubernetes.io/tags: Environment=dev,Team=test
         spec:
             rules:
-            - host: echoserver.example.com
-                http:
+            - http:
                 paths:
```
johngmyers, line 148: The indentation is off here. Could you fix it?

johngmyers (changes requested):

docs/examples/echo_server.md
prasadkatti: It was actually off before. I fixed it here. Reference - https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#ingressrule-v1-networking-k8s-io

kishorj (approved):

PR #2955, opened 2023-01-03
Build image with ko

johngmyers:
### Issue

N/A

### Description

Uses ko to build the LBC image.

Reduces build time significantly and increases build reliability. Previously builds would take 15 minutes to over 2 hours and would fail about 3 out of 4 times, both because of the `go mod download` in the Dockerfile.

### Checklist
- N/A Added tests that cover your change (if possible)
- N/A Added/modified documentation as required (such as the `README.md`, or the `docs` directory)
- [x] Manually tested


olemarkus:
I strongly support this! Building LBC is a bit prohibitive right now.

kishorj:
Thanks for the contribution, I will look into this changes post the v2.4.6 patch release.

hack/install-ko.sh
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if ! command -v ko &> /dev/null; then
+  cd "$(dirname "${BASH_SOURCE[0]}")" || exit 1
+  go install github.com/google/ko@v0.12.0
```
kishorj, line 23: It failed for me with the following error:
```
go: downloading github.com/go-openapi/analysis v0.21.3
../../go/pkg/mod/github.com/go-openapi/validate@v0.22.0/spec.go:23:2: github.com/go-openapi/analysis@v0.21.3: verifying module: checksum mismatch
	downloaded: h1:vR88pR69D/jGh02vPbT4qoxiG+p9e5uT75JJG3O1JIU=
	sum.golang.org: h1:CPEa+B2oYCkb+lIKB4xP6Ork8Gvh0GNg9dm/twI3+QA=
```
I was able to install v0.11.2 and run the tests

helm/aws-load-balancer-controller/templates/deployment.yaml
```diff
@@ -153,8 +153,6 @@ spec:
           value: "{{ $value }}"
         {{- end }}
         {{- end }}
-        command:
-        - /controller
```
kishorj, line 157: Is there a standard entrypoint for ko built container images?

scripts/ci_e2e_test.sh
```diff
@@ -72,21 +72,7 @@ build_push_controller_image() {
...
-  until [ "$n" -ge 2 ]; do
-    DOCKER_CLI_EXPERIMENTAL=enabled docker buildx build . --target bin \
-      --tag "${CONTROLLER_IMAGE_NAME}" \
-      --push \
-      --progress plain \
-      --platform linux/amd64 && break
-    n=$((n + 1))
-    sleep 2
-  done
+  make docker-push IMG=${CONTROLLER_IMAGE_NAME} IMG_PLATFORM=linux/amd64
```
kishorj, line 75: would this not affect the multi-arch tests?

kishorj (commented):
I have two items I'm trying to work through at the moment:
1. ability to save to tarball, the `--tarball` flag to build command appears to be the way to go
2. Configuring a base image. The ko build appears to use a debian based image, I'm not certain if there is a fixed image per ko version, or it depends on the latest one available at the time of the build. I'm also interested to see how we can address the base image CVEs.

hack/install-ko.sh
johngmyers: You're apparently not the first to run across this. I found https://github.com/go-openapi/analysis/issues/81

johngmyers: Filed ko-build/ko#940

helm/aws-load-balancer-controller/templates/deployment.yaml
johngmyers: Yes. See the last sentence of https://ko.build/get-started

scripts/ci_e2e_test.sh
johngmyers: The old code specified `--platform linux/amd64` so I preserved that behavior.

johngmyers:
The base image is `cgr.dev/chainguard/static` and it takes what is the latest at the time of build.

Information on overriding the base image is at https://ko.build/configuration/#overriding-base-images

I doubt we'd be faster at fixing CVEs than Chainguard.

hack/install-ko.sh
johngmyers: Shall I back off to v0.11.2?

johngmyers:
Did you want a make target to create a tarball? Should that be instead of or in addition to pushing to a registry?

hack/install-ko.sh
kishorj: lets back off to v0.11.2, we can update in future

helm/aws-load-balancer-controller/templates/deployment.yaml
kishorj: Thanks. The entrypoint change will introduce backwards incompatibility with the existing manifests, we will release note it.

kishorj:
> The base image is `cgr.dev/chainguard/static` and it takes what is the latest at the time of build.
> 
> Information on overriding the base image is at https://ko.build/configuration/#overriding-base-images
> 
> I doubt we'd be faster at fixing CVEs than Chainguard.

Lets use `public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot:2022-07-27-1658910674.2` as the base image. Since the base image is internal, it will help us meet the SLAs for patching CVEs.

kishorj:
> Did you want a make target to create a tarball? Should that be instead of or in addition to pushing to a registry?

This is a separate requirement for publishing the official ECR images. We can revisit it when it is time to publish the v2.5.0 images.

kishorj:
Please preserve the existing ldflags and the build flags to the extent possible/necessary, especially the version information.

johngmyers:
Is there any particular reason to use CGO? It doesn't appear to be needed.

johngmyers:
This configuration removes a whole swath of attack surface by disabling CGO.

kishorj:
> Is there any particular reason to use CGO? It doesn't appear to be needed.

We don't need, lets remove it

kishorj (approved):

PR #2967, opened 2023-01-10
Ability to reconfigure NLB target group health check

kishorj:
### Issue

Add the ability to reconfigure NLB target group health check settings without recreating the target group.

### Description

Due to limitations on the AWS NLB side, the AWS LB controller needed to delete and create a new NLB target group in case the health check settings were modified. With recent capability enhancements, the AWS NLB now allows reconfiguration of all of the target group health check attributes. This PR enables the controller to modify the NLB healthcheck configuration without having to delete the target groups.

We've introduced a feature flag `NLBHealthCheckAdvancedConfiguration` which is set to `true` by default. This feature flag can be used to revert to the legacy behavior if the NLB feature is not available in the given AZ.

Manual Tests:
- verified controller modifies the target group health check configuration without deleting existing target group
- for http/https health check, default http matcher is 200-399, and can be overridden via annotation
- controller re-uses existing target group with matching tags

### Checklist
- [x] Manually tested


M00nF1sh (approved):
/lgtm
/approve

Issue #2939, opened 2022-12-21
Allow disabling ALB/Ingress management completely (and NLB management)


robinkb:
**Is your feature request related to a problem?**

We deployed the aws-load-balancer-controller in our organization in order to create Network Load Balancers in our EKS clusters. Mainly because we had trouble with the deprecated Classic Elastic Load Balancer provisioning with the in-tree Kubernetes controller. However, we already use the NGINX Ingress Controller to control our Ingress resources, and do not want this functionality from aws-load-balancer-controller.

Unfortunately, it is impossible to do this in the current Helm chart implementation. I can disable some functionality related to Ingresses, but it's impossible to disable the ValidatingWebhook. Because aws-load-balancer-controller deploys a ValidatingWebhook for a component that it doesn't even manage, it's interrupting our flow and I had to fork the chart.

**Describe the solution you'd like**

I would like a simple toggle (or a number of toggles) in the Helm chart that allows me to disable Ingress/ALB Management completely, including extra resources like the ValidatingWebhook.

**Describe alternatives you've considered**

I forked the chart and added a toggle to disable the ValidatingWebhook, but I don't want to maintain a fork.

**Extra notes**

I'm willing to put in some time to implement the changes in the Helm chart, if necessary. But I need a design first.

kishorj:
@robinkb, we already have a feature request #2801 for this issue. We plan to provide a helm template value to disable ingress validation webhook.

kishorj:
I wasn't able to get much about "NLB management" from the issue details, could you explain further?

robinkb:
Ah, my apologies, I looked for existing issues but didn't find that one. Disabling management of Load Balancer Services (by fulfilling them with Network Load Balancers) is not something that I personally want, but maybe that's something that someone else would require. The AWS Load Balancer Controller currently does two very different things, and as a user, it makes sense to me that either of those things can be disabled.

But I have no personal requirement for that, so this issue can be closed in favor of #2801. Though I would love an easier way to disable Ingress management completely by just toggling one boolean, instead of digging through all the configuration options and guessing which ones I should disable.

kishorj:
@robinkb, the management of NLB can be disabled via the feature gate `EnableServiceController` - please refer to the [live doc](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/deploy/configurations/#feature-gates) for further details. I'm closing this issue.

nuno-silva:
[...]

PR #3011, opened 2023-01-30
feat(helm): allow disabling ingress validation

### LINKED ISSUE #2801: Option to disable ingress validation webhook

I am already using ingress-nginx for my ingress objects. Sometimes I encounter some issues where the LB-controller wants to validate ingress objects while its not even handling them 

I would like to be able to disable the ingress validating part of the lb-controller validating webhook

none just want to be able to disable lb-controller webhook from validating ingress objects.
I also notived the feature-gate to let the LB-controller only look at service of type loadbalancer but i see that the webhook is installed via helm and there is no bool to enable or disable the ingress part so this would also not fix it.

mhulscher:
### Issue

fixes #2801

### Description

I exclusively use the aws-lb-controller to reconcile Services of type LoadBalancer. I don't want my Ingress resources to be validated.

### Checklist
- [x] Manually tested


johngmyers:
This change doesn't make a lot of sense to me. It looks like the wrong approach. LBC should always validate Ingresses that it is going to act on.

If the problem is that LBC is rejecting validation for Ingresses that it is not going to act on, that is a bug in the validator. The correct approach would be to fix the validator to ignore Ingresses that LBC is not going to act on.

mhulscher:
@johngmyers well see that's the thing, I don't want any ingresses acted upon or validated by the LBC. Even if I wanted to scope the webhook config, there is no way to select only ingresses that have a specific ingressclass.

As an example of what I am running into, the LBC is rejecting ingresses that configure an ingresClassName for which an IngressClass resource does not (yet) exist. And this is not the LBCs business to interfere in.

helm/aws-load-balancer-controller/values.yaml
```diff
@@ -213,6 +213,9 @@ syncPeriod:
 # Namespace the controller watches for updates to Kubernetes objects, If empty, all namespaces are watched.
 watchNamespace:
 
+# disableIngressValidation disables the validation of resources of kind Ingress, false by default
```
kishorj, line 216: Lets group this configuration under webhookConfig, that way we can group the webhook related config together. Something like this:
```
# webhookConfig contains configurations specific to the validating and mutating webhooks
webhookConfig:
    # disableIngressValidation disables the validation of ingress resources, false by default
    disableIngressValidation: false
```

johngmyers:
An ingress with an ingressClassName for which no IngressClass exists seems a perfectly reasonable thing for any ingress controller's validating webhook to act on. That's just plainly a broken configuration.

helm/aws-load-balancer-controller/values.yaml
mhulscher: Done. I placed the webhookConfig above the existing webhook* keys.

mhulscher:
@johngmyers respectfully, I disagree with your assessment. In my opinion, no ingress-controller should deny Ingress resources that do not specifically have a class configured that the controller watches for. Especially not when I don't want the LBC to validate any ingresses.

I use the LBC to create Services of type LoadBalancer because the in-tree AWS cloudprovider code is deprecated and will be removed in the future. I do not wish to use any of the ingress functionality of the LBC.

johngmyers:
It would make sense to have a mode which disables all ingress functionality of LBC, but it does not make sense to have LBC create ingresses but not admission-validate them.

/hold for provider-aws meeting

mhulscher:
That sounds reasonable. Looks like there was a related issue that may or may not have covered this exact scope: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2939 . It looks like there is already a way to disable the reconciliation of Services, but not yet of Ingresses.

Angelin01:
Let me add my current issue with this validation.

I deploy AWS LB Controller with Terraform right after the cluster is created. Similarly, I want to deploy ArgoCD so it can deploy the rest of the applications.

However, I'd like to specify the Ingress Class that ArgoCD's Ingress will use immediately, even if it doesn't exist. That way, when ArgoCD itself deploys the Ingress Controller, it will act upon the existing Ingress and everything will self-configure.

However, the webhook validation stops this from happening: it wants the Ingress resource to exist even if it is not watching for Ingress.

Honestly, such a feature is nice, but being able to disable it, be it globally or on a _per ingress basis_ is a must.

joaocc:
Hi. Any news with regards to this PR? Thx

johngmyers:
We made ingress validation ignore ingresses managed by other controllers in #3272. Any more is unlikely to move forward.

joaocc:
I'm still trying to understand if it's possible to configure the AWS LBC to restrict itself to the role of LBC, disabling all behaviours related to processing of ingresses, while keeping all processing on services and LB.
Compared with other cloud providers, it feel strange not having that option, as it effectively forces developers/engineers to have an extra component in their system when they actively decided it is not needed/wanted.
For instance, we are migrating from another provider where we have ingress done by nginx-ingress, with LB managed by the provider via annotation. in AWS we are already having to deploy another component (AWS LBC) as the in-tree controller seems to be path to deprecation. Not only that, but the current lack of flag forces another component to be listening for ingresses.
I have seen references to creating a fork to this chart and adding a variable to disable the ingress part of the chart, but it would be much easier for everyone if that type of feature was made available in the main chart, leaving it for each developer/engineer the decision to disable it according to their requirements.
Being new to AWS LBC, maybe this was already discussed elsewhere (could not find specific issue).
Thanks

Angelin01:
@joaocc We disabled most things here, using the Helm Chart. Maybe with the changes in #3272 [my use case](https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3011#issuecomment-1470800328) would no longer be problem, but I hope this is enough for your side of things:

```yaml
createIngressClassResource: false
enableShield: false
enableWaf: false
enableWafv2: false
disableIngressClassAnnotation: true
disableIngressGroupNameAnnotation: true
```

johngmyers:
Disabling all processing of Ingresses would be a different feature than this PR does.

Angelin01:
> Disabling all processing of Ingresses would be a different feature than this PR does.

I mean, yes, but this PR disables _one more_ component of the Ingress part of AWS LB Controller. In fact, I can't remember out of the top of my head anything else that needs disabling. Nonetheless, I agree the PR should probably be redone, but I still agree with the ideia.

johngmyers:
@Angelin01 I believe your use case would be handled: you would deploy the ArgoCD `IngressClass` with a `spec.controller` value indicating the other ingress controller. The LBC admission controller would then ignore Ingresses with that IngressClass.

I could envision a feature to disable all Ingress processing entirely, but that strikes me as being a bit niche.

joaocc:
@johngmyers we have a different case from @Angelin01 but I believe the main issue here is the fact that AWS LBC forces us to keep a component active in the system even after we determined we don't need/want it in that particular scenario. From several design perspectives, we try to minimize the amount of code running on our systems, as kubernetes world is already complex enough :)

johngmyers:
You're generally going to have little choice but to deploy LBC into an AWS cluster because the AWS CCM's service controller is deprecated and unlikely to get much further maintenance. LBC's service controller relies on webhooks, especially if you want it to be the default controller for Service.

So disabling LBC's ingress controller isn't likely to save much in either resources or complexity.

PR #3026, opened 2023-02-05
Use apimachinery conventions for ingress validation errors

johngmyers:
### Issue

N/A

### Description

Use the apimachinery error types for validation failures. These allow returning multiple failures at once and integrate better with tooling such as `kubectl edit`.

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/annotations/parser.go
+++ b/pkg/annotations/parser.go
@@ -3,9 +3,10 @@ package annotations
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/pkg/errors"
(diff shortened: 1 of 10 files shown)


johngmyers:
* Needs to do an earlier return when it determines the `Ingress` is for a different controller.
* Should return a `field.NotFound` when the `IngressClass` doesn't exist.
* The validation of the link from `IngressClass` to `IngressClassParams` should be in a webhook on the `IngressClass` (this can be a separate PR)

PR #3028, opened 2023-02-06
Remove constrains of multiple TLS on certificate auto-discovery

jerryhe1999:
Co-authored-by: LechG <lech.gorlewicz@gmail.com>

### Issue
Fixes: #1598 

### Description
This PR removes the constraint of single TLS certificate per host during auto-discovery. The controller will now unconditionally auto-discover all matching TLS certificates for the given host.  Auto-discovery can still be disabled by specifying the certificates manually via annotation.

The controller is backwards compatible since existing ingresses/groups that are treated valid have only one TLS certificate.

Manual test has been done by creating multiple certificates on the ACM for the same host: mijerryapp.fortest.com, create an ingress for this host, verify ALB gets created and all of the certs get added to the listener. The old controller did not work with the multiple certificates in place.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ingress/cert_discovery.go
+++ b/pkg/ingress/cert_discovery.go
@@ -80,9 +80,6 @@ func (d *acmCertDiscovery) Discover(ctx context.Context, tlsHosts []string) ([]s
 			}
 		}
 
-		if len(certARNsForHost) > 1 {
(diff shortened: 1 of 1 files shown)


kishorj (commented):
Thanks for the contribution. Could you also document the manual tests?

kishorj (approved):

PR #3029, opened 2023-02-06
Add script to create kOps cluster and run tests

jaydeokar:
### Issue

N/A

### Description 
This script add the ability to spin up a Kops cluster and run the test suite. 

### Checklist
- [X] Manually tested

### DIFF
--- a/scripts/lib/common.sh
+++ b/scripts/lib/common.sh
@@ -6,6 +6,7 @@ SECONDS=0
 
 # Set IP Family
 IP_FAMILY="${IP_FAMILY:-IPv4}"
+USE_EKSCTL="${USE_EKSCTL:-true}"
 
 echo "Running AWS Load Balancer Controller e2e tests with the following variables
 KUBE_CONFIG_PATH: $KUBE_CONFIG_PATH
@@ -13,7 +14,8 @@ CLUSTER_NAME: $CLUSTER_NAME
 REGION: $REGION
 IP_FAMILY: $IP_FAMILY
 OS_OVERRIDE: $OS_OVERRIDE
-ENDPOINT: $ENDPOINT"
+ENDPOINT: $ENDPOINT
+USE_EKSCTL: $USE_EKSCTL"
 
 if [[ -z "${OS_OVERRIDE}" ]]; then
   OS_OVERRIDE=linux
@@ -32,15 +34,18 @@ if [[ $REGION == "cn-north-1" || $REGION == "cn-northwest-1" ]]; then
   go env -w GOSUMDB=sum.golang.google.cn
 fi
 
-GET_CLUSTER_INFO_CMD="aws eks describe-cluster --name $CLUSTER_NAME --region $REGION"
+if [[ $USE_EKSCTL == true ]]; then
+  GET_CLUSTER_INFO_CMD="aws eks describe-cluster --name $CLUSTER_NAME --region $REGION"
 
-if [[ -z "${ENDPOINT}" ]]; then
-  CLUSTER_INFO=$($GET_CLUSTER_INFO_CMD)
-else
-  CLUSTER_INFO=$($GET_CLUSTER_INFO_CMD --endpoint $ENDPOINT)
+  if [[ -z "${ENDPOINT}" ]]; then
+    CLUSTER_INFO=$($GET_CLUSTER_INFO_CMD)
+  else
+    CLUSTER_INFO=$($GET_CLUSTER_INFO_CMD --endpoint $ENDPOINT)
+  fi
+
+  VPC_ID=$(echo $CLUSTER_INFO | jq -r '.cluster.resourcesVpcConfig.vpcId')
+
+  echo "VPC ID: $VPC_ID"
 fi
 
-VPC_ID=$(echo $CLUSTER_INFO | jq -r '.cluster.resourcesVpcConfig.vpcId')
 ACCOUNT_ID=$(aws sts get-caller-identity | jq -r '.Account')
-
-echo "VPC ID: $VPC_ID"
--- a/scripts/lib/install.sh
+++ b/scripts/lib/install.sh
@@ -1,0 +1,63 @@
+install_aws_lb_controller() {
+    local vpc_id="${1}"
+    local image="${2}"
+    local create_service_account="${3}"
+    local cluster_name="${CLUSTER_NAME}"
+    local region="${REGION}"
+    local optional_args=""
+
+    if [[ "${create_service_account}" == "false" ]] ; then
+        optional_args="--set serviceAccount.create=false --set serviceAccount.name=aws-load-balancer-controller"
+    fi
+
+    printf "Get update from eks-charts hub \n"
+    helm repo add eks https://aws.github.io/eks-charts
+    helm repo update
+
+    printf "Install aws-load-balancer-controller Helm chart \n"
+    helm upgrade -i aws-load-balancer-controller eks/aws-load-balancer-controller \
+        -n kube-system \
+        --set clusterName=$cluster_name --set region=$region --set vpcId=$vpc_id --set image.repository=$image $optional_args
+
+    wait_until_deployment_ready aws-load-balancer-controller
+
+}
+
+wait_until_deployment_ready(){
+
+    local deployment_name="${1}"
+    local attempts=0
+    local retries=60
+    printf "Checking if deployment ${deployment_name} is ready \n"
+
+    while [ ${attempts} -le ${retries} ]
+    do
+        desiredReplicas=$(kubectl get deployments.apps ${deployment_name} -n kube-system -ojsonpath="{.spec.replicas}")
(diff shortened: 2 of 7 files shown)


scripts/lib/kops.sh
```diff
@@ -0,0 +1,49 @@
+create_kops_cluster(){
```
kishorj, line 1: kops::create_cluster

scripts/lib/kops.sh
```diff
@@ -0,0 +1,49 @@
...
+
+    printf "Using ${KOPS_S3_BUCKET} as kops state store \n"
```
kishorj, line 3: echo instead of printf

scripts/lib/kops.sh
```diff
@@ -0,0 +1,49 @@
...
+    printf "Starting cluster creation: ${CLUSTER_NAME} \n"
+    "${KOPS_BIN}/kops" update cluster --name "${CLUSTER_NAME}" --yes --admin
+
+    printf "Waiting for cluster to be up: ${CLUSTER_NAME} \n"
+    "${KOPS_BIN}/kops" validate cluster --wait 15m
+
+    printf "Cluster creation complete \n"
+}
+
+delete_kops_cluster(){
```
kishorj, line 45: kops::delete_cluster

scripts/run_kops_test.sh
```diff
@@ -0,0 +1,100 @@
...
+    fi
+  fi
+
+  if ! delete_kops_cluster; then
+    printf "Delete cluster failed.. \n"
+  fi
+}
+
+: "${TEST_ID:=$RANDOM}"
+: "${REGION:='us-west-2'}"
```
kishorj, line 35: You could use :- to assign default value value only if it is missing. This provides a useful way to override the defaults when you invoke the scripts. For example,
REGION=${REGION:-"us-west-2"}

scripts/run_kops_test.sh
```diff
@@ -0,0 +1,100 @@
...
+: "${SKIP_MAKE_TEST_BINARIES:=false}"
+: "${EXTRA_GINKGO_FLAGS:=''}"
+: "${FOCUS:=''}"
+: "${PRINT_CONTROLLER_LOGS:=false}"
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+GINKGO_TEST_BUILD="$SCRIPT_DIR/../test/build"
+CONTROLLER_IAM_POLICY_FILE="$(dirname "${BASH_SOURCE[0]}")/../docs/install/iam_policy.json"
+CONTROLLER_IAM_POLICY_NAME="aws-lb-controller-kops-${TEST_ID}"
+KOPS_S3_BUCKET="kops-alb-eks-${AWS_ACCOUNT_ID}"
```
kishorj, line 51: Is AWS_ACCOUNT_ID always going to be set? If not, you might have to do something like 
  AWS_ACCOUNT_ID=$(aws sts get-caller-identity --region ${REGION} --query Account --output text)

scripts/run_kops_test.sh
```diff
@@ -0,0 +1,100 @@
...
+KOPS_STATE_STORE=s3://$KOPS_S3_BUCKET
+CLUSTER_NAME="kops-alb-test-cluster-${TEST_ID}.k8s.local"
+KOPS_BIN="~/kops_bin"
+SSH_KEYS=~/.ssh/kopstempkey
+USE_EKSCTL=false
+policy_arn=""
```
kishorj, line 57: use uppercase for variable names

scripts/run_kops_test.sh
```diff
@@ -0,0 +1,100 @@
...
+cleanup(){
+
+  exitcode=$?
+  if [[ ${exitcode} -gt 0 ]]; then
+    printf "Error occurred. Attempting to clean setup.. \n"
+  else
+    printf "Tests finished. Attempting to clean setup.. \n"
+  fi
+
+  if ! uninstall_aws_lb_controller; then
```
kishorj, line 15: did you verify further cleanup steps will get executed if this step fails?

scripts/run_kops_test.sh
```diff
@@ -0,0 +1,100 @@
...
+elif [[ $REGION == "cn-northwest-1" ]];then
+  image="961992271922.dkr.ecr.cn-northwest-1.amazonaws.com.cn/amazon/aws-load-balancer-controller"
+else
+  image="602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-load-balancer-controller"
+fi
+
+trap cleanup EXIT
+
+# Step 1: Create kOps cluster and wait for it to complete
+create_kops_cluster
```
kishorj, line 79: Please put this in a function, for example main, and invoke the main function

scripts/lib/install.sh
```diff
@@ -0,0 +1,63 @@
+install_aws_lb_controller() {
```
kishorj, line 1: Is this not handled by the ginkgo tests?

kishorj, line 1: This file involves helm chart installation/deletion. Maybe rename it to helm.sh. Then the functions like:
helm::install...

scripts/lib/common.sh
```diff
@@ -32,15 +34,18 @@ if [[ $REGION == "cn-north-1" || $REGION == "cn-northwest-1" ]]; then
...
-  CLUSTER_INFO=$($GET_CLUSTER_INFO_CMD --endpoint $ENDPOINT)
+  if [[ -z "${ENDPOINT}" ]]; then
+    CLUSTER_INFO=$($GET_CLUSTER_INFO_CMD)
+  else
+    CLUSTER_INFO=$($GET_CLUSTER_INFO_CMD --endpoint $ENDPOINT)
+  fi
+
+  VPC_ID=$(echo $CLUSTER_INFO | jq -r '.cluster.resourcesVpcConfig.vpcId')
+
+  echo "VPC ID: $VPC_ID"
```
kishorj, line 48: Are these not required for kops?

scripts/run_kops_test.sh
jaydeokar: Yes, the if statement catches the error. We are logging with echo and moving forward with other cleanup execution

scripts/lib/common.sh
jaydeokar: I didn't quite catch that.. VPC_ID is required by ginkgo tests which we pass as a parameter

scripts/lib/install.sh
jaydeokar: I don't think ginkgo installs it by default. It's similar to how we test using eksctl (e2e test)

scripts/run_kops_test.sh
```diff
@@ -0,0 +1,106 @@
...
+main(){
+
+  # Step 1: Create kOps cluster and wait for it to complete
+  kops::create_kops_cluster
+
+  # Step 2: Create policy and attach to nodes role created by kOps
+  POLICY_ARN=$(iam::create_policy "${CONTROLLER_IAM_POLICY_NAME}" "${CONTROLLER_IAM_POLICY_FILE}" "${REGION}")
+  iam::attach_policy_to_role "${POLICY_ARN}" "nodes.${CLUSTER_NAME}"
+
+  # Step 3: Set hop limit to 2 for IMDSv2
```
kishorj, line 88: You've specified the vpc-id and the region during helm installation, IMDS need not be enabled. Did you run into any errors?

scripts/run_kops_test.sh
```diff
@@ -0,0 +1,106 @@
...
+: "${KOPS_BINARY_VERSION:=v1.25.3}"
+: "${KUBERNETES_VERSION:=1.26.0}"
+: "${NODE_SIZE:=c5.xlarge}"
+: "${NODE_COUNT:=3}"
+: "${OS_OVERRIDE:=linux}"
+: "${ENDPOINT:=""}"
+: "${SKIP_MAKE_TEST_BINARIES:=false}"
+: "${EXTRA_GINKGO_FLAGS:=''}"
+: "${FOCUS:=''}"
+: "${PRINT_CONTROLLER_LOGS:=false}"
```
kishorj, line 45: lets print controller logs by default

scripts/lib/kops.sh
```diff
@@ -0,0 +1,49 @@
+kops::create_kops_cluster(){
```
kishorj, line 1: kops::create_cluster

scripts/run_kops_test.sh
jaydeokar: Yeah if I don't set hop limit to 2, this is what we see in the logs
```
"NoCredentialProviders: no valid providers in chain. Deprecated.\n\tFor verbose messaging see aws.Config.CredentialsChainVerboseErrors
```

jaydeokar: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1747#issuecomment-752215264
It looked similar to this

kishorj: got it. We are not using IRSA here, thats why.

PR #3085, opened 2023-03-01
Use the VPC ID from the remote target group to check pod IPs against its CIDR ranges

### LINKED ISSUE #3084: Cannot register nodes with target group outside host VPC
**Describe the bug**
When trying to register the IP of a pod with a target group that is outside of the current VPC, the controller don't set the `all` value for `AvailabilityZone` resulting in the request to register the instance failing.

**Steps to reproduce**
1. Have a IP target group in a peered VPC
2. Set up a target group binding against the target group of type IP
3. Wait for the controller to reconcile
4. Get an error message in the logs

```json
{"level":"error","ts":1677695808.508565,"logger":"controller.targetGroupBinding","msg":"Reconciler error","reconciler group":"elbv2.k8s.aws","reconciler kind":"TargetGroupBinding","name":"external-app-lb","namespace":"projectcontour","error":"ValidationError: You must specify a valid Availability Zone or select 'all' for all enabled Availability zones, since the IP address '10.130.6.97' is outside the VPC\n\tstatus code: 400, request id: e9f1daa7-be4a-41fb-907f-9d10ffa6781c"}
{"level":"info","ts":1677695972.3855486,"msg":"registering targets","arn":"...","targets":[{"AvailabilityZone":null,"Id":"10.130.175.52","Port":8080},{"AvailabilityZone":null,"Id":"10.130.6.97","Port":8080},{"AvailabilityZone":null,"Id":"10.130.80.237","Port":8080}]}
```

**Expected outcome**
I expect that the request to register the ip with the target group sets the `AvailabilityZone` parameter to all if the pod IP address is outside the CIDR range of the VPC of the target group instead of the VPC of the host machine.

**Environment**

* AWS Load Balancer controller version
`2.4.7`

* Kubernetes version
`1.24`

* Using EKS (yes/no), if so version?
Yes (`1.24`)

**Additional Context**:
This is where in the code the `AvailabilityZone` is set. My suggestion is to look at the CIDR range for the target group's VPC rather than inferring it from the VPC of the host before defaulting back to this behaviour.

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/d9c2a774b7a83eb42de0af2c087e8e3ef36186fa/pkg/targetgroupbinding/resource_manager.go#L396-L398

zeevallin:
### Issue
Attempts to fix #3084 

### Description
In this change we derive the VPC ID from the destination target group to determine if the pod IP is outside it's CIDR ranges or default back to the current behaviour if it's not.

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/targetgroupbinding/resource_manager.go
+++ b/pkg/targetgroupbinding/resource_manager.go
@@ -42,7 +42,8 @@ func NewDefaultResourceManager(k8sClient client.Client, elbv2Client services.ELB
 	podInfoRepo k8s.PodInfoRepo, sgManager networking.SecurityGroupManager, sgReconciler networking.SecurityGroupReconciler,
 	vpcInfoProvider networking.VPCInfoProvider,
 	vpcID string, clusterName string, failOpenEnabled bool, endpointSliceEnabled bool, disabledRestrictedSGRulesFlag bool,
-	eventRecorder record.EventRecorder, logger logr.Logger) *defaultResourceManager {
+	eventRecorder record.EventRecorder, logger logr.Logger,
+) *defaultResourceManager {
 	targetsManager := NewCachedTargetsManager(elbv2Client, logger)
 	endpointResolver := backend.NewDefaultEndpointResolver(k8sClient, podInfoRepo, failOpenEnabled, endpointSliceEnabled, logger)
 
@@ -235,7 +236,8 @@ func (m *defaultResourceManager) cleanupTargets(ctx context.Context, tgb *elbv2a
 // updateTargetHealthPodCondition will updates pod's targetHealth condition for matchedEndpointAndTargets and unmatchedEndpoints.
 // returns whether further probe is needed or not
 func (m *defaultResourceManager) updateTargetHealthPodCondition(ctx context.Context, targetHealthCondType corev1.PodConditionType,
-	matchedEndpointAndTargets []podEndpointAndTargetPair, unmatchedEndpoints []backend.PodEndpoint) (bool, error) {
+	matchedEndpointAndTargets []podEndpointAndTargetPair, unmatchedEndpoints []backend.PodEndpoint,
+) (bool, error) {
 	anyPodNeedFurtherProbe := false
 
 	for _, endpointAndTarget := range matchedEndpointAndTargets {
@@ -271,7 +273,8 @@ func (m *defaultResourceManager) updateTargetHealthPodCondition(ctx context.Cont
 // updateTargetHealthPodConditionForPod updates pod's targetHealth condition for a single pod and its matched target.
 // returns whether further probe is needed or not.
 func (m *defaultResourceManager) updateTargetHealthPodConditionForPod(ctx context.Context, pod k8s.PodInfo,
-	targetHealth *elbv2sdk.TargetHealth, targetHealthCondType corev1.PodConditionType) (bool, error) {
+	targetHealth *elbv2sdk.TargetHealth, targetHealthCondType corev1.PodConditionType,
+) (bool, error) {
 	if !pod.HasAnyOfReadinessGates([]corev1.PodConditionType{targetHealthCondType}) {
 		return false, nil
 	}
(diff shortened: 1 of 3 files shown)


M00nF1sh:
@zeevallin 
There were a similar PR in the past for this: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/2736/files
We had concerns that this will add a lot describe calls. (Seems your PR addressed the concern by caching it).

A better idea in my mind is store the VPC ID of targetGroups in the spec in TargetGroupBinding during TGB creation. Similar to TargetType(https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/apis/elbv2/v1beta1/targetgroupbinding_types.go#L132), if a VPC ID is not specified, the mutating webhook will sets the VPC ID directly. 

What do you think? we are fine to merge this PR as it is(after reviewed the detailed changes) and do above change ourself as well.

zeevallin:
I agree that if we can store the VPC ID in the binding and have that take precedence, that adds an extra layer of control for the user.

Both would probably be a good idea.

@M00nF1sh

rodrigorfk:
Hi there, I've just opened #3109 which is exactly related to what is being address here, for reference, as I've described there, we are running a patched version of the controller using a similar approach of what is being introduced here, meaning that we introduced a similar process of describing the TargetGroup and getting the right VPC id out of it, our implementation is also backed by a cache and after running the solution for more than a year on cluster with 450+ TargetGroupBindings, we haven't faced any side effect of such change.

ikosenn:
@M00nF1sh anything I can do to help get this across the finish line

johngmyers:
@ikosenn the recommendation was to add a `vpc` field to `TargetGroupBinding`. Perhaps you could open a competing PR to do that?

(though I wonder if the new field should go into the status instead of the spec)

ikosenn:
> @ikosenn the recommendation was to add a `vpc` field to `TargetGroupBinding`. Perhaps you could open a competing PR to do that?
> 
> (though I wonder if the new field should go into the status instead of the spec)

@johngmyers I'll work on getting a competing PR. I can think of good arguments for having it in spec or status. 
Do we want to stay consistent and follow patterns already established with TargetType? Trying to think of a case where the TargetGroup's TargetType should be handled any differently from the TargetGroup's VPC.

nathanpreen:
Hi @ikosenn @johngmyers is there any further progress on this at all? This is functionality we'd like to make use of so our controller can add targets to a Target Group in a different VPC to the cluster.

ikosenn:
> Hi @ikosenn @johngmyers is there any further progress on this at all? This is functionality we'd like to make use of so our controller can add targets to a Target Group in a different VPC to the cluster.

I haven't had time to work on this but it's definitely on my radar as I need it too for our setup. Should be able to start looking into it this weekend.

nathanpreen:
> > Hi @ikosenn @johngmyers is there any further progress on this at all? This is functionality we'd like to make use of so our controller can add targets to a Target Group in a different VPC to the cluster.
> 
> I haven't had time to work on this but it's definitely on my radar as I need it too for our setup. Should be able to start looking into it this weekend.

Thanks @ikosenn. Keen to get this progressed if I can help in anyway. We're looking to get this solved ASAP on our side to mitigate an architectural flaw that we have.

nathanpreen:
@ikosenn do you have an update on this at all by any chance?

ikosenn:
> @ikosenn do you have an update on this at all by any chance?

@nathanpreen  Nothing concrete still in the initial steps of understanding the project, creating test env e.t.c. Kindly, feel free to take up the issue if you need it resolved with some urgency. At the moment I am only able to work on this during my free time.

ikosenn:
@nathanpreen see https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3479

nathanpreen:
@ikosenn apologies for the delayed response, had other priorities at work and didn't check this. PR looks great! Hopefully it can get into a release candidate soon

Issue #3071, opened 2023-02-22
AWS LB Controller with EndpointSlices enabled not compatible with Kubernetes v1.25


mikestef9:
**Describe the bug**
Updated EKS cluster to v1.25, running latest version of LB controller v2.4.6, with following config

```
# enableEndpointSlices enables k8s EndpointSlices for IP targets instead of Endpoints (default false)
enableEndpointSlices: true
```

LB controller pod is crash looping, with following logs, seems we may have missed this https://kubernetes.io/docs/reference/using-api/deprecation-guide/#endpointslice-v125
```
{"level":"error","ts":1677101790.503112,"logger":"controller.targetGroupBinding","msg":"Could not wait for Cache to sync","reconciler group":"[elbv2.k8s.aws](http://elbv2.k8s.aws/)","reconciler kind":"TargetGroupBinding","error":"failed to wait for targetGroupBinding caches to sync: no matches for kind \"EndpointSlice\" in version \"[discovery.k8s.io/v1beta1\](http://discovery.k8s.io/v1beta1/)""}
{"level":"info","ts":1677101790.503688,"logger":"controller.service","msg":"Starting workers","worker count":3}
{"level":"info","ts":1677101790.5037487,"logger":"controller.service","msg":"Shutdown signal received, waiting for all workers to finish"}
{"level":"info","ts":1677101790.5042624,"logger":"controller.service","msg":"All workers finished"}
{"level":"error","ts":1677101790.5041447,"logger":"controller.ingress","msg":"Could not wait for Cache to sync","error":"failed to wait for ingress caches to sync: timed out waiting for cache to be synced"}
{"level":"error","ts":1677101790.5045505,"msg":"error received after stop sequence was engaged","error":"failed to wait for ingress caches to sync: timed out waiting for cache to be synced"}
{"level":"info","ts":1677101790.504128,"logger":"controller-runtime.webhook","msg":"shutting down webhook server"}
{"level":"error","ts":1677101790.504633,"logger":"setup","msg":"problem running manager","error":"failed to wait for targetGroupBinding caches to sync: no matches for kind \"EndpointSlice\" in version \"[discovery.k8s.io/v1beta1\](http://discovery.k8s.io/v1beta1/)""}
```

**Steps to reproduce**
Run Kubernetes v1.25 cluster with endpointSlices: true in the controller config. 

**Expected outcome**
Controller should work as normal

**Environment**

[...]

PR #3119, opened 2023-03-23
add test coverage for endPointSlices

oliviassss:
### Issues
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3071

### Description

Add the e2e test coverage with the feature endPointSlices enabled.

#### Tests
- Ran the e2e test on k8s 1.25 LBC v2.4.7, verified the test case passed.
- Negative test: ran the e2e test on k8s 1.25 LBC v2.4.6, where the API version `discovery.k8s.io/v1beta1` is not supported, and the test case failed.

### Checklist
- [x] Manually tested

### DIFF
--- a/test/framework/controller/installation_manager.go
+++ b/test/framework/controller/installation_manager.go
@@ -12,7 +12,7 @@ import (
 // InstallationManager is responsible for manage controller installation in cluster.
 type InstallationManager interface {
 	ResetController() error
-	UpgradeController(controllerImage string) error
+	UpgradeController(controllerImage string, enableEndPointSlices bool) error
 }
 
 // NewDefaultInstallationManager constructs new defaultInstallationManager.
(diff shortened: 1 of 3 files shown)


test/framework/controller/installation_manager.go
```diff
@@ -81,6 +84,7 @@ func (m *defaultInstallationManager) computeDefaultHelmVals() map[string]interfa
 		"create": false,
 		"name":   m.controllerSAName,
 	}
+	vals["enableEndpointSlices"] = false
```
kishorj, line 87: do we need this?

oliviassss: Removed this line, we don't need it since the default value of enableEndpointSlices is false. Thanks for pointing out.

kishorj (approved):

PR #3131, opened 2023-03-30
docs: add some context to the "RBAC config for secrets" doc

prasadkatti:
### Description

Attempting to provide some more context on why the new Role could be needed.

### Checklist
- [ ] Manually tested


johngmyers:
Doesn't everyone use projected tokens these days? Why would one get an OIDC token through a Secret?

PR #3139, opened 2023-04-04
generate manifest with service mutator and modify webhook

jerryhe1999:
### Issue

N/A

### Description

Originally if created a service type of LoadBalancer without any annotation, those services will get reconciled by the ccm. After installing the webhook, the webhook should add the spec.loadBalancerClass to new LoadBalancer service without annotation and then let it reconciled by the lb controller. 

When you create a service of type LoadBalancer without annotation, it is usually the ccm that creates the load balancer for that service. However, by using the AWS Load Balancer controller's webhook, you can choose to use a different load balancer by specifying the loadBalancerClass field. This can give you better control over the configuration and behavior of the load balancer to meet your specific needs.

### Checklist
- [x] Manually tested


helm/aws-load-balancer-controller/templates/webhook.yaml
```diff
@@ -56,6 +56,28 @@ webhooks:
     resources:
     - pods
   sideEffects: None
+- clientConfig:
```
kishorj, line 59: could we make this webhook conditional based on a configuration value. For example, `defaultControllerForService` set to `true` by default. So if you don't want to make the lbc as default controller for service of type LoadBalancer, you can install the controller with `--set defaultControllerForService=false`

helm/aws-load-balancer-controller/values.yaml
```diff
@@ -338,3 +338,6 @@ clusterSecretsPermissions:
 # ingressClassConfig contains configurations specific to the ingress class
 ingressClassConfig:
   default: false
+
+# disable MutatorWebhook for NLB (default false)
```
kishorj, line 342: update the comment to something similar to the following:
```
# disableServiceMutatorWebhook allows you disable the webhook which makes this controller the default for all new services of type LoadBalancer
```

helm/aws-load-balancer-controller/templates/webhook.yaml
```diff
@@ -56,6 +56,30 @@ webhooks:
     resources:
     - pods
   sideEffects: None
+{{ if .Values.disableServiceMutatorWebhook }}
```
kishorj, line 59: To chomp the whitespace to the left,
```
{{- if
```

helm/aws-load-balancer-controller/templates/webhook.yaml
```diff
@@ -56,6 +56,30 @@ webhooks:
...
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    resources:
+    - services
+  sideEffects: None
+{{ end }}
```
kishorj, line 82:
```
{{- end }}
```

kishorj (approved):

Issue #2367, opened 2021-11-16
Modify all SecurityGroups tagged with `kubernetes.io/cluster/$name`


ellistarn:
**Is your feature request related to a problem?**
Currently, the controller fails if there are multiple security groups tagged with `kubernetes.io/cluster/$name`.

**Describe the solution you'd like**
It's common to tag both controlplane and dataplane resources with `kubernetes.io/cluster/$name`. This is a powerful discovery mechanism to simplify the end user configuration burden. Right now, the lb controller assumes that there is only one security group with this tag, and that it's owned by the control plane. This creates compatibility issues like https://github.com/weaveworks/eksctl/issues/4054.

**Describe alternatives you've considered**
We could potentially use a different tag for the control plane security groups, but this may be challenging to coordinate with all providers (kops, capi, eks, etc)

ellistarn:
@M00nF1sh any progress on prioritization here?

armujahid:
I am also getting this issue if we use default karpenter provisioner in a newly created eks cluster https://karpenter.sh/v0.7.1/getting-started/getting-started-with-eksctl/#provisioner

ellistarn:
@armujahid, we made karpenter flexible to arbitrary tags, so you should be able to work around this.

armujahid:
@ellistarn yes that's what I have done by using another nodegroup selector like kubernetes.io/cluster/${CLUSTER_NAME}

mkotsalainen:
@armujahid @ellistarn Can you tell me what security group selector that I can use to fix this? I also have a plain vanilla eksctl cluster.

armujahid:
@mkotsalainen i am using  

kubernetes.io/cluster/${CLUSTER_NAME}: owned

Security group selector for my vanilla karpenter cluster

mkotsalainen:
Ok thanks. Before I saw your answer I tried 
```
provider:
    securityGroupSelector:
      Name: "*eks-cluster-sg-${CLUSTER_NAME}*"`
```
and that seems to fix my problem (getting traefik ingress controller to work with karpenter)

ellistarn:
@M00nF1sh any word on this one?

hitsub2:
It does not get fixed, right?

carflo:
My clusters ran into the same issue when we migrated over to Karpenter for node management (multiple SGs tagged with the cluster tag). 

[...]

PR #3147, opened 2023-04-12
Enables providing multiple tags for worker node security group discovery

carflo:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2367

### Description

* Adds a new flag `--service-target-eni-security-group-tags` to allow users to specify additional tags that should be used when the controller looks for the security group to use when adding ingress rules for NLB targets. This addresses the limitation called out in the documentation [here](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/service/nlb/#worker-node-security-groups-selection).

Example: Using `--service-target-eni-security-group-tags=Name=aws-lbc-nlb-sg` will look for a security group that has the required `kubernetes.io/cluster/${cluster-name}` tag AND a `Name=aws-lbc-nlb-sg` tag. 
 
Other:
* Fixes a couple typos in documentation.
* Removes trailing whitespace in documentation.

### Checklist
- [x] Manually tested


Pacobart:
Looking forward to this much needed functionality!

ricardorqr:
Any update on this PR?

carflo:
@kishorj @johngmyers is there anything else I can do to get a review from you folks? Thanks in advance 🙇

johngmyers:
It's in my queue

docs/guide/service/nlb.md
```diff
@@ -182,6 +182,8 @@ The controller automatically selects the worker node security groups that it mod
 
     `${cluster-name}` is the name of the Kubernetes cluster.
 
+In the case that you have more than one matching security group with the tag `kubernetes.io/cluster/${cluster-name}`, you may specify additional tags with the `endpoint-security-group-tags` to further specify the security group that should be used.
```
johngmyers, line 185:
```suggestion
In the case that you have more than one matching security group with the tag `kubernetes.io/cluster/${cluster-name}` on a target ENI, you may specify additional tags with the `--endpoint-security-group-tags` flag to further specify the security group that should be used.
```

pkg/targetgroupbinding/networking_manager.go
```diff
@@ -518,19 +520,36 @@ func (m *defaultNetworkingManager) resolveEndpointSGForENI(ctx context.Context,
...
 	}
 	clusterResourceTagKey := fmt.Sprintf("kubernetes.io/cluster/%s", m.clusterName)
-	sgIDsWithClusterTag := sets.NewString()
+	sgIDsWithMatchingEndpointSGTags := sets.NewString()
 	for sgID, sgInfo := range sgInfoByID {
+		isMatch := true
 		if _, ok := sgInfo.Tags[clusterResourceTagKey]; ok {
-			sgIDsWithClusterTag.Insert(sgID)
+			for endpointSGTagKey, endpointSGTagValue := range m.endpointSGTags {
+				if sgInfo.Tags[endpointSGTagKey] != endpointSGTagValue {
```
johngmyers, line 528: Doesn't correctly handle the case of `endpointSGTagValue == ""`. In that case, should fail to match SGs which don't have the tag key present.

johngmyers (changes requested):
The TargetGroupBinding reconciler is not an area I'm overly familiar with, so I'm learning as I go.

The name of the flag doesn't say much about what it does:

* "endpoint" is ambiguous. Perhaps use something more like `target-eni-security-group-tags`
* This is documented in the section specific to `Service`s and the description suggests that it is not relevant to `Ingress`es. I haven't yet found where in the code this gets bypassed for the ALB case. If it is specific to `Service`, then the flag name should start with `service-`.

docs/guide/service/nlb.md
johngmyers, line 185: By my reading of the code, it does not further refine only in the case where there is more than one matching security group, it requires all security groups to have the additional tags in order to be considered.

carflo:
> The TargetGroupBinding reconciler is not an area I'm overly familiar with, so I'm learning as I go.
> 
> The name of the flag doesn't say much about what it does:
> 
> * "endpoint" is ambiguous. Perhaps use something more like `target-eni-security-group-tags`
> * This is documented in the section specific to `Service`s and the description suggests that it is not relevant to `Ingress`es. I haven't yet found where in the code this gets bypassed for the ALB case. If it is specific to `Service`, then the flag name should start with `service-`.

Correct, this only applies to `Service` resources. How does `service-target-eni-security-group-tags` sound?

johngmyers:
> Correct, this only applies to `Service` resources. How does `service-target-eni-security-group-tags` sound?

That seems reasonable.

docs/guide/service/nlb.md
carflo: Correct. I worded it this way because there should only be one matching security group in the end, but maybe this is better wording?

```
To further refine which security groups should be considered, use the `--service-target-eni-security-group-tags` flag. The controller will then only consider security groups that contain all the tags defined in this flag.
``` 

Feel free to suggest something better though :)

pkg/targetgroupbinding/networking_manager.go
carflo: Ah good catch! I've updated the if statement to also check if the key exists and added [2 test cases](https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3147/files#diff-2da03592457251d5626b5e89577539ef29eced083f9a326b5923a91fcea9c11cR1676-R1755) to cover empty tag values.

carflo:
It looks like there's timeouts happening when the tests run but I'm not entirely sure these are related to the code changes.

docs/guide/service/nlb.md
```diff
@@ -182,6 +182,8 @@ The controller automatically selects the worker node security groups that it mod
 
     `${cluster-name}` is the name of the Kubernetes cluster.
 
+In the case that you have more than one matching security group with the tag `kubernetes.io/cluster/${cluster-name}` on a target ENI, you may specify additional tags with the `--service-target-eni-security-group-tags` flag to further specify the security group that should be used.
```
johngmyers, line 185: Something like 

```suggestion
If it is possible for multiple security groups with the tag `kubernetes.io/cluster/${cluster-name}` to be on a target ENI, you may use the `--service-target-eni-security-group-tags` flag to specify additional tags that must also match in order for a security group to be used.
```

carflo:
@johngmyers thanks for the feedback! I think I've made all the requested changes and I'm just now needing the PR checks to pass.

ketan-jpmc:
Looking forward to this much needed functionality!

arunslm:
> Looking forward to this much needed functionality!

+1

ssudhir123:
+1. Much needed one.

pkg/config/controller_config.go
```diff
@@ -74,6 +75,9 @@ type ControllerConfig struct {
 	// List of Tag keys on AWS resources that will be managed externally.
 	ExternalManagedTags []string
 
+	// AWS Tags that will be used by the controller to find the worker node security group to add inbound rules from NLBs.
```
johngmyers, line 78:
```suggestion
	// ServiceTargetENISGTags are AWS tags, in addition to the cluster tags, for finding the target ENI security group to which to add inbound rules from NLBs.
```

pkg/config/controller_config.go
```diff
@@ -128,7 +132,8 @@ func (cfg *ControllerConfig) BindFlags(fs *pflag.FlagSet) {
 		"Enable EndpointSlices for IP targets instead of Endpoints")
 	fs.BoolVar(&cfg.DisableRestrictedSGRules, flagDisableRestrictedSGRules, defaultDisableRestrictedSGRules,
 		"Disable the usage of restricted security group rules")
-
+	fs.StringToStringVar(&cfg.EndpointSGTags, flagServiceTargetENISGTags, nil,
+		"AWS Tags that will be used by the controller to find the worker node security group to add inbound rules from NLBs")
```
johngmyers, line 136:
```suggestion
		"AWS Tags, in addition to cluster tags, for finding the target ENI security group to which to add inbound rules from NLBs")
```

pkg/config/controller_config.go
```diff
@@ -74,6 +75,9 @@ type ControllerConfig struct {
 	// List of Tag keys on AWS resources that will be managed externally.
 	ExternalManagedTags []string
 
+	// AWS Tags that will be used by the controller to find the worker node security group to add inbound rules from NLBs.
+	EndpointSGTags map[string]string
```
johngmyers, line 79:
```suggestion
	ServiceTargetENISGTags map[string]string
```

johngmyers (changes requested):

barathnagarajan:
+1

raj-mehta:
+1

carflo:
@johngmyers the `throttle` package appears to have some flaky tests. These pass locally on my machine and my changes don't include any changes to that package. I'll keep running `/retest` in the meantime but do you have any other suggestions?

johngmyers:
My current PR to attempt to fix this is kubernetes/test-infra#30160

Pacobart:
When will this be available. We've been waiting for this feature for a few months now. Please prioritize merging this much needed functionality.

pkg/config/controller_config.go
```diff
@@ -18,6 +18,7 @@ const (
 	flagDefaultTags                                  = "default-tags"
 	flagDefaultTargetType                            = "default-target-type"
 	flagExternalManagedTags                          = "external-managed-tags"
+	flagServiceTargetENISGTags                       = "service-target-eni-security-group-tags"
```
M00nF1sh, line 21: how about just `endpoint-eni-security-group-tags`

M00nF1sh:
left some comments.

this seems an alternative solution to https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2367, but requires more manual intervention(extra tags on SGs).

I'm good with this change as a temporary solution and it works for customers that only want one SG to be modified.

pkg/targetgroupbinding/networking_manager.go
```diff
@@ -518,19 +520,36 @@ func (m *defaultNetworkingManager) resolveEndpointSGForENI(ctx context.Context,
 		return "", err
 	}
 	clusterResourceTagKey := fmt.Sprintf("kubernetes.io/cluster/%s", m.clusterName)
-	sgIDsWithClusterTag := sets.NewString()
+	sgIDsWithMatchingEndpointSGTags := sets.NewString()
 	for sgID, sgInfo := range sgInfoByID {
```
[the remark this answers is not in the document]
carflo: Much cleaner. Thanks!

pkg/config/controller_config.go
johngmyers: The flag is specific to Service, so the "service-" prefix is appropriate.

pkg/targetgroupbinding/networking_manager_test.go
```diff
@@ -1412,3 +1414,368 @@ func Test_defaultNetworkingManager_computeRestrictedIngressPermissionsPerSG(t *t
...
+	type args struct {
+		ctx     context.Context
+		eniInfo networking.ENIInfo
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		want    string
+		wantErr bool
```
johngmyers, line 1438: A better pattern is to have `wantErr string` with the value being the expected error message.

johngmyers (changes requested):

pkg/targetgroupbinding/networking_manager_test.go
```diff
@@ -1435,7 +1435,7 @@ func Test_defaultNetworkingManager_resolveEndpointSGForENI(t *testing.T) {
 		fields  fields
 		args    args
 		want    string
-		wantErr bool
+		wantErr error
```
johngmyers, line 1438:
```suggestion
		wantErr string
```

There's little point to constructing an `error` just to unpack its message in the test.

carflo: There's other tests in the same file that do the same so I was trying to keep it consistent.

M00nF1sh (approved):
/lgtm
/approve
Thanks for making this contribution :D

Issue #3167, opened 2023-04-20
Throttling: Rate exceeded error when reconcile ALB


lmbaschiera:
**Describe the bug**
We're currently upgrading our alb ingress controller to v2.4.4 from v1.2.0-alpha.1 in preparation for a kubernetes upgrade.
Hundreds of `DescribeTargetHealth`, `DescribeTargetGroups`, `DescribeTags` and `RegisterTargets` events are making us hit the API rate limits impacting all clusters - even the ones not upgraded to the new controller version.

How can I fix this issue? I've asked for an API rate limit increase but that's just a band-aid for this issue.

`{"level":"error","ts":1682003556.0227876,"logger":"controller-runtime.manager.controller.ingress","msg":"Reconciler error","name":"release-8611","namespace":"","error":"Throttling: Rate exceeded\n\tstatus code: 400, request id: a450ce32-cf10-48a4-9736-aa5de377aa93"}`

We're having issue creating and destroying ingresses because we hit the limits even when destroying the load balancers, leaving a lot of `Terminating` resources in our clusters.

**Environment**

* AWS Load Balancer controller version v2.4.4
* Kubernetes version v1.21.14
* Using EKS (yes/no), if so version? yes

**Additional Context**:
We use these clusters to provision whole sets of applications in isolated environments for testing purposes, so we create and destroy dozens of environments per day, and each environment has between 20 and 30 ingress resources.

oliviassss:
@lmbaschiera The fix has been shipped with the v2.5.2 release, I'm closing this issue as of now. Please feel free to reach out if you're still having this issue.

Issue #1871, opened 2021-03-10
Throttling: Rate exceeded error when reconcile ALB


thanhma:
I'm getting the Rate exceeded error in v2.1.3
```
{"level":"error","ts":1615309687.1415963,"logger":"controller","msg":"Reconciler error","reconcilerGroup":"elbv2.k8s.aws","reconcilerKind":"TargetGroupBinding","controller":"targetGroupBinding","name":"k8s-default-alias4bh-d5c9e3xxxx","namespace":"default","error":"Throttling: Rate exceeded\n\tstatus code: 400, request id: 4eef24db-b4b3-46bf-88da-5ed3eb4xxxxx"}
```
The `targetgroupbindingMaxConcurrentReconciles` is left to None (as by default).

After query the `ThrottlingException` as guided in https://aws.amazon.com/premiumsupport/knowledge-center/cloudtrail-rate-exceeded/, I found that exception thrown on `DescribeTargetHealth`, `DescribeTargetGroups`, `DescribeTags`, `RegisterTargets`... events

The number of ingress is big (about 10 ingresses per cluster, 200 ingresses in a single VPC). Target type is `instance`, each cluster has about 60 instances and the number of target per ingress may reach to 5,000 since rules are numerous as well.

M00nF1sh:
@thanhma 
By default, the controller didn't set any throttling for the AWS API calls, and we provide a flag to set it by customers.
let me find the recommended settings and reply back here

thanhma:
@M00nF1sh 
In CloudTrail log, I can see many `DescribeTags`, `DescribeTargetHealth`, `DescribeTargetGroups` calls getting `ThrottlingException`. Do these failures affect reconcile with `instance` target type? May it cause target register failed, or performance degrade?

Do we have a rate limit configuration?

thanhma:
@M00nF1sh 
My cluster stopped reconciling ALB because of `RegisterTarget` facing `Throttling Exception: Rate exceeded` error. 
Is there a way to optimize API calls to prevent throttling?

M00nF1sh:
@thanhma 

Yes, you can set via controller flag: ```--aws-api-throttle=Elastic Load Balancing v2:^RegisterTarget=20:3,Elastic Load Balancing v2:^Describe=40:10```
Above is the default throttle for normal AWS accounts. If you have multiple controller/cluster, they are sharing the same throttle per account/region, and you'll need to lower the throttle settings above.

[...]

PR #3186, opened 2023-05-02
Support AWS RGT APIs with feature flag

oliviassss:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3167
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1871

### Description

- This PR adds the support of AWS Resource Group API calls to `ListTargetGroups` and `ListLoadBalancers` of tagging manager. Also, it provids a feature gates flag `EnableRGTAPI` with default value to `false`, so no behavior change for existing users. 
- In order to enable the RGT API calls
    - set `--feature-gates=EnableRGTAPI=true` in command line flag
    - add`tag:GetResources` permission to the IAM policy used by the controller.

#### Tests
- Unit tests, and e2e tests with `EnableRGTAPI=true`
- Installed and deleted ingresses/services with `EnableRGTAPI=true`, verified the controller functioned as expected, and the resources were up and running

### Checklist
- [x] Manually tested

### DIFF
--- a/controllers/ingress/group_controller.go
+++ b/controllers/ingress/group_controller.go
@@ -51,7 +51,7 @@ func NewGroupReconciler(cloud aws.Cloud, k8sClient client.Client, eventRecorder 
 	enhancedBackendBuilder := ingress.NewDefaultEnhancedBackendBuilder(k8sClient, annotationParser, authConfigBuilder)
 	referenceIndexer := ingress.NewDefaultReferenceIndexer(enhancedBackendBuilder, authConfigBuilder, logger)
 	trackingProvider := tracking.NewDefaultProvider(ingressTagPrefix, controllerConfig.ClusterName)
-	elbv2TaggingManager := elbv2deploy.NewDefaultTaggingManager(cloud.ELBV2(), cloud.VpcID(), controllerConfig.FeatureGates, logger)
+	elbv2TaggingManager := elbv2deploy.NewDefaultTaggingManager(cloud.ELBV2(), cloud.VpcID(), controllerConfig.FeatureGates, cloud, logger)
 	modelBuilder := ingress.NewDefaultModelBuilder(k8sClient, eventRecorder,
 		cloud.EC2(), cloud.ACM(),
 		annotationParser, subnetsResolver,
--- a/controllers/ingress/group_controller.go
+++ b/controllers/ingress/group_controller.go
@@ -51,7 +51,7 @@ func NewGroupReconciler(cloud aws.Cloud, k8sClient client.Client, eventRecorder 
 	enhancedBackendBuilder := ingress.NewDefaultEnhancedBackendBuilder(k8sClient, annotationParser, authConfigBuilder)
 	referenceIndexer := ingress.NewDefaultReferenceIndexer(enhancedBackendBuilder, authConfigBuilder, logger)
 	trackingProvider := tracking.NewDefaultProvider(ingressTagPrefix, controllerConfig.ClusterName)
-	elbv2TaggingManager := elbv2deploy.NewDefaultTaggingManager(cloud.ELBV2(), cloud.VpcID(), controllerConfig.FeatureGates, cloud, logger)
+	elbv2TaggingManager := elbv2deploy.NewDefaultTaggingManager(cloud.ELBV2(), cloud.VpcID(), controllerConfig.FeatureGates, cloud.RGT(), logger)
 	modelBuilder := ingress.NewDefaultModelBuilder(k8sClient, eventRecorder,
 		cloud.EC2(), cloud.ACM(),
 		annotationParser, subnetsResolver,
--- a/controllers/service/service_controller.go
+++ b/controllers/service/service_controller.go
@@ -40,7 +40,7 @@ func NewServiceReconciler(cloud aws.Cloud, k8sClient client.Client, eventRecorde
 
 	annotationParser := annotations.NewSuffixAnnotationParser(serviceAnnotationPrefix)
 	trackingProvider := tracking.NewDefaultProvider(serviceTagPrefix, controllerConfig.ClusterName)
-	elbv2TaggingManager := elbv2.NewDefaultTaggingManager(cloud.ELBV2(), cloud.VpcID(), controllerConfig.FeatureGates, cloud, logger)
+	elbv2TaggingManager := elbv2.NewDefaultTaggingManager(cloud.ELBV2(), cloud.VpcID(), controllerConfig.FeatureGates, cloud.RGT(), logger)
 	serviceUtils := service.NewServiceUtils(annotationParser, serviceFinalizer, controllerConfig.ServiceConfig.LoadBalancerClass, controllerConfig.FeatureGates)
 	modelBuilder := service.NewDefaultModelBuilder(annotationParser, subnetsResolver, vpcInfoProvider, cloud.VpcID(), trackingProvider,
 		elbv2TaggingManager, controllerConfig.FeatureGates, controllerConfig.ClusterName, controllerConfig.DefaultTags, controllerConfig.ExternalManagedTags, controllerConfig.DefaultSSLPolicy, controllerConfig.FeatureGates.Enabled(config.EnableIPTargetType), serviceUtils)
--- a/controllers/service/service_controller.go
+++ b/controllers/service/service_controller.go
@@ -40,7 +40,7 @@ func NewServiceReconciler(cloud aws.Cloud, k8sClient client.Client, eventRecorde
 
 	annotationParser := annotations.NewSuffixAnnotationParser(serviceAnnotationPrefix)
 	trackingProvider := tracking.NewDefaultProvider(serviceTagPrefix, controllerConfig.ClusterName)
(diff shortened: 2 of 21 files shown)


controllers/ingress/group_controller.go
```diff
@@ -51,7 +51,7 @@ func NewGroupReconciler(cloud aws.Cloud, k8sClient client.Client, eventRecorder
 	enhancedBackendBuilder := ingress.NewDefaultEnhancedBackendBuilder(k8sClient, annotationParser, authConfigBuilder)
 	referenceIndexer := ingress.NewDefaultReferenceIndexer(enhancedBackendBuilder, authConfigBuilder, logger)
 	trackingProvider := tracking.NewDefaultProvider(ingressTagPrefix, controllerConfig.ClusterName)
-	elbv2TaggingManager := elbv2deploy.NewDefaultTaggingManager(cloud.ELBV2(), cloud.VpcID(), controllerConfig.FeatureGates, logger)
+	elbv2TaggingManager := elbv2deploy.NewDefaultTaggingManager(cloud.ELBV2(), cloud.VpcID(), controllerConfig.FeatureGates, cloud, logger)
 	modelBuilder := ingress.NewDefaultModelBuilder(k8sClient, eventRecorder,
```
M00nF1sh, line 55: should be passing cloud.RGT() instead of entire cloud .

controllers/service/service_controller.go
```diff
@@ -40,7 +40,7 @@ func NewServiceReconciler(cloud aws.Cloud, k8sClient client.Client, eventRecorde
 
 	annotationParser := annotations.NewSuffixAnnotationParser(serviceAnnotationPrefix)
 	trackingProvider := tracking.NewDefaultProvider(serviceTagPrefix, controllerConfig.ClusterName)
-	elbv2TaggingManager := elbv2.NewDefaultTaggingManager(cloud.ELBV2(), cloud.VpcID(), controllerConfig.FeatureGates, logger)
+	elbv2TaggingManager := elbv2.NewDefaultTaggingManager(cloud.ELBV2(), cloud.VpcID(), controllerConfig.FeatureGates, cloud, logger)
```
M00nF1sh, line 43: same as above

docs/deploy/configurations.md
```diff
@@ -157,5 +157,6 @@ They are a set of kye=value pairs that describe AWS load balance controller feat
 | EndpointsFailOpen                     | string                          | false          | Enable or disable allowing endpoints with `ready:unknown` state in the target groups. |
 | EnableServiceController               | string                          | true           | Toggles support for `Service` type resources. |
 | EnableIPTargetType                    | string                          | true           | Used to toggle support for target-type `ip` across `Ingress` and `Service` type resources. |
+| EnableRGTCalls                        | string                          | false          | If enabled, the tagging manager will describe resource tags via RGT APIs, otherwise via ELB APIs. |
```
M00nF1sh, line 160: maybe `EnableRGTAPI` ?

docs/install/iam_policy.json
```diff
@@ -236,6 +236,13 @@
                 "elasticloadbalancing:ModifyRule"
             ],
             "Resource": "*"
+        },
+        {
+            "Effect": "Allow",
```
M00nF1sh, line 241: let's not add to default policy since customers don't need this by default.
we can specify in documentation that `tag:GetResources` is needed when enable the feature flag

pkg/aws/services/rgt.go
```diff
@@ -17,6 +27,42 @@ func NewRGT(session *session.Session) RGT {
...
+
+func ParseRGTTags(tags []*resourcegroupstaggingapi.Tag) map[string]string {
+	result := make(map[string]string, len(tags))
+	for _, tag := range tags {
+		result[aws.StringValue(tag.Key)] = aws.StringValue(tag.Value)
+	}
+	return result
+}
+
+func NewRGTTagFilters(tags []tracking.TagFilter) []*resourcegroupstaggingapi.TagFilter {
```
M00nF1sh, line 57: this `NewRGTTagFilters ` function should be put into tracking package. the `pkg/aws/service` should be free of our business logic, so it don't know "tracking" package at all.

pkg/deploy/elbv2/tagging_manager.go
```diff
@@ -254,7 +259,12 @@ func (m *defaultTaggingManager) ListLoadBalancers(ctx context.Context, tagFilter
 		lbARNsWithinVPC = append(lbARNsWithinVPC, lbARN)
 		lbByARNWithinVPC[lbARN] = lb
 	}
-	tagsByARN, err := m.describeResourceTags(ctx, lbARNsWithinVPC)
+	tagsByARN := make(map[string]map[string]string)
+	if m.featureGates.Enabled(config.EnableRGTCalls) {
+		tagsByARN, err = m.describeResourceTagsViaRGT(ctx, services.ResourceTypeELBLoadBalancer, tagFilters)
+	} else {
```
M00nF1sh, line 265: this is not correct.
the purpose of our change is to replace the entire client-side tag filter of `DescribeLoadBalancersAsList+ describeResourceTags ` to a single `RGT.GetResources` call.

With this change, we are still list all LoadBalancer & TargetGroups, which is not expected

pkg/deploy/elbv2/tagging_manager.go
```diff
@@ -237,13 +242,69 @@ func (m *defaultTaggingManager) ListListenerRules(ctx context.Context, lsARN str
...
+	if err != nil {
+		return nil, err
+	}
+	return matchedLBs, nil
+}
+
+func (m *defaultTaggingManager) ListTargetGroups(ctx context.Context, stackTags map[string]string, tagFilters ...tracking.TagFilter) ([]TargetGroupWithTags, error) {
+	var matchedTGs []TargetGroupWithTags
+	var err error
+	if m.featureGates.Enabled(config.EnableRGTAPI) {
```
M00nF1sh, line 262: same as above.

pkg/deploy/elbv2/load_balancer_synthesizer.go
```diff
@@ -114,6 +114,7 @@ func (s *loadBalancerSynthesizer) findSDKLoadBalancers(ctx context.Context) ([]L
 	stackTags := s.trackingProvider.StackTags(s.stack)
 	stackTagsLegacy := s.trackingProvider.StackTagsLegacy(s.stack)
 	return s.taggingManager.ListLoadBalancers(ctx,
+		stackTags,
```
M00nF1sh, line 117: i think we don't pass this additional stackTags.

why not directly convert TagFilter to RGTFilter? we can add a method to [`TagFilter `](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/deploy/tracking/tag_filter.go#LL7C8-L7C8) to convertItToRGTFilter.

since this ListLoadBalancers's interface accepts a list of tag filters, we can do as following:
```
lbARNS = sets.NewString().          # need to use a set here to avoid conflicts since same arn can be returned multiple times.
for _, tagFilter in tagFilters {
   resources, err = GetResources(...tagFilter)
  if err != nil {
     return err;
   }
   for _, resource in resources {
     lbARNS.insert(resource.arn)
  }
}
for _, lbARN := lbARNS {
......original other logic...
}
```
This way we will still discover the resources tagged with "legacyTags"(v1 controller), so remains our backwards compability as well. (only downside is we will be doing two RGT calls but are fine to me, we can remove legacyTags from the input once we decide to drop this v1->v2 migration support)

oliviassss: Reverted the addition of stackTags, thanks for the hint! However I just keeps the resources in the updated PR, because we still need the `resource.Tags` in addtion to the `resource.ResourceARN`

pkg/deploy/elbv2/tagging_manager.go
```diff
@@ -238,12 +243,70 @@ func (m *defaultTaggingManager) ListListenerRules(ctx context.Context, lsARN str
 }
 
 func (m *defaultTaggingManager) ListLoadBalancers(ctx context.Context, tagFilters ...tracking.TagFilter) ([]LoadBalancerWithTags, error) {
+	var matchedLBs []LoadBalancerWithTags
+	var err error
+	if m.featureGates.Enabled(config.EnableRGTAPI) {
+		matchedLBs, err = m.listLoadBalancersRGT(ctx, tagFilters)
+	}
```
M00nF1sh, line 250: if m.featureGates.Enabled(config.EnableRGTAPI) {
   return matchedLBs, err = m.listLoadBalancersRGT(ctx, tagFilters);
}
return m.listLoadBalancersNative(ctx, tagFilters);

M00nF1sh (approved):
/lgtm
/approve

ejholmes:
The release notes state:

> RGT feature is not available for private clusters

Is there a good explanation somewhere for why this is? We'd love to test this, but our EKS clusters are private. Reading the code here, I can't see any reason why that'd matter though.

rittneje:
@oliviassss @M00nF1sh We have the same question with regards to what a "private cluster" is. Based on the code it sounds like it has something to do with PrivateLink, but I'm unclear what that has to do with the load balancer.

Issue #3185, opened 2023-05-02
Optional Certs not being attached when adding cert annotation for the first time


nickzelei:
**Describe the bug**
When adding the SSL cert annotation for the first time with multiple certificates, only the default certificate is associated with the load balancer.
After restarting the deployment of the aws load balancer controller, it properly adds the second certificate.

**Steps to reproduce**
I create a service of type LoadBalancer and have it spin up initially without any SSL certificates attached.
This spins up the LB in standard HTTP mode.

Initial Annotations:
```
    service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
    service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
    service.beta.kubernetes.io/aws-load-balancer-ssl-negotiation-policy: ELBSecurityPolicy-TLS13-1-2-2021-06
    service.beta.kubernetes.io/aws-load-balancer-type: external
```

I then create two ACM certificates.
Once they have been properly issued, I then attach them by adding the ssl annotation:
```
    service.beta.kubernetes.io/aws-load-balancer-ssl-cert: arn:aws:acm:us-west-2:295461250191:certificate/ea714dcd-1dcd-47b9-a750-f6fdbe771d35,arn:aws:acm:us-west-2:295461250191:certificate/a4ce1483-bb9a-41e6-a203-804c4d40cf5f
```

This updates the NLB to use only the first certificate. The second cert remains unused.
I roll the aws-lbc pods, then the second certificate is added.

I have been able to consistently repro this by:
1. removing the ssl-cert annotation
2. waiting for ACM to mark the certs as unused.
3. Re-adding the ssl-cert annotation with both certs.
4. aws-lbc consistently only adds the first certificate
5. rolling the deployment via `kubectl`
6. aws-lbc associates the missing certificate

To workaround this, when adding the cert, only add a single one with the annotation. Then update the service to include the second cert.
This isn't ideal as automated systems are adding both of these at the same time.

**Expected outcome**
I expect each certificate to be associated with the load balancer regardless of how or when they are added to the annotation.

**Environment**

* AWS Load Balancer controller version
  * v2.5.1 (via helm chart 1.5.2)
* Kubernetes version
  * 1.25
* Using EKS (yes/no), if so version?
  * yes - eks.2

**Additional Context**:

Here is a link to the logs. The first file is the booting up of the AWS-LBC. You can see at the bottom when it detects the addition of the cert annotation.
[...]

PR #3196, opened 2023-05-11
check both sdkLS and resLS sslpolicy for nil

oliviassss:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3185

### Description

Check both sdkLS and resLS sslpolicy for nil when updating extra certificates for listeners. This PR fixes the edge case that when adding ssl certs to an existing service that has no ssl cert originally, the controller will not skip adding extra certs because of the sslpolicy is nil in sdkLS.

#### Tests
- Created a NLB with no ssl cert specified, then added multiple certs via annotation `service.beta.kubernetes.io/aws-load-balancer-ssl-cert`, verified all the certs got added to the listeners of the NLB successfully
- Created a ALB with no ssl cert specified, then added multiple certs via annotation `alb.ingress.kubernetes.io/certificate-arn`, verified all the certs got added to the listeners of the ALB successfully
### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/deploy/elbv2/listener_manager.go
+++ b/pkg/deploy/elbv2/listener_manager.go
@@ -158,8 +158,8 @@ func (m *defaultListenerManager) updateSDKListenerWithSettings(ctx context.Conte
 func (m *defaultListenerManager) updateSDKListenerWithExtraCertificates(ctx context.Context, resLS *elbv2model.Listener,
 	sdkLS ListenerWithTags, isNewSDKListener bool) error {
 	// if TLS is not supported, we shouldn't update
-	if sdkLS.Listener.SslPolicy == nil {
-		m.logger.V(1).Info("SDK Listener doesn't have SSL Policy set, we skip updating extra certs for non-TLS listener.")
+	if resLS.Spec.SSLPolicy == nil && sdkLS.Listener.SslPolicy == nil {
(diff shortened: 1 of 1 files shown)


pkg/deploy/elbv2/listener_manager.go
```diff
@@ -158,8 +158,8 @@ func (m *defaultListenerManager) updateSDKListenerWithSettings(ctx context.Conte
 func (m *defaultListenerManager) updateSDKListenerWithExtraCertificates(ctx context.Context, resLS *elbv2model.Listener,
 	sdkLS ListenerWithTags, isNewSDKListener bool) error {
 	// if TLS is not supported, we shouldn't update
-	if sdkLS.Listener.SslPolicy == nil {
-		m.logger.V(1).Info("SDK Listener doesn't have SSL Policy set, we skip updating extra certs for non-TLS listener.")
+	if resLS.Spec.SSLPolicy == nil && sdkLS.Listener.SslPolicy == nil {
+		m.logger.Info("Res and Sdk Listener don't have SSL Policy set, we skip updating extra certs for non-TLS listener.")
```
kishorj, line 162: This always prints out the message for non-TLS listeners, I feel the log should still be V(1)

oliviassss: revert back to V(1), thanks

PR #3230, opened 2023-06-07
Remove policy/v1beta1 since the min supported k8s version supports policy/v1

rdrgmnzs:
### Description

Remove policy/v1beta1 since as of version 2.5.x the min supported k8s version 1.22 supports policy/v1.

### Checklist
- [X] Manually tested

### DIFF
--- a/helm/aws-load-balancer-controller/templates/pdb.yaml
+++ b/helm/aws-load-balancer-controller/templates/pdb.yaml
@@ -1,9 +1,5 @@
 {{- if and .Values.podDisruptionBudget (gt (int .Values.replicaCount) 1) }}
-{{- if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" }}
 apiVersion: policy/v1
-{{- else }}
-apiVersion: policy/v1beta1
-{{- end }}
 kind: PodDisruptionBudget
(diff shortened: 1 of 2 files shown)


johngmyers:
I don't see how this "fixes" the check, but it does make it easier to tell `helm template` that the capability exists.

/ok-to-test
/lgtm

kishorj:
policy/v1 is available since k8s 1.21. The min version we support with the latest release v2.5.x is k8s 1.22 or later. We can remove the policy/v1beta1 completely from the helm template.

helm/aws-load-balancer-controller/templates/pdb.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if and .Values.podDisruptionBudget (gt (int .Values.replicaCount) 1) }}
-{{- if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" }}
```
kishorj, line 2: We can remove this capability check completely and explicitly use policy/v1

rdrgmnzs:
Ok, went ahead and removed the conditional so we explicitly rely on policy/v1.

kishorj (approved):

Issue #3243, opened 2023-06-13
Make `enable-leader-election` toggle-able via Helm Chart


rblaine95:
**Is your feature request related to a problem?**
The AWS Load Balancer Controller currently has a [command line argument](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.5/deploy/configurations/#controller-command-line-flags) for enabling/disabling leader election.

Would it be possible to be able to set this to `false` via the helm chart?
Is there a specific reason why this hasn't already been done?

**Describe the solution you'd like**
Able to set `--enable-leader-election=true|false` via the Helm Chart like other CLI arguments.

dabcoder:
I have a branch for this in my fork, will open a PR for review very soon.

PR #3248, opened 2023-06-16
Provide toggle to enable or disable leader election in Helm chart

dabcoder:
### Issue

Fixes https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3243.

### Checklist
- [ ] Manually tested


helm/aws-load-balancer-controller/templates/deployment.yaml
```diff
@@ -149,6 +149,9 @@ spec:
         {{- if ne .Values.defaultTargetType "instance" }}
         - --default-target-type={{ .Values.defaultTargetType }}
         {{- end }}
+        {{- if kindIs "bool" .Values.enableLeaderElection }}
+        - --enable-leader-election={{ .Values.enableLeaderElection}}
```
kishorj, line 153: disabling leader election is not a good idea given the default replica count of 2

dabcoder: I see, that's a fair point. However, if you're providing a flag in the controller's code I would also expect to be able to use it in the Helm chart (the [Nginx Ingress controller ](https://docs.nginx.com/nginx-ingress-controller/installation/installation-with-helm/)has this option for instance). 
I can close this PR @kishorj, your call. If so, a comment in the related issue would be good too.

kishorj: @dabcoder, what is your use case for disabling leader election? One way is to make the replica count 1 when le is disabled in chart, but I feel it is unsafe it we introduce the possibility of multiple controllers working on the same set of resources.

dabcoder: > @dabcoder, what is your use case for disabling leader election? 

If or when one needs to only run 1 replica. @rblaine95 might want to add some additional info about this request.

rblaine95: Are there any risks associated with having replicas >1 with leader election disabled?

We did some testing in our development environment where everything was running on a single node and tested a catastrophic failure (spot interruption request on our single node) and we noticed that, if the current LB Controller leader received a termination request at the same time as an Nginx Ingress Controller, there would be a very brief moment where the secondary LB Controller wouldn't become the leader in time to deregister the terminating Nginx Ingress Controller which would cause _some_ traffic to be routed to a pod that's no longer available.

The thought process was that, if we disable leader election, we'd have an Active/Active setup and can avoid this kind of situation.

rblaine95: > Just to make sure I understand this use case too, are you running the latter controller as a back up for the former (the LB controller)?

Yes.

In our tests we ran (all on a single spot instance):
* 2 [Ingress Nginx Controller](https://github.com/kubernetes/ingress-nginx/) replicas
  * Pod Disruption Budget allowing for a single interruption
* 2 AWS LB Controller replicas
  * Pod Disruption Budget allowing for a single interruption
* An nginx deployment with `helm create nginx; helm install nginx -n default --set replicaCount=5 ./nginx`, Pod Disruption Budget allowing for a single interruption, and an ingress
* [AWS Node Termination Handler](https://github.com/aws/aws-node-termination-handler) to gracefully drain nodes on Spot Interruption Requests
* [Kubernetes Cluster Autoscaler](https://github.com/kubernetes/autoscaler) to automatically scale out a new node as quickly as possible

We then ran [bombardier](https://github.com/codesenberg/bombardier): `bombardier -c 30 -d 30m https://nginx.example.com` and triggered a Spot Interruption Request.

The result was that pod terminations occurred according to the configured Pod Disruption Budgets.
1 AWS LB Controller, 1 Nginx Controller, and 1 Nginx Deployment received termination requests.
The replacement node started up and the now Pending replicas scheduled successfully.
The Interrupted node was gracefully drained.
Exactly as expected.

The unexpected result was when, occasionally, the **AWS LB Controller Leader** was the one that got interrupted at the same time as one of the Ingress Nginx Controllers.
When this happened, the fail-over AWS LB Controller would not become the active leader in time to deregister the now no longer available Ingress Nginx Controller.
_Sidenote: We resolved some additional graceful termination of the Nginx Controllers by adding `sleep 15` and `SIGQUIT` pre-stop lifecycle hooks for properly graceful deregistration of Nginx Controllers._
We suspected that adding a `sleep 15` to a pre-stop lifecycle hook to the AWS LB Controller would work, but `sleep` and `sh` do not exist in the docker image 🤷 

It's a niche issue that only happens in a single node environment.
This is _**not**_ an issue in production (multi-node, correctly configured high-availability) environments.
We observed this in an unimportant dev/test environment where we were purposefully trying to break things to test HA and failure points.

We were only curious about risks associated with disabling leader election on the AWS LB Controller if `replicas >1`

PR #3261, opened 2023-06-28
feat: change pod mutating webhook to fail open

### LINKED ISSUE #2711: Allow configuration of the MutatingWebhook failure policy
**Describe the bug**
I ran into issues with TLS certs being regenerated due to these bugs: 

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2312
https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/2264

Once the TLS certs changed, the MutatingWebhook for PodReadinessGate started failing and blocking the rollout of pods on services using this feature.

This was the error:

```log
Error creating: Internal error occurred: failed calling webhook "mpod.elbv2.k8s.aws": Post "https://aws-lb-controller-webhook-service.kube-system.svc:443/mutate-v1-pod?timeout=10s": x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "aws-load-balancer-controller-ca")
```

I think this exposes an availability concern because if all the pods backing a service get rescheduled while the mutatingwebhook is broken, the service will go down. My understanding is the PodReadinessGate is a bonus feature to make rollouts more smooth in Kubernetes and I think it'd be preferable for the feature to just not work rather than block rollouts all together.

**Steps to reproduce**

Break TLS certs on the LB controller while using PodReadinessGates, then reschedule pods backing an LB in that namespace.

**Expected outcome**
I'd like to either be able to configure the webhooks failure policy or set it to fail open.

**Environment**

* AWS Load Balancer controller version: 2.4.1
* Kubernetes version: 1.21
* Using EKS (yes/no), if so version? Yes, platform version 7

**Additional Context**:

mikutas:
### Issue

closes #2711 

### Checklist
- [x] Manually tested

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -1,4 +1,3 @@
----
 apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
@@ -11,7 +10,7 @@ webhooks:
         name: webhook-service
         namespace: system
         path: /mutate-v1-pod
-    failurePolicy: Fail
+    failurePolicy: Ignore
     name: mpod.elbv2.k8s.aws
     rules:
       - apiGroups:
--- a/helm/aws-load-balancer-controller/templates/webhook.yaml
+++ b/helm/aws-load-balancer-controller/templates/webhook.yaml
@@ -19,7 +19,7 @@ webhooks:
       name: {{ template "aws-load-balancer-controller.webhookService" . }}
       namespace: {{ $.Release.Namespace }}
       path: /mutate-v1-pod
-  failurePolicy: Fail
+  failurePolicy: Ignore
   name: mpod.elbv2.k8s.aws
   admissionReviewVersions:
   - v1beta1
--- a/webhooks/core/pod_mutator.go
+++ b/webhooks/core/pod_mutator.go
@@ -44,7 +44,7 @@ func (m *podMutator) MutateUpdate(ctx context.Context, obj runtime.Object, oldOb
 	return obj, nil
 }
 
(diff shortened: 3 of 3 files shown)


johngmyers (changes requested):
The linked issue #2711 is specific to the PodReadinessGate mutating webhook for Pods. It makes sense to have that webhook fail open, as failing closed can easily cause persistent lack of availability and failing open can only cause an intermittent lack of availability.

This PR, however, creates a single configuration option for the failure policy of all webhooks, not just the PodReadinessGate. Other webhooks, especially the validating ones, are much more critical. So that's a NACK from me.

The Helm chart should just configure the PodReadinessGate mutating webhook to fail open. There is  no need for it to be configurable.

johngmyers:
Why does this have to be configurable? The chart should just configure `failurePolicy: Ignore`.

johngmyers:
@mikutas please remove the second commit.

johngmyers:
@mikutas please update the comment at https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/webhooks/core/pod_mutator.go#L46 and run `make crds`.

/lgtm cancel

kawai-kobold:
@johngmyers It looks like the approval label wasn't applied, can you check? We'd love to have this fix in our environment, the webhook failures make most of our GitOps deploys into EKS a real pain.

seanturner026:
I am truly sorry to be a pest here, but, this is a small PR and was approved previously per the below comment. Is there anything needed to get this out? This issue just caused an outage

https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3261#issuecomment-1667112548

@johngmyers @kishorj @M00nF1sh

seanturner026:
Apologies, looks like this is already implemented https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3653

mikutas:
This issue is about pod mutating webhook
https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3653 is about service mutating webhook

mikutas:
implemented in newer PR #4062 😕

Issue #3265, opened 2023-06-30
Disable backend SG rule management


nakamume:
**Is your feature request related to a problem?**
We'd like to manage all backend security-group(SG) rules manually and outside of the controller. This would allow -
- To limit the number of rules that are created for backends. This has already been greatly improved with the introduction of `EnableBackendSecurityGroup` for ingresses but services still create a lot of rules. If the users manually create the rules, they can further optimize it.
- Platform teams to control which ports should be opened. Currently product teams can use whatever ports they want and the controller creates the corresponding rules. If the rules are managed outside of controller (manually) - platform teams would have better controls on what ports should be opened.
- We (personally) manage our rules using terraform and general allowed some ports from sources within our VPC. So, in almost all cases, the rules added by the controller are redundant for us.

**Describe the solution you'd like**
- A controller flag to enable/disable backend SG rule management.
- At least for ingresses, a param can be introduced in `IngressClassParam`. But not sure if something like this exists for services.

**Describe alternatives you've considered**
Both ingress and service support annotations to disable the backend SG rule management, but that's repetitive and doesn't provide platform teams to enforce the restriction on controller level.

johngmyers:
Platform teams can control ports with a policy-enforcing admission controller such as Gatekeeper.

PR #3266, opened 2023-06-30
Controller cfg for backend sg rule management

nakamume:
### Issue
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3265

### Description
This PR introduces a controller config flag called `ManageBackendSGRules`. If the config is true (default) - controller behaves as before. If the config is `false` - controller wont manage the backend SG rules for both ingresses and services. Rules would still be created for controller managed frontend SGs for ingresses.

### Checklist
- [x] Manually tested

### DIFF
--- a/controllers/ingress/group_controller.go
+++ b/controllers/ingress/group_controller.go
@@ -61,7 +61,7 @@ func NewGroupReconciler(cloud aws.Cloud, k8sClient client.Client, eventRecorder 
 		authConfigBuilder, enhancedBackendBuilder, trackingProvider, elbv2TaggingManager, controllerConfig.FeatureGates,
 		cloud.VpcID(), controllerConfig.ClusterName, controllerConfig.DefaultTags, controllerConfig.ExternalManagedTags,
 		controllerConfig.DefaultSSLPolicy, controllerConfig.DefaultTargetType, backendSGProvider, sgResolver,
-		controllerConfig.EnableBackendSecurityGroup, controllerConfig.EnableBackendSGRuleManagement, controllerConfig.DisableRestrictedSGRules,
+		controllerConfig.EnableBackendSecurityGroup, controllerConfig.ManageBackendSGRules, controllerConfig.DisableRestrictedSGRules,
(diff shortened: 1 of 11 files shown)


nakamume:
@kishorj @M00nF1sh can I get some love ❤️ here?

johngmyers:
This looks like it interacts with #3277. It's possible that #2277 addresses this.

The interaction between `--enable-backend-security-group` and `--manage-backend-sg-rules` is unclear and confusing. The documentation is not at all clear.

nakamume:
> This looks like it interacts with https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3277. It's possible that https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/2277 addresses this.

Not sure how #2277 addresses this? Maybe wrong PR link? But #3277 seems to be going into the same direction except that it only covers the SG rules for ALBs(ingresses) and not NLBs (LB services).

> The interaction between --enable-backend-security-group and --manage-backend-sg-rules is unclear and confusing.

Yes, I think it's a "bit". First is about weather a dedicated Security Group be created/used for backends and second is about rules for backend communication.

johngmyers:
Why would one create a managed security group if one isn't going to create rules? Which of the combinatorics make sense for which use cases?

nakamume:
I touched upon it in the issue desc https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3265. Note that the feature `BackendSGManagement` is only for ingresses and doesn't cover the LB services.
With `BackendSGManagement` users can create (manually) or let controller create for them a single dedicated backend SG, this SG would be shared by all the ingresses in that cluster (ingresses managed by the LBC).
Then, they can either let the LBC control the rules for backends <- current/default behavior (unless users put an annotation not to do so) or manage the rules themselves <- disable rule management by the controller (feature added by this PR).

Even with `manageBackendSG: false` - controller creates the backend rules. `BackendSG` is just a single dedicated SG shared by all ingresses, without it controller creates a backend SG per ingress group.

PR #3272, opened 2023-07-03
Make Ingress validating webhook ignore ingresses not managed by AWS LBC

johngmyers:
### Issue

None

### Description

Makes the validating admission webhook for Ingress permit creation and modification all ingresses for which it is able to determine that it is not going to manage.

Also fixes a bug where LBC would incorrectly manage Ingresses with an IngressClassName specifying a different controller when LBC was configured to manage Ingresses without either annotation or IngressClassName.

### Checklist
- [ ] Manually tested

### DIFF
--- a/webhooks/networking/ingress_validator.go
+++ b/webhooks/networking/ingress_validator.go
@@ -25,12 +25,13 @@ const (
 // NewIngressValidator returns a validator for Ingress API.
 func NewIngressValidator(client client.Client, ingConfig config.IngressConfig, logger logr.Logger) *ingressValidator {
 	return &ingressValidator{
-		annotationParser:              annotations.NewSuffixAnnotationParser(annotations.AnnotationPrefixIngress),
-		classAnnotationMatcher:        ingress.NewDefaultClassAnnotationMatcher(ingConfig.IngressClass),
-		classLoader:                   ingress.NewDefaultClassLoader(client),
-		disableIngressClassAnnotation: ingConfig.DisableIngressClassAnnotation,
-		disableIngressGroupAnnotation: ingConfig.DisableIngressGroupNameAnnotation,
-		logger:                        logger,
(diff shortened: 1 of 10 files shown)


johngmyers:
@kishorj @M00nF1sh can I please get a review?  This has been open a month and I have other PRs stacked behind it.

webhooks/networking/ingress_validator.go
```diff
@@ -88,6 +98,21 @@ func (v *ingressValidator) ValidateDelete(ctx context.Context, obj runtime.Objec
 	return nil
 }
 
+// checkIngressClass checks to see if this ingress is handled by this controller.
+func (v *ingressValidator) checkIngressClass(ctx context.Context, ing *networking.Ingress) (bool, error) {
```
kishorj, line 102: nit: the `checkIngressClass` name is confusing. Since the purpose here is to check if the webhook should skip validating further, we could rename to either of `checkIngressSkipValidation`, `shouldSkipFurtherValidation`, `checkIngressUnmanaged`.

Changes look good otherwise.

johngmyers: This function is validating the `kubernetes.io/ingress.class` annotation, the `IngressClassName` field, and the `spec.controller` of the `IngressClass` referenced by the `IngressClassName` field. Skipping further validation is only one possible outcome; it can fail validation if the `IngressClassName` references a nonexistent `IngressClass`.

For naming these sub-validation methods, I tend to prefer using the name of the field or subsection that is being validated.  The existing naming pattern isn't entirely consistent; this could be fixed in follow-up.

It looks like `checkIngressClassUsage()` is being made redundant; it could be removed in this or the followup PR.

kishorj: lets refactor in a follow up PR

kishorj (approved):

PR #3275, opened 2023-07-06
docs update for using annotations

jaydeokar:
### Issue
N/A

### Description
Added doc update about the use of annotations and expected behavior of the controller

### Checklist
- [ ] Manually tested

### DIFF
--- a/docs/guide/ingress/annotations.md
+++ b/docs/guide/ingress/annotations.md
@@ -12,6 +12,7 @@ You can add annotations to kubernetes Ingress and Service objects to customize t
     - Annotations that configures LoadBalancer / Listener behaviors have different merge behavior when IngressGroup feature is been used. `MergeBehavior` column below indicates how such annotation will be merged.
         - Exclusive: such annotation should only be specified on a single Ingress within IngressGroup or specified with same value across all Ingresses within IngressGroup.
         - Merge: such annotation can be specified on all Ingresses within IngressGroup, and will be merged together.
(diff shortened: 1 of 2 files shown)


johngmyers:
I cannot figure out what this update is trying to say.

oliviassss:
@johngmyers, we got some internal tickets regarding the behavior that the controller reverts some manual edit on ingress/services/other resources during reconciliation, which is an expected behavior because the controller reconciles based on the manifest. So just to call it out in our doc to use annotation as best practice.

johngmyers:
For one, annotations are not the best practice. For annotations which have a corresponding field in `IngressClassParams`, the field is best practice. For the `group.name` annotation, `IngressClass` is best practice. For the `target-type` annotation, the `--default-target-type` flag is best practice.

The sentence as written is not very comprehendable. What you're trying to say is that the controller manages the configuration of the LBs that it creates and if one wants to change the configuration of the LB one needs to do that through the configuration given to LBC.

oliviassss:
@johngmyers, thanks for the feedback, maybe we add a high-level note in `aws-load-balancer-controller/docs/deploy/configurations.md` - 
The controller manages the configurations of the resources it creates, we do not recommend out-of-band modifications to these resources because the controller may revert the manual changes during reconciliation. We recommend to use configuration options provided as best practice, such as ingress and service annotations, controller command line flags and IngressClassParams.

johngmyers:
That is much better wording, thanks. Consider `how-it-works.md` as the location; `configurations.md` is more about the command-line flags.

PR #3277, opened 2023-07-09
Add securityGroups field to IngressClassParams

johngmyers:
### Issue

#2920

Competes with #2919.

### Description

Adds a `securityGroups` field to `IngressClassParams`, to allow overriding the `alb.ingress.kubernetes.io/security-groups` and `alb.ingress.kubernetes.io/manage-backend-security-group-rules` annotations for all Ingresses in an IngressClass.

### Checklist
- [ ] Manually tested


johngmyers:
I'm not entirely sure of the naming of the `managedInbound` and `managedBackend` fields. Suggestions?

config/crd/bases/elbv2.k8s.aws_ingressclassparams.yaml
```diff
@@ -146,6 +148,43 @@ spec:
...
+              securityGroups:
+                description: SecurityGroups defines the security groups to attach
+                  to all Ingresses that belong to IngressClass with this IngressClassParams.
+                properties:
+                  ids:
+                    description: IDs specify the resource IDs of security groups.
+                      Exactly one of this, `managedInbound`, or `tags` must be specified.
+                    items:
+                      description: SecurityGroupID specifies a security group ID.
+                      pattern: sg-[0-9a-f]+
```
oliviassss, line 160: Can it also accept SG names just to keep consistent with the `alb.ingress.kubernetes.io/security-groups` annotation

pkg/ingress/model_build_load_balancer.go
```diff
@@ -269,78 +269,120 @@ func (t *defaultModelBuildTask) buildLoadBalancerSubnetMappings(ctx context.Cont
...
-		}
-		t.logger.Info("SG configured via annotation", "LB SGs", lbSGTokens, "backend SG", t.backendSGIDToken)
-	}
-	return lbSGTokens, nil
-}
+			for _, sgID := range frontendSGIDs {
+				lbSGTokens = append(lbSGTokens, core.LiteralStringToken(sgID))
+			}
 
-func (t *defaultModelBuildTask) buildFrontendSGNameOrIDsFromAnnotation(ctx context.Context) ([]string, error) {
```
oliviassss, line 325: I'm just thinking would it be more maintainable and readable if we have separate functions like `buildFrontendSGNameOrIDsFromAnnotation()` and `buildFrontendSGNameOrIDsFromSelector()` and call them in the `buildLoadBalancerSecurityGroups()`?

pkg/ingress/model_build_load_balancer_test.go
```diff
@@ -682,6 +683,924 @@ func Test_defaultModelBuildTask_buildLoadBalancerName(t *testing.T) {
...
+									Namespace: "awesome-ns",
+									Name:      "ing-1",
+								},
+							},
+						},
+					},
+				},
+				scheme: elbv2.LoadBalancerSchemeInternetFacing,
+			},
+			want: []string{"sg-managed"},
```
oliviassss, line 950: Should the default value of `--enable-backend-security-group` is true? The controller should always create both frontend and backend SGs to ALB if there's no annotation or spec.SecurityGroups specified?

pkg/networking/security_group_resolver.go
```diff
@@ -102,3 +109,102 @@ func (r *defaultSecurityGroupResolver) splitIntoSgNameAndIDs(sgNameOrIDs []strin
...
+		if tagKey == clusterResourceTagKey {
+			return true
+		}
+		if strings.HasPrefix(tagKey, clusterResourceTagPrefix) {
+			// If the cluster tag is for a different cluster, keep track of it and exclude
+			// the security group if no matching tag found for the current cluster.
+			hasClusterResourceTagPrefix = true
+		}
+	}
+	if hasClusterResourceTagPrefix {
```
oliviassss, line 206: Maybe it's better to add a log here just to indicate we skip some SG because it's tagged for another cluster `cluster-xxx`

webhooks/elbv2/ingressclassparams_validator.go
```diff
@@ -89,6 +92,63 @@ func validateCIDR(cidr string, fieldPath *field.Path) field.ErrorList {
...
+		if securityGroups.IDs != nil {
+			if len(icp.Spec.InboundCIDRs) > 0 {
+				allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "inboundCIDRs"), "May not have both `inboundCIDRs` and `securityGroups.ids`"))
+			}
+			if securityGroups.ManagedInbound {
+				allErrs = append(allErrs, field.Forbidden(fieldPath.Child("managedInbound"), "may not have both `ids` and `managedInbound` set"))
+				return allErrs
+			}
+			if securityGroups.Tags != nil {
+				allErrs = append(allErrs, field.Forbidden(fieldPath.Child("tags"), "may not have both `ids` and `tags` set"))
```
oliviassss, line 113: I'm just wondering why would the `ids` and `tags` be exclusive? Let's say if a user specifies both (for fine-grained filtering purpose, or just by accident), can we just filter the SG by both field instead of error out?

config/crd/bases/elbv2.k8s.aws_ingressclassparams.yaml
johngmyers: No, the field should be strongly typed. If you want to search by the `Name` tag, use the `tags` field using a tag key of `Name`

webhooks/elbv2/ingressclassparams_validator.go
johngmyers: What is the use case for specifying both ids and tags? An admin would typically have one way of communicating security groups from the IaC provisioning system to the deploy-to-Kubernetes system. Either they'd propagate a list of IDs or they'd provision the groups with tags according to some convention. Why would someone use one mechanism for some groups and another mechanism for others?

pkg/ingress/model_build_load_balancer.go
johngmyers: Extracted two receivers.

pkg/networking/security_group_resolver.go
johngmyers: I incorporated code corresponding to #3292

apis/elbv2/v1beta1/ingressclassparams_types.go
```diff
@@ -100,9 +133,14 @@ type IngressClassParamsSpec struct {
 	Scheme *LoadBalancerScheme `json:"scheme,omitempty"`
 
 	// InboundCIDRs specifies the CIDRs that are allowed to access the Ingresses that belong to IngressClass with this IngressClassParams.
+	// If this and `securityGroups` are both specified, `securityGroups.managed` must be `true`.
```
oliviassss, line 136: nit: should it be `securityGroups.ManageInbound` must be `true`? Can we also call this out in ingress_class.md and also helm chart definition?

webhooks/elbv2/ingressclassparams_validator.go
Aleksei-Poliakov: I don't see much point in having both. If you have id, there isn't much value in having the name (e.g. for the sake of documentation one can always put this information in comments or in some non-actionable annotation). Allowing both means you either have to make one take precedence when these don't agree; or still fail at runtime - but this is way more complicated to explain what's wrong (e.g. the message would have to read something like "sg-123 has name 'name-1', but the annotation specified 'name-2' - you have to either only use name/id; or if both are used - they should point to the same entity" - but having such a reasonable message is somewhat a security concern because you can now map security-group-id to a name which might have sensitive info)

At the same time clearly having a behavior to fail when both are specified allows using things like admission controllers to explicitly fail when resource are provisioned, rather than at runtime, so you can't even deploy incorrectly configured ingress.

PR #3279, opened 2023-07-11
Update subnetResolver to better support dualstack LBs

nakamume:
### Description

Current subnet auto-discovery logic can return the IPv4 only subnets for dualstack LBs. LB would use these subnets but then would fail to provision. This PR improves the subnet-resolver for dualstack LBs by 
- discovering the subnets which have both IPv4+IPv6 CIDRs.
- validating the discovered (when explicitly given or for existing LBs) subnets whether they can support dualstack LBs.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ingress/model_build_load_balancer.go
+++ b/pkg/ingress/model_build_load_balancer.go
@@ -48,7 +48,7 @@ func (t *defaultModelBuildTask) buildLoadBalancerSpec(ctx context.Context, liste
 	if err != nil {
 		return elbv2model.LoadBalancerSpec{}, err
 	}
-	subnetMappings, err := t.buildLoadBalancerSubnetMappings(ctx, scheme)
+	subnetMappings, err := t.buildLoadBalancerSubnetMappings(ctx, scheme, ipAddressType)
 	if err != nil {
 		return elbv2model.LoadBalancerSpec{}, err
 	}
@@ -186,7 +186,7 @@ func (t *defaultModelBuildTask) buildLoadBalancerIPAddressType(_ context.Context
 	}
 }
 
-func (t *defaultModelBuildTask) buildLoadBalancerSubnetMappings(ctx context.Context, scheme elbv2model.LoadBalancerScheme) ([]elbv2model.SubnetMapping, error) {
+func (t *defaultModelBuildTask) buildLoadBalancerSubnetMappings(ctx context.Context, scheme elbv2model.LoadBalancerScheme, ipAddressType elbv2model.IPAddressType) ([]elbv2model.SubnetMapping, error) {
 	var explicitSubnetSelectorList []*v1beta1.SubnetSelector
 	var explicitSubnetNameOrIDsList [][]string
 	for _, member := range t.ingGroup.Members {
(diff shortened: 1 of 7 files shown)


johngmyers:
Where are the documentation changes?

Why don't you just tag the subnets appropriately? Why not provision IPv4-only loadbalancers out of dual-stack subnets?

nakamume:
> Where are the documentation changes?

I don't think this warrants any doc changes. It doesn't bring in any "new" behavior - it just corrects an existing behavior. But happy to add, if you have something specific in mind.


> Why don't you just tag the subnets appropriately? 

As in to avoid LBC from picking the wrong subnets? Not sure how that would work? If I'm not mistaken, by this logic we can ask everyone to pass the subnet IDs appropriately and remove the auto-discovery logic altogether.

I plan to use only the "discovery" tags on the subnets and then let the LBC discover them for my end users. If they plan to provision a v4 LB, any subnet out of the ones with discovery tags could be used. But for dualstack LBs - only the subnets with both v4 & v6 must be used.


> Why not provision IPv4-only loadbalancers out of dual-stack subnets?

We are going through a migration phase, where lots of subnets and existing clusters still use pure IPv4 while the newer ones would be dualstack. I think this could be achieved if we make use of the `cluster-tag` check but we don't plan to use that for auto-discovery (as we have lots of clusters and we'd hit AWS tag limit if we start tagging subnets with those).

johngmyers:
> I don't think this warrants any doc changes. It doesn't bring in any "new" behavior - it just corrects an existing behavior. But happy to add, if you have something specific in mind.

It is new behavior: it changes the subnet selection behavior to no longer match what is described in the documentation.

I don't see how this "correcting" the existing behavior.

> As in to avoid LBC from picking the wrong subnets? Not sure how that would work?

Put the tags on the dualstack subnets and not the IPv4-only subnets?

> If I'm not mistaken, by this logic we can ask everyone to pass the subnet IDs appropriately and remove the auto-discovery logic altogether.

I'm not following the steps to this conclusion.

If one really has a need to allocate the IPv4-only load balancers out of IPv4-only subnets, one could use selectors with a different set of tags.

> I plan to use only the "discovery" tags on the subnets and then let the LBC discover them for my end users. If they plan to provision a v4 LB, any subnet out of the ones with discovery tags could be used. But for dualstack LBs - only the subnets with both v4 & v6 must be used.

Why put the discovery tags on the IPv4-only subnets?

johngmyers:
Can you not make your existing tagged subnets be dual-stack?

This change adds complexity, increasing the risk of a bug and increasing the maintenance cost.

johngmyers:
Another suggestion would be to use the `spec.subnets.tags` field of `IngressClassParams` to specify a different tag to use to discover subnets. You'd want to use `spec.ipAddressType` to specify the address type in any case.

nakamume:
We were really hoping for not making any changes to existing subnets. IngressClassParams can definitely be used in our case and that's what we've been working on now. So, this PR isn't a blocker for us.
It's just that the current implementation of subnet auto-discovery is buggy - it could discover wrong (IPv4 only) subnets for dualstack LBs. And it'd be great if it gets corrected.

johngmyers:
My position is that the subnets are configured and/or tagged incorrectly, not that the resolver is buggy. The resolver is behaving exactly as documented and intended.

I really don't see a use case, outside of testing, for an IPv4-only subnet in a dual-stack VPC.

Issue #3264, opened 2023-06-29
Controller runtime flag sync-period is not working as expected


oliviassss:
**Describe the bug**
The controller runtime config flag `--sync-period` is not working as expected, the controller is not reconciling every 1hr as specified by default. Could be a bug.

**Steps to reproduce**

**Expected outcome**
A concise description of what you expected to happen.

**Environment**

* AWS Load Balancer controller version
* Kubernetes version
* Using EKS (yes/no), if so version?

**Additional Context**:

oliviassss:
The `manager.Options.SyncPeriod` has been deprecated in favor of `manager.Options.Cache.SyncPeriod` from upstream controller-runtime since v1.14.5, we probably need to modify the controller runtime flag for AWS LBC accordingly.
ref:
https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/manager/manager.go#L169C13-L169C13
https://github.com/kubernetes-sigs/controller-runtime/releases

oliviassss:
related issues:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2800
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2515

oliviassss:
The root cause of the issue with the controller not reconciling at the specified interval per `--sync-period` is due to the eventhandler ignoring the update event. The event handler currently checks for differences in annotation and spec between the old and new ingresses. If there is no difference, the update event is ignored to prevent unnecessary reconciliations that would result in AWS API calls. Therefore, reconciliation is not actually triggered in this situation. 

We can utilize the Kubernetes resourceVersion as a fix instead. By checking if the resourceVersion is the same, we can skip the check for differences in annotation and spec between the old and new ingresses. We can incorporate this fix in our upcoming patch release.

Ref:

    https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/controllers/ingress/eventhandlers/ingress_events.go#L43
    https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/deploy/wafv2/web_acl_association_manager.go#L13

oliviassss:
resolve as the fix has been shipped.

Issue #2800, opened 2022-09-14
Self Healing: Allow AWS Controller to Detect and Fix AWS Resource Changes on Interval


lucas-howard-macmillan:
**Is your feature request related to a problem?**

If a load balancer is deleted through the AWS Console, the AWS load balancer does not notice or re-create the load balancer.

The AWS load balancer controller must be restarted, and then the missing load balancer is recreated.

**Describe the solution you'd like**

An argument that could be passed into the controller indicating that it should do a full scan of AWS on a certain interval in an attempt to detect and fix drift within AWS from the expected state.

This would basically emulate the behavior that the AWS load balancer controller does when it starts up.

Potentially, for large deployments, you might also need a segment size argument as well
i.e.

Every 5 minutes scan AWS for 100 ingresses, then the next 5 minutes the next 100 ingresses etc..

**Describe alternatives you've considered**

I have used all existing arguments, such as sync period, but none of them cause the load balancer to be re-created.

kishorj:
/assign @M00nF1sh 
Investigate further on the periodic sync issue
This is similar to https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2515

lukonjun:
Experienced the same behaviour. I assumed when I delete the LB via the AWS Console the ALB Controller would automatically recreate, however it did not.

0xF4D3C0D3:
I'm experiencing the same issue. It only gets recovered when the number of replicas behind the service is changed.

ChrisV78:
Experienced the same by accident, removed a ALB from the AWS console and the lb-controller only recreates the ALB the moment you restart the lb-controller deployment.

lucas-howard-macmillan:
@oliviassss Thank You!

Issue #2515, opened 2022-02-17
Autorecover aws-load-balancer-controller managed SG inbound rules


stevo-f3:
**Is your feature request related to a problem?**
aws-load-balancer-controller managed security group inbound rules for LB Service `loadBalancerSourceRanges` definition are not being recovered, unless there is a change to `loadBalancerSourceRanges`. Currently (aws-load-balancer-controller v2.3.1, chart v1.3.3) accidental/unwanted deletion of controller managed SG rule requires detecting rule is missing + changing `loadBalancerSourceRanges` (back and forth) to trigger controller to reconcile/recover the rule.

**Describe the solution you'd like**
I expect aws-load-balancer-controller to reconcile (compare desired and actual) and recover all actual resources it manages (including inbound SG rules) even when there's no change event on declared desired state (Service definition, loadBalancerSourceRanges).

**Describe alternatives you've considered**
No effective alternatives come to mind

kishorj:
@stevo-f3, when you make changes to the ingress/service, controller will restore the changes. We will investigate further why the sync-period setting did not work in this case.

oliviassss:
Hi, we have shipped the fix in v2.5.4, please check the details in our release note: https://github.com/kubernetes-sigs/aws-load-balancer-controller/releases/tag/v2.5.4. 
I'm closing this ticket as of now, please feel free to reach out or reopen if you have any issues. Thanks

PR #3280, opened 2023-07-12
fix the bug that evenhanlder ignores the update per sync-period

oliviassss:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3264
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2800
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2515

### Description

Currently, there is a conflict between the controller runtime flag `--sync-period`, which is to force reconcile every fix interval, and the eventhandlers, which are to avoid unnecessary reconciles to optimize the AWS API usage. When the controller runtime sends an update event to the eventhandler per interval in `--sync-period`, the eventhandler will ignore the update event if there is no difference in the annotations/specs of the ingress or service. Therefore, if the end users made some manual modification to the resources managed by the controller as mentioned in the issues above, the controller could not revert since it does not reconcile under this situation.

We fix the bug by leveraging k8s resourceVersions -
- if the resourceVersions are the same between old and new objects, the controller will continue to reconcile; 
- if the resourceVersions do change, then the eventhandler continues to check the annotation and specs to see if it is an update event with real change in resources.

However, as the best practice, we do not recommend manual modification, nor should the users depend on the controller auto-reconciliation to heal the resources they changed, or to mitigate any security risks. Since no matter how frequent the controller reconciles, there should always be a gap.

### Checklist
- [ ] Manually tested

### DIFF
--- a/controllers/ingress/eventhandlers/ingress_events.go
+++ b/controllers/ingress/eventhandlers/ingress_events.go
@@ -44,10 +44,12 @@ func (h *enqueueRequestsForIngressEvent) Update(e event.UpdateEvent, queue workq
 	//	1. Ingress annotation updates
 	//	2. Ingress spec updates
 	//	3. Ingress deletion
-	if equality.Semantic.DeepEqual(ingOld.Annotations, ingNew.Annotations) &&
-		equality.Semantic.DeepEqual(ingOld.Spec, ingNew.Spec) &&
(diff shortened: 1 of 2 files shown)


M00nF1sh (approved):
/lgtm
/approve

PR #3281, opened 2023-07-12
doc enhancement for waf addons and reconciliation

oliviassss:
### Description

Doc improvement for the controller's exact behavior regarding reconciliation and waf addons.
following up this [PR](https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3280). 
Checked the rendering by `make docs-preview`

### Checklist
- [ ] Manually tested

### DIFF
--- a/docs/deploy/configurations.md
+++ b/docs/deploy/configurations.md
@@ -98,7 +98,7 @@ Currently, you can set only 1 namespace to watch in this flag. See [this Kuberne
 |log-level                              | string                          | info            | Set the controller log level - info, debug |
 |metrics-bind-addr                      | string                          | :8080           | The address the metric endpoint binds to |
 |service-max-concurrent-reconciles      | int                             | 3               | Maximum number of concurrently running reconcile loops for service |
-|[sync-period](#sync-period)                            | duration                        | 1h0m0s          | Period at which the controller forces the repopulation of its local object stores|
+|[sync-period](#sync-period)                            | duration                        | 10h0m0s          | Period at which the controller forces the repopulation of its local object stores|
 |targetgroupbinding-max-concurrent-reconciles | int                       | 3               | Maximum number of concurrently running reconcile loops for targetGroupBinding |
 |targetgroupbinding-max-exponential-backoff-delay | duration              | 16m40s          | Maximum duration of exponential backoff for targetGroupBinding reconcile failures |
 |watch-namespace                        | string                          |                 | Namespace the controller watches for updates to Kubernetes objects, If empty, all namespaces are watched. |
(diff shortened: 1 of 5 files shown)


docs/deploy/configurations.md
```diff
@@ -127,6 +127,16 @@ Once disabled:
 * you can no longer create Ingresses with the `alb.ingress.kubernetes.io/group.name` annotation.
 * you can no longer alter the value of an `alb.ingress.kubernetes.io/group.name` annotation on an existing Ingress.
 
+### sync-period
+`--sync-period` defines a fix interval for the controller to auto reconcile, default to 1hr. Please be mindful that frequent reconciliations may incur unnecessary AWS API usage.
```
M00nF1sh, line 131: s/fix/fixed

docs/deploy/configurations.md
```diff
@@ -127,6 +127,16 @@ Once disabled:
...
+
+As best practice, we do not recommend users to manually modify the resources managed by the controller. And users should not depend on the controller auto-reconciliation to revert the manual modification, or to mitigate any security risks. 
+Since no matter how frequent the controller reconciles, there should always be a gap.
```
M00nF1sh, line 134: remove this line `Since no matter how frequent the controller reconciles, there should always be a gap.`

docs/deploy/configurations.md
```diff
@@ -127,6 +127,16 @@ Once disabled:
...
+
+### waf-addons
+By default, the controller assumes solo ownership of the WAF addons associated to the provisioned ALBs, via the flag `--enable-waf` and `--enable-wafv2`.
```
M00nF1sh, line 137: s/solo/sole

docs/deploy/configurations.md
M00nF1sh, line 131: let's also change the default sync-period to 10 hour instead 1 hour, which matches controller-runtime's default value: https://github.com/kubernetes-sigs/controller-runtime/blob/56a973c46efe463ef7b5fbecd8495e8378b50d20/pkg/cache/cache.go#L43

need to update documentation here as well: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/helm/aws-load-balancer-controller/values.yaml#L218

docs/deploy/configurations.md
```diff
@@ -98,7 +98,7 @@ Currently, you can set only 1 namespace to watch in this flag. See [this Kuberne
 |log-level                              | string                          | info            | Set the controller log level - info, debug |
 |metrics-bind-addr                      | string                          | :8080           | The address the metric endpoint binds to |
 |service-max-concurrent-reconciles      | int                             | 3               | Maximum number of concurrently running reconcile loops for service |
-|sync-period                            | duration                        | 1h0m0s          | Period at which the controller forces the repopulation of its local object stores|
+|[sync-period](#sync-period)                            | duration                        | 1h0m0s          | Period at which the controller forces the repopulation of its local object stores|
```
M00nF1sh, line 101: should be 10h0m0s

johngmyers:
A 10-hour resync period means that if someone does something to cause drift, the time the controller reverts it is almost guaranteed to be outside of office hours. This is much more dangerous.

docs/deploy/configurations.md
oliviassss: done.

M00nF1sh:
> A 10-hour resync period means that if someone does something to cause drift, the time the controller reverts it is almost guaranteed to be outside of office hours. This is much more dangerous.

Hi John, since this sync-period feature never worked before, it's not a regression no matter how large this sync-period value is.   

The reason we want this to be a large value is due to we don't want to suddenly introduce a large amount of AWS APIs in a patch version.
Also, users shouldn't rely on this sync-period feature to override their custom made settings at all. IIRC, the sync-period exists in early days of Kubernetes when the informer mechanism wasn't reliable. Newer controllers in Kubernetes eco all have a large sync-period settings.

M00nF1sh (approved):
/lgtm
/approve

johngmyers:
I would have gone for something like 4-6 hours. It might push the revert out of business hours, but unlikely to go into sleep time.

Assuming things that cause the controller to do a reconciliation are more likely to happen during business hours, a 10-hour sync period s increases the probability of the reversion happening during sleep time. But I suppose a few stars would have to align.

PR #3290, opened 2023-07-21
[WIP] Skip processing an ingress that does not have a valid certificate for its hosts

seanseth7:
… 
### Description

If a single ALB ingress in an ingress group includes a host that doesn't have a corresponding certificate in ACM, it will block successful deployment of the entire ingress group. Instead of this behavior, we can simply ignore the "bad ingress" and include log a message explaining that the ingress doesn't have a host that corresponds to a certificate. 

With the recent introduction of the fix for the syncPeriod parameter, I think it's safe to make this change because the load balancer controller will eventually try to perform certificate discovery and attachment after at most 10 hours (the default period), so if the "bad ingress" becomes an ingress that has a valid certificate for its hosts, it would eventually get processed successfully. Additionally, users can make the syncPeriod more frequent if needed.

### Checklist
- [ ] Manually tested
  - Integration testing
    - Test Case 1: Multiple ingresses in one group
      - ALB Ingress with no valid cert for host
        - image
      - ALB ingress with valid cert for host
        - image
      - Bad ingress is ignored and good ingress is successfully processed and model is deployed
        - image
    - Test Case 2: Single ingress in a group
      - ALB Ingress with no valid cert for host
        - 

### DIFF
--- a/pkg/ingress/cert_discovery.go
+++ b/pkg/ingress/cert_discovery.go
@@ -2,18 +2,18 @@ package ingress
 
 import (
 	"context"
+	"strings"
+	"sync"
+	"time"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/acm"
 	"github.com/go-logr/logr"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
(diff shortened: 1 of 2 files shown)


seanseth7:
I'll be making this PR from my organization's fork instead.

PR #3295, opened 2023-07-26
Skip processing an ingress that does not have a valid certificate for its hosts

seanseth7:
### Issue

N/A (I don't see any related issues, but feel free to let me know and I'll link them here)

### Description

If a single ALB ingress in an ingress group includes a host that doesn't have a corresponding certificate in ACM, it will block successful deployment of the entire ingress group. Instead of this behavior, we can simply ignore the "bad ingress" and include log a message explaining that the ingress doesn't have a host that corresponds to a certificate. 

With the recent introduction of the fix for the syncPeriod parameter, I think it's safe to make this change because the load balancer controller will eventually try to perform certificate discovery and attachment after at most 10 hours (the default period), so if the "bad ingress" becomes an ingress that has a valid certificate for its hosts, it would eventually get processed successfully. Additionally, users can make the syncPeriod more frequent if needed.

### Checklist
- [x] Manually tested
  - Integration testing
    - Test Case 1: Multiple ingresses in one group
      - ALB Ingress with no valid cert for host
        - image
      - ALB ingress with valid cert for host
        - image
      - Bad ingress is ignored and good ingress is successfully processed and model is deployed
        - image
    - Test Case 2: Single ingress in a group
      - ALB Ingress with no valid cert for host
        - image
      - Load balancer is constructed/model deployed without invalid port config
        - image

### DIFF
--- a/pkg/ingress/cert_discovery.go
+++ b/pkg/ingress/cert_discovery.go
@@ -2,18 +2,18 @@ package ingress
 
 import (
 	"context"
+	"strings"
+	"sync"
+	"time"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/acm"
 	"github.com/go-logr/logr"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
-	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/util/cache"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/aws/services"
-	"strings"
-	"sync"
-	"time"
 )
 
 const (
@@ -81,7 +81,8 @@ func (d *acmCertDiscovery) Discover(ctx context.Context, tlsHosts []string) ([]s
 		}
 
 		if len(certARNsForHost) == 0 {
-			return nil, errors.Errorf("no certificate found for host: %s", host)
+			d.logger.Info("no certificate found for host", "host", host)
+			continue
 		}
 		certARNs.Insert(certARNsForHost...)
 	}
--- a/pkg/ingress/model_build_listener.go
+++ b/pkg/ingress/model_build_listener.go
@@ -140,6 +140,10 @@ func (t *defaultModelBuildTask) computeIngressListenPortConfigByPort(ctx context
 			inboundCIDRv6s: inboundCIDRV6s,
 		}
 		if protocol == elbv2model.ProtocolHTTPS {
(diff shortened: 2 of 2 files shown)


seanseth7:
I'm currently in the process of getting the CLA approved by my org's admin

johngmyers:
I don't think it's okay to just skip over the ingress. This could cause the ALB to be torn down if its cert gets deleted.

The right way to address this is through the ingress admission controller. I'm not going to be reviewing ingress admission PRs until #3026 lands, though.

seanseth7:
> I don't think it's okay to just skip over the ingress. This could cause the ALB to be torn down if its cert gets deleted.
> 
> The right way to address this is through the ingress admission controller. I'm not going to be reviewing ingress admission PRs until #3026 lands, though.

@johngmyers
Do you mind clarifying how the ALB would get torn down? I was under the impression that if there are no valid certs for the HTTPS listener, that listener would simply not be configured for the ALB.

Ah I see. Just to give you more context on why I was trying to make this change, my team was running into issues whenever a certificate expires for an ingress in an ingress group. The controller currently just gets into an error state because no cert exists for an ingress' hosts and doesn't successfully process any new ingresses in that group, so I was just trying to find the simplest approach to getting around that issue. 

Also, is that ingress admission controller something that still needs to be built? Do you mind linking the code if it already exists? I know that this issue https://github.com/kubernetes-sigs/aws-load-balancer-controller/projects/11#card-88834664 was slated for the 2.6.0 release, which seems to solve my team's issue as well, and if you still think that's accurate, do you have an idea as to an ETA for that release? All good if not. 

Thanks!

johngmyers:
The admission controller is part of LBC. I cannot predict when things would be released.

seanseth7:
> The admission controller is part of LBC. I cannot predict when things would be released.
@johngmyers 
Got it. Thanks for clarifying.

seanseth7:
@johngmyers I took some more time to look at the PR you mentioned https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3026 , and just to be sure that we're on the same page, you think it's best to simply update that ingress validator to add another check for if an ingress has a valid certificate in ACM for its hosts, correct? I'm happy to draft up another PR based on the changes you already made in that PR with new certificate validation logic if you agree.

johngmyers:
@seanseth7 yes, that is the shape of it. I haven't explored the edge cases, such as what happens if the cert expires after the LB is created.

hmehta:
Upvote, as we would be needing this as well.

However, I think it should be configurable with an annotation, for example `alb.ingress.kubernetes.io/tls-autodiscovery-strategy` which would default to `match-all` which would enforce all the hosts to have same certificate, and to enable this, it would have `match-any` strategy available too.

Issue #3285, opened 2023-07-18
AWS Load Balancer Controller should not block "TargetGroupBinding" update if invalid Security Group ID assign


guessi:
**Describe the bug**

According to [TargetGroupBinding Specification](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.5/guide/targetgroupbinding/spec/), user could define [securityGroup](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.5/guide/targetgroupbinding/spec/#elbv2.k8s.aws/v1beta1.SecurityGroup) for **TargetGroupBinding**. However, if user provide an invalid Security Group ID, `Security Group ID does exists, but it's belongs to another AWS Account, or maybe there's a typo`, AWS Load Balancer Controller will fail to manage ***ALL TargetGroupBindings***. ( Tested with v2.4.2, 2.4.4, v2.4.7, v2.5.3, all the same )

**Steps to reproduce**

```
apiVersion: elbv2.k8s.aws/v1beta1
kind: TargetGroupBinding
metadata:
  name: ...
  namespace: ...
spec:
  ipAddressType: ...
  serviceRef:
    name: ...
    port: ...
  networking:
    ingress:
      from:
        securityGroup: sg-xxxxxxxxxx # <------------ put an invalid SG-ID, or unreachable SG-ID.
  targetGroupARN: ...
  targetType: ...
```

AWS Load Balancer Controller will showing error like this,

```
{"level":"error","ts":... ,"logger":"controller-runtime.manager.controller.targetGroupBinding","msg":"Reconciler error","reconciler group":"[elbv2.k8s.aws ](http://elbv2.k8s.aws/)","reconciler kind":"TargetGroupBinding","name":"...","namespace":"...","error":"InvalidGroup.NotFound: The security group 'sg-XXXXXXXXX' does not exist\n\tstatus code: 400, request id: ..."
```

But the error message is actually showing unrelated TargetGroupBindings, since all TargetGroupBindings are blocked by problematic one, and all errors pointed to the same Security Group ID.

**Expected outcome**

Even one of the ***TargetGroupBinding*** is invalid, it should not block the rest of others.

For example, if there have A, B, C, D four TargetGroupBindings, and C have invalid securityGroup assigned. A, B, D should still reconcile-able. But not they all unable to get update.

**Environment**

* AWS Load Balancer controller version: `All version could reproduce this issue easily.`
[...]

PR #3296, opened 2023-07-26
don't block TGB reconciliation loop on failed SG ingress reconciliation

michaelsaah:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3285

### Description

the controller performs an SG reconciliation step for all (cluster-wide) SGs gathered from all TGBs during the TGB reconciliation loop, before TG endpoints are reconciled: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/d177c898ddd86071eecc2fd918d72ebfb0af7892/pkg/targetgroupbinding/resource_manager.go#L139-L141

the way the code is currently written, this means that any failure during SG reconciliation blocks reconciliation of all targets across the whole cluster. such a failure can be caused by something as innocuous as a SG being deleted before the associated TGB is deleted, or a SG being entered on a TGB erroneously. this can easily lead to severe outages if not remediated quickly.

this commit changes the method `reconcileWithIngressPermissionsPerSG` to not exit on a single failed SG ingress reconciliation - instead it will log the offending error and continue through the loop.

I tested this by setting up two TGBs, modifying one to have a non-existent ingress SG, deleting a target pod, then ensuring that the new target is reconciled. the new error message was also logged:
```
{"level":"error","ts":"2023-07-26T19:39:03Z","msg":"Security group reconciliation failed for SG","SecurityGroupID":"sg-0d904125da9f258a7","error":"InvalidGroup.NotFound: The security group 'sg-0b4935e438dcca2ee' does not exist\n\tstatus code: 400, request id: 34e0001c-c85e-4f6e-b7e1-c32093afe402"}
```

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/targetgroupbinding/networking_manager.go
+++ b/pkg/targetgroupbinding/networking_manager.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"strings"
 	"sync"
+	libErrors "errors"
 
 	awssdk "github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -199,11 +200,12 @@ func (m *defaultNetworkingManager) reconcileWithIngressPermissionsPerSG(ctx cont
 	aggregatedIngressPermissionsPerSG := m.computeAggregatedIngressPermissionsPerSG(ctx)
 
 	permissionSelector := labels.SelectorFromSet(labels.Set{tgbNetworkingIPPermissionLabelKey: tgbNetworkingIPPermissionLabelValue})
+	sgReconciliationErrors := make([]error, 0)
 	for sgID, permissions := range aggregatedIngressPermissionsPerSG {
 		if err := m.sgReconciler.ReconcileIngress(ctx, sgID, permissions,
 			networking.WithPermissionSelector(permissionSelector),
 			networking.WithAuthorizeOnly(!computedForAllTGBs)); err != nil {
-			m.logger.Error(err, "Security group reconciliation", "SecurityGroupID", sgID)
+			sgReconciliationErrors = append(sgReconciliationErrors, err)
 			continue
 		}
 	}
@@ -212,6 +214,11 @@ func (m *defaultNetworkingManager) reconcileWithIngressPermissionsPerSG(ctx cont
 		if err := m.gcIngressPermissionsFromUnusedEndpointSGs(ctx, aggregatedIngressPermissionsPerSG); err != nil {
 			return err
 		}
+	}
+
+	if len(sgReconciliationErrors) > 0 {
+		err := libErrors.Join(sgReconciliationErrors...)
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


johngmyers (changes requested):
I cannot support what-me-worry ignoring of errors. At the very least, the error needs to be surfaced as a Kubernetes event and the handler needs to mark the work item for retry.

michaelsaah:
thanks @johngmyers. surfacing the error via an Event sounds fine to me.

I'm not sure what you mean by "mark the work item for retry." can you point me towards any examples in the code where this is performed?

michaelsaah:
@johngmyers @M00nF1sh just bumping this, it's an ongoing operational issue for us (Twilio) on EKS. any help appreciated.

michaelsaah:
just another bump, any eyes appreciated

M00nF1sh:
@michaelsaah 

Thanks for this contribution. I agree with @johngmyers that the error cannot be ignored without retry. Since such SG modification errors can be transient errors such as EC2 api throttle and shall be retried to avoid issues in security group settings.

How about below change:
1. inside reconcileWithIngressPermissionsPerSG, when reconcile for each SG(aggregatedIngressPermissionsPerSG), we don't fail upon error but aggregate all errors into a single error and return it. (e.g. use errors.Join)
2. in https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/d177c898ddd86071eecc2fd918d72ebfb0af7892/pkg/targetgroupbinding/resource_manager.go#L139, if ReconcileForPodEndpoints failed with error, we don't return the error immediately, instead we continue target registration/deregistration and return `runtime.NewRequeueNeeded("networking reconciliation")` if the ReconcileForPodEndpoints returned a error before.

michaelsaah:
thanks for the review @M00nF1sh, addressed your feedback. have a good weekend!

michaelsaah:
just bumping this for a re-review, thanks

michaelsaah:
morning, just a monday review bump, thanks

pkg/targetgroupbinding/networking_manager.go
```diff
@@ -199,11 +200,13 @@ func (m *defaultNetworkingManager) reconcileWithIngressPermissionsPerSG(ctx cont
 	aggregatedIngressPermissionsPerSG := m.computeAggregatedIngressPermissionsPerSG(ctx)
 
 	permissionSelector := labels.SelectorFromSet(labels.Set{tgbNetworkingIPPermissionLabelKey: tgbNetworkingIPPermissionLabelValue})
+	sgReconciliationErrors := make([]error, 0)
```
johngmyers, line 203:
```suggestion
	var sgReconciliationErrors []error
```

johngmyers (approved):
Just a minor style nit

/lgtm

michaelsaah:
thanks for the re-review @johngmyers, committed your style suggestion

michaelsaah:
looks like that test might be flaky: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3289#issuecomment-1646227496

M00nF1sh (approved):
Sorry for late review. This change looks good to me now, thanks a lot for your contribution :D 

/lgtm
/approve

Issue #2457, opened 2022-01-11
security-groups annotation attachment not working


antigenius0910:
**Describe the bug**
I have my aws-loadbalancer-controller create with follow arg
```
  - args:
    - --cluster-name=XXXX
    - --ingress-class=alb
    - --disable-restricted-sg-rules=true
```
based on documentation here. https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/ingress/annotations/#security-groups

when I add annotation 
```
  annotations:
    alb.ingress.kubernetes.io/security-groups: sg-0248f7f1feed2xxxx
```
then this particular security group should be attach to my loadbalancer, however, a new security group has been created
```
ᐅ aws --profile=xxxx-staging elb describe-load-balancers --load-balancer-names ab6d68f8da95d499e9048758743fce76 | grep -A2 SecurityGroups
            "SecurityGroups": [
                "sg-09062ba8f62540221"
            ],
```

**Steps to reproduce**
```
ᐅ kubectl apply -f allowed_example.yaml

apiVersion: v1
kind: Service
metadata:
  name: example-service
  annotations:
    alb.ingress.kubernetes.io/security-groups: sg-0248f7f1feed2xxxx
spec:
  selector:
    app: example
  ports:
    - port: 1111
      targetPort: 1111
  type: LoadBalancer
```

```
Name:                     example-service
Namespace:                default
Labels:                   <none>
Annotations:              alb.ingress.kubernetes.io/security-groups: sg-0248f7f1feed2xxxx
Selector:                 app=example
Type:                     LoadBalancer
IP Family Policy:         SingleStack
IP Families:              IPv4
IP:                       172.20.234.196
IPs:                      172.20.234.196
LoadBalancer Ingress:     ab6d68f8da95d499e9048758743fce76-573891456.us-east-2.elb.amazonaws.com
Port:                     <unset>  1111/TCP
TargetPort:               1111/TCP
NodePort:                 <unset>  30359/TCP
Endpoints:                <none>
Session Affinity:         None
External Traffic Policy:  Cluster
Events:
  Type    Reason                Age   From                Message
  ----    ------                ----  ----                -------
  Normal  EnsuringLoadBalancer  28m   service-controller  Ensuring load balancer
  Normal  EnsuredLoadBalancer   28m   service-controller  Ensured load balancer

```

```
[...]

PR #3313, opened 2023-08-07
webhook: new flag to configure the webhook server's minimum TLS version

alebedev87:
### Issue

N/A

### Description

The goal of this PR is to support environments where TLS versions other than `1.3` are still used, mainly FIPS enabled systems. The boringcrypto enabled builds check for the maximum FIPS TLS version [which is set to TLS1.2](https://github.com/golang/go/blob/0184fe5ece4f84fda9db04d2472b76efcaa8ef55/src/crypto/tls/boring.go#L24-L28).

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/config/runtime_config.go
+++ b/pkg/config/runtime_config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/spf13/pflag"
@@ -26,6 +27,7 @@ const (
 	flagWebhookCertDir          = "webhook-cert-dir"
 	flagWebhookCertName         = "webhook-cert-file"
 	flagWebhookKeyName          = "webhook-key-file"
+	flagWebhookTLSMinVersion    = "webhook-tls-min-version"
 
 	defaultKubeconfig              = ""
 	defaultLeaderElectionID        = "aws-load-balancer-controller-leader"
@@ -40,10 +42,18 @@ const (
 	defaultQPS = 1e6
 	// High enough Burst to fit all expected use cases. Burst=0 is not set here, because
 	// client code is overriding it.
-	defaultBurst           = 1e6
-	defaultWebhookCertDir  = ""
-	defaultWebhookCertName = ""
-	defaultWebhookKeyName  = ""
+	defaultBurst                = 1e6
+	defaultWebhookCertDir       = ""
+	defaultWebhookCertName      = ""
+	defaultWebhookKeyName       = ""
+	defaultWebhookTLSMinVersion = "1.3"
+)
+
+var (
+	supportedTLSVersions = map[string]struct{}{
+		"1.2": {},
+		"1.3": {},
+	}
 )
 
 // RuntimeConfig stores the configuration for the controller-runtime
@@ -61,6 +71,7 @@ type RuntimeConfig struct {
 	WebhookCertDir          string
 	WebhookCertName         string
 	WebhookKeyName          string
+	WebhookTLSMinVersion    string
 }
 
 // BindFlags binds the command line flags to the fields in the config object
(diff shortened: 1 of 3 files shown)


pkg/config/runtime_config.go
```diff
@@ -40,10 +42,21 @@ const (
...
+	defaultWebhookCertName      = ""
+	defaultWebhookKeyName       = ""
+	defaultWebhookTLSMinVersion = "1.3"
+)
+
+var (
+	supportedTLSVersions = map[string]struct{}{
+		"":    {},
+		"1.0": {},
+		"1.1": {},
```
johngmyers, line 56: We definitely shouldn't support 1.0 or 1.1

johngmyers:
I think we should revert #2457 and make 1.2 the minimum version across the board. @oliviassss ?

johngmyers:
My opinion is that we should not be exposing this as configuration; administrators should not have to bother with this kind of stuff.

On the other hand, we probably should cater somewhat for people who are constrained to work within the idiotic FIPS-140 stuff.

If there are problematic ciphers in the Go's default TLS 1.2 support (I would suggest we check) then we can just hardcode-disable those.

pkg/config/runtime_config.go
alebedev87: Yeah, these ones are too outdated for now. I added them since the controller-runtime still supports them. I'm fine to remove them along with `""` as it turns into `1.0`.

oliviassss:
@johngmyers, yeah I agree 1.0 and 1.1 are outdated, AWS is also deprecating the 1.0 and 1.1 for the API endpoints: https://aws.amazon.com/blogs/security/tls-1-2-required-for-aws-endpoints/. And I agree that the exposed configuration is not necessary here, we could make 1.2 the minimum. 

But I'm just a bit concerned about the 1.2 compatibility with k8s 1.22+ (since we are supporting k8s 1.22+). By checking the [changelogs](https://github.com/kubernetes/kubernetes/blob/1620473a9a01dd6bbef3398c0acb2e581d0a13c3/CHANGELOG/CHANGELOG-1.3.md?plain=1#L377), seems like TLS 1.2 is the min supported version since k8s 1.3?

alebedev87:
> But I'm just a bit concerned about the 1.2 compatibility with k8s 1.22+ (since we are supporting k8s 1.22+). By checking the [changelogs](https://github.com/kubernetes/kubernetes/blob/1620473a9a01dd6bbef3398c0acb2e581d0a13c3/CHANGELOG/CHANGELOG-1.3.md?plain=1#L377), seems like TLS 1.2 is the min supported version since k8s 1.3?

Looking at `KubeleConfiguration`, the minimum TLS version is set to `""` by default in 1.27 ([API ref](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-KubeletConfiguration)) which is later turn into TLS1.2 thanks to [kubernetes/component-base/cli/flag#TLSVersion](https://github.com/kubernetes/component-base/blob/854e515d6d9cdba26f2da339c3f92a1e3be81b4a/cli/flag/ciphersuites_flag.go#L131-L147) used by [kubelet here](https://github.com/kubernetes/kubernetes/blob/1620473a9a01dd6bbef3398c0acb2e581d0a13c3/cmd/kubelet/app/server.go#L1117C38-L1117C38). Which explains why I didn't see any problems with other controller-runtime based workloads which don't set `TLSMinVersion` on the latest Kubernetes releases.

So, TLS1.2 seems to be a safe default however it's indeed only [intermediate security TLS profile](https://wiki.mozilla.org/Security/Server_Side_TLS). Maybe it still makes some sense to be in the modern profile by default (TLS1.3) and let some exotic environments like FIPS set TLS1.2 (not strong opinion, revert to L1.2 by default would do the thing for me too).

alebedev87:
`""`, `1.0`, `1.1` versions removed, just in case we would like to move on with the configurable approach.

pkg/config/runtime_config.go
alebedev87: `1.0`, `1.1` removed.

oliviassss:
@alebedev87, thanks, I think it's better to set TLS default version as 1.2, so it fits all the environments including FIPS, and allow the configuration for 1.3 for users who want stricter TLS version.

johngmyers:
Filed #3318

alebedev87:
Closing since https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3318 reset the minimum TLS version to 1.2.

PR #3329, opened 2023-08-10
Add support for NLB security group

oliviassss:
### Description

Added support of Security Groups for NLB. With the security group support, it is feasible to forward the NLB traffic to the EC2 instances without having to open up the instances for global access. For backwards compatibility, NLBs created without the security groups or the existing NLBs will continue to provide the legacy behavior. Similar to ALB, there are two sets of SGs for NLB - frontend and backend SGs:

- The controller will automatically create and attach the frontend SG to the NLB provisioned, and add rules for `inbound-cidrs` and `listen-ports`. If the users want to attach existing frontend SG to the NLB, they can explicitly specify via annotation `service.beta.kubernetes.io/aws-load-balancer-security-groups`
- The Backend SG controls the traffic between the NLB and the EC2 instances/ENIs, and it gets attached to the NLB similar to the frontend SG. In case of auto-generated frontend SG, the controller automatically adds Node/ENI SG rules to allow egress traffic from the NLB. The rule management is disabled by default if the frontend SG is specified via annotation. We provide an annotation to configure controller’s management on backend SG rules regardless of the frontend SG type `service.beta.kubernetes.io/aws-load-balancer-manage-backend-security-group-rules: true/false`

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/service/model_build_load_balancer.go
+++ b/pkg/service/model_build_load_balancer.go
@@ -10,15 +10,17 @@ import (
 	"sort"
 	"strconv"
 
-	"sigs.k8s.io/aws-load-balancer-controller/pkg/algorithm"
-	elbv2deploy "sigs.k8s.io/aws-load-balancer-controller/pkg/deploy/elbv2"
-
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/service/ec2"
+	awssdk "github.com/aws/aws-sdk-go/aws"
+	ec2sdk "github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/algorithm"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/annotations"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/config"
+	elbv2deploy "sigs.k8s.io/aws-load-balancer-controller/pkg/deploy/elbv2"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/deploy/tracking"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/k8s"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/model/core"
 	elbv2model "sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/networking"
 )
@@ -33,7 +35,11 @@ const (
 )
 
 func (t *defaultModelBuildTask) buildLoadBalancer(ctx context.Context, scheme elbv2model.LoadBalancerScheme) error {
-	spec, err := t.buildLoadBalancerSpec(ctx, scheme)
+	existingLB, err := t.fetchExistingLoadBalancer(ctx)
+	if err != nil {
+		return err
+	}
+	spec, err := t.buildLoadBalancerSpec(ctx, scheme, existingLB)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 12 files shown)


pkg/service/model_build_load_balancer.go
```diff
@@ -67,13 +78,85 @@ func (t *defaultModelBuildTask) buildLoadBalancerSpec(ctx context.Context, schem
...
 }
 
+func (t *defaultModelBuildTask) buildLoadBalancerSecurityGroups(ctx context.Context, existingLB *elbv2deploy.LoadBalancerWithTags,
+	ipAddressType elbv2model.IPAddressType) ([]core.StringToken, error) {
+	if existingLB != nil && len(existingLB.LoadBalancer.SecurityGroups) == 0 {
+		return nil, nil
+	}
+	if !t.featureGates.Enabled(config.NLBSecurityGroup) {
+		if existingLB != nil && len(existingLB.LoadBalancer.SecurityGroups) != 0 {
+			return nil, errors.New("conflicting security groups configuration")
```
johngmyers, line 96: This is not a terribly helpful error message, as it gives no clue as to what is conflicting. Better to say something like "existing NLB has security groups attached, but NLBSecurityGroup feature gate is disabled"

pkg/service/model_build_load_balancer.go
```diff
@@ -67,13 +78,85 @@ func (t *defaultModelBuildTask) buildLoadBalancerSpec(ctx context.Context, schem
...
 		LoadBalancerAttributes: lbAttributes,
 		Tags:                   tags,
 	}
 	return spec, nil
 }
 
+func (t *defaultModelBuildTask) buildLoadBalancerSecurityGroups(ctx context.Context, existingLB *elbv2deploy.LoadBalancerWithTags,
+	ipAddressType elbv2model.IPAddressType) ([]core.StringToken, error) {
+	if existingLB != nil && len(existingLB.LoadBalancer.SecurityGroups) == 0 {
+		return nil, nil
```
johngmyers, line 92: Should it not be possible to add a SG to an existing NLB by explicitly setting the relevant annotation?

pkg/service/model_builder_test.go
```diff
@@ -2930,53 +3006,3424 @@ func Test_defaultModelBuilderTask_Build(t *testing.T) {
...
 			}
-			builder := NewDefaultModelBuilder(annotationParser, subnetsResolver, vpcInfoProvider, "vpc-xxx", trackingProvider, elbv2TaggingManager, featureGates,
-				"my-cluster", nil, nil, "ELBSecurityPolicy-2016-08", defaultTargetType, enableIPTargetType, serviceUtils)
+			builder := NewDefaultModelBuilder(annotationParser, subnetsResolver, vpcInfoProvider, "vpc-xxx", trackingProvider, elbv2TaggingManager, ec2Client, featureGates,
+				"my-cluster", nil, nil, "ELBSecurityPolicy-2016-08", defaultTargetType, enableIPTargetType, serviceUtils,
+				backendSGProvider, sgResolver, tt.enableBackendSG, tt.disableRestrictedSGRules)
 			ctx := context.Background()
-			stack, _, err := builder.Build(ctx, tt.svc)
+			stack, _, _, err := builder.Build(ctx, tt.svc)
 			if tt.wantError {
```
johngmyers, line 6423: `wantError` should be more specific: it should be a `string` and should `assert.EqualError()`

pkg/service/model_builder_test.go
```diff
@@ -2930,53 +3006,3424 @@ func Test_defaultModelBuilderTask_Build(t *testing.T) {
...
 				enableIPTargetType = *tt.enableIPTargetType
 			}
-			builder := NewDefaultModelBuilder(annotationParser, subnetsResolver, vpcInfoProvider, "vpc-xxx", trackingProvider, elbv2TaggingManager, featureGates,
-				"my-cluster", nil, nil, "ELBSecurityPolicy-2016-08", defaultTargetType, enableIPTargetType, serviceUtils)
+			builder := NewDefaultModelBuilder(annotationParser, subnetsResolver, vpcInfoProvider, "vpc-xxx", trackingProvider, elbv2TaggingManager, ec2Client, featureGates,
+				"my-cluster", nil, nil, "ELBSecurityPolicy-2016-08", defaultTargetType, enableIPTargetType, serviceUtils,
+				backendSGProvider, sgResolver, tt.enableBackendSG, tt.disableRestrictedSGRules)
 			ctx := context.Background()
-			stack, _, err := builder.Build(ctx, tt.svc)
+			stack, _, _, err := builder.Build(ctx, tt.svc)
```
johngmyers, line 6422: Should probably be doing asserts on the returned value of `backendSGRequired`

Issue #3330, opened 2023-08-10
Expose configuration for tolerateNonExistentBackendService and tolerateNonExistentBackendAction


kc9ddi:
**Is your feature request related to a problem?**
When deploying an ingress that is misconfigured with a missing service or action, the ingress will deploy successfully. However, when serving requests destined for the missing service/action, those requests will fail with an HTTP 503 error. This behavior was introduced intentionally in https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/1967 - in order to ensure that a single misconfiguration doesn't block all ingress reconciliation. However, it also means that a category of misconfiguration is more difficult to discover prior to deploying it and running traffic against the misconfigured ingress.

**Describe the solution you'd like**
The above PR mentions that this behavior should be configurable, to allow user to opt-in to the behavior as it was prior to the PR making that change. It would be nice if this configuration were exposed.

**Describe alternatives you've considered**
Manual validation of ingress configuration prior to deployment using custom / 3rd party tools.

M00nF1sh:
I think it’s reasonable to allow configuration of this，maybe a creature flag?

PR #3342, opened 2023-08-16
Expose ingress configuration options for missing backends

kc9ddi:
### Issue

This PR fixes https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3330

### Description

This PR adds command line flags to specify whether the the controller should tolerate missing backend services and actions. If the flags are not specified, the default value will be to tolerate missing backend actions and services (which is the current behavior.)

The code changes here are simply to thread these values through to `enhanced_backend_builder.go`, and rely on them rather than using the `defaultTolerateNonExistentBackendService` and `defaultTolerateNonExistentBackendAction` constant fields. The logic here is already unit tested, so no new tests have been added.

### Checklist
- [x] Manually tested

### DIFF
--- a/helm/aws-load-balancer-controller/templates/deployment.yaml
+++ b/helm/aws-load-balancer-controller/templates/deployment.yaml
@@ -122,6 +122,12 @@ spec:
         {{- if kindIs "bool" .Values.disableIngressGroupNameAnnotation }}
         - --disable-ingress-group-name-annotation={{ .Values.disableIngressGroupNameAnnotation }}
         {{- end }}
+        {{- if kindIs "bool" .Values.tolerateNonExistentBackendService }}
+        - --tolerate-non-existent-backend-service={{ .Values.tolerateNonExistentBackendService }}
+        {{- end }}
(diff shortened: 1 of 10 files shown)


kc9ddi:
Hi @M00nF1sh and @oliviassss, I just wanted to confirm that review of this PR isn't waiting on me for anything?

helm/aws-load-balancer-controller/templates/deployment.yaml
```diff
@@ -122,6 +122,12 @@ spec:
         {{- if kindIs "bool" .Values.disableIngressGroupNameAnnotation }}
         - --disable-ingress-group-name-annotation={{ .Values.disableIngressGroupNameAnnotation }}
         {{- end }}
+        {{- if kindIs "bool" .Values.tolerateNonExistentBackendService }}
```
oliviassss, line 125: Could you also add an entry in the `values.yaml`?

kc9ddi: Yep, updated.

M00nF1sh (approved):
/lgtm
/approve

PR #3343, opened 2023-08-16
Add check for LB availability to e2e

oliviassss:
### Description

Add a step to wait for the NLB to be available before proceeding to any resource deletions, to some test suites that were lacking this check. For the below reasons:

- There's a limit from ELB side that if we delete the SG before NLB configuration finishes, it will cause an issue and may put the internal metrics on fire, and ELB is working on adding an error message to this.
- We should check the LB status no matter whether we are doing traffic test or just validating the configurations. 

#### Test:

- Ran the e2e test in my dev cluster and verified all test suites passed.

### Checklist
- [ ] Manually tested


johngmyers:
The suite is timing out. Perhaps the timeout in scripts/run-e2e-test.sh needs to be increased?

oliviassss:
@johngmyers, thank you! I was just aware that ginkgo v2 change the default timeout from 24h to 1h. I have updated the timeout to 2h, let's see how it works this time.

test/e2e/service/nlb_instance_target_test.go
```diff
@@ -3,14 +3,13 @@ package service
...
 	"fmt"
-	"strings"
-
 	awssdk "github.com/aws/aws-sdk-go/aws"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/aws-load-balancer-controller/test/framework/http"
 	"sigs.k8s.io/aws-load-balancer-controller/test/framework/utils"
+	"strings"
```
johngmyers, line 12: This is an undesirable change. Please format with "goimports"

johngmyers (changes requested):

scripts/ci_e2e_test.sh
```diff
@@ -210,7 +210,7 @@ test_controller_image() {
   CERT_ID3="1001570b-1779-40c3-9b49-9a9a41e30058"
   CERTIFICATE_ARNS=${CERTIFICATE_ARNS:-"${CERTIFICATE_ARN_PREFIX}/${CERT_ID1},${CERTIFICATE_ARN_PREFIX}/${CERT_ID2},${CERTIFICATE_ARN_PREFIX}/${CERT_ID3}"}
 
-  ginkgo -v -r test/e2e -- \
+  ginkgo -v -r test/e2e -timeout 2h -- \
```
johngmyers, line 213:
```suggestion
  ginkgo -v -r test/e2e --timeout 2h -- \
```

Issue #1572, opened 2020-10-27
Global LoadBalancer support


M00nF1sh:
Global LoadBalancer support

guan840912:
any update 😊

jwenz723:
Bump to keep this from closing.

nvermalulu:
any update .. Must need feature  for DR

ALB Ingress --> Global Load Balancer (aws global accelerator ) --> external DNS     automation  like we have currently ALB Ingress --> external DNS

olemarkus:
Being able to have LB controller manage either instance or IP targets to a global accelerator would be really useful.

IliaGe:
+1 for needed use cases :)

xeivieni:
+1 needed here to 🙏

tien1214:
+1

dbs-gong:
+1

goruha:
+1

apatapniou:
+1

M00nF1sh:
We acknowledge this is a important feature request, and will start work on it post our v2.4.0 release.

vyanhursky:
+1

danthegoodman1:
+1

mihaela1992:
+1

ChristianOps:
+1

carlcauchi:
+1

Eran-BenShahar:
+1

GuilleAmutio:
+1

olivernaaris:
+1

ShonL:
Any updates? 2.4 is out

mreardonx:
+1

WaffleThief123:
+1 here as well

acrazing:
+1 again, any updates?

anilerencelik:
+1

thalescosta:
+1

fabianberisha:
+1

regmicmahesh:
+1

fjanicki:
bump, +1

sgrotz-cepres:
No update on this?

gbhosal:
+1

jwenz723:
I was told by an AWS tam that this will likely not get worked on during 2023.

ChrisLahaye:
You could provision the load balancer with appropriate tags and global accelerator resources manually. I know this is not an ideal or fully dynamic solution but it at least lets you use ingresses with global accelerator.

[...]

PR #3405, opened 2023-09-27
feat: add support for global accelerator endpoint groups

kaessert:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1572

### Description

It's a first try to get some support to register and deregister endpoints to an existing endpointgroup.
Because users are limited to one endpointgroup per region, i also think it makes the most sense to approach it that way. Later down the line one could work with weights even which would give users a mechanism to switch traffic between clusters.

The approach i chose for now has an ugly limitation of being: You can't delete an endpoint by deleting the listener. It looks to me like this is a design-limitation because as far as i see it:
- logic to destroy stuff is omitting it's output in the generated stack per ingress group
- ususally the external state is pictured with tags 
- endpoints on ga don't support tags
- this means to me that the current logic of deleting resources is not capable of capturing the deletion case :(

What i tried next is to implement a custom cleanup logic, inspired by the way we cleanup security groups but that has the pitfall that we do the cleanup AFTER the stack is deployed (deleted in that case) which leaves us with no way of identifying the endpoint we created earlier on because we can't access the arn of the load-balancer anymore. 

Maybe someone who knows the code better knows how to work around that. I considered adding labels as a workaround but wanted to get feedback first :)

### Checklist
- [x ] Added/modified documentation as required (such as the `README.md`, or the `docs` directory)
- [x] Manually tested

### DIFF
--- a/pkg/aws/services/globalaccelerator.go
+++ b/pkg/aws/services/globalaccelerator.go
@@ -1,0 +1,25 @@
+package services
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/globalaccelerator"
+	"github.com/aws/aws-sdk-go/service/globalaccelerator/globalacceleratoriface"
+)
+
+type GlobalAccelerator interface {
+	globalacceleratoriface.GlobalAcceleratorAPI
+}
+
+// NewGlobalAccelerator constructs new GlobalAccelerator implementation.
+func NewGlobalAccelerator(session *session.Session) GlobalAccelerator {
+	return &defaultGlobalAcceleratorAPI{
+		// global accelerator always needs `us-west-2`-region
+		GlobalAcceleratorAPI: globalaccelerator.New(session, aws.NewConfig().WithRegion("us-west-2")),
+	}
+}
+
+// default implementation for Global Accelerator.
+type defaultGlobalAcceleratorAPI struct {
+	globalacceleratoriface.GlobalAcceleratorAPI
+}
--- a/pkg/deploy/globalaccelerator/endpoint_manager.go
+++ b/pkg/deploy/globalaccelerator/endpoint_manager.go
@@ -1,0 +1,85 @@
+package globalaccelerator
+
+import (
+	"context"
+
+	awssdk "github.com/aws/aws-sdk-go/aws"
+	gasdk "github.com/aws/aws-sdk-go/service/globalaccelerator"
+
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/aws/services"
+)
+
+type EndpointManager interface {
+	// AddEndpoint add an endpoint to a globalaccelerator endpoint group
+	AddEndpoint(ctx context.Context, endpointGroupARN string, lbArn string) error
+
+	// GetEndpoint gets the existing endpoint for an endpointgroup and load balancer
+	GetEndpoint(ctx context.Context, endpointGroupARN string, lbArn string) (string, error)
+
+	// DeleteEndpoints deletes an existing endpoint for an endpointgroup and load balancer
+	DeleteEndpoint(ctx context.Context, endpointGroupARN string, lbArn string) error
+}
+
+func NewDefaultEndpointManager(gaClient services.GlobalAccelerator, logger logr.Logger) *defaultEndpointManager {
+	return &defaultEndpointManager{
+		gaClient: gaClient,
+	}
+}
+
+var _ EndpointManager = &defaultEndpointManager{}
+
+type defaultEndpointManager struct {
+	gaClient services.GlobalAccelerator
+}
+
+func (m *defaultEndpointManager) AddEndpoint(ctx context.Context, endpointGroupARN string, lbArn string) error {
+	_, err := m.gaClient.AddEndpoints(&gasdk.AddEndpointsInput{
+		EndpointGroupArn: &endpointGroupARN,
+		EndpointConfigurations: []*gasdk.EndpointConfiguration{
+			{
+				EndpointId: awssdk.String(lbArn),
+			},
+		},
+	})
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (m *defaultEndpointManager) GetEndpoint(ctx context.Context, endpointGroupARN string, lbArn string) (string, error) {
+	epResponse, err := m.gaClient.DescribeEndpointGroup(&gasdk.DescribeEndpointGroupInput{
+		EndpointGroupArn: &endpointGroupARN,
+	})
+
+	if err != nil {
+		return "", err
+	}
+
+	for _, endpoint := range epResponse.EndpointGroup.EndpointDescriptions {
+		if *endpoint.EndpointId == lbArn {
+			return lbArn, nil
+		}
+	}
+
+	return "", nil
+}
+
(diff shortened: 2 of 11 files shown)


johngmyers:
I don't think we should use an annotation for the ARN; that's in the domain of the cluster operator, not the service developer. This should instead be a field in `IngressClassParams`. It should also support searching by tag selector.

johngmyers:
Documentation of the new annotations are missing from `docs/guide/ingress/annotations.md`.

johngmyers:
If we can't find anywhere in the AWS API to stash ownership tags, we could at least store them somewhere in the Kubernetes API.

pkg/aws/services/globalaccelerator.go
```diff
@@ -0,0 +1,25 @@
...
+
+// NewWAFv2 constructs new WAFv2 implementation.
+func NewGlobalAccelerator(session *session.Session) GlobalAccelerator {
+	return &defaultGlobalAcceleratorAPI{
+		// global accelerator always needs `us-west-2`-region
+		GlobalAcceleratorAPI: globalaccelerator.New(session, aws.NewConfig().WithRegion("us-west-2")),
+	}
+}
+
+// default implementation for WAFv2.
```
johngmyers, line 22:
```suggestion
// default implementation for Global Accelerator.
```

pkg/deploy/globalaccelerator/endpoint_manager.go
```diff
@@ -0,0 +1,86 @@
+package globalaccelerator
+
+import (
+	"context"
+
+	//networking "k8s.io/api/networking/v1"
```
johngmyers, line 6:
```suggestion
```

pkg/deploy/globalaccelerator/endpoint_synthesizer.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	"context"
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/model/core"
+	elbv2model "sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2"
+	gamodel "sigs.k8s.io/aws-load-balancer-controller/pkg/model/globalaccelerator"
+)
+
+// NewProtectionSynthesizer constructs new protectionSynthesizer
```
johngmyers, line 12:
```suggestion
// NewEndpointSynthesizer constructs new endpointSynthesizer
```

pkg/deploy/globalaccelerator/endpoint_synthesizer.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (*endpointSynthesizer) PostSynthesize(ctx context.Context) error {
+	// nothing to do here.
+	return nil
+}
+
+func (s *endpointSynthesizer) synthesizeGAEndpoints(ctx context.Context, lbARN string, resEndpoints []*gamodel.Endpoint) error {
+	if len(resEndpoints) > 1 {
+		return errors.Errorf("[should never happen] multiple Global Accelerator Endpoints desired on LoadBalancer: %v", lbARN)
```
johngmyers, line 63:
```suggestion
		return fmt.Errorf("[should never happen] multiple Global Accelerator Endpoints desired on LoadBalancer: %v", lbARN)
```

johngmyers (commented):
For cleanup, can't you delete the endpoints to the ARNs of all ALBs in the stack which aren't requesting to be attached to GA? This would include all ARNs of all ALBs which are going to be deleted.

johngmyers:
You could also put a tag on the ALBs listing any endpoint group each is registered with.

atarax:
> I don't think we should use an annotation for the ARN; that's in the domain of the cluster operator, not the service developer. This should instead be a field in `IngressClassParams`. It should also support searching by tag selector.

That sound great. Shouldn't we not support both ways though like we do it for the ingress group name? I don't really get that last sentence about searching by tag selector though, sorry :/

atarax:
> If we can't find anywhere in the AWS API to stash ownership tags, we could at least store them somewhere in the Kubernetes API.

What do you suggest? Would the status of the ingress object be an option?

atarax:
> For cleanup, can't you delete the endpoints to the ARNs of all ALBs in the stack which aren't requesting to be attached to GA?

Sounds like a nice idea. Do you mean doing this by the same logic as in the load balancer synthesizer? I'm struggling to find a good way to abstract this without making methods of it public.

johngmyers:
> Shouldn't we not support both ways though like we do it for the ingress group name?

Other fields are done both ways because they started out when IngressClassParams didn't exist. My opinion is that new things that affect the ALB as a whole should only go into IngressClassParams, but I don't know what the approvers or other reviewers think.

> I don't really get that last sentence about searching by tag selector though, sorry :/

I was referring to the `tags` fields in the selectors added in #2945 and #3277. One should follow that pattern for being able to discover AWS resources by either ID or tags. 

> What do you suggest? Would the status of the ingress object be an option?

I don't think the ingress status has a place to put that. You could perhaps put it in an annotation and rely on a finalizer, but I think the tag-on-ALB idea is better.

atarax:
@johngmyers maybe we can start here because there is many open point. The main thing i'm stuck with is the fact that i don't know how to identify the load balancers which are about to get deleted in the `Synthesize` method. 
In the current setup, when in arriving in `Synthesize` for the endpoint the load balancer is already deleted.

atarax:
> Other fields are done both ways because they started out when IngressClassParams didn't exist. My opinion is that new things that affect the ALB as a whole should only go into IngressClassParams, but I don't know what the approvers or other reviewers think.

Oh, thanks a lot for that explanation. I'll work it that way then and i also see what you mean with searchable now :) 

> I don't think the ingress status has a place to put that. You could perhaps put it in an annotation and rely on a finalizer, but I think the tag-on-ALB idea is better.

I don't really like the idea of putting an annotation either, that's why i decided to get some feedback first before working on that. But the whole idea with keeping the state on the lb-tags kind of crumbles for me on the fact that in the endpoint `Synthesize` method, the lb is already deleted when a delete operation is taking place.

johngmyers:
I think you might need to add to the ALB tear-down code. But I haven't delved into the back-end code yet; I've been working on the front-end so far. @M00nF1sh has more expertise there.

atarax:
> I think you might need to add to the ALB tear-down code.
would be doable but would highly violate the modularity of the current design though

atarax:
@johngmyers @M00nF1sh , any news here?

hhamalai:
Great that you're working on this issue allowing proper lifecycle management of GA+AWS LB integration. 

Have you thought should e.g. `status.loadBalancer.ingress` contain the GA IPs instead of the LB hostname?
This could make this change interoperable with other projects like external-dns.

Plus, when GA is used with the LB the traffic probably should be routed via GA-only, and LB type should be internal so that GA cannot be skipped.

atarax:
@hhamalai regarding the load-balancer ip, yes i thought about implementing that later down the line. But there is still an architecutal problem because current logic completely relies on the fact that all objects are taggable which GA endpoints are not. We don't delete lb-dependencies first either, so i wanted feedback from architects how they would approach this.
This line for reference: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/f38fe4911cc17e96e8179eaf2b63ddec8fc36aff/pkg/deploy/elbv2/load_balancer_synthesizer.go#L62

Regarding second point, i like the idea but not sure, need to think about it :)

jukie:
Hi is this PR still being considered and is @atarax still working on it? I'd like to contribute and see this feature implemented so if not I'll open another PR based on their work with the suggested changes and other enhancements.

kaessert:
Sad

PR #3416, opened 2023-10-03
enable ingress instance e2e test for ipv6

oliviassss:
### Description

1. Enable ingress e2e test suites for instance ipv6 target type
2. Update prow script to run ingress and service instance test suites for ipv6 clusters
### Checklist
- [x] Manually tested


johngmyers:
It might be useful to test dualstack load balancers on IPv4 clusters, but that's orthogonal.

/lgtm

johngmyers:
One wrinkle with IPv6 instance targets is that the instances must have a **primary** IPv6 address or the targets won't register. Primary IPv6 addresses are a relatively newish concept, so not all tools support them yet. For example, the Terraform AWS provider doesn't support it yet: hashicorp/terraform-provider-aws#33733

Issue #3418, opened 2023-10-03
[NLB] Resolve health check port when a port name is provided


fad3t:
Hello,

We're having a service with multiple ports: a UDP port that's receiving the traffic, and a TCP port for the health check.
Both ports are exposed as (dynamic) nodePorts. For now we have to statically defined the nodePort for the TCP health check, and reference that port in the `service.beta.kubernetes.io/aws-load-balancer-healthcheck-port` annotation.

Instead, we would like to pass the port name to the annotation, and leave the LB controller resolve it.
I saw this has been implemented for the ALB (https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/799) but not for the NLB yet. I guess the same logic could be applied.

Thanks,
Fred

oliviassss:
Close this issue as the PR as shipped with v2.6.2 release.

javorkolev123:
Hello,

Do I understand correctly that this feature has already been added? It does not seem to be documented though - https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.8/guide/service/annotations/#healthcheck-port.

fad3t:
Hi,
That's correct, feature is there but I missed the doc.

PR #3419, opened 2023-10-03
feat: resolve health check port name for NLB

fad3t:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3418

### Description

This PR adds support for health check named ports. This is useful when the nodePort is dynamically assigned by the controller.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/service/model_build_listener.go
+++ b/pkg/service/model_build_listener.go
@@ -20,7 +20,7 @@ func (t *defaultModelBuildTask) buildListeners(ctx context.Context, scheme elbv2
 	}
 
 	for _, port := range t.service.Spec.Ports {
-		_, err := t.buildListener(ctx, port, *cfg, scheme)
+		_, err := t.buildListener(ctx, t.service, port, *cfg, scheme)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


pkg/service/model_build_listener.go
```diff
@@ -20,17 +20,17 @@ func (t *defaultModelBuildTask) buildListeners(ctx context.Context, scheme elbv2
 	}
 
 	for _, port := range t.service.Spec.Ports {
-		_, err := t.buildListener(ctx, port, *cfg, scheme)
+		_, err := t.buildListener(ctx, t.service, port, *cfg, scheme)
 		if err != nil {
```
M00nF1sh, line 24: we don't need to pass t.service down the call stack, we can just access t.service within `buildTargetGroupHealthCheckPort` :D 
(the reason we have this defaultModelBuildTask is to save states(e.g. service object been handled) as variables into defaultModelBuildTask instead of pass them around).

M00nF1sh:
thanks for contributing.
Left some minor comments, the rest lgtm :D

fad3t:
> thanks for contributing. Left some minor comments, the rest lgtm :D

Ahah, thanks for the review - I didn't realize about `defaultModelBuildTask` being available to the function 😄 
I've removed the service argument, should be OK now.

M00nF1sh (approved):
/lgtm
/approve

Issue #3082, opened 2023-03-01
Respects Single AZ Whitelist of ELB in v2


hitsub2:
**Is your feature request related to a problem?**
If we got a whitelist of single az of ELB, the v2 controller does not respects that, according to the document:

<p>
At least two subnets in different Availability Zones. The AWS Load Balancer Controller chooses one subnet from each Availability Zone. When multiple tagged subnets are found in an Availability Zone, the controller chooses the subnet whose subnet ID comes first lexicographically. Each subnet must have at least eight available IP addresses.
</p>

Currently error comes up from the v2 controller

```json
{"level":"error","ts":1677568163.5564663,"logger":"controller-runtime.manager.controller.ingress","msg":"Reconciler error","name":"ingress-default","namespace":"default","error":"subnets count less than minimal required count: 1 < 2"}
```

**Describe the solution you'd like**
Maybe adding a flag to turn off the check and respects the whitelist.

**Describe alternatives you've considered**
 Adding another annotation of the non-target subnet, for example, **subnet-0987654321** .

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
 name: ingress-default
 annotations:
   alb.ingress.kubernetes.io/scheme: internet-facing
   alb.ingress.kubernetes.io/target-type: ip
   alb.ingress.kubernetes.io/target-group-attributes: deregistration_delay.timeout_seconds=10
   alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}]'
   alb.ingress.kubernetes.io/backend-protocol: HTTP
   alb.ingress.kubernetes.io/subnets: subnet-123456789,subnet-0987654321
```

M00nF1sh:
we can add a new featureGate such `ALBSingleSubnet`, once enabled, the minimal count for [ALB can be one here](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/networking/subnet_resolver.go#L357).

/kind good-first-issue

Sajiyah-Salat:
I would like to take this up. I got the issue. but I am confused as what would you like to see in ``ALBSingleSubnet``.  also I am new to this so if there would be any resources which help me to do so would be welcome.
/assign

hitsub2:
It is a legacy issue that we got the single AZ for ALB to acquire low latency for our workfload. Maybe it is not best choice to solve latecy issue behind ALB, but migration from Single to Different is quite not possible for now.

PR #3437, opened 2023-10-18
Feature alb subnets

jerryhe1999:
### Issue

#3082 

### Description

Added a new feature gate named "ALBSingleSubnet" with default value false, once it set to true, the user who get whitelisted by AWS ELB team for using only one subnet for their application load balancer could be processed as expected.

Manual Test Items:

For account which doesn't get whitelisted, creating the ingress resource with only one subnet attached. 
Comes up with LBC error message while ALBSIngleSubnet is set to false.
Comes up with ELB error message while ALBSIngleSubnet is set to true.

For account which get whitelisted, creating the ingress resource with only one subnet attached. 
Comes up with LBC error message while ALBSIngleSubnet is set to false.
ALB is successfully provisioning while ALBSIngleSubnet is set to true.

### Checklist
- [X] Manually tested

### DIFF
--- a/pkg/networking/subnet_resolver.go
+++ b/pkg/networking/subnet_resolver.go
@@ -373,10 +373,7 @@ func (r *defaultSubnetsResolver) validateSubnetsMinimalCount(subnets []*ec2sdk.S
 // computeSubnetsMinimalCount returns the minimal count requirement for subnets.
 func (r *defaultSubnetsResolver) computeSubnetsMinimalCount(subnetLocale subnetLocaleType, resolveOpts SubnetsResolveOptions) int {
 	minimalCount := 1
-	if resolveOpts.ALBSingleSubnet && resolveOpts.LBType == elbv2model.LoadBalancerTypeApplication {
-		return minimalCount
-	}
-	if resolveOpts.LBType == elbv2model.LoadBalancerTypeApplication && subnetLocale == subnetLocaleTypeAvailabilityZone {
+	if resolveOpts.LBType == elbv2model.LoadBalancerTypeApplication && subnetLocale == subnetLocaleTypeAvailabilityZone && !resolveOpts.ALBSingleSubnet {
 		minimalCount = 2
 	}
 	return minimalCount
--- a/pkg/networking/subnet_resolver.go
+++ b/pkg/networking/subnet_resolver.go
@@ -48,7 +48,7 @@ type SubnetsResolveOptions struct {
 	AvailableIPAddressCount int64
 	// whether to check the cluster tag
 	SubnetsClusterTagCheck bool
-	// Disable subnet minimal count restriction
+	// whether to allow using only 1 subnet for provisioning ALB, default to false
 	ALBSingleSubnet bool
 }
 
@@ -97,7 +97,7 @@ func WithSubnetsClusterTagCheck(SubnetsClusterTagCheck bool) SubnetsResolveOptio
 	}
 }
 
-// WithALBSingleSubnet generate an option that foncigure ALBSingleSubnet
+// WithALBSingleSubnet generate an option that configures ALBSingleSubnet
 func WithALBSingleSubnet(ALBSingleSubnet bool) SubnetsResolveOption {
 	return func(opts *SubnetsResolveOptions) {
 		opts.ALBSingleSubnet = ALBSingleSubnet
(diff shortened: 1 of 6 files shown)


pkg/networking/subnet_resolver.go
```diff
@@ -48,6 +48,8 @@ type SubnetsResolveOptions struct {
 	AvailableIPAddressCount int64
 	// whether to check the cluster tag
 	SubnetsClusterTagCheck bool
+	// Disable subnet minimal count restriction
```
oliviassss, line 51: Maybe reword: whether to allow using only 1 subnet for provisioning ALB, default to `false`

pkg/networking/subnet_resolver.go
```diff
@@ -95,6 +97,13 @@ func WithSubnetsClusterTagCheck(SubnetsClusterTagCheck bool) SubnetsResolveOptio
 	}
 }
 
+// WithALBSingleSubnet generate an option that foncigure ALBSingleSubnet
```
oliviassss, line 100: nit: foncigure or configure?

pkg/networking/subnet_resolver.go
```diff
@@ -364,6 +373,9 @@ func (r *defaultSubnetsResolver) validateSubnetsMinimalCount(subnets []*ec2sdk.S
 // computeSubnetsMinimalCount returns the minimal count requirement for subnets.
 func (r *defaultSubnetsResolver) computeSubnetsMinimalCount(subnetLocale subnetLocaleType, resolveOpts SubnetsResolveOptions) int {
 	minimalCount := 1
+	if resolveOpts.ALBSingleSubnet {
```
oliviassss, line 376: Maybe add a check that the LBType is `elbv2model.LoadBalancerTypeApplication`

oliviassss:
Thanks for the contribution. Please also add:
1. doc updates regarding the new feature gate flag: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/deploy/configurations.md#feature-gates
2. doc updates re subnet discovery: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/deploy/subnet_discovery.md#subnet-auto-discovery
3. Please add the manual tests in the PR descriptions.

pkg/networking/subnet_resolver.go
jerryhe1999: I am not sure whether is necessary to add the check of elbv2model.LoadBalancerTypeApplication here since if the type is NLB the minimalCount will be returned with 1 anyway, but it might make this code change more readable?

pkg/networking/subnet_resolver.go
```diff
@@ -364,6 +373,9 @@ func (r *defaultSubnetsResolver) validateSubnetsMinimalCount(subnets []*ec2sdk.S
 // computeSubnetsMinimalCount returns the minimal count requirement for subnets.
 func (r *defaultSubnetsResolver) computeSubnetsMinimalCount(subnetLocale subnetLocaleType, resolveOpts SubnetsResolveOptions) int {
 	minimalCount := 1
+	if resolveOpts.ALBSingleSubnet && resolveOpts.LBType == elbv2model.LoadBalancerTypeApplication {
+		return minimalCount
+	}
 	if resolveOpts.LBType == elbv2model.LoadBalancerTypeApplication && subnetLocale == subnetLocaleTypeAvailabilityZone {
```
johngmyers, line 379: Would be much more straightforward as:

```suggestion
	if resolveOpts.LBType == elbv2model.LoadBalancerTypeApplication && subnetLocale == subnetLocaleTypeAvailabilityZone && !resolveOpts.ALBSingleSubnet {
```

johngmyers:
Is there any particular reason this needs to be feature-flagged?

johngmyers:
To answer my own question, I think this is so that LBC can give better and earlier error messages in the common case where the ALB service requires a minimum of 2.

oliviassss:
@johngmyers, it's an opt-in feature from ELB side. They are working on lifting the minimal subnet constraint for ALB. AFAIK, in some region like KIX they only require 1 subnet for ALB, but most regions still require 2 subnets as minimal unless the account is allowlisted.

pkg/networking/subnet_resolver.go
jerryhe1999: Thank you for reviewing @johngmyers, the change has been made and submitted.

PR #3467, opened 2023-11-02
Drop the Security disclosures that point to AWS security

dims:
A community repository SHOULD NOT be pointing to specific vendor's security process


M00nF1sh (approved):
/lgtm
/approve

PR #3522, opened 2024-01-03
Add indention to forwardConfig

jacurtis:
This clarifies to the reader that `targetGroups` and `TargetGroupStickinessConfig` are both within the `forwardConfig`, which was not clear before due to lack of indentation.

### Issue

General improvement for docs readability, not related to an issue

### Description

When configuring the `forwardConfig`, it was not initially clear to me that these two properties were part of the `forwardConfig` block since they were indented to the same level as it, instead of indented a layer deeper to show nesting.

This change simply provides the extra indention in the example code of the docs which should make it clearer to other readers to prevent them from making the same mistake.

### DIFF
--- a/docs/guide/ingress/annotations.md
+++ b/docs/guide/ingress/annotations.md
@@ -462,22 +462,22 @@ Traffic Routing can be controlled with following annotations:
                 {
                     "type":"forward",
                     "forwardConfig":{
-                    "targetGroups":[
-                        {
-                        "serviceName":"service-1",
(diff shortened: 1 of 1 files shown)


AndrewCharlesHay:
LGTM ✅

oliviassss:
@AndrewCharlesHay, can you please rebase the main branch to pick up the fix to e2e tests failure?
https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3539
thanks

Issue #3499, opened 2023-11-28
support for (mTLS) Mutual Transport Layer Security


berry2012:
**Is your feature request related to a problem?**
_Originally posted by @kishorj in https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1973#issuecomment-841568062_

**Describe the solution you'd like**
I'd like to see ALB controller add support for mTLS.
The [Application Load Balancer now reliably verifies certificate-based client identities](https://aws.amazon.com/blogs/aws/mutual-authentication-for-application-load-balancer-to-reliably-verify-certificate-based-client-identities) hence the ask if this can now be effected in the ALB controller settings.

**Describe alternatives you've considered**
1. With Nginx Ingress controller by adding additional annotations to your Ingress Resource to specify Client-Certificate Authentication.
https://github.com/kubernetes/ingress-nginx/blob/main/docs/examples/auth/client-certs/ingress.yaml

2. With ALB controller provisioning an ALB after creating the ingress resource then manually editing the created ALB from EC2 console to enable the mTLS (not smooth, but it is a workaround for now). Having a way to use annotations to specify the certificate stored in secret will give better experience.

arununzer:
+1 , Can we have the support for mTLS for ALB controller as an ingress Annotations like we have it for nginx ingress controller

shraddhabang:
Thank you for bringing this up. We are currently working on the implementation to support mTLS configuration through controller. We are planning to release this in coming month in next minor release.

JCachada:
Hey 😄 No pressure, but is this still planned for release in the next minor? Have there been any updates to the plan in the meantime?

kangadrewie:
> Hey 😄 No pressure, but is this still planned for release in the next minor? Have there been any updates to the plan in the meantime?

+1 -- I see the ticket is still in todo, this would be a great addition

arununzer:
Hello Folks, Any idea on when it will be released ..an appropriate release date ?

oliviassss:
@JCachada, @kangadrewie, @arununzer, hi all, I think the team are working on releasing this feature with v2.7.0, targetting on mid Jan. Thanks

suman-ganta:
Hi @oliviassss, is v2.7.0 still on track to release this week?

shraddhabang:
@suman-ganta I have raised a PR for this feature. The release for v2.7.0 is set by end of January. Thank you for your patience.

[...]

PR #3532, opened 2024-01-17
Implement mutual TLS authentication support for Ingress

shraddhabang:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3499
### Description

Support for (mTLS) Mutual Transport Layer Security on Ingress through AWS LB Controller

This PR adds the support for mutual TLS authentication support for Ingress. Its delivers mTLS feature by integrating the Trust Stores into listener management. The customer will be able to set the mTLS mode and will be able to provide the existing Trust Store Name/ARN (they have created through CLI/Console) through new annotations for Ingress.

#### New Ingress Annotation 

```
alb.ingress.kubernetes.io/mutual-authentication
Type: string
Default Value: '[{"port": 443, "mode": "off"}]'
# port: listen port
# mode: "off" (deafult) | "passthrough" | "verify"
# trustStore: "arn:aws:elasticloadbalancing:trustStoreArn" or "my-trust-store"// required when the mode is "verify"
# ignoreClientCertificateExpiry : true | false (default)
```
##### Example
```
# Listen port configuration
alb.ingress.kubernetes.io/listen-ports: '[{"HTTPS": 80}, {"HTTPS": 443}, {"HTTPS": 8080}, {"HTTPS": 8443}]'


# mTLS configuration
alb.ingress.kubernetes.io/mutual-authentication: '[{"port": 80, "mode": "passthrough"}, 
                                    {"port": 443, "mode": "verify", "trustStore": "arn:aws:elasticloadbalancing:trustStoreArn", "ignoreClientCertificateExpiry" : true}]'
```
We will identify the HTTPS listener based on its port and set the specified the mTLS configuration. So in the above example, we specified four HTTPS listeners with ports 80, 443, 8080, 8443. As per the annotation, the listener HTTPS:80 will be set to "passthrough” mode, the listener HTTPS:443 will be set to “verify” mode and will be associated with the provided trust-store-arn “arn:aws:elasticloadbalancing:trustStoreArn”. The remaining listeners HTTPS:8080 and HTTPS:8443 will be set to default mTLS mode i.e “off”.  

#### New load balancer attributes for  enabling connection logs  for Ingresses and IngressGroup
We are also adding the support for connection logs support similar to what we support for access logs today for ALB. 
##### Example on Ingress annotation
```
alb.ingress.kubernetes.io/load-balancer-attributes: connection_logs.s3.enabled=true,connection_logs.s3.bucket=my-connection-log-bucket,access_logs.s3.prefix=my-app
```

##### Example on IngressClassParams specification
```
kind: IngressClassParams
metadata:
    name: awesome-class
spec:
  loadBalancerAttributes:
  - key: connection_logs.s3.enabled
    value: "true"
  - key: connection_logs.s3.bucket
    value: "my-connection-logs"  
  - key: connection_logs.s3.prefix
    value: "myapp"
```

More : https://docs.aws.amazon.com/elasticloadbalancing/latest/application/mutual-authentication.html

### Checklist
- [x] Added tests that cover your change (if possible) -  Added some unit tests
- [ ] Added/modified documentation as required (such as the `README.md`, or the `docs` directory) - Will raise separate PR for this
- [x] Manually tested
   * Verified the config set on ingress annotation for mTLS is reflected on the associated ALB for all the modes while creating and modifying listener
   * Verified the trustStore name is accepted through ingress annotation and controller was able to fetch the associated trust-store-arn and create and modify the listener properly
   * Verified that the ingress class param is accepted and given precedence over ingress annotation
   * Manually tested input validations
   * Verified the new connection logs attribute is being updated through ingress annotation


go.mod
```diff
@@ -1,9 +1,9 @@
 module sigs.k8s.io/aws-load-balancer-controller
 
-go 1.20
+go 1.21.5
```
oliviassss, line 3: Need to keep it in minor version, eg `1.21`

pkg/ingress/model_build_listener.go
```diff
@@ -247,3 +257,199 @@ func (t *defaultModelBuildTask) computeIngressExplicitSSLPolicy(_ context.Contex
...
+
+	if ing.IngClassConfig.IngClassParams != nil && ing.IngClassConfig.IngClassParams.Spec.MutualAuthentication != nil {
+		rawMtlsConfigStringFromIngressClassParam = string(*ing.IngClassConfig.IngClassParams.Spec.MutualAuthentication)
+		fromIngressClassParams = true
+	}
+	// If both Ingress and IngressClassParam is missing mutualAuthentication config, return default mutualAuthentication mode
+	if exists := t.annotationParser.ParseStringAnnotation(annotations.IngressSuffixMutualAuthentication, &rawMtlsConfigString, ing.Ing.Annotations); !exists {
+		if !fromIngressClassParams {
+			return map[int64]*elbv2model.MutualAuthenticationAttributes{443: {
+				Mode: string(elbv2model.MutualAuthenticationOffMode),
```
oliviassss, line 281: What if the users do not have port 443 in the listeners (like they have port 8080), and neither do they specify the mtls setting?

oliviassss: also need to call out the default mtls in the ingress annotation and ingress class param doc.

shraddhabang: This will be ignored then while building the Listener SDK request since we only map the port and their respective setting in mTLS.

apis/elbv2/v1beta1/ingressclassparams_types.go
```diff
@@ -107,6 +109,10 @@ type IngressClassParamsSpec struct {
 	// +optional
 	SSLPolicy string `json:"sslPolicy,omitEmpty"`
 
+	// MutualAuthentication specifies the mutual TLS authentication config for all Ingresses that belong to IngressClass with this IngressClassParams.
+	// +optional
+	MutualAuthentication *MutualAuthentication `json:"mutualAuthentication,omitempty"`
```
M00nF1sh, line 114: type MutualAuthentication struct {
 Mode ...
 TrustStoreArn...
}
type IngressClassParamsSpec struct {
	....

	// MutualAuthentication specifies the mutual TLS authentication config for all Ingresses that belong to IngressClass with this IngressClassParams.
	// +optional
	MutualAuthentication *MutualAuthentication `json:"mutualAuthentication,omitempty"`
}

shraddhabang: As discussed, removed the support for IngressClassParam from this release. We will add a documentation update which will specify that warning to use mTLS config on IngressGroups. Thanks for the review. 
We will add the IngressClassParam support for mTLS in follow up patch release.

M00nF1sh (commented):
/lgtm
/approve

PR #3567, opened 2024-02-09
feat: allowed ACM cert discovery to filter on CA ARNs (#3565)

### LINKED ISSUE #3565: CertDiscovery: Ability to restrict/filter available certs
Our intent is to replace the certificates with newly issued ones from a different PCA issuer (centrally managed via Terraform).
I already read through https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1598 to clarify what options we have.

I think it would be helpful if we can enhance this function here with some filtering methods:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/32a3339f7d2a48897a5548424a2859c583b0428b/pkg/ingress/cert_discovery.go#L110-L132

I am thinking about filtering for the `CertificateAuthorityArn`(s) as we have a new CA. Another option would be to filter for `Issued at` newer than `<date XYZ>`.

Maybe something like this
```go
		if len(allowedCertificateAuthorityArns) > 0 {
			resp, err := acmClient.DescribeCertificate(&acm.DescribeCertificateInput{
				CertificateArn: certSummary.CertificateArn,
			})
			if err == nil && slices.Contains(allowedCertificateAuthorityArns, aws.StringValue(resp.Certificate.CertificateAuthorityArn)) {
				certARN := aws.StringValue(certSummary.CertificateArn)
				certARNs = append(certARNs, certARN)
			}
		} else {
			certARN := aws.StringValue(certSummary.CertificateArn)
			certARNs = append(certARNs, certARN)
		}
```

(`allowedCertificateAuthorityArns` is type `[]string{}`)

Downside of this is that the `CertificateAuthorityArn` is not part of the summary and requires an extra API call to get the details of the certificate. But the cert details are already retrieved inside the func `loadDomainsForCertificate`, so maybe we can cache these in an efficient way.

Do the migration via the already available option (using multiple certificates mentioned above)

/cc: @the-technat, @blanchardma, @aescrob, @asdfgugus, @js-315385995

the-technat:
### Issue

Resolves #3565

### Description

This PR adds a new flag that is passed to the ACMCertDiscovery type in order to filter auto-discovered certificates for a list of CA ARNs. This is usefull in situations where one needs to switch PCA while ensuring the aws-load-balancer-controller should explicitly pick the new CA to allow the old one to be removed (which is not possible when certificates are in use).

### Checklist
- [ ] Added tests that cover your change (if possible) -> there's no mock for the AWS ACM api in the code right now
- [x] Manually tested


the-technat:
Any news on this?

mkilchhofer (approved):

pkg/ingress/cert_discovery.go
```diff
@@ -102,7 +106,10 @@ func (d *acmCertDiscovery) loadDomainsForAllCertificates(ctx context.Context) (m
 		if err != nil {
 			return nil, err
 		}
-		domainsByCertARN[certARN] = certDomains
+		if len(certDomains.List()) > 0 {
```
M00nF1sh, line 109: you can directly use `len(certDomains)` which is more efficient than convert it into list.

pkg/ingress/cert_discovery.go
```diff
@@ -143,14 +150,20 @@ func (d *acmCertDiscovery) loadDomainsForCertificate(ctx context.Context, certAR
...
 	certDetail := resp.Certificate
-	domains := sets.NewString(aws.StringValueSlice(certDetail.SubjectAlternativeNames)...)
-	switch aws.StringValue(certDetail.Type) {
-	case acm.CertificateTypeImported:
-		d.certDomainsCache.Set(certARN, domains, d.importedCertDomainsCacheTTL)
-	case acm.CertificateTypeAmazonIssued, acm.CertificateTypePrivate:
-		d.certDomainsCache.Set(certARN, domains, d.privateCertDomainsCacheTTL)
+
+	// check if cert is issued from an allowed CA
+	if len(d.allowedCAARNs) == 0 || slices.Contains(d.allowedCAARNs, *certDetail.CertificateAuthorityArn) {
```
M00nF1sh, line 155: will `certDetail.CertificateAuthorityArn` guaranteed to be `non-nil`? let's use `awssdk.StringValue(certDetail.CertificateAuthorityArn)` to be safe.

pkg/ingress/cert_discovery.go
the-technat: thx for the hint, I updated it

the-technat:
@M00nF1sh, @mkilchhofer somehow I accidentially closed this PR and mixed up git. Please refer to 3591.

PR #3479, opened 2023-11-12
feat: Support TargetGroupBinding on targets outside the cluster's VPC

ikosenn:
### Issue

fixes https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3084

### Description
This PR adds VPC ID to the TargetGroupBinding Spec so as to allow registration in target groups that are created with in a VPC that is different from that which the controller is running in. 

### Checklist
- [ ] Added tests that cover your change (if possible)
- [x] Added/modified documentation as required (such as the `README.md`, or the `docs` directory)
- [x] Manually tested
- [x] Made sure the title of the PR is a good description that can go into the release notes

### BONUS POINTS checklist: complete for good vibes and maybe prizes?! :exploding_head:
- [ ] Backfilled missing tests for code in same general area :tada:
- [ ] Refactored something and made the world a better place :star2:

PR #3570, opened 2024-02-11
fix: targetGroupBinding now works if the LBs is in a different VPCs than the POD

### LINKED ISSUE #3462: Targetgroupbinding doesn't register target pods in a peered VPC
Describe the bug
I created a target group in VPC 1 where my current load balancer is. My target pods are in a different VPC hosting the k8s cluster. When I created the targetgroupbinding, the pods are not registered. When I describe the targetgroupbinding, there's no event logged. Is this a missing feature in the CRD or did I do something wrong here?

I've tried this in the same VPC and it worked as intended.

Steps to reproduce
Create a target group in a VPC that's different from the k8s VPC.
Create the targetgroupbinding.

Expected outcome
The pods would be registered as targets under the target group.

marcosdiez:
closes #3462

Details on the solution is here: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3462#issuecomment-1935877452

*** VERY IMPORTANT NOTE ***

for this to work, one need to add one extra IAM permission to the ALB controller: `elasticloadbalancing:DescribeLoadBalancers`

One can test the implementation using the following docker container: `marcosdiez/aws-load-balancer-controller:v2.7.1-m1`

### DIFF
--- a/pkg/targetgroupbinding/resource_manager.go
+++ b/pkg/targetgroupbinding/resource_manager.go
@@ -51,6 +51,7 @@ func NewDefaultResourceManager(k8sClient client.Client, elbv2Client services.ELB
 	networkingManager := NewDefaultNetworkingManager(k8sClient, podENIResolver, nodeENIResolver, sgManager, sgReconciler, vpcID, clusterName, endpointSGTags, logger, disabledRestrictedSGRulesFlag)
 	return &defaultResourceManager{
 		k8sClient:         k8sClient,
+		elbv2Client:       elbv2Client,
 		targetsManager:    targetsManager,
 		endpointResolver:  endpointResolver,
 		networkingManager: networkingManager,
@@ -69,6 +70,7 @@ var _ ResourceManager = &defaultResourceManager{}
 // default implementation for ResourceManager.
 type defaultResourceManager struct {
 	k8sClient         client.Client
+	elbv2Client       services.ELBV2
 	targetsManager    TargetsManager
 	endpointResolver  backend.EndpointResolver
 	networkingManager NetworkingManager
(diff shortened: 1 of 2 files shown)


pkg/targetgroupbinding/resource_manager.go
```diff
@@ -405,14 +400,73 @@ func (m *defaultResourceManager) registerPodEndpoints(ctx context.Context, tgARN
...
+		if !networking.IsIPWithinCIDRs(podIP, albCIDRs) {
+			m.logger.Info(fmt.Sprintf("On %s, the podIP %s does not belong to any of it's LB's CIDRs %v, hence we are setting AvailabilityZone=all", tgARN, podIP, albCIDRs))
 			target.AvailabilityZone = awssdk.String("all")
 		}
 		sdkTargets = append(sdkTargets, target)
 	}
 	return m.targetsManager.RegisterTargets(ctx, tgARN, sdkTargets)
 }
 
+func (m *defaultResourceManager) getAlbVpcCidr(ctx context.Context, tgARN string) ([]netip.Prefix, error) {
```
oliviassss, line 413: Thanks for the contribution. My concern is for every reconciliation on IP target group, this PR adds 2 ELBv2 API calls, `DescribeTargetGroupsAsList` and `DescribeLoadBalancersAsList`. If the user has a large scale of resources, this may significantly increase the API calls or controller reconciliation latency during the situation like controller restart/auto-reconcile.

oliviassss:
@marcosdiez, hi, would this PR work for your use case too?
https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3479

marcosdiez:
Hello @oliviassss . #3479 is similar to this PR. More complex, but does less AWS API calls, which is great. That being said, after writing this PR I found out I need to manipulate an ALB in a different AWS account and that VPC sharing is not an option. So... #3479 does replace this PR but neither solve my new problem :/

Issue #3505, opened 2023-12-04
Lack of ExternalName Service support


maciejmaz:
I'm creating this feature request to reopen the issue described here: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1223

**Is your feature request related to a problem?**
When you set up ALB Ingress and the service linked with this ingress is `ExternalName` type, the Target Group in AWS doesn't have any targets registered and the ingress returns HTTP 503 error.

**Describe the solution you'd like**
Adding an ingress with an `ExternalName` service type should properly route traffic to the service.
This would allow creating cross-namespace relations between Ingress and the services.

Example: 
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx-pod
  namespace: namespace1
  labels:
    app: nginx
spec:
  containers:
  - name: nginx-container
    image: nginx:latest
    ports:
    - containerPort: 80
      protocol: TCP

---

kind: Service
apiVersion: v1
metadata:
  name: nginx-service
  namespace: namespace1
spec:
  selector:
    app: nginx
  type: ClusterIP
  ports:
  - port: 80

---

kind: Service
apiVersion: v1
metadata:
  name: nginx-service-external
  namespace: namespace2
spec:
  type: ExternalName
  externalName: nginx-service.namespace1.svc.cluster.local

---

apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test-ingress
  namespace: namespace2
  annotations:
    kubernetes.io/ingress.class: alb
    alb.ingress.kubernetes.io/scheme: internet-facing
    alb.ingress.kubernetes.io/target-type: ip
    alb.ingress.kubernetes.io/group.name: ingress
    alb.ingress.kubernetes.io/certificate-arn: <placeholder>
    alb.ingress.kubernetes.io/listen-ports: '[{"HTTPS":443}]'
spec:
  rules:
    - host: example.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: nginx-service-external
                port:
                  number: 80
```

Should result hosting nginx from namespace1 on ingress deployed in namespace2

**Describe alternatives you've considered**
1. Switching to Nginx ingress controller
2. Adding some nginx pod doing the proxy_pass as suggested here: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1223#issuecomment-652510197

yalattas:
[...]

PR #3571, opened 2024-02-12
feat: adding support for creating targets from ExternalName

msvticket:
### Issue

 #3505

### Description

I added support for ip target groups when service is of type ExternalName.

I have not updated any documentation, but I'm happy to do that in case you want the code change to begin with.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/backend/endpoint_resolver.go
+++ b/pkg/backend/endpoint_resolver.go
@@ -3,15 +3,23 @@ package backend
 import (
 	"context"
 	"fmt"
+	"math"
+	"reflect"
+	"sort"
+	"sync"
+	"time"
 
 	awssdk "github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/go-logr/logr"
+	"github.com/miekg/dns"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	discovery "k8s.io/api/discovery/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	elbv2api "sigs.k8s.io/aws-load-balancer-controller/apis/elbv2/v1beta1"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/k8s"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -27,6 +35,8 @@ type EndpointResolver interface {
 	// ResolveNodePortEndpoints will resolve endpoints backed by nodePort.
 	ResolveNodePortEndpoints(ctx context.Context, svcKey types.NamespacedName, port intstr.IntOrString,
 		opts ...EndpointResolveOption) ([]NodePortEndpoint, error)
+
+	Cleanup(ctx context.Context, svcKey *elbv2api.TargetGroupBinding)
 }
 
 // NewDefaultEndpointResolver constructs new defaultEndpointResolver
@@ -37,6 +47,10 @@ func NewDefaultEndpointResolver(k8sClient client.Client, podInfoRepo k8s.PodInfo
 		failOpenEnabled:      failOpenEnabled,
 		endpointSliceEnabled: endpointSliceEnabled,
 		logger:               logger,
+
+		externalNameServices:    make(map[types.NamespacedName]string),
+		externalNameReconcilers: make(map[string]externalNameReconciler),
+		externalNameMutex:       sync.Mutex{},
 	}
 }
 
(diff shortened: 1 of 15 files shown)


msvticket:
Could you have a look at this @oliviassss ?

Leen15:
any update on this?

Reyalsorik:
Any update? I would appreciate this functionality.

nousot-cloud-guy:
Is there anything that's holding up this PR that someone could help finish? This would be really useful functionality to have.

felipegouveiae:
@johngmyers @M00nF1sh please review and merge! I'm begging you!

pkg/backend/endpoint_resolver.go
```diff
@@ -27,12 +28,17 @@ var ErrNotFound = errors.New("backend not found")
 type EndpointResolver interface {
 	// ResolvePodEndpoints will resolve endpoints backed by pods directly.
 	// returns resolved podEndpoints and whether there are unready endpoints that can potentially turn ready in future reconciles.
-	ResolvePodEndpoints(ctx context.Context, svcKey types.NamespacedName, port intstr.IntOrString,
-		opts ...EndpointResolveOption) ([]PodEndpoint, bool, error)
+	ResolvePodEndpoints(ctx context.Context, svckey types.NamespacedName, svc *corev1.Service, port intstr.IntOrString, opts ...EndpointResolveOption) ([]IpEndpoint, bool, error)
```
ryantate13, line 31: There's no need for a breaking API change here. The variadic list of `EndpointResolveOption` argument should eliminate that necessity, can just create a new one to the effect of `WithService(*corev1.Service)`. This package is currently at v2, merging this would require tagging a v3 and that creates more churn for adoption than necessary.

pkg/backend/endpoint_types.go
```diff
@@ -12,18 +12,18 @@ type Endpoint interface {
 	GetIdentifier(includeTimestamp bool) string
 }
 
-// An endpoint provided by pod directly.
-type PodEndpoint struct {
+// IpEndpoint is an endpoint for an ip address
+type IpEndpoint struct {
```
ryantate13, line 16: Name here should remain stable

pkg/backend/endpoint_types.go
```diff
@@ -12,18 +12,18 @@ type Endpoint interface {
...
 	// Pod's IP.
 	IP string
 	// Pod's container port.
 	Port int32
 	// Pod that provides this endpoint.
-	Pod k8s.PodInfo
+	Pod *k8s.PodInfo
```
ryantate13, line 22: This needs to remain backwards compatible

ryantate13 (changes requested):
I'm excited about this functionality getting built. I do think we need to be cognizant of dependents on this package and maintain a stable API surface for v2 however. Seems like some simple fix ups should get is back into that territory.

msvticket:
> I'm excited about this functionality getting built. I do think we need to be cognizant of dependents on this package and maintain a stable API surface for v2 however. Seems like some simple fix ups should get is back into that territory.

I don't mind putting some more time in this PR if it could be merged. But since this PR is ignored by the maintainers I don't see any point in spending that time now.

pkg/backend/endpoint_types.go
msvticket: Why?

pkg/backend/endpoint_types.go
msvticket: Why?

deens-cyera:
Hey, any update in the pr ?

chris-codaio:
Another vote to get this reviewed & committed - just ran into this limitation this week!

pkg/ingress/model_build_target_group.go
```diff
@@ -228,12 +228,16 @@ func (t *defaultModelBuildTask) buildTargetGroupName(_ context.Context,
...
-func (t *defaultModelBuildTask) buildTargetGroupTargetType(_ context.Context, svcAndIngAnnotations map[string]string, classCfg ClassConfiguration) (elbv2model.TargetType, error) {
+func (t *defaultModelBuildTask) buildTargetGroupTargetType(_ context.Context, svcAndIngAnnotations map[string]string, classCfg ClassConfiguration, svc *corev1.Service) (elbv2model.TargetType, error) {
 	rawTargetType := string(t.defaultTargetType)
 	_ = t.annotationParser.ParseStringAnnotation(annotations.IngressSuffixTargetType, &rawTargetType, svcAndIngAnnotations)
 	if classCfg.IngClassParams != nil && classCfg.IngClassParams.Spec.TargetType != "" {
 		rawTargetType = string(classCfg.IngClassParams.Spec.TargetType)
 	}
+	if svc.Spec.Type == corev1.ServiceTypeExternalName && rawTargetType != string(elbv2model.TargetTypeIP) {
+		t.logger.Info("Target type will be ip since service is an ExternalName", "service", k8s.NamespacedName(svc))
+		rawTargetType = string(elbv2model.TargetTypeIP)
```
zac-nixon, line 266: I'd recommend throwing an error rather then overwriting this value.

msvticket: I think returning an error if something else than ip is set in an annotation makes sense, but still default to ip. What do you think about that?

msvticket: I have now pushed a commit with that change

zac-nixon:
Hi. I'd like to update everyone on the conversation we had in Slack.

```
Zac Nixon
  [1:55 PM](https://kubernetes.slack.com/archives/D09ARVB64TF/p1755723355218379)
I reviewed the PR, the code your wrote makes sense but I have a concern. How does the controller know that the DNS record has been updated, so that it can update the target group? Right now, you're relying the sync time (~10 hours) to do the DNS updates. It seems that we would need some background process that is querying dns periodically and enqueuing events when the record changes.


Mårten Svantesson
  [8:51 AM](https://kubernetes.slack.com/archives/D09ARVB64TF/p1755791499599059)
Well, my thought is that this is an initial implementation that I tried to keep as simple as possible. You are right in that it probably wont suffice for everybody, but I think it is better to base improvements on feedback rather than guesses.


Zac Nixon
  [7:53 AM](https://kubernetes.slack.com/archives/D09ARVB64TF/p1755874427108159)
Yeah, I see what you're saying. I really appreciate the PR, but generally we don't accept features that aren't complete. Not having a way to timely update the LoadBalancer when the DNS entries changes seems like a big problem. Can you explain how this feature helps you out?


Mårten Svantesson
  [7:09 AM](https://kubernetes.slack.com/archives/D09ARVB64TF/p1756130961627789)
I use it to implement this solution: https://aws.amazon.com/blogs/networking-and-content-delivery/hosting-internal-https-static-websites-with-alb-s3-and-privatelink/
By using an ExternalName for s3 I don’t need a separate ALB but can use the same that is used for accessing internal application running in EKS.
Amazon Web ServicesAmazon Web Services
[Hosting Internal HTTPS Static Websites with ALB, S3, and PrivateLink | Amazon Web Services](https://aws.amazon.com/blogs/networking-and-content-delivery/hosting-internal-https-static-websites-with-alb-s3-and-privatelink/)
Amazon Simple Storage Service (Amazon S3) is a powerful platform that enables you to do various tasks. One notable feature is the ability to create a bucket with an FQDN, point an alias record to the bucket website endpoint, and immediately get up-and-running with an HTTP static website. If you want to serve HTTPS traffic […]
Dec 30th, 2022
https://aws.amazon.com/blogs/networking-and-content-delivery/hosting-internal-https-static-websites-with-alb-s3-and-privatelink/



Zac Nixon
  [8:36 AM](https://kubernetes.slack.com/archives/D09ARVB64TF/p1756136188696239)
Thanks for the link. I suspect that this works for your case because AFAIK, Private Link uses a static IP address.
[8:37](https://kubernetes.slack.com/archives/D09ARVB64TF/p1756136238044719)
Generally, we would like to supply more general solutions, this one has a pretty sharp edge where Target Group changes aren't propagated very fast.
```

TL;DR - There is some more work to be done here to accept this change. I don't really have the time until ~October to put more stuff on my plate.

msvticket:
@zac-nixon I have now added support for regularly checking DNS (based on TTL). I have also removed any non-essential modifications. Please review!

pkg/backend/endpoint_resolver.go
```diff
@@ -336,3 +382,173 @@ func convertCoreEndpointAddressToDiscoveryEndpoint(endpoint corev1.EndpointAddre
 	}
 	return ep
 }
+
+func (r *defaultEndpointResolver) watchExternalName(ctx context.Context, ttl uint32, svc *corev1.Service, port intstr.IntOrString) {
```
zac-nixon: This probably warrants some abstraction and tests.

pkg/backend/endpoint_resolver.go
```diff
@@ -43,6 +52,8 @@ func NewDefaultEndpointResolver(k8sClient client.Client, podInfoRepo k8s.PodInfo
 		failOpenEnabled:      failOpenEnabled,
 		endpointSliceEnabled: endpointSliceEnabled,
 		logger:               logger,
+
+		externalNameReconcilers: make(map[string]context.CancelFunc),
```
zac-nixon: We'll need a mutex here. It's possible that there are threads competing to mutate this.

pkg/backend/endpoint_resolver.go
```diff
@@ -336,3 +382,173 @@ func convertCoreEndpointAddressToDiscoveryEndpoint(endpoint corev1.EndpointAddre
...
+	case <-ctx.Done():
+		err = ctx.Err()
+	case _ = <-ticker.C:
+		var minTTL uint32
+		minTTL, err = r.reconcileExternalNameEndpointSlice(ctx, svc, port)
+		if minTTL == math.MaxUint32 || minTTL < 10 {
+			// Try again in 10 seconds if no valid TTL was found. Don't query in less than 10 seconds to avoid churn
+			minTTL = 10
+		}
+		go r.watchExternalName(ctx, minTTL, svc, port)
```
zac-nixon: IMO, we'll want to bound how many threads we spin up here. It would be great to reduce DNS load by preventing duplicate resolution threads if the same dns name is used.

zac-nixon:
I am curious, did you consider just a pure in-memory cache instead of using endpoint slices?

msvticket:
> I am curious, did you consider just a pure in-memory cache instead of using endpoint slices?

Yes, but that made the code way more complex.

pkg/backend/endpoint_resolver.go
msvticket: OK

pkg/backend/endpoint_resolver.go
msvticket: I have added tests

pkg/backend/endpoint_resolver.go
msvticket: OK

ksawerykarwacki:
Any chance that this will be merged ever? This PR is open for over two years, majority fo people just opted for dual ingress approach with alb pointing to nginx and nginx pointing to service. WIth EOL of nginx ingress controller next month, there will be no better time to introduce this feature.

PR #3575, opened 2024-02-13
feat: enable/disable automountServiceAccountToken at Pod level

darkweaver87:
### Description

Some people may want to disable mounting API credentials at ServiceAccount level but still want controller Pod to have automatically access to resources defined in RBACs. 
This PR makes it customizable.

### Checklist
- [x] Manually tested


helm/aws-load-balancer-controller/values.yaml
```diff
@@ -388,3 +388,6 @@ enableServiceMutatorWebhook: true
 
 # serviceTargetENISGTags specifies AWS tags, in addition to the cluster tags, for finding the target ENI SG to which to add inbound rules from NLBs.
 serviceTargetENISGTags:
+
+# Automount API credentials for a Service Account at Pod level.
+automountServiceAccountToken: true
```
oliviassss, line 393: I would have the default value as `false` to avoid changing existing user behavior.

darkweaver87: OK :-) done

oliviassss: Sorry, I've taken a further look here, looks like the `automountServiceAccountToken` is an opt-out for both pod and SA: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#opt-out-of-api-credential-automounting. And if `automountServiceAccountToken` is specified both on pod and SA level, the pod level one will take precedence. 

I'm just trying to avoid breaking change to existing users here. Do you mind changing the `automountServiceAccountToken` at pod level to an if condition? If it's specified in the values.yaml, we populate in the manifest, otherwise leave it blank. And in the `values.yaml` let's just comment out this entry.

darkweaver87: Yes you're right. I applied your recommendation without thinking. The default in kubernetes, if not explicitly set, is `true` for both `Pod` and `ServiceAccount`.
Usually I set it to `false` at `ServiceAccount` level to avoid automatic mount and `true` at `Pod` level to allow pods which really need it. It's particularly useful for the default service account in a namespace.
Some security tools check for this: https://hub.armosec.io/docs/c-0034.
Actually, at first, I wanted to make it not configurable. What do you think about removing the value and just set it hardly to  `false` at `ServiceAccount` level  and `true` at `Pod` level ? It won't be breaking in anyway except if people is using `aws-load-balancer-controller` `ServiceAccount` for another purpose that this helm chart ...
Or maybe another option might be to have a `setAutomountServiceAccountToken` which if is `true` set it to the values I was mentioning previously.

darkweaver87: @oliviassss do you have an opinion on this ?

PR #3585, opened 2024-02-20
subnet discovery: new flag to enable prioritization by availability-zone locale

alebedev87:
### Issue

N/A

### Description
Draft PR to gather opinions about adding a flag to enable different algorithms for the subnet selection from multiple qualified ones when they all come from the same availability zone. The sorting by subnet IDs remains the default. A new sorting alogrithm `azfirst` is added to prioritize `availability-zone` locale over other locales. This change may help to reduce the random aspect of the subnet selection when multiple locales are used in a cluster.

Example of where this can be helpful:
- AZ1: has availability-zone an outpost subnets, all tagged with cluster and role tags, so all eligible for the discovery
- AZ2: has only availability-zone subnets also tagged for the subnet discovery
- The discovery may fail with "subnets in multiple locales" error if outposts subnets from AZ1 have lower subnet IDs
 
### Checklist
- [ ] Manually tested


alebedev87:
@M00nF1sh  @oliviassss: what do you think about this proposal?

Issue #3386, opened 2023-09-13
Support for optionally enforcing NLB security groups on PrivateLink traffic


omallo:
**Is your feature request related to a problem?**
I would like to be able to leverage [this AWS feature](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-security-groups.html#update-security-settings) which allows me to configure whether NLB security groups should block PrivateLink traffic or not, so that I can easily limit incoming traffic on an NLB to PrivateLink connections while blocking any other traffic. Achieving this is not easily possible without the mentioned feature, as described [here](https://medium.com/@fabian_lober/internal-server-error-api-gateway-with-network-load-balancer-92ab1aa2bd7b).

**Describe the solution you'd like**
I would like the AWS Load Balancer controller to support a new annotation on a Kubernetes LoadBalancer service which allows me to enable/disable the enforcement of NLB security groups on PrivateLink traffic. Such a boolean valued annotation would map 1:1 to the corresponding AWS functionality.

**Describe alternatives you've considered**
No satisfying alternative solution to the described problem was found.

lteixeira-dock:
Up

jawanjalShiva:
We are facing the same issue. We are using VPCLink to route traffic from apigw to my internal NLB. AWS LB Controller is creating NLB and when I use "service.beta.kubernetes.io/load-balancer-source-ranges" annotation (https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/annotations/#lb-source-ranges).  I'm using internal NLB and whitelisting apigw IPs is not helping here.  

I had to set "Enforce inbound rules on PrivateLink traffic" to "off" in NLB Security configuration to allow traffic from apigw.

Puneeth-n:
facing the same issue here

veekaly:
@oliviassss I'd like to work on this feature

shraddhabang:
@veekaly Thank you for showing interest in picking this issue? Do you have any updates on this?

lteixeira-dock:
Any update?

shraddhabang:
Delivered in v2.8.0

PR #3594, opened 2024-02-29
Add support for optionally enforcing NLB security groups on PrivateLi…

wweiwei-li:
…nk traffic

### Issue
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3386

### Description
This is a feature to support optionally enforce NLB security groups on PrivateLink traffic.  For the change, added a new annotation `aws-load-balancer-inbound-sg-rules-on-private-link-traffic` to configure whether to apply security group rules to traffic sent to the load balancer through AWS PrivateLink.  The value of it can be set to "off" and "on".  After creating new NLB, will check if the annotation is specified. If yes, will call updateSDKLoadBalancerWithSecurityGroups, where added the logic to check if we need to call setSecurityGroup to update EnforceSecurityGroupInboundRulesOnPrivateLinkTraffic.  

### Checklist
- [x] Manually tested


pkg/service/model_build_load_balancer.go
```diff
@@ -173,6 +179,22 @@ func (t *defaultModelBuildTask) buildLoadBalancerIPAddressType(_ context.Context
...
+		return "", nil
+	}
+
+	switch securityGroupsInboundRulesOnPrivateLink {
+	case string(elbv2model.SecurityGroupsInboundRulesOnPrivateLinkOn):
+		return elbv2model.SecurityGroupsInboundRulesOnPrivateLinkOn, nil
+	case string(elbv2model.SecurityGroupsInboundRulesOnPrivateLinkOff):
+		return elbv2model.SecurityGroupsInboundRulesOnPrivateLinkOff, nil
+	default:
+		return "", errors.Errorf("unknown securityGroupsInboundRulesOnPrivateLink status: %v", securityGroupsInboundRulesOnPrivateLink)
```
shraddhabang, line 194: We should modify this error message to give them suggestion to use the allowed values as well. Something like ``errors.Errorf("Invalid value for securityGroupsInboundRulesOnPrivateLink status: %v, value must be one of [%v, %v]", securityGroupsInboundRulesOnPrivateLink), string(elbv2model.SecurityGroupsInboundRulesOnPrivateLinkOn), string(elbv2model.SecurityGroupsInboundRulesOnPrivateLinkOff)``

oliviassss:
@wweiwei-li, thanks for the contribution. I left some comments. 
>The value of it can be set to "off" and "on". It is optional and if it is default to "on"
For backward compatibility, I think we shouldn't have "on" as default.
If the user doesn't specify the annotation, we skip adding any spec to their NLBs, nothing changed; if the annotation exists, we validate the value and add the spec accordingly.

pkg/model/elbv2/load_balancer.go
```diff
@@ -86,6 +87,13 @@ const (
 	IPAddressTypeDualStack IPAddressType = "dualstack"
 )
 
+type SecurityGroupsInboundRulesOnPrivateLinkStatus string
+
+const (
+	SecurityGroupsInboundRulesOnPrivateLinkOn  SecurityGroupsInboundRulesOnPrivateLinkStatus = "on"
+	SecurityGroupsInboundRulesOnPrivateLinkOff SecurityGroupsInboundRulesOnPrivateLinkStatus = "off"
+)
+
```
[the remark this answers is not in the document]
wweiwei-li: Yes, it is okay to move these to pkg/model/ec2/security_group.go.  One concern is this field is actually a load balancer field, even though it is related to security groups

pkg/service/model_build_load_balancer.go
```diff
@@ -173,6 +179,22 @@ func (t *defaultModelBuildTask) buildLoadBalancerIPAddressType(_ context.Context
 	}
 }
 
+func (t *defaultModelBuildTask) buildSecurityGroupsInboundRulesOnPrivateLink(_ context.Context) (elbv2model.SecurityGroupsInboundRulesOnPrivateLinkStatus, error) {
+	securityGroupsInboundRulesOnPrivateLink := ""
```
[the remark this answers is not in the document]
wweiwei-li: Sure, we can just use `var securityGroupsInboundRulesOnPrivateLink string`

pkg/service/model_build_load_balancer.go
```diff
@@ -173,6 +179,22 @@ func (t *defaultModelBuildTask) buildLoadBalancerIPAddressType(_ context.Context
...
+	if exists := t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixEnforceSGInboundRulesOnPrivateLinkTraffic, &securityGroupsInboundRulesOnPrivateLink, t.service.Annotations); !exists {
+		return "", nil
```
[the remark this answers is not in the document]
wweiwei-li: Good point, I agree if user does not specify the annotation, we should just skip adding it as a spec in the LB.

pkg/service/model_build_load_balancer.go
```diff
@@ -173,6 +179,22 @@ func (t *defaultModelBuildTask) buildLoadBalancerIPAddressType(_ context.Context
...
+	}
+
+	switch securityGroupsInboundRulesOnPrivateLink {
+	case string(elbv2model.SecurityGroupsInboundRulesOnPrivateLinkOn):
+		return elbv2model.SecurityGroupsInboundRulesOnPrivateLinkOn, nil
+	case string(elbv2model.SecurityGroupsInboundRulesOnPrivateLinkOff):
+		return elbv2model.SecurityGroupsInboundRulesOnPrivateLinkOff, nil
+	default:
+		return "", errors.Errorf("Invalid value for securityGroupsInboundRulesOnPrivateLink status: %v, value must be one of [%v, %v]", securityGroupsInboundRulesOnPrivateLink, string(elbv2model.SecurityGroupsInboundRulesOnPrivateLinkOn), string(elbv2model.SecurityGroupsInboundRulesOnPrivateLinkOff))
+	}
```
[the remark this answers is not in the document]
wweiwei-li: It won't allow empty string "".  An empty string will also fall into the condition of an invalid value.

pkg/deploy/elbv2/load_balancer_manager.go
```diff
@@ -75,6 +76,12 @@ func (m *defaultLoadBalancerManager) Create(ctx context.Context, resLB *elbv2mod
 		return elbv2model.LoadBalancerStatus{}, err
 	}
 
+	if resLB.Spec.Type == elbv2model.LoadBalancerTypeNetwork && string(*resLB.Spec.SecurityGroupsInboundRulesOnPrivateLink) != "" {
```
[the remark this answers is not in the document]
wweiwei-li: As you suggested if user does not specify the annotation, we will skip adding it as a spec. In this case, we will no longer have resLB.Spec.SecurityGroupsInboundRulesOnPrivateLink = "", so we can just check if it is nil.

pkg/service/model_build_load_balancer.go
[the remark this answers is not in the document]
shraddhabang: Just note : Default value for this is "On".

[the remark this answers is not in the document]
oliviassss: we need unit tests added for the cases: 1)annotation setting to "on", 
2)annotation setting to "off"
3)no annotation at all.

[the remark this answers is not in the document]
wweiwei-li: I added some code in pkg/service/model_builder_test.go.  When the annotation is set to "on", the build model will have "securityGroupsInboundRulesOnPrivateLink":"on". If the annotation is set to "off", the build model will have "securityGroupsInboundRulesOnPrivateLink":"off". If there is no annotation, "securityGroupsInboundRulesOnPrivateLink" will not be included. Is this sufficient?

pkg/annotations/constants.go
```diff
@@ -50,39 +50,40 @@ const (
...
+	SvcLBSuffixSubnets                                   = "aws-load-balancer-subnets"
+	SvcLBSuffixEIPAllocations                            = "aws-load-balancer-eip-allocations"
+	SvcLBSuffixPrivateIpv4Addresses                      = "aws-load-balancer-private-ipv4-addresses"
+	SvcLBSuffixIpv6Addresses                             = "aws-load-balancer-ipv6-addresses"
+	SvcLBSuffixALPNPolicy                                = "aws-load-balancer-alpn-policy"
+	SvcLBSuffixTargetNodeLabels                          = "aws-load-balancer-target-node-labels"
+	SvcLBSuffixLoadBalancerAttributes                    = "aws-load-balancer-attributes"
+	SvcLBSuffixLoadBalancerSecurityGroups                = "aws-load-balancer-security-groups"
+	SvcLBSuffixManageSGRules                             = "aws-load-balancer-manage-backend-security-group-rules"
+	SvcLBSuffixEnforceSGInboundRulesOnPrivateLinkTraffic = "aws-load-balancer-inbound-sg-rules-on-private-link-traffic"
```
M00nF1sh, line 88: should this be something like
```
aws-load-balancer-enforce-sg-inbound-rules-on-pl-traffic
```
or 
```
aws-load-balancer-enforce-sg-inbound-rules-on-private-link-traffic
```

pkg/deploy/elbv2/load_balancer_manager.go
```diff
@@ -75,6 +76,12 @@ func (m *defaultLoadBalancerManager) Create(ctx context.Context, resLB *elbv2mod
 		return elbv2model.LoadBalancerStatus{}, err
 	}
 
+	if resLB.Spec.Type == elbv2model.LoadBalancerTypeNetwork && resLB.Spec.SecurityGroupsInboundRulesOnPrivateLink != nil {
```
M00nF1sh, line 79: is this `EnforceSecurityGroupInboundRulesOnPrivateLinkTraffic` can only be set with this SetSecurityGroups call after LB creation?

pkg/deploy/elbv2/load_balancer_manager.go
```diff
@@ -186,20 +193,52 @@ func (m *defaultLoadBalancerManager) updateSDKLoadBalancerWithSecurityGroups(ctx
 	}
 	desiredSecurityGroups := sets.NewString(awssdk.StringValueSlice(securityGroups)...)
 	currentSecurityGroups := sets.NewString(awssdk.StringValueSlice(sdkLB.LoadBalancer.SecurityGroups)...)
-	if desiredSecurityGroups.Equal(currentSecurityGroups) {
+
+	if desiredSecurityGroups.Equal(currentSecurityGroups) && (resLB.Spec.Type != elbv2model.LoadBalancerTypeNetwork || resLB.Spec.SecurityGroupsInboundRulesOnPrivateLink == nil) {
```
M00nF1sh, line 197: is there an "Describe" API that shows the "EnforceSecurityGroupInboundRulesOnPrivateLinkTraffic" on a existing LB? 
If not, we may need to make this call on every Ingress reconcile, which feels unnecessary. 
Shall we consider use a cache to overcome it? i.e. cache the state of "EnforceSecurityGroupInboundRulesOnPrivateLinkTraffic" per LB if ELB didn't offer a describe API.

pkg/deploy/elbv2/load_balancer_manager.go
wweiwei-li: Yes, we can only set it after LB creation

pkg/annotations/constants.go
wweiwei-li: Good point,  I like `aws-load-balancer-enforce-sg-inbound-rules-on-private-link-traffic` but annotation must be 63 characters or less, so we decided to use `aws-load-balancer-inbound-sg-rules-on-private-link-traffic`

pkg/service/model_build_load_balancer.go
```diff
@@ -87,6 +92,22 @@ func (t *defaultModelBuildTask) buildLoadBalancerSpec(ctx context.Context, schem
 		LoadBalancerAttributes: lbAttributes,
 		Tags:                   tags,
 	}
+
+	if securityGroupsInboundRulesOnPrivateLink != "" {
```
M00nF1sh, line 96: i'd prefer to use below:
```
if securityGroupsInboundRulesOnPrivateLink != ""  {
   spec.SecurityGroupsInboundRulesOnPrivateLink = &securityGroupsInboundRulesOnPrivateLink
}
```

pkg/deploy/elbv2/load_balancer_manager.go
```diff
@@ -186,20 +193,41 @@ func (m *defaultLoadBalancerManager) updateSDKLoadBalancerWithSecurityGroups(ctx
...
 	desiredSecurityGroups := sets.NewString(awssdk.StringValueSlice(securityGroups)...)
 	currentSecurityGroups := sets.NewString(awssdk.StringValueSlice(sdkLB.LoadBalancer.SecurityGroups)...)
-	if desiredSecurityGroups.Equal(currentSecurityGroups) {
+
+	isEnforceSGInboundRulesOnPrivateLinkUpdated, currentEnforceSecurityGroupInboundRulesOnPrivateLinkTraffic, desiredEnforceSecurityGroupInboundRulesOnPrivateLinkTraffic := isEnforceSGInboundRulesOnPrivateLinkUpdated(resLB, sdkLB)
+	if desiredSecurityGroups.Equal(currentSecurityGroups) && !isEnforceSGInboundRulesOnPrivateLinkUpdated {
 		return nil
 	}
 
+	if !desiredSecurityGroups.Equal(currentSecurityGroups) {
```
M00nF1sh, line 202: 1. Nit, since there is only one change call technically, i'd prefer to use a single log call
var changeDescriptions []string
if !desiredSecurityGroups.Equal(currentSecurityGroups) {
   changeDescriptions = append(changeDescriptions, "changeSecurityGroups", changeSecurityGroupsDesc)
}
req := &elbv2sdk.SetSecurityGroupsInput{
		LoadBalancerArn: sdkLB.LoadBalancer.LoadBalancerArn,
		SecurityGroups:  securityGroups,
}

if ! isEnforceSGInboundRulesOnPrivateLinkUpdated {
   changeDescriptions = append(changeDescriptions, "changeEnforceSecurityGroupInboundRulesOnPrivateLinkTraffic", changeEnforceSecurityGroupInboundRulesOnPrivateLinkTrafficDesc)
   req.EnforceSecurityGroupInboundRulesOnPrivateLinkTraffic: awssdk.String(desiredEnforceSecurityGroupInboundRulesOnPrivateLinkTraffic),
}
m.logger.Info("modifying loadBalancer security group settings",
			"stackID", resLB.Stack().StackID(),
			"resourceID", resLB.ID(),
			"arn", awssdk.StringValue(sdkLB.LoadBalancer.LoadBalancerArn),
			...changeDescriptions)
2. per the current log, `req.EnforceSecurityGroupInboundRulesOnPrivateLinkTraffic` is only set when there is a change. So if the security group changed but `isEnforceSGInboundRulesOnPrivateLinkUpdated` is false, then req.EnforceSecurityGroupInboundRulesOnPrivateLinkTraffic will be nil, what's the API behavior under such case, will it untouch the existing EnforceSecurityGroupInboundRulesOnPrivateLinkTraffic settings on LB or reset it to default value(something we don't want).

wweiwei-li: 1. Yes, we can switch to use a single log call.

wweiwei-li: 2. Good question, it will untouch the existing EnforceSecurityGroupInboundRulesOnPrivateLinkTraffic settings

pkg/service/model_build_load_balancer.go
wweiwei-li: Yes, good point, will update

M00nF1sh (approved):
/lgtm
/approve

vumdao:
Hi, can we know when this ticket is released?

elaineclsouza:
Any predictions on when this will be published in the release?

VanessaDiniz:
I'm also waiting

william-araujo-dock:
When will you implement this feature?

wweiwei-li:
Hey, This feature will be released in V2.8.0 targeting the end of this week.

PR #3608, opened 2024-03-11
fix: new ca-filter causing expontentially more api-calls

the-technat:
### Issue

Sort of related to #3565 

### Description

The recently merged #3591 contained a bug where on large clusters the initial reconcile loop after startup would take hours instead of minutes (on clusters with a lot of ingresses/certificates). This is due to the missing domains-cache of certificates that were filtered by this newly introduced flag. 

This PR fixes this behaviour by setting an empty cache value for these filtered certificates. At least in our (@swisspost) testing environment this reduced duration of the reconcile loop significantly.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ingress/cert_discovery.go
+++ b/pkg/ingress/cert_discovery.go
@@ -2,7 +2,6 @@ package ingress
 
 import (
 	"context"
-	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -16,6 +15,7 @@ import (
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/util/cache"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/utils/strings/slices"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/aws/services"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/ingress/cert_discovery.go
```diff
@@ -153,18 +153,18 @@ func (d *acmCertDiscovery) loadDomainsForCertificate(ctx context.Context, certAR
...
-	return sets.String{}, nil
-
+	switch aws.StringValue(certDetail.Type) {
+	case acm.CertificateTypeImported:
+		d.certDomainsCache.Set(certARN, domains, d.importedCertDomainsCacheTTL)
+	case acm.CertificateTypeAmazonIssued, acm.CertificateTypePrivate:
+		d.certDomainsCache.Set(certARN, domains, d.privateCertDomainsCacheTTL)
+	}
+	return domains, nil
 }
```
mkilchhofer, line 168: Real diff for this feature can be viewed here:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/compare/v2.7.1...the-technat:aws-load-balancer-controller:main

Or on this screenshot:
image

M00nF1sh:
taking a look, wondering how the original PR would cause such regression when the new flag is not used given it had the `len(d.allowedCAARNs) == 0` check..

pkg/ingress/cert_discovery.go
```diff
@@ -153,18 +153,18 @@ func (d *acmCertDiscovery) loadDomainsForCertificate(ctx context.Context, certAR
 	certDetail := resp.Certificate
 
 	// check if cert is issued from an allowed CA
+	// otherwise empty-out the list of domains
+	domains := sets.String{}
 	if len(d.allowedCAARNs) == 0 || slices.Contains(d.allowedCAARNs, awssdk.StringValue(certDetail.CertificateAuthorityArn)) {
```
M00nF1sh, line 158: just for my understanding, the originally PR didn't introduce any regression right?
And you faced this issue due to you used the new "allowedCAARNs" feature?
If so, this fix looks good to me. However, i'd like to change this code to be like 

```
domains := sets.NewString(aws.StringValueSlice(certDetail.SubjectAlternativeNames)...)
switch aws.StringValue(certDetail.Type) {
case acm.CertificateTypeImported:
	d.certDomainsCache.Set(certARN, domains, d.importedCertDomainsCacheTTL)
case acm.CertificateTypeAmazonIssued, acm.CertificateTypePrivate:
	d.certDomainsCache.Set(certARN, domains, d.privateCertDomainsCacheTTL)
}
if len(d.allowedCAARNs) == 0 || slices.Contains(d.allowedCAARNs, awssdk.StringValue(certDetail.CertificateAuthorityArn)) {
   return domains, nil
}
return sets.String{}, nil
```
technically there is no functional difference since `allowedCAARNs` is a controller-level flag which is immutable given the controller's lifetime. However, from coding perspective, the cache shall be for the "domains" before the "CA filter logic" and this make the code more robust(e.g. works even `allowedCAARNs` can be dynamically updated somehow).

mkilchhofer: Our PR introduce this behavior only if both conditions are met:
1. use the `--allowed-certificate-authority-arns=...` parameter
2. there exists certificates inside AWS ACM which are not issued from a authority in the allowed list.

I/we can test your proposal tomorrow in our staging environment. But I assume that your proposed code change does not work properly as we then also cache the domain(s) of a certificate which is not issued from an allowed CA.

And the function first tries to load the domains from the cache before it reaches this filtering code block:
```go
func (d *acmCertDiscovery) loadDomainsForCertificate(ctx context.Context, certARN string) (sets.String, error) {
	if rawCacheItem, ok := d.certDomainsCache.Get(certARN); ok {
		return rawCacheItem.(sets.String), nil
	}
	// only continues in case we didn't find it inside the cache
```

M00nF1sh: @mkilchhofer 
You are right, the cache shall be the domains after the filter. 
ideally we should refactor the code such that the cache stores certificate details from AWS. (but for now it only caches domains and your filter logic is on cert CA), thus filter logic has to be run before the cache.

I'll approve this, and maybe refactor this in the future if we ever need to make the ca list mutable.

oliviassss:
govulncheck failing is unrelated, will merge it.

Issue #2638, opened 2022-05-03
LoadBalancer Controller should not try to delete LB resource which is not created by itself


lqychange:
**Is your feature request related to a problem?**

Have project "A "deployed in EKS "A" with service type of "LoadBalancer", alone with  LoadBalancer controller "A". The LB controller have provisioned an internal NLB, as well as a TargetGroup registered in this NLB as a listener.  And for this project, a VPC Link also being provisioned alone with the NLB, and being used in API GW.

Now we have project "B" deployed in EKS "B", and have LoadBalancer controller "B"  deployed on top of this EKS cluster as well. In project "B", we also have a service with type of "LoadBalancer", to re-use existing resources, we specific the NLB name which created from project "A" in the service manifest yaml, and now LB controller "B" in EKS "B" are only creating new TargetGroup and add a listener to it under existing NLB, and everything running fine.

The issue is, when I try to delete service from EKS "B" for Project "B", the deletion get stuck, and I saw the log from LB controller pods, that this LB controller trying to delete the NLB as well, which are not provisioned by this LB controller

**Describe the solution you'd like**
LB controller should only deleting/deregister the resource provisioned by itself. 

**Describe alternatives you've considered**
No

kishorj:
@lqychange,
In the current design, we don't support using an externally created load balancer. You could label an existing load balancer with the appropriate label, and the controller picks up the load balancer, however, controller will not be able to determine whether the resource is external. Using an externally created load balancer is a feature request.

You've also mentioned EKS "A" and EKS "B" and I assume they are different kubernetes clusters. As of now, controller doesn't support load balancer across multiple clusters.

Would you be able to share how you made the controller re-use the load balancer, and also shared across multiple clusters. If you share across multiple clusters, the controllers in each cluster will be in a race condition updating the target groups.

junzebao:
> You could label an existing load balancer with the appropriate label, and the controller picks up the load balancer

Hi @kishorj , I'm wondering what labels are necessary for the LB to be picked up by the controller? is it these ones defined here? https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/annotations/#resource-tags

RoryCrispin:
Hi @kishorj ,

Plus one to this;
I find myself in an unsupported use case which I'd have thought was quite common;

The in-tree controller is now deprecated and many are planning a migration to this controller, but doing an all-in-one leap is sadly not possible in all cases,

[...]

PR #3633, opened 2024-04-03
Controller does not delete listeners having specific tag (e.g. indicating listeners are externally managed)

paulzhang97:
### Issue

The controller removes listeners not recognized by itself. In some deployments, listeners are created outside of the controller and managed externally. We need a way to tell the controller not to delete such listeners during reconciliation.

Related [issue](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2638)

### Description

Listener synthesizer get the env variable `PREVENT_DELETION_TAG_NAME` specifying the tag name for not deleting listeners. It does not delete listeners tagged with the one set by `PREVENT_DELETION_TAG_NAME`.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/deploy/elbv2/listener_synthesizer.go
+++ b/pkg/deploy/elbv2/listener_synthesizer.go
@@ -2,6 +2,8 @@ package elbv2
 
 import (
 	"context"
+	"os"
+
 	awssdk "github.com/aws/aws-sdk-go/aws"
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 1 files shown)


paulzhang97:
@kishorj @johngmyers This is my first PR. Just want to check whether you would need any action from me.

paulzhang97:
Can we have some review on this PR?

modevops:
I have the same issue as the user above.  ALB removes listeners not recognized by itself.   This makes it impossible to share aws alb controller with a non k8's service or ec2 instance.  I am doing path based routing and I have resource that uses to much memory and cpu and I do not want to migrate it to k8s.  I am stuck between a  rock and hard place because resources is used in so many parts they cannot change it.  Can this be escalated!!!! Please!!!!!

paulzhang97:
@modevops I implemented it at the forked [repo](https://github.com/redpanda-data/aws-load-balancer-controller). Check it out to see whether it meets your needs.

Issue #3617, opened 2024-03-18
mservice.elbv2.k8s.aws mutating webhook blocks Helm rollbacks


diversario:
**Describe the bug**
When using Helm to deploy a Service object (of type LoadBalancer) into a cluster, the installation and subsequent updates succeed, but a rollback with `--force` fails.

This is because the webhook only activates for `CREATE` and not for `UPDATE` operations - meaning, when Helm attempts to roll back and update the Service to its previous state, this previous state does not include the `spec.loadBalancerClass` field set by the webhook. This results in

```
[helm rollback - release: echoserver]  Error: failed to replace object:
 Service "echoserver-nlb" is invalid:
 spec.loadBalancerClass:
 Invalid value: "null": may not change once set
```

because (afaik) Helm sends the entire Service object to the API server due to the `--force` flag.

**Steps to reproduce**
Deploy any Helm chart with an LB service. I'll use AWS LBC chart as a placeholder for this example.
1. Deploy the chart 2-3 times so there's a release history:
  `helm upgrade --install --wait --wait-for-jobs --timeout 10m --atomic aws-lbc eks/aws-load-balancer-controller`
2. Run a rollback with `--force`:
     `helm rollback aws-lbc 1 --force --wait --no-hooks --timeout 600s`

**Expected outcome**
Not an error; I expect the webhook to modify the service objects appropriately, and the rollback to complete.

**Environment**

* AWS Load Balancer controller version
6.7.1
* Kubernetes version
1.28.3
* Using EKS (yes/no), if so version?
no

**Additional Context**:
Using Helm 3.10.3

PowerSurj:
> * AWS Load Balancer controller version
>   6.7.1

are you sure? do you mean Helm Chart version 2.7.1

diversario:
Yes, sorry. The AWS LBC version 2.7.1, chart version is the one that sets that appVersion.

shraddhabang:
Delivered in v2.8.0

PR #3641, opened 2024-04-10
Preserve loadBalancerClass on Service updates

diversario:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3617

### Description

See the linked issue for the observed problem.

This PR enables service mutator to run on updates to service objects (the `UPDATE` operation needs to be added to the webhook configuration for this to work).

Update mutator will only act on service objects that have `spec.loadBalancerClass` set to a non-empty value. When this is true, the incoming (updated) Service object will be updated to have the same `spec.loadBalancerClass` as the existing Service object.

Considering that `spec.loadBalancerClass`, once set, cannot be unset or modified, this change prevents unexpected validation errors in cases of `UPDATE` operation with Service object that omits the `spec.loadBalancerClass` field. This is the case for `helm rollback --force`, described in the linked issue, as well as for `kubectl replace` command.

Documentation may need to be updated depending on the outcome of https://github.com/aws/eks-charts/issues/1063 - that PR changes some chart values that are being referenced in the docs.

### Testing
This change was tested using the helm chart built from https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3653.

For testing, I had MWC with no UPDATE operation first (as it is currently configured), then I enabled it in the chart.

Tested by creating a Service via `kubectl apply`, updating it via `kubectl apply`, and then replacing it via `kubectl replace`. `kubectl replace` previously would fail with validation error due to attempt to modify spec.loadBalancerClass. No longer fails.

I submitted Service manifests with and without the `spec.loadBalancerClass` field.

### Checklist
- [x] Manually tested

### DIFF
--- a/webhooks/core/service_mutator.go
+++ b/webhooks/core/service_mutator.go
@@ -52,7 +52,41 @@ func (m *serviceMutator) MutateCreate(ctx context.Context, obj runtime.Object) (
 }
 
 func (m *serviceMutator) MutateUpdate(ctx context.Context, obj runtime.Object, oldObj runtime.Object) (runtime.Object, error) {
-	return obj, nil
+	// this mutator only cares about Service objects
+	newSvc, ok := obj.(*corev1.Service)
+	if !ok {
+		return obj, nil
+	}
+
+	oldSvc, ok := oldObj.(*corev1.Service)
+	if !ok {
+		return obj, nil
+	}
+
+	if newSvc.Spec.Type != corev1.ServiceTypeLoadBalancer {
+		return obj, nil
+	}
+
+	// does the old Service object have spec.loadBalancerClass?
+	if oldSvc.Spec.LoadBalancerClass != nil && *oldSvc.Spec.LoadBalancerClass != "" {
+		// if so, let's inspect the incoming object for the same field
+
+		// does the new Service object lack spec.loadBalancerClass?
+		// if so, set it to the old value
+		// if yes, then leave it be because someone wanted it that way, let the user deal with the error
+		if newSvc.Spec.LoadBalancerClass == nil || *newSvc.Spec.LoadBalancerClass == "" {
+			newSvc.Spec.LoadBalancerClass = oldSvc.Spec.LoadBalancerClass
+
+			m.logger.Info("preserved loadBalancerClass", "service", newSvc.Name, "loadBalancerClass", *newSvc.Spec.LoadBalancerClass)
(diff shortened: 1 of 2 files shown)


webhooks/core/service_mutator.go
```diff
@@ -52,7 +52,41 @@ func (m *serviceMutator) MutateCreate(ctx context.Context, obj runtime.Object) (
...
+	// does the old Service object have spec.loadBalancerClass?
+	if oldSvc.Spec.LoadBalancerClass != nil && *oldSvc.Spec.LoadBalancerClass != "" {
+		// if so, let's inspect the incoming object for the same field
+
+		// does the new Service object lack spec.loadBalancerClass?
+		// if so, set it to the old value
+		// if yes, then leave it be because someone wanted it that way, let the user deal with the error
+		if newSvc.Spec.LoadBalancerClass == nil || *newSvc.Spec.LoadBalancerClass == "" {
+			newSvc.Spec.LoadBalancerClass = oldSvc.Spec.LoadBalancerClass
+
```
oliviassss, line 79: I was just thinking some cornor cases - would this change cause some issue when this field in the newSvc is intended to be nil or ""?

diversario: Since this handler is for updates to an _existing_ service only, there's a couple of reasons the field would be missing/empty:

1. User is updating the service object via `patch`, submitting only modified fields. In this case, tacking the LBC field onto the update is a noop, and there's no change in behavior.
2. User is updating the service object via `put` (this is the `kubectl replace`) and omits the field. The normal response would be an error, as the field is immutable.
With this handler, the error does not occur as we putting the field back. This is a change in behavior.
Whether this change is unexpected - I think it is not, because presumably user has enabled the webhook, is aware of the webhook effects, and expects LB Services in their cluster to be managed by AWS LBC. So this change, then, just prevents the user from running into an error (and this is the entire reason behind this PR because our deployment tooling - Helm - runs into this very error).
Importantly, because the handler inspects the existing Service, it will only touch service that are known to be managed by AWS LBC, and not something else.

oliviassss:
@diversario, thanks for the contribution. I'm good with the changes, can you please add the manual tests you've done in the PR descriptions?
/lgtm

diversario:
> @diversario, thanks for the contribution. I'm good with the changes, can you please add the manual tests you've done in the PR descriptions? /lgtm

Updated description.

PR #3643, opened 2024-04-11
Identify VPC by resource tags

cathal-c:
### Issue

Addresses https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3644.

### Description

This PR introduces a new runtime argument, `aws-vpc-tag-key` which is used to identify VPCs, from which we can infer the VPC ID.

This is useful for cases where access to AWS metadata is blocked and the VPC ID is not known at deploy time, options to infer the VPC ID are limited.

### Checklist
- [ ] Manually tested


pkg/aws/cloud_config.go
```diff
@@ -12,9 +12,11 @@ const (
 	flagAWSAPIEndpoints  = "aws-api-endpoints"
 	flagAWSAPIThrottle   = "aws-api-throttle"
 	flagAWSVpcID         = "aws-vpc-id"
+	flagAWSVpcTagKey     = "aws-vpc-tag-key"
```
M00nF1sh, line 15: shall we consider to have `aws-vpc-infer-from-tags` and make it a string map? 
seems that's would be more general as in my sample vpcs, i only have a `name`: `vpc-name` tag.

jeswinkoshyninan: @M00nF1sh thank you for your comments, it looks good to me. Please note that I am from team of Cathal and since he is currently away, I am taking over this request and changes based on your comments are now added into a new PR: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3656/files. please take a look into the same.

Issue #2897, opened 2022-11-25
AWS load balancer controller continues to provide high cardinality unbounded metrics to prometheus endpoint 


yodaflomaster:
**Describe the bug**
After upgrading version of aws load balancer controller using helm, i keep seeing rest_client_request_latency_seconds histogram metric exposed on the Prometheus metrics endpoint. It includes a `URL` tag containing the URI of all API versions. It's about ~900 metrics. I've delete chart, check dependencies and redeploy. But the problem didn't go away
https://github.com/kubernetes-sigs/controller-runtime/issues/1423
https://github.com/kubernetes-sigs/controller-runtime/pull/1587

```
...
rest_client_request_latency_seconds_bucket{url="https://172.20.0.1:443/api/v1/endpoints?limit=%7Bvalue%7D&resourceVersion=%7Bvalue%7D",verb="GET",le="0.001"} 0
rest_client_request_latency_seconds_bucket{url="https://172.20.0.1:443/api/v1/endpoints?limit=%7Bvalue%7D&resourceVersion=%7Bvalue%7D",verb="GET",le="0.002"} 0
rest_client_request_latency_seconds_bucket{url="https://172.20.0.1:443/api/v1/endpoints?limit=%7Bvalue%7D&resourceVersion=%7Bvalue%7D",verb="GET",le="0.004"} 0
rest_client_request_latency_seconds_bucket{url="https://172.20.0.1:443/api/v1/endpoints?limit=%7Bvalue%7D&resourceVersion=%7Bvalue%7D",verb="GET",le="0.008"} 0
rest_client_request_latency_seconds_bucket{url="https://172.20.0.1:443/api/v1/endpoints?limit=%7Bvalue%7D&resourceVersion=%7Bvalue%7D",verb="GET",le="0.016"} 1
rest_client_request_latency_seconds_bucket{url="https://172.20.0.1:443/api/v1/endpoints?limit=%7Bvalue%7D&resourceVersion=%7Bvalue%7D",verb="GET",le="0.032"} 1
rest_client_request_latency_seconds_bucket{url="https://172.20.0.1:443/api/v1/endpoints?limit=%7Bvalue%7D&resourceVersion=%7Bvalue%7D",verb="GET",le="0.064"} 1
[...]

PR #3645, opened 2024-04-11
feat(chart): Added additional service monitor functionality

stevehipwell:
### Issue

This PR will allow manual remediation of #2897.

### Description

This PR adds the ability to set the Helm chart `ServiceMonitor` endpoint values for `scrapeTimeout`, `relabelings` & `metricRelabelings`. It also fixes the `jobLabel` functionality (it should be a label name and not a value) and refactors the `ServiceMonitor` template for easier readability.

### Checklist
- [x] Manually tested


stevehipwell:
@kishorj @oliviassss could you please take a look at this PR? The failing workflow looks to have been failing for a while and the fix PR (#3600) was closed.

stevehipwell:
@kishorj @oliviassss could you please take a look at this PR?

helm/aws-load-balancer-controller/templates/servicemonitor.yaml
```diff
@@ -29,7 +25,19 @@ spec:
   endpoints:
     - port: metrics-server
       path: /metrics
-      {{- with .Values.serviceMonitor.interval }}
+      scheme: http
```
oliviassss, line 28: just for my understanding, why we specify `http` explicitly here?

stevehipwell: I always set the `scheme` so it's very clear what the intent is even for someone who's starting at the `ServiceMonitor` and working up through the other resources. From an implementation point of view this is something where even though the default is correct we 100% care about the value, so we should explicitly set it.

stevehipwell:
@oliviassss it looks like the dependency review job is currently failing across the board (also this PR isn't making any changes to the deps), are you planning on manually merging?

PR #3653, opened 2024-04-18
Provide more customization options for the service mutator webhook

diversario:
### Issue

- https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3641
- https://github.com/aws/eks-charts/issues/1063

### Description of changes

For https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3641, the webhook configuration needs to support at least `operations` customization, but being able to exclude certain services by a selector is important for https://github.com/aws/eks-charts/issues/1063.

This PR adds a new section for service mutator webhook configuration. Defaults are set to match the existing configuration, so that this change does not unexpectedly enable service mutation behavior from https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3641.

I've not updated the README, want to get feedback first.

### Checklist
- [ ] Added/modified documentation as required (such as the `README.md` for modified charts)
- [x] Incremented the chart `version` in `Chart.yaml` for the modified chart(s)
- [x] Manually tested. Describe what testing was done in the testing section below
- [x] Make sure the title of the PR is a good description that can go into the release notes

### Testing

This change was tested with AWS LBC v2.7.2, and also with AWS LBC built from https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3641.

For testing, I had no `UPDATE` operation first, then I enabled it in the chart.

Tested by creating a Service via `kubectl apply`, updating it via `kubectl apply`, and then replacing it via `kubectl replace`. `kubectl replace` previously would fail with validation error due to attempt to modify `spec.loadBalancerClass`. No longer fails.


helm/aws-load-balancer-controller/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: aws-load-balancer-controller
 description: AWS Load Balancer Controller Helm chart for Kubernetes
-version: 1.7.1
-appVersion: v2.7.0
+version: 1.8.0
+appVersion: v2.7.2
```
oliviassss, line 5: We will bump up the version number during release, please revert this change, thanks.

helm/aws-load-balancer-controller/test.yaml
```diff
@@ -6,7 +6,7 @@ replicaCount: 2
 
 image:
   repository: public.ecr.aws/eks/aws-load-balancer-controller
-  tag: v2.7.0
+  tag: v2.7.2
```
oliviassss, line 9: same here, no need to change version number.

helm/aws-load-balancer-controller/values.yaml
```diff
@@ -8,7 +8,7 @@ revisionHistoryLimit: 10
 
 image:
   repository: public.ecr.aws/eks/aws-load-balancer-controller
-  tag: v2.7.0
+  tag: v2.7.2
```
oliviassss, line 11: same here.

helm/aws-load-balancer-controller/values.yaml
```diff
@@ -389,5 +389,25 @@ ingressClassConfig:
 # enableServiceMutatorWebhook allows you enable the webhook which makes this controller the default for all new services of type LoadBalancer
 enableServiceMutatorWebhook: true
 
+# serviceMutatorWebhook contains configurations specific to the service mutator webhook
+serviceMutatorWebhook:
+  # Specifies whether a service mutator webhook should be created
+  enabled: true
```
oliviassss, line 395: I think the `serviceMutatorWebhook.enabled` is redundant to  `enableServiceMutatorWebhook`, can you just use `enableServiceMutatorWebhook`? it needs to be true if someone wants to config `serviceMutatorWebhook`

helm/aws-load-balancer-controller/templates/webhook.yaml
```diff
@@ -56,7 +56,7 @@ webhooks:
     resources:
     - pods
   sideEffects: None
-{{- if .Values.enableServiceMutatorWebhook }}
+{{- if and .Values.enableServiceMutatorWebhook .Values.serviceMutatorWebhook.enabled }}
```
oliviassss, line 59: I don't think it's necessary to have `serviceMutatorWebhook.enabled`, probably just use `enableServiceMutatorWebhook`. it needs to be true if someone wants to config service mutator webhook.

oliviassss:
thanks for the contribution. the change lgtm. can you please add the manual testing you've done on the service mutator webhook in the PR description? thanks.

diversario:
> thanks for the contribution. the change lgtm. can you please add the manual testing you've done on the service mutator webhook in the PR description? thanks.

Updated description.

Issue #3644, opened 2024-04-11
Support identifying VPC by tags


cathal-c:
**Is your feature request related to a problem?**
For cases where access to AWS metadata is blocked and the VPC ID is not known at deploy time, options to infer the VPC ID are limited.

**Describe the solution you'd like**
I would like to request a third option, to specify tags used to filter VPC IDs via the EC2 `DescribeVpcs..` API set.

**Describe alternatives you've considered**
N/A

PR #3656, opened 2024-04-22
Fetch VPC ID from runtime using VPC tags provided via controller flags

jeswinkoshyninan:
### Issue

This PR will address the issue raised: #3644
 
### Description

This PR introduces a new runtime argument, aws-vpc-tags which is used to identify VPCs from the list of tags from which we can infer the VPC ID. There is also an optional argument called aws-vpc-name-tag-key incase if the "Name" will not be key-name for the VPC name in tag. 

This is useful for cases where access to AWS metadata is blocked and the VPC ID is not known at deploy time, options to infer the VPC ID are limited

Manual Tests Details

```

Dev Build

➜  ~ kubectl get deploy aws-load-balancer-controller  -n aws-lb-controller -o=jsonpath="{.spec.template.spec.containers[0].args}" 
["--cluster-name=j119106042","--ingress-class=alb","--aws-vpc-tags=Name=j119106042"]%                                                                                                                                                                ➜  ~ 

When I have added the log to display the vpc_id, it got properly fetched the VPC(hidden) based on the given flag --aws-vpc-tags=Name="". 

➜  ~ kubectl logs aws-load-balancer-controller-566f9f556b-dhlbr -n aws-lb-controller 
{"level":"info","ts":"2024-05-09T20:12:03Z","msg":"version","GitVersion":"","GitCommit":"","BuildDate":""}
VPC ID:  vpc-XXXXXXXXXXXX
{"level":"info","ts":"2024-05-09

Current Build

➜  ~ kubectl logs aws-load-balancer-controller-6776c547-7mfj4  -n aws-lb-controller 
{"level":"info","ts":"2024-05-14T10:29:47Z","msg":"version","GitVersion":"","GitCommit":"","BuildDate":""}
{"level":"info","ts":"2024-05-14T10:29:47Z","logger":"controller-runtime.metrics","msg":"Metrics server is starting to listen","addr":":8080"}
{"level":"info","ts":"2024-05-14T10:29:47Z","logger":"setup","msg":"adding health check for controller"}
{"level":"info","ts":"2024-05-14T10:29:47Z","logger":"setup","msg":"adding readiness check for webhook"}
{"level":"info","ts":"2024-05-14T10:29:47Z","logger":"controller-runtime.webhook","msg":"Registering webhook","path":"/mutate-v1-pod"}
{"level":"info","ts":"2024-05-14T10:29:47Z","logger":"controller-runtime.webhook","msg":"Registering webhook","path":"/mutate-v1-service"}
{"level":"info","ts":"2024-05-14T10:29:47Z","logger":"controller-runtime.webhook","msg":"Registering webhook","path":"/validate-elbv2-k8s-aws-v1beta1-ingressclassparams"}
{"level":"info","ts":"2024-05-14T10:29:47Z","logger":"controller-runtime.webhook","msg":"Registering webhook","path":"/mutate-elbv2-k8s-aws-v1beta1-targetgroupbinding"}
{"level":"info","ts":"2024-05-14T10:29:47Z","logger":"controller-runtime.webhook","msg":"Registering webhook","path":"/validate-elbv2-k8s-aws-v1beta1-targetgroupbinding"}
{"level":"info","ts":"2024-05-14T10:29:47Z","logger":"controller-runtime.webhook","msg":"Registering webhook","path":"/validate-networking-v1-ingress"}
{"level":"info","ts":"2024-05-14T10:29:47Z","logger":"setup","msg":"starting podInfo repo"}
{"level":"info","ts":"2024-05-14T10:29:49Z","logger":"controller-runtime.webhook.webhooks","msg":"Starting webhook server"}
{"level":"info","ts":"2024-05-14T10:29:49Z","msg":"Starting server","kind":"health probe","addr":"0.0.0.0:61779"}
{"level":"info","ts":"2024-05-14T10:29:49Z","msg":"Starting server","path":"/metrics","kind":"metrics","addr":"0.0.0.0:8080"}
{"level":"info","ts":"2024-05-14T10:29:49Z","logger":"controller-runtime.certwatcher","msg":"Updated current TLS certificate"}
{"level":"info","ts":"2024-05-14T10:29:49Z","logger":"controller-runtime.certwatcher","msg":"Starting certificate watcher"}
{"level":"info","ts":"2024-05-14T10:29:49Z","logger":"controller-runtime.webhook","msg":"Serving webhook server","host":"","port":9443}

```

### Checklist
- [ ] Manually tested


jeswinkoshyninan:
@oliviassss @johngmyers can you please take look into this same ?

jeswinkoshyninan:
Thank you @shraddhabang , seems like test looks good to us. Can we merge this change?

jeswinkoshyninan:
@shraddhabang @johngmyers @oliviassss tagging again incase if we lost the track of this PR, seems like the tests looks good and okay to merge ?

oliviassss:
@jeswinkoshyninan thanks for the contribution, this change looks good to me overall, just a few questions:
1. iiuc, we are supporting filtering the vpc by tags, it does not have to be "Name". So can we change the flag `aws-vpc-name-tag-key` to `aws-vpc-tag-key` to make it more general, and by default it's `Name`.
2. I think maybe we need a validation that, if users use both flag `--aws-vpc-id` and `--aws-vpc-tags`, the controller should error out. since it should be either specify a vpc-id, or filter by vpc tag.

jeswinkoshyninan:
@oliviassss thank you for the comments 
Regarding your first point, I have made the changes and the second point, currently if VPCID is set, then that will be returned instead of fetching VPCID using tags. In this case, is this required to add a condition since this will not cause any problems. Please let me know your thoughts ?

oliviassss:
yeah it's not causing any problem, since it's just return the vpc id if specified. But I think we should call out for this bevior:
1. in `docs/deploy/configurations.md`, add a call out that `--aws-vpc-id` will take precedence if both `--aws-vpc-id` and `--aws-vpc-tag` are specified.
2. maybe add a debug log line in the `func getVpcID()`, if `cfg.VpcID != nil`, we can log out that vpc-id is specified, and the controller just used it. It should be V(1) log, so only visible when users enable debug log level.

jeswinkoshyninan:
@oliviassss make sense, have pushes the changes, please take a look.

docs/deploy/configurations.md
```diff
@@ -64,51 +64,53 @@ Currently, you can set only 1 namespace to watch in this flag. See [this Kuberne
...
+| aws-max-retries                                                                 | int                             | 10                                         | Maximum retries for AWS APIs                                                                                                                   |
+| aws-region                                                                      | string                          | [instance metadata](#instance-metadata)    | AWS Region for the kubernetes cluster                                                                                                          |
+| aws-vpc-id                                                                      | string                          | [instance metadata](#instance-metadata)    | AWS VPC ID for the Kubernetes cluster                                                                                                          |
+| aws-vpc-tags                                                                    | stringMap                       |                                            | Tags for the Kubernetes cluster VPC
```
oliviassss, line 74: Maybe also add a callout here:
>Note that if both flags `--aws-vpc-id` and `--aws-vpc-tags` are specified, the controller will ignore the `--aws-vpc-tags` flag, and use the value in`--aws-vpc-id` to fetch the VPC info.

docs/deploy/installation.md
```diff
@@ -37,7 +37,7 @@ You can set the IMDSv2 as follows:
 aws ec2 modify-instance-metadata-options --http-put-response-hop-limit 2 --http-tokens required --region <region> --instance-id <instance-id>
 ```
 
-Instead of depending on IMDSv2, you can specify the AWS Region and the VPC via the controller flags `--aws-region` and `--aws-vpc-id`.
+Instead of depending on IMDSv2, you can specify the AWS Region via the controller flag `--aws-region`, and the AWS VPC via controller flag `--aws-vpc-id` or by specifying vpc tags via the flag `--aws-vpc-tags` and an optional flag `--aws-vpc-tag-key` if you have a different key for the tag other than "Name". Note that if you specify flags `--aws-vpc-id` and `--aws-vpc-tags`, then value given to `--aws-vpc-id` will be taken by controller.
```
oliviassss, line 40: nit: reword a bit
>Note that if both flags `--aws-vpc-id` and `--aws-vpc-tags` are specified, the controller uses the value in `--aws-vpc-id` to fetch the VPC info and ignores the other flag.

docs/deploy/configurations.md
oliviassss: Another question, does it accept multiple key:value pairs for the tags filter?

oliviassss:
/ok-to-test
can you also add the manual testing you've done in the PR description? thanks

docs/deploy/configurations.md
jeswinkoshyninan: yes, it is a string Map. It will accept multiple values.

docs/deploy/installation.md
jeswinkoshyninan: refactor the sentence. thanks

docs/deploy/configurations.md
jeswinkoshyninan: Added the details here as well.

jeswinkoshyninan:
@oliviassss PR desc updated with the manual test details.

jeswinkoshyninan:
@oliviassss just for transparency, seems like when enabling debug logs, the logs which we set to message user "vpcid is specified" is not working when I am testing. This will not affect the functionality. But just letting you know that I am checking.

jeswinkoshyninan:
@oliviassss identified the problem with logger and fixed the changes. I can see the logs when we set debs

```
➜  .kube kubectl logs aws-load-balancer-controller-796b7c7f76-wrvdx -n aws-lb-controller -f
{"level":"info","ts":"2024-05-14T21:41:20Z","msg":"version","GitVersion":"","GitCommit":"","BuildDate":""}
{"level":"debug","ts":"2024-05-14T21:41:20Z","msg":"vpcid is specified using flag --aws-vpc-id, controller will use the value","vpc: ":"vpc-XXXXXXXXX"}
```

oliviassss:
@jeswinkoshyninan, discussed internally that we will ship this commit in next minor release, which should be v2.9.0, for a feature like this. Thanks for your patience.

jeswinkoshyninan:
@oliviassss thank you for the updates, please let me know once we confirm the release v2.9.0.

pkg/aws/cloud.go
```diff
@@ -132,6 +131,20 @@ func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud,
...
 
+func getVpcID(cfg CloudConfig, ec2Service services.EC2, metadata services.EC2Metadata, logger logr.Logger) (string, error) {
+
+	if cfg.VpcID != "" {
+		logger.V(1).Info("vpcid is specified using flag --aws-vpc-id, controller will use the value", "vpc: ", cfg.VpcID)
+		return cfg.VpcID, nil
+	}
+
+	if cfg.VpcTags != nil {
+		return inferVPCIDFromTags(ec2Service, cfg.VpcNameTagKey, cfg.VpcTags[cfg.VpcNameTagKey])
```
alfredkrohmer, line 142: @jeswinkoshyninan I don't really get this part: why does this allow to pass a map with an arbitrary number of items for `--aws-vpc-tags` if we only ever pass a single key-value pair (the one with the key specified in `aws-vpc-tag-key`) as the tag filter? Shouldn't it rather just convert the map from `--aws-vpc-tags` into a list of tag filters, one for each item?

pinkavaj: Related to #3889

PR #3684, opened 2024-05-08
Add `load-balancer-name-prefix` annotation

hfern:
### Description

This adds a `alb.ingress.kubernetes.io/load-balancer-name-prefix` annotation. This is useful for being able to control the name of the created load balancer without needing to specify a fully unique name.

We have several similar clusters in one AWS account, all with similar namespace/resources. This leads to this situation, where it's hard to tell at a glance (without digging into tags) which load balancer goes to which cluster:

image

Already the ALBC will provide uniqueness for you if you do not specify a name. This is nice because you can be reasonable certain there will be no name collisions. That means you can copy/paste (and we can vend) a static piece of yaml into different clusters or namespaces IF you do not specify the `load-balancer-name`. 

However, if you want to get more descriptive with the name via `load-balancer-name`, now you need to worry about collision (even potentially with resources in other clusters).

`load-balancer-name-prefix` is a happy middle ground where you can be more specific without having to own avoiding naming collisions. 

### Checklist
- [x] Manually tested - make test and reviewed generated docs so they look right

### DIFF
--- a/docs/guide/ingress/annotations.md
+++ b/docs/guide/ingress/annotations.md
@@ -17,6 +17,7 @@ You can add annotations to kubernetes Ingress and Service objects to customize t
 | Name                                                                                                  | Type                        |Default|Location|MergeBehavior|
 |-------------------------------------------------------------------------------------------------------|-----------------------------|-------|--------|------|
 | [alb.ingress.kubernetes.io/load-balancer-name](#load-balancer-name)                                   | string                      |N/A|Ingress|Exclusive|
+| [alb.ingress.kubernetes.io/load-balancer-name-prefix](#load-balancer-name-prefix)                     | string                      |N/A|Ingress|Exclusive|
 | [alb.ingress.kubernetes.io/group.name](#group.name)                                                   | string                      |N/A|Ingress|N/A|
 | [alb.ingress.kubernetes.io/group.order](#group.order)                                                 | integer                     |0|Ingress|N/A|
 | [alb.ingress.kubernetes.io/tags](#tags)                                                               | stringMap                   |N/A|Ingress,Service|Merge|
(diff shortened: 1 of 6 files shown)


pkg/ingress/model_build_load_balancer.go
```diff
@@ -88,31 +89,28 @@ func (t *defaultModelBuildTask) buildLoadBalancerSpec(ctx context.Context, liste
...
-			return "", errors.New("load balancer name cannot be longer than 32 characters")
-		}
-		return name, nil
+	explicitName, err := t.getNameAnnotation(annotations.IngressSuffixLoadBalancerName, maximumLoadBalancerNameLength)
+
+	if explicitName != "" || err != nil {
+		return explicitName, err
 	}
-	if len(explicitNames) > 1 {
-		return "", errors.Errorf("conflicting load balancer name: %v", explicitNames)
```
hfern, line 108: Since this block of code would have to be duplicated for the prefix annotation, I just moved it to a new `getNameAnnotation` and parameterized the annotation name.

docs/guide/ingress/annotations.md
```diff
@@ -176,6 +177,20 @@ Traffic Routing can be controlled with following annotations:
         alb.ingress.kubernetes.io/load-balancer-name: custom-name
         ```
 
+- <a name="load-balancer-name-prefix">`alb.ingress.kubernetes.io/load-balancer-name-prefix`</a> specifies the custom name prefix to use for the load balancer. The provided value will have 11 characters appended to it for uniqueness. Prefixes longer than 21 characters will be treated as an error.
```
oliviassss, line 180: just for my understanding why is it 11 as the limit?

pkg/ingress/model_build_load_balancer.go
```diff
@@ -88,31 +89,28 @@ func (t *defaultModelBuildTask) buildLoadBalancerSpec(ctx context.Context, liste
...
+
+	prefixName, err := t.getNameAnnotation(annotations.IngressSuffixLoadBalancerNamePrefix, maximumLoadBalancerNameLength-11)
```
oliviassss, line 98: nit: maybe change the `11` to a constant instead of hard coded here.

pkg/ingress/model_build_load_balancer.go
```diff
@@ -123,6 +121,30 @@ func (t *defaultModelBuildTask) buildLoadBalancerName(_ context.Context, scheme
 	return fmt.Sprintf("k8s-%.8s-%.8s-%.10s", sanitizedNamespace, sanitizedName, uuid), nil
 }
 
+func (t *defaultModelBuildTask) getNameAnnotation(annotationName string, maxLength int) (string, error) {
```
oliviassss, line 124: nit: maybe rename to `parseNameOrNamePrefixFromAnnotation()`

Issue #3715, opened 2024-05-23
Outposts and local zone do not support mTLS feature


oliviassss:
**Describe the bug**
This an issue introduced from v2.7.0 with the mTLS feature, that the controller builds a `MutualAuthenticationconfig` with default value to `off` and pass to the listener config regardless of whether users enable the feature via the annotation `alb.ingress.kubernetes.io/mutual-authentication`. But the ELB API does not support this `MutualAuthenticationconfig` object in outposts or local zone, we should fix this from controller side.
Reference of code:

1. https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/ingress/model_build_listener.go#L56
2. https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/ingress/model_build_listener.go#L278

As a workaround, if the users do not need this feature at all, and if the subnets are in local zone or outposts, please roll back to v2.6.2 where we do not have this feature, and upgrade later after we have a fix.

The fix:
If the annotation `alb.ingress.kubernetes.io/mutual-authentication` is not specified by the user, the controller should return nil and do nothing on the `MutualAuthentication`. If the user enabled it via annotation and want to disable it later, they should specify explicitly via annotation. In this way, we could avoid unnecessary API calls from the controller, and also reduce the overhead in the controller. 

**Steps to reproduce**

**Expected outcome**
A concise description of what you expected to happen.

**Environment**

* AWS Load Balancer controller version: v2.7.0+
* Kubernetes version
* Using EKS (yes/no), if so version?

**Additional Context**:

oliviassss:
close as the fix shipped with v2.8.1

PR #3717, opened 2024-05-24
fixed the mTLS bug https://github.com/kubernetes-sigs/aws-load-balanc…

shethyogita83:
Fixing mTLS bug : https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3715

### Checklist
- [x] Manually tested


oliviassss:
Thanks for the fix, look good to me overall, some comments:
1. https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/ingress/model_build_listener.go#L56, I think we need to avoid pass the `MutualAuthentication` obj with the API even when the value is `nil`
2. https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/ingress/model_build_listener.go#L161. Need to add a nil check for `mutualAuthenticationAttributes`

shethyogita83:
> Thanks for the fix, look good to me overall, some comments:
> 
> 1. https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/ingress/model_build_listener.go#L56, I think we need to avoid pass the `MutualAuthentication` obj with the API even when the value is `nil`
> 2. https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/ingress/model_build_listener.go#L161. Need to add a nil check for `mutualAuthenticationAttributes`
> 3. can you pls also test this manually and add the test cases in the PR description? Thanks.

Added tests for the same and also did manual testing by creating an alb without mutual auth annotation and with secure listeners port and protocol. Everything is working as expected.

oliviassss:
@shethyogita83, thanks, how about these 2 comments?
>1. https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/ingress/model_build_listener.go#L56, I think we need to avoid pass the MutualAuthentication obj with the API even when the value is nil
2. https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/ingress/model_build_listener.go#L161. Need to add a nil check for mutualAuthenticationAttributes

I think right now we are still passing the `MutualAuthentication` param (with nil value) to the API, if my understanding is correct, the param itself is not supported in the API for outposts and local zone? So it will still have issue when users deploy the controller in these zones? Unless I'm missing anything

docs/guide/ingress/annotations.md
```diff
@@ -790,16 +790,19 @@ TLS support can be controlled with the following annotations:
...
-        - `port: listen port ` 
-            - Must be a HTTPS port specified by [listen-ports](#listen-ports).
-        - `mode: "off" (default) | "passthrough" | "verify"`
-            - `verify` mode requires an existing trust store resource.
-            -  See [Create a trust store](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/mutual-authentication.html#create-trust-store) in the AWS documentation for more details.
-        - `trustStore: ARN (arn:aws:elasticloadbalancing:trustStoreArn) | Name (my-trust-store)`
-            - Both ARN and Name of trustStore are supported values.
-            - `trustStore` is required when mode is `verify`.
-        - `ignoreClientCertificateExpiry : true | false (default)`
+    !!!note
```
shraddhabang, line 793: Can you please update this line as well? https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/guide/ingress/annotations.md?plain=1#L61
We want to default value here to N/A since we are not setting from our side.

Issue #3714, opened 2024-05-23
Cannot successfully register targets if VPC subnets are different from VPC CIDR block associations


nullren:
**Describe the bug**
the controller cannot register targets in targetgroupbinding with aws api error

```
ValidationError: The Availability Zone is required for IP address '10.4.40.87' because it is not in the VPC
```

this is because the [registerPodEndpoints](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/9e9432b654461a3bbc92638afcd8f4abc5de977f/pkg/targetgroupbinding/resource_manager.go#L386-L422) function is looking up cidr blocks associated to the vpc and not the actual subnets which is what the aws api compares.

**Steps to reproduce**

create vpc and associate more cidr blocks than subnets

```
$ aws ec2 describe-vpcs --vpc-ids vpc-0ea5e9af6c340f01f | jq -r '.Vpcs[] | .CidrBlockAssociationSet[] | .CidrBlock'
10.1.0.0/16
10.2.0.0/16
10.3.0.0/16
10.4.0.0/16
```

then add a few subnets to the vpc

```
$ aws ec2 describe-subnets --filters "Name=vpc-id,Values=vpc-0ea5e9af6c340f01f" | jq -r '.Subnets[] | .CidrBlock'
10.1.0.0/19
10.2.0.0/19
```

try to manually add a target to the target group

```
$ aws elbv2 register-targets --target-group-arn arn:aws:elasticloadbalancing:us-east-1:000000000000:targetgroup/k8s-default-targetgroupname-0000000/0000000000000 --targets Id=10.4.46.99,Port=80

An error occurred (ValidationError) when calling the RegisterTargets operation: The Availability Zone is required for IP address '10.4.46.99' because it is not in the VPC
```

**Expected outcome**
this is fixed when adding `AvailabilityZone=all` to the request and what should happen in the initial check

**Environment**

* AWS Load Balancer controller version: 2.7.2
* Kubernetes version
* Using EKS (yes/no), if so version? yes

**Additional Context**:

nullren:
i've started a [pr](https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3722) to draft out some ideas here but to also on my end to unblock my work.

one thing i'm ensure of is the "source of truth" for what the problem is, it seems there are a few things going on here:
- does the `register-targets` api actually look to just the subnets for discovery?
[...]

PR #3722, opened 2024-05-28
Targetgroup subnets discovery

nullren:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3714

### Description

we have ips that are covered by the "describe vpcs" aws api list of cidr blocks. however, when the "register target" api is then used to add those ips, they fail because they're not found in any of the subnets in "describe subnets" api for the vpc. they successfully register when given the `"all"` az.

### Checklist
- [ ] Manually tested


pkg/targetgroupbinding/resource_manager.go
```diff
@@ -413,7 +429,7 @@ func (m *defaultResourceManager) registerPodEndpoints(ctx context.Context, tgARN
 		if err != nil {
 			return err
 		}
-		if !networking.IsIPWithinCIDRs(podIP, vpcCIDRs) {
+		if !networking.IsIPWithinCIDRs(podIP, vpcCIDRs) || !networking.IsIPWithinCIDRs(podIP, subnetCIDRs) {
```
ArtfulCoder, line 432: do we need similar logic in `deregisterTargets`  func ?

PR #3775, opened 2024-07-22
fixup(sg-resolver): Allow multiple SGs with the same Name tag

alloveras:
### Issue

I looked but couldn't find any open issues about the problem this pull request intends to solve.

### Description

The current implementation of [`ResolveViaNameOrID`](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/networking/security_group_resolver.go#L35) incorrectly [assumes that searching security groups by the `Name` tag will always yield a single resource per searched name](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/networking/security_group_resolver.go#L56-L58). This is incorrect because AWS allows multiple security groups in the same VPC to share the same `Name` tag.

In practice, this issue manifests in the form of a rather cryptic runtime error:

```
Events:
  Type     Reason             Age                    From       Message
  ----     ------             ----                   ----       -------
  Warning  FailedBuildModel   9m (x207 over 2d2h)    ingress    Failed build model due to couldn't find all securityGroups, nameOrIDs: [some-sg-name], found: [sg-xxxxx sg-yyyyy sg-zzzzz]"
```

Our particular use case is the following:

We have a large set of IP ranges that we need to allow ingress from in our ALBs. Unfortunately, the number of CIDR ranges we need to allow lists is bigger than the [default maximum quota of 60 ingress rules per security group](https://docs.aws.amazon.com/vpc/latest/userguide/amazon-vpc-limits.html#vpc-limits-security-groups). To remediate that, we created multiple security groups all with different IDs but with the same `tag:Name=allow-ingress-from-xyz`, and spread the CIDR ranges across them. 

We have some automation in place to create/delete these security group resources when the number of allowed CIDR ranges changes. Hence, we cannot hard-code the security group IDs in the `alb.ingress.kubernetes.io/security-groups` annotation. Instead, we would like the controller to dynamically resolve the security group IDs to attach to the ALB from the specified security group names in the annotation.

In the current implementation, the approach described above works if and only if all the security group names specified in the annotation resolve to a **single** security group ID (1-to-1 mapping). Otherwise, the controller errors out with the error message shown above.

The approach I've taken to solve this problem is the following:

- For ID-based lookups, it is correct to assume that the number of security group IDs requested needs to match with the number of security group IDs found. That is because, within a VPC, security group IDs are unique.

- For name-based lookups, each given name needs to match one or more security groups.

To implement that, I moved the "response correctness" logic from the `ResolveViaNameOrID` to the respective `resolveViaGroupID` and `resolveViaGroupName` so that it can be slightly different on each search strategy.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/networking/security_group_resolver.go
+++ b/pkg/networking/security_group_resolver.go
@@ -7,6 +7,7 @@ import (
 	awssdk "github.com/aws/aws-sdk-go/aws"
 	ec2sdk "github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/pkg/errors"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/algorithm"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/aws/services"
 )
 
@@ -35,6 +36,7 @@ type defaultSecurityGroupResolver struct {
 func (r *defaultSecurityGroupResolver) ResolveViaNameOrID(ctx context.Context, sgNameOrIDs []string) ([]string, error) {
 	sgIDs, sgNames := r.splitIntoSgNameAndIDs(sgNameOrIDs)
 	var resolvedSGs []*ec2sdk.SecurityGroup
+
 	if len(sgIDs) > 0 {
 		sgs, err := r.resolveViaGroupID(ctx, sgIDs)
 		if err != nil {
@@ -42,6 +44,7 @@ func (r *defaultSecurityGroupResolver) ResolveViaNameOrID(ctx context.Context, s
 		}
 		resolvedSGs = append(resolvedSGs, sgs...)
 	}
+
 	if len(sgNames) > 0 {
 		sgs, err := r.resolveViaGroupName(ctx, sgNames)
 		if err != nil {
@@ -49,13 +52,12 @@ func (r *defaultSecurityGroupResolver) ResolveViaNameOrID(ctx context.Context, s
 		}
 		resolvedSGs = append(resolvedSGs, sgs...)
 	}
+
 	resolvedSGIDs := make([]string, 0, len(resolvedSGs))
 	for _, sg := range resolvedSGs {
 		resolvedSGIDs = append(resolvedSGIDs, awssdk.StringValue(sg.GroupId))
 	}
-	if len(resolvedSGIDs) != len(sgNameOrIDs) {
-		return nil, errors.Errorf("couldn't find all securityGroups, nameOrIDs: %v, found: %v", sgNameOrIDs, resolvedSGIDs)
-	}
+
 	return resolvedSGIDs, nil
 }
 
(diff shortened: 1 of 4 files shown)


shraddhabang:
Thank for your contribution. We will start reviewing this. 
/ok-to-test

shraddhabang (approved):

pkg/networking/security_group_resolver.go
```diff
@@ -83,10 +102,31 @@ func (r *defaultSecurityGroupResolver) resolveViaGroupName(ctx context.Context,
...
 		},
 	}
+
 	sgs, err := r.ec2Client.DescribeSecurityGroupsAsList(ctx, req)
 	if err != nil {
 		return nil, err
 	}
+
+	resolvedSGNames := make([]string, 0, len(sgs))
+	for _, sg := range sgs {
```
oliviassss, line 112: we need to add nil check for the `sgs` and `sg.Tags`

pkg/networking/security_group_resolver_test.go
```diff
@@ -205,13 +258,34 @@ func Test_defaultSecurityGroupResolver_ResolveViaNameOrID(t *testing.T) {
 			wantErr: errors.New("Describe.Error: unable to describe security groups"),
 		},
 		{
-			name: "unable to resolve all security groups",
+			name: "unable to resolve all security group ids",
 			args: args{
 				nameOrIDs: []string{
-					"sg group one",
```
oliviassss, line 211: can we keep a negative test case for the mix name&ids ?

pkg/networking/security_group_resolver.go
```diff
@@ -35,42 +36,60 @@ type defaultSecurityGroupResolver struct {
...
 		GroupIds: awssdk.StringSlice(sgIDs),
 	}
+
 	sgs, err := r.ec2Client.DescribeSecurityGroupsAsList(ctx, req)
 	if err != nil {
 		return nil, err
 	}
+
+	resolvedSGIDs := make([]string, 0, len(sgs))
+	for _, sg := range sgs {
```
oliviassss, line 75: same here, I think we need a nil check for the `sgs`

pkg/networking/security_group_resolver.go
alloveras: I think the Golang specification tells us that these would be unnecessarily redundant.

> A "for" statement with a "range" clause iterates through all entries of an array, slice, string or map, values received on a channel, or integer values from zero to an upper limit [[Go 1.22](https://go.dev/ref/spec#Go_1.22)]. For each entry it assigns iteration values to corresponding iteration variables if present and then executes the block. 
> ...
> For an array, pointer to array, or slice value a, the index iteration  [...] _**For a nil slice, the number of iterations is 0.**_

Source: https://go.dev/ref/spec#For_range

Example: https://go.dev/play/p/vBe3S9lbhvh

That said, if you feel strongly about having them or is a common pattern in the codebase I am more than happy to add an extra commit with the additional `nil` checks 😉

oliviassss: yeah for the `sgs` it was fine, but how about `sg.Tags`? I know in this function we are filtering by tags, so the returned sgs should have `Tags` field, but just in case if some code changes in the future and we miss this, it will error out because the controller does not check whether there's a Tag spec.

alloveras: > but how about sg.Tags?

I think for `sg.Tags` is the same. If the code changes and the `SecurityGroup` struct removes/renames the `Tags` field, that will produce a compiler error and, hence, we will detect it immediately. Or am I missing something?

pkg/networking/security_group_resolver.go
alloveras: Let's continue the discussion in: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3775#discussion_r1697739321

pkg/networking/security_group_resolver_test.go
alloveras: I pushed https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3775/commits/12a7953a3112c0478a3f5178bf16b5650910e14d to add the negative test for mix name & ids as well as improve error reporting a little bit.

oliviassss:
/lgtm
/approve
this needs a minor release, and we need to call out in the release note too. Thanks. CC: @shraddhabang

alloveras:
> this needs a minor release, and we need to call out in the release note too.

Hey @oliviassss,
I just noticed that a new patch release got recently issued and it does **not** include the fix in this PR. Do I need to do anything (Github comment, others) to ensure that this change will be included in future releases?

Thank you!

shraddhabang:
Hey @alloveras , The patch release v2.8.3 was done just to address a CVE. We have planned to inclde these changes in v2.9.0.

Issue #3706, opened 2024-05-17
Security Group Inbound is automatically set to open 0.0.0.0


Rishabh-Hupr:
Hi Team

Based on the below line,
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/9b4999b63ca90c53fb3dbe3c823b8ca7fa84ba3d/docs/guide/service/annotations.md?plain=1#L431
it mentions that if the service Object has the annotation of `service.beta.kubernetes.io/aws-load-balancer-scheme: internal` , then the source ranges for the LB will be from VPC CIDR, it's the same for v2.7 and v2.6.
However, it doesn't seem to be working.

I created a service to type `LoadBalancer` and had the same annotation in there(tried with both v2.7.2 and v2.6.2 LBC version).
I spun an internal NLB up, however one of the SG related to NLB, was allowing traffic from `0.0.0.0/0`, which is not desirable. As one would think that the doc mentioning 👇🏻, will have it's effect, but certainly not working in this case.
>- The VPC CIDR will be used if `service.beta.kubernetes.io/aws-load-balancer-scheme` is `internal` 

``service.yaml 👇🏻``
```
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: null
  namespace: kube-system
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-scheme: internal
  labels:
    app: web
  name: web
spec:
  ports:
  - port: 8080
    protocol: TCP
    targetPort: 80
  selector:
    app: web
  type: LoadBalancer
```

## Can we get a clarification why this is not working as expected? And if so, can we get a fix?
I can provide more details if needed.

uditsidana:
I can confirm the same behaviour.

oliviassss:
@Rishabh-Hupr, @uditsidana, can you check if the annotation `service.beta.kubernetes.io/load-balancer-source-ranges` or `spec.loadBalancerSourceRanges` works in your case to restrict the range?
check our live doc for more details: https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.7/guide/service/annotations/#access-control

uditsidana:
@oliviassss No issues applying the ranges using the annotation or spec.loadBalancerSourceRanges. Works perfectly, we just wanted to highlighted a particular behaviour as mentioned in the docs.

oliviassss:
thanks, looks like a flaw in doc, I will double check and fix
/kind documentation

[...]

PR #3781, opened 2024-07-24
Add Annotation Check for Load-Balancer Scheme in SG Source Ranges

yash97:
### Issue

solves #3706 
### Description
This pull request adds a new check for the Kubernetes service annotation "service.beta.kubernetes.io/aws-load-balancer-scheme": "internal".

Key changes:
1. Implemented a check for the internal load balancer scheme annotation.
2. Modified the default inbound CIDR logic:
   - If source range is not provided:
     a. For internal scheme: Default inbound CIDR will be vpcCIDR
     b. For internet-facing scheme: Default inbound CIDR remains 0.0.0.0/0

A good, clear description == a faster review :)
-->

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/service/model_build_managed_sg.go
+++ b/pkg/service/model_build_managed_sg.go
@@ -21,8 +21,8 @@ const (
 	resourceIDManagedSecurityGroup = "ManagedLBSecurityGroup"
 )
 
-func (t *defaultModelBuildTask) buildManagedSecurityGroup(ctx context.Context, ipAddressType elbv2model.IPAddressType) (*ec2model.SecurityGroup, error) {
-	sgSpec, err := t.buildManagedSecurityGroupSpec(ctx, ipAddressType)
+func (t *defaultModelBuildTask) buildManagedSecurityGroup(ctx context.Context, ipAddressType elbv2model.IPAddressType, scheme elbv2model.LoadBalancerScheme) (*ec2model.SecurityGroup, error) {
+	sgSpec, err := t.buildManagedSecurityGroupSpec(ctx, ipAddressType, scheme)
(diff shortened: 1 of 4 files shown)


pkg/service/model_build_managed_sg.go
```diff
@@ -132,9 +133,22 @@ func (t *defaultModelBuildTask) buildCIDRsFromSourceRanges(_ context.Context, ip
 		if prefixListsConfigured {
 			return cidrs, nil
 		}
-		cidrs = append(cidrs, "0.0.0.0/0")
-		if ipAddressType == elbv2model.IPAddressTypeDualStack {
-			cidrs = append(cidrs, "::/0")
+		var scheme string
+		ok := t.annotationParser.ParseStringAnnotation(annotations.SvcLBSuffixScheme, &scheme, t.service.Annotations)
+		if ok && (scheme == string(elbv2model.LoadBalancerSchemeInternal) || scheme == "") {
```
M00nF1sh, line 138: when will scheme be ""? it shouldn't be empty.

M00nF1sh (commented):
/lgtm
would you help draft a change note into the overview of this PR.
This one is a behavior change(even though it makes it matches our documentation). 
Customers that explicitly relied on previous behavior could be impacted on their internal-facing lBs and need to use sourceRanges instead.

alexku7:
should be careful with that change.
If I understand correctly, people rely on that bug for a few last years. 
likely most people even are unaware that the SG is open by that bug.
Once you fix it , the LB controller might close currently opened rule with 0.0.0.0/0 and cause outage.

PR #3785, opened 2024-07-28
[helm chart] Only create ServiceMonitor if CRD/api resource is available

asherf:
### Description

It is best practice to make sure a given api resource is defined in the cluster before trying to create it.
Example: https://github.com/fluent/helm-charts/blob/154127f95f40ea53f7955826a937340cf2b48dad/charts/fluent-bit/templates/servicemonitor.yaml#L1C1-L1C106


asherf:
@kishorj Any feedback here ?

Issue #1608, opened 2020-11-04
Dual TCP/UDP NLB shared across multiple services


hongkongkiwi:
Hi there,

I would like to share a single NLB across two services with both UDP and TCP ports open.

For example:
serviceA - Port 550 UDP
serivceB - Port 8899 TCP 

I couldn't seem to find a way to do this unless using an application load balancer and ingress routes.

Is there a way to do this in the v2.0.0 release?

The major blocker was that the targetgroup annotation was only supported at an ingress level (not a service level) so there just seems no way to share a LB.

kishorj:
The service resource models closely to the AWS NLB resource, so we don't intend to support single LB across multiple service resources. See if specifying multiple service ports helps your use case.

If you are not able to specify both TCP and UDP ports on a service resource of type `LoadBalancer`, you can try using service of type `NodePort`. The only limitation currently is that the TCP and UDP ports cannot be the same value.

hongkongkiwi:
Thanks @kishorj, that's a pretty clear explanation. I will try the NodePort approach.

philomory:
I also would love to see something akin to `alb.ingress.kubernetes.io/group.name` on NLBs (presumably as something akin to `service.beta.kubernetes.io/aws-load-balancer-group-name`).

philomory:
As far as I'm aware this shouldn't be stale. I'd still love to see this, at least.

M00nF1sh:
@philomory 
This controller currently supports NodePort service as well. If you use the NLB-IP mode annotation `service.beta.kubernetes.io/aws-load-balancer-type: "nlb-ip"` on services, even it's a nodePort service, the controller will manage the loadbalancer and targetGroups for you automatically.

Also, there have been proposal in upstream to allow dual TCP/UDP.

TBBle:
[Mixed-protocol (TCP/UDP) Service is alpha in k8s 1.20](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancers-with-mixed-protocol-types), is this the feature ticket to track for support for the `MixedProtocolLBService` feature gate?

Or is this for multiple Services contributing to a single NLB, similar to #1707 and #1545?

kishorj:
@TBBle, this is the correct issue for mixed protocol support.

Once the MixedProtocolLBService feature gate is enabled, service of type LoadBalancer with mixed protocols should work fine without further changes with the following limitations - 
[...]

PR #3807, opened 2024-08-13
Add support for TCP_UDP to NLB TargetGroups and Listeners (rebase)

lyda:
This is a work in progress - I need to test it.

### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1608#issuecomment-937346660

And based on this PR: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/2275

### Description

Previously, aws-load-balancer-controller ignored extra overlapping
ServicePorts defined in the Kubernetes Service spec if the external port
numbers were the same even if the protocols were different (e.g. TCP:53,
UDP:53).

This behavior prevented users from exposing services that support TCP
and UDP on the same external load balancer port number.

This patch solves the problem by detecting when a user defines multiple
ServicePorts for the same external load balancer port number but using
TCP and UDP protocols separately. In such situations, a TCP_UDP
TargetGroup and Listener are created and SecurityGroup rules are
updated accordingly. If more than two ServicePorts are defined, only the
first two mergeable ServicePorts are used. Otherwise, the first
ServicePort is used.

### Checklist
- [x] Manually tested

### DIFF
--- a/apis/elbv2/v1beta1/targetgroupbinding_types.go
+++ b/apis/elbv2/v1beta1/targetgroupbinding_types.go
@@ -87,6 +87,9 @@ const (
 
 	// NetworkingProtocolUDP is the UDP protocol.
 	NetworkingProtocolUDP NetworkingProtocol = "UDP"
+
+	// NetworkingProtocolTCP_UDP is the TCP_UDP protocol.
+	NetworkingProtocolTCP_UDP NetworkingProtocol = "TCP_UDP"
 )
 
 // NetworkingPort defines the port and protocol for networking rules.
--- a/apis/elbv2/v1beta1/targetgroupbinding_types.go
+++ b/apis/elbv2/v1beta1/targetgroupbinding_types.go
@@ -77,7 +77,7 @@ type NetworkingPeer struct {
 	SecurityGroup *SecurityGroup `json:"securityGroup,omitempty"`
 }
 
-// +kubebuilder:validation:Enum=TCP;UDP
+// +kubebuilder:validation:Enum=TCP;UDP;TCP_UDP
 // NetworkingProtocol defines the protocol for networking rules.
 type NetworkingProtocol string
 
--- a/pkg/service/model_build_listener.go
+++ b/pkg/service/model_build_listener.go
@@ -3,6 +3,7 @@ package service
 import (
 	"context"
 	"fmt"
+	"slices"
 	"strconv"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -35,7 +36,10 @@ func (t *defaultModelBuildTask) buildListeners(ctx context.Context, scheme elbv2
 		key := port.Port
 		if vals, exists := portMap[key]; exists {
 			if len(vals) > 1 {
-				port = mergeServicePortsForListener(vals)
+				port, err = mergeServicePortsForListener(vals)
+				if err != nil {
+					return err
+				}
 			} else {
 				port = vals[0]
 			}
@@ -257,23 +261,19 @@ func (t *defaultModelBuildTask) buildListenerAttributes(ctx context.Context, svc
 	return attributes, nil
 }
 
-func mergeServicePortsForListener(ports []corev1.ServicePort) corev1.ServicePort {
-	port0 := ports[0]
-	mergeableProtocols := map[corev1.Protocol]bool{
-		corev1.ProtocolTCP: true,
-		corev1.ProtocolUDP: true,
+func mergeServicePortsForListener(ports []corev1.ServicePort) (corev1.ServicePort, error) {
+	if len(ports) != 2 {
+		return corev1.ServicePort{}, fmt.Errorf("Can only merge two ports, not %d (%+v)", len(ports), ports)
 	}
-	if _, ok := mergeableProtocols[port0.Protocol]; !ok {
-		return port0
-	}
-	for _, port := range ports[1:] {
-		if _, ok := mergeableProtocols[port.Protocol]; !ok {
-			continue
-		}
-		if port.NodePort == port0.NodePort && port.Protocol != port0.Protocol {
-			port0.Protocol = corev1.Protocol("TCP_UDP")
-			break
+	for _, port := range ports {
+		if !slices.Contains([]string{"TCP", "UDP"}, string(port.Protocol)) {
+			return corev1.ServicePort{}, fmt.Errorf("Unsupported protocol for merging: %s", port.Protocol)
 		}
 	}
-	return port0
+	if ports[0].Protocol == ports[1].Protocol {
+		return corev1.ServicePort{}, fmt.Errorf("Protocols can't match for merging: %s", ports[0].Protocol)
+	}
+	port := ports[0]
+	port.Protocol = corev1.Protocol("TCP_UDP")
+	return port, nil
 }
--- a/pkg/service/model_build_listener.go
+++ b/pkg/service/model_build_listener.go
@@ -19,16 +19,38 @@ func (t *defaultModelBuildTask) buildListeners(ctx context.Context, scheme elbv2
 		return err
 	}
 
+	// group by listener port number
+	portMap := make(map[int32][]corev1.ServicePort)
 	for _, port := range t.service.Spec.Ports {
-		_, err := t.buildListener(ctx, port, *cfg, scheme)
-		if err != nil {
-			return err
+		key := port.Port
+		if vals, exists := portMap[key]; exists {
+			portMap[key] = append(vals, port)
+		} else {
+			portMap[key] = []corev1.ServicePort{port}
 		}
 	}
+
+	// execute build listener
+	for _, port := range t.service.Spec.Ports {
+		key := port.Port
+		if vals, exists := portMap[key]; exists {
+			if len(vals) > 1 {
+				port = mergeServicePortsForListener(vals)
+			} else {
+				port = vals[0]
+			}
+			_, err := t.buildListener(ctx, port, cfg, scheme)
+			if err != nil {
+				return err
+			}
+			delete(portMap, key)
+		}
+	}
+
 	return nil
 }
 
-func (t *defaultModelBuildTask) buildListener(ctx context.Context, port corev1.ServicePort, cfg listenerConfig,
+func (t *defaultModelBuildTask) buildListener(ctx context.Context, port corev1.ServicePort, cfg *listenerConfig,
 	scheme elbv2model.LoadBalancerScheme) (*elbv2model.Listener, error) {
 	lsSpec, err := t.buildListenerSpec(ctx, port, cfg, scheme)
 	if err != nil {
@@ -39,7 +61,7 @@ func (t *defaultModelBuildTask) buildListener(ctx context.Context, port corev1.S
 	return ls, nil
 }
 
-func (t *defaultModelBuildTask) buildListenerSpec(ctx context.Context, port corev1.ServicePort, cfg listenerConfig,
+func (t *defaultModelBuildTask) buildListenerSpec(ctx context.Context, port corev1.ServicePort, cfg *listenerConfig,
 	scheme elbv2model.LoadBalancerScheme) (elbv2model.ListenerSpec, error) {
 	tgProtocol := elbv2model.Protocol(port.Protocol)
 	listenerProtocol := elbv2model.Protocol(port.Protocol)
@@ -149,7 +171,7 @@ func validateTLSPortsSet(rawTLSPorts []string, ports []corev1.ServicePort) error
 	return nil
 }
 
-func (t *defaultModelBuildTask) buildTLSPortsSet(_ context.Context) (sets.String, error) {
+func (t *defaultModelBuildTask) buildTLSPortsSet(_ context.Context) (sets.Set[string], error) {
 	var rawTLSPorts []string
 
 	_ = t.annotationParser.ParseStringSliceAnnotation(annotations.SvcLBSuffixSSLPorts, &rawTLSPorts, t.service.Annotations)
@@ -160,7 +182,7 @@ func (t *defaultModelBuildTask) buildTLSPortsSet(_ context.Context) (sets.String
 		return nil, err
 	}
 
-	return sets.NewString(rawTLSPorts...), nil
+	return sets.New[string](rawTLSPorts...), nil
 }
 
 func (t *defaultModelBuildTask) buildBackendProtocol(_ context.Context) string {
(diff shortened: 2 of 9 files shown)


lyda:
OK, I think I've addressed the issues with the rebase.  But the verify step is now failing.  Any ideas why?

lyda:
@M00nF1sh / @oliviassss : Hey there, I think this is done and reviewable.  I've run it in a test EKS system and it does what I needed it to do.  However I'm not an expert in this code base or with load balancers so would love feedback from yourselves.  Thank you!

pkg/service/model_build_target_group_test.go
z0rc, line 1: There are a lot of mixed tab/space indentations introuduced in this file. It's quite prominent when viewing this file here in gitlab (your text editor may mask this).

lyda: The new commit should fix that.

lyda: Done.

In general it would likely be better to put json objects in json files so you'd avoid this mix of formatting; just have vim pipe the buffer to jq. But that sort of change seemed out of scope.

lyda:
I've done manual testing and it worked fine.  There really aren't docs to update.  The information that the current aws-load-balancer-controller doesn't support listening to UDP and TCP on the same port number is information you'll find in the issues.

z0rc:
> Is there a slack/discord/irc channel I can chat about this to address any remaining issues?

I believe developers (I'm not one of them) are somewhat reachable via https://slack.k8s.io/, channel `#provider-aws`.

apis/elbv2/v1beta1/targetgroupbinding_types.go
```diff
@@ -87,6 +87,9 @@ const (
 
 	// NetworkingProtocolUDP is the UDP protocol.
 	NetworkingProtocolUDP NetworkingProtocol = "UDP"
+
+	// NetworkingProtocolTCP_UDP is the TCP_UDP protocol.
+	NetworkingProtocolTCP_UDP NetworkingProtocol = "TCP_UDP"
```
M00nF1sh, line 92: we don't need this.
the networking rules here is meant for security groups(which don't have a tcp_udp rule protocol)

instead of TCP_UDP, we should just generate two rules in TGB's networking, one for TCP and another one for UDP.

pkg/service/model_build_listener.go
```diff
@@ -19,16 +19,38 @@ func (t *defaultModelBuildTask) buildListeners(ctx context.Context, scheme elbv2
...
+		key := port.Port
+		if vals, exists := portMap[key]; exists {
+			portMap[key] = append(vals, port)
+		} else {
+			portMap[key] = []corev1.ServicePort{port}
 		}
 	}
+
+	// execute build listener
+	for _, port := range t.service.Spec.Ports {
```
M00nF1sh, line 34: nit: i prefer to loop over `portMap` instead of `t.service.Spec.Ports`

pkg/service/model_build_listener.go
```diff
@@ -234,3 +256,24 @@ func (t *defaultModelBuildTask) buildListenerAttributes(ctx context.Context, svc
 	}
 	return attributes, nil
 }
+
+func mergeServicePortsForListener(ports []corev1.ServicePort) corev1.ServicePort {
```
M00nF1sh, line 260: this works but have two issues:
1. we are hacking the k8s `corev1.ServicePort` concept by introduce a new `corev1.Protocol("TCP_UDP")`, which don't exists in k8s, and in theory is an malformed `corev1.ServicePort. (TCP_UDP is not a validate Protocol enum)
2. this don't generate a warning/error if unsupported protocol/protocol conbination is found.
3. this exists as a "glue" code to make the existing `buildListeners` function happy, but in the long run would make the code harder to understand and maintain.


Instead, i think a better option would be:
1. modify buildListeners to accept a list of corev1.ServicePort. inside buildListeners, it first validate whether this list of corev1.ServicePort can form a single listener(and error out when cannot), and generate listener of correct protocol.

TBBle:
Does it make sense to mark this PR as closing #1608? That should cause GitHub to link to this PR from that ticket, rather than relying on the comments. (It might also keep k8s-ci-robot from closing that issue as stale while this PR is in-progress... maybe?)

I note that this PR isn't solving the actual use-case for which that ticket was opened, but the bulk of the discussion in that ticket was for this use-case, so tying them together makes sense to me. A solution for the original use-case would revolve around some kind of new `LoadBalancerInstance` CRD feature and is basically orthogonal to this PR.

lyda:
Awesome, thanks for the review.  I'll work through these this week.

pkg/service/model_build_listener.go
lyda: I'm confused, that's what `buildListeners` now does. It loops through the `corev1.ServicePort`s and determines if there's a case where the same port number exists for `TCP` and `UDP`. It then synthesises a single `corev1.ServicePort` from the original two that uses the `TCP_UDP` protocol that AWS load balancers support.

Either way it then calls `buildListener`.

I could change everything from `buildListener` on down to use an internal data structure that includes a slice of `corev1.ServicePort`s as well as the protocol being used for the load balancer.  That seems like a rather complex change. And it will still have the rats nest of conditionals the current code has. Plus all `corev1.ServicePort` stuff would operate off the first element - so might as well just include a single `corev1.ServicePort`. And at that point all I've done is avoided abusing an enum.

That said, point number two is definitely valid and I'm now returning errors for a few scenarios up the chain.

lyda:
> Does it make sense to mark this PR as closing #1608? That should cause GitHub to link to this PR from that ticket, rather than relying on the comments. (It might also keep k8s-ci-robot from closing that issue as stale while this PR is in-progress... maybe?)
> 
> I note that this PR isn't solving the actual use-case for which that ticket was opened, but the bulk of the discussion in that ticket was for this use-case, so tying them together makes sense to me. A solution for the original use-case would revolve around some kind of new `LoadBalancerInstance` CRD feature and is basically orthogonal to this PR.

Good point.  I'll turn that off because this is completely different to the original issue and just addresses a comment within. Sorry I missed that.

lyda (commented):
OK, I think I've gone through them all.  I see your point on `buildListener` but I don't think the change will make it any better. The code will still have to pick through the three "protocols" (I really do dislike AWS's `TCP_UDP` "feature" here so I have to put quotes around protocols at least once) and that's going to make the config translation code more complex than just the binary choices before.

The real decision is: do you want to support this feature for the few of us who need it and pay the complexity price or do you not. I completely understand either choice. I would prefer the former obviously because I have zero interest in maintaining a fork, but that's a me problem, not a you problem.

pkg/service/model_build_listener_test.go
```diff
@@ -301,3 +301,220 @@ func Test_defaultModelBuilderTask_buildListenerAttributes(t *testing.T) {
 		})
 	}
 }
+
+func Test_mergeServicePortsForListener(t *testing.T) {
```
TBBle, line 305: Are these tests out-of-date? They don't seem to match the function signature of `mergeServicePortsForListener`.

pkg/service/model_build_listener.go
```diff
@@ -234,3 +260,20 @@ func (t *defaultModelBuildTask) buildListenerAttributes(ctx context.Context, svc
...
+	for _, port := range ports {
+		if !slices.Contains([]string{"TCP", "UDP"}, string(port.Protocol)) {
+			return corev1.ServicePort{}, fmt.Errorf("Unsupported protocol for merging: %s", port.Protocol)
+		}
+	}
+	if ports[0].Protocol == ports[1].Protocol {
+		return corev1.ServicePort{}, fmt.Errorf("Protocols can't match for merging: %s", ports[0].Protocol)
+	}
+	port := ports[0]
+	port.Protocol = corev1.Protocol("TCP_UDP")
```
TBBle, line 277: Given that we're taking just the first listed port's details, should we also validate that targetPort and/or nodePort match? (Which one matters I guess depends on whether it's an Instance or IP target group, which we can't see in this function)

pkg/service/model_build_listener.go
```diff
@@ -39,7 +65,7 @@ func (t *defaultModelBuildTask) buildListener(ctx context.Context, port corev1.S
 	return ls, nil
 }
 
-func (t *defaultModelBuildTask) buildListenerSpec(ctx context.Context, port corev1.ServicePort, cfg listenerConfig,
+func (t *defaultModelBuildTask) buildListenerSpec(ctx context.Context, port corev1.ServicePort, cfg *listenerConfig,
```
TBBle, line 68: I can't comment directly on it, but I think the logic for switching to `ProtocolTLS` immediately below probably needs to be specific to `ProtocolTCP` rather than `!ProtocolUDP` as I don't think turning a `TCP_UDP` port into a TLS port is a valid conversion. (Maybe it should fail if we get a `TCP_UDP` port where the TCP side would match the TLS configuration? (This might be the only place we use `port.Name`, from a brief check, which would also resolve the concern that the merged `TCP_UDP` port gains the name of the first-processed `ServicePort`, and loses the second's name.)

pkg/service/model_build_target_group.go
```diff
@@ -484,6 +484,8 @@ func (t *defaultModelBuildTask) buildTargetGroupBindingNetworking(_ context.Cont
 				Protocol: &protocolTCP,
 				Port:     &tgPort,
 			})
+		case corev1.Protocol("TCP_UDP"):
+			fallthrough
```
TBBle, line 488: Do we not need to build a port-list with both TCP and UDP here? `buildTargetGroupBindingNetworkingLegacy` does it that way, but I admit I'm not familiar with the difference in these two functions.

I also note that the test for this function has not had a `TCP_UDP` test added, while `buildTargetGroupBindingNetworkingLegacy`'s test has been updated.

apis/elbv2/v1beta1/targetgroupbinding_types.go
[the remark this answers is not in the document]
lyda: Ah.  Yes, added that.

lyda: Hm, this type is used for a lot of things. For some `TCP`, `UDP` and `TCP_UDP` are allowed. For others, only `TCP` and `UDP`.

Should I split out the types across the code base.  It makes the PR bigger, but it would work better with kubebuilder.

lyda (commented):
Ugh, I hate Github's UI. OK, comments for some of what's been addressed.

pkg/service/model_build_listener_test.go
lyda: Sorry, yes they were.  Fixed. Thanks, nice catch!

pkg/service/model_build_listener.go
TBBle: It only grouped by port, but that doesn't ensure the targetPort and/or nodePort are identical. If I recall correctly, there was a test-case which showed this, and is currently a `success:true` case, but once that merged port is given to the rest of the code, it'll produce a firewall entry that is sending TCP traffic to the right nodePort, but then the UDP traffic would be sent to the _same_ nodePort even though the service's nodePort for UDP was different.

We can't resolve that case with TCP_UDP, so it probably should fail in this function, as we can't catch that case in code that receives the merged `ServicePort`.

It's complicated a little because if we're using IP target group mode, then the nodePorts won't be used, and so it won't matter if they don't match, and similarly for the targetPort values in instance target group mode. So maybe this function needs to know if we're working with an instance target group or IP target group, so it knows which of `nodePort` and `targetPort` to validate.

pkg/service/model_build_listener.go
TBBle: At this point, I don't see the difference in the current loop and this:

```go
// execute build listener
for _, vals := range portMap {
	var port corev1.ServicePort
	if len(vals) > 1 {
		port, err = mergeServicePortsForListener(vals)
		if err != nil {
			return err
		}
	} else {
		port = vals[0]
	}
	_, err := t.buildListener(ctx, port, cfg, scheme)
	if err != nil {
		return err
	}
}
```

except that this doesn't modify `t.service.Spec.Ports` in-place. Is that in-place modification an expected side-effect here? If so, then the current loop will not remove the second of the two merged ports after replacing the first one with the result of `mergeServicePortsForListener`. (That in-place modification is also an example of the leakage of the `TCP_UDP` protocol name outside this code, but I haven't checked to see if this actually has any effect).

`portMap` in this version of the loop could probably be `map[int32][]*corev1.ServicePort` if we're not modifying `t.service.Spec.Ports` in-place, saving a bunch of string-copying. That `var port corev1.ServicePort` could probably also be a pointer, if I remember my Go lifetime rules correctly, and we could even pass a pointer to the `ServicePort` down into `buildListener`, as I see that was already done with `cfg` in this PR. (That said, I'm speculating on GitHub; I haven't asked the compiler to validate this particular idea.)

zarko-a:
Any chance this can move forward? We were looking to implement HTTP/3 and lack of support for TCP_UDP in AWS LBC is a blocker.

devusb:
yeah same — would love to see this merged.

lyda:
I've tried.  I have unanswered questions in the PR.

The feedback I got was fine, but in order to do it this PR will get a fair bit larger.  And It's not clear to me that it's possible to split up. In addition, the speed at which I get feedback on PRs means that splitting it would take years to get the work done.

So I'm kinda stuck.  I have other things I need to work on at the moment and my work project that needs this PR is on hold at the moment. If folks have ideas on how to get the project maintainers to engage, that would be great.

zarko-a:
Thanks @lyda . I definitely feel your frustration. My comment was primarily aimed at maintainers in hopes of getting them engaged. 

Multiple protocols are supported by both K8S and AWS NLB for a long time, it's LB Controller that can't link the two. 

@M00nF1sh @TBBle this is second attempt at the community implementing this feature. [First one](https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/2275) dates back to 2021. Which indicates that there are multiple customers that are looking for this so much they are willing to contribute their own time to help make this happen. 

What can we do get this on some reasonable timeline?

zac-nixon:
Hi. I'm sorry about the delayed responses. I will make sure to review this today.

apis/elbv2/v1beta1/targetgroupbinding_types.go
zac-nixon: I agree that adding this is not needed. I'm submitting a refactor commit for the proposed change.

zac-nixon:
I had to rebase and fix multiple things. I opened a new PR here.
https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4161

LMK if you want to work off those PR, if not, I will close it.

Thank you for your patience and getting the ball rolling on this feature!

lyda:
Looks good, closing this one.

PR #3808, opened 2024-08-13
New Feature: Multi Cluster TargetGroupBindings

Alex-Waring:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2173

### Description

This PR attempts to implement the feature requested in the issue linked.

Problem: When running TargetGroupBinding resources on target groups that are managed elsewhere as well (either by another cluster or by manual attachment) the controller will remove any targets it did not add.

Reason: In `defaultResourceManager.reconcileWithIPTargetType`, the targets in an TG are listed, compared with the expected from within the cluster, and those that are not expected are removed. This is so that pods or instances that have been removed from the filter will be deregistered as well.

Solution: I have created a new function called ListOwnedTargets, gated behind an option on the controller, that will only return targets that the cluster manages. This is done by checking the IP is within the EKS subnets. To do this I have implemented a new EKS client and info resolver to return that information from the cluster. 

This doesn't cover instances, details on how this can be done are outlined in https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3478, but https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3478#discussion_r1714157235 from @shraddhabang asked that I not add instances as this is just a stop gap.

I have not added docs yet, as I would like input that that is a valid solution before I write docs.

The suggested solution in the linked issue was to use a config map to store the targets. Configmaps are limited to 1Mb, and are not designed for large volumes of data. That means that the solution is not scalable and does not work for larger clients.

### Checklist
- [x] Manually tested


Alex-Waring:
@shraddhabang As promised

Bharat23:
> This is done by checking the IP is within the EKS subnets
I think that a great technique to solve a subset of the problem, but if other targets (not managed by this TGB) are in the same subnet, then this logic will fall short. 
I think a better logic would be to tap into service event and register and deregister target (in target group) based on when it is registered and registered in svc.

Alex-Waring:
Hi @Bharat23, that's a valid point but there are a couple of reasons I went with this approach:

- The aim of this feature is multi cluster target groups, not to support other targets. Adding in logic for lambdas and other types of targets makes this a lot more complicated. It can be expanded on but it's not a problem I'm trying to solve.
- EKS clusters can share subnets but it's very much bad practice to do so as it leads to IP exhaustion.
- querying a list of all service endpoints puts a strain on the API server that doesn't scale very well with large clusters, in a way that looking up IPs doesn't.

This very much isn't a perfect solution, but it's one that meets a use case until a better one, that also addresses instances and is stateful, can be created.

shraddhabang:
We will start taking a look at this actively. Thanks.

micahnoland:
> * EKS clusters can share subnets but it's very much bad practice to do so as it leads to IP exhaustion.

We want to do blue/green EKS cluster cutovers within the same subnets, and would like to use Multi Cluster TargetGroupBindings to achieve this.

Robsta86:
> > * EKS clusters can share subnets but it's very much bad practice to do so as it leads to IP exhaustion.
> 
> 
> 
> We want to do blue/green EKS cluster cutovers within the same subnets, and would like to use Multi Cluster TargetGroupBindings to achieve this.
> 
> 

Same here :-)

Bharat23:
> EKS clusters can share subnets but it's very much bad practice to do so as it leads to IP exhaustion.

IP exhaustion depends on how big of a subnet has been created.

Our requirement is to support multiple sources, i.e., both EKS and non-EKS. E.g., EC2 instance through ASG and EKS regardless of the subnet they are in. 
AWS products like ASG/ECS services manage this scenario really and only manipulate their own targets.

M00nF1sh:
Sorry for the late review, i have a few high level comments:
1. it couples the controller with EKS API, which i prefer not to as it supposed to work on non-eks clusters like KOPS. If we really wants to differs by subnets, maybe use an optional flag to specify the subnetIDs or better CIDRs for this feature?
    * the resourcesVpcConfig.SubnetIDs from EKS API is meant to specify subnets used for EKS control-plane's cross-account ENIs. Semantically, it's not necessarily all subnets of work nodes, thus i prefer to use a flag to pass-in subnets/cidrs instead of couple with this inaccurate info from EKS API.
    * also, from the comments in this issue, seems there are users wants to reuse same subnet. For this case, this implementation won't work and requires some state tracking like configMap/endpoints.  But i'm ok with this implementation as it's for a unique use case that subnets/cidrs differ per cluster. Maybe we can extend this implementation in the future.
2. Nit, Instead of add a `ListOwnedTargets` , maybe we can extend the original `ListTargets` to ListTargets(ctx context.Context, tgARN string, opts  ...ListTargetsOption) ([]TargetInfo, error). (similar to https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/backend/endpoint_resolver.go#L30, where we can have ListTargetsOption that did CIDR based filter, and maybe in the future can extend to add filter based on "known added" targets for the use case of share subnets). This way we can share some of the implementation in ListTargets.

Alex-Waring:
Thanks @M00nF1sh that all makes sense, I'll look to implement those points and I definitely think this is a step towards target tracking. Unfortunately I won't have time for a week or so.

To all the comments on this, I hear you and understand this isn't the final solution you want, but hopefully it's a step in the right direction.

pkg/targetgroupbinding/targets_manager.go
```diff
@@ -145,6 +154,60 @@ func (m *cachedTargetsManager) ListTargets(ctx context.Context, tgARN string) ([
...
+	if rawTargetsCacheItem, exists := m.targetsCache.Get(tgARN); exists {
+		targetsCacheItem := rawTargetsCacheItem.(*targetsCacheItem)
+		targetsCacheItem.mutex.Lock()
+		defer targetsCacheItem.mutex.Unlock()
+		refreshedTargets, err := m.refreshUnhealthyTargets(ctx, tgARN, targetsCacheItem.targets)
+		if err != nil {
+			return nil, err
+		}
+		targetsCacheItem.targets = refreshedTargets
+		return cloneTargetInfoSlice(refreshedTargets), nil
```
zac-nixon, line 172: There is a slight bug here, when we use take targets from the cache we do not perform filtering and end up regressing on the multi-cluster behavior.

zac-nixon (commented):
Hi! I am new to the project, I apologize for the random review. Please let me know what you think, or if the bug I call out is not really a bug. I tested out your PR in my own clusters.

Issue #2688, opened 2022-06-13
Support for referencing existing security groups for inbound traffic


vincentheet:
**Is your feature request related to a problem?**
Currently it is possible to reference CIDR's using the `alb.ingress.kubernetes.io/inbound-cidrs` annotation. This is great for allowing traffic from the public IP space. But for internal ALB's it would be great to have the ability to allow inbound traffic from specific security groups. For example allowing traffic from the API Gateway (via VPC Link) to an ALB without opening up the ALB to the whole subnet or VPC. I would rather reference the security group of the VPC Link. But one can think of other use cases where a specific EC2 instance (not part of the EKS cluster) should be able to connect to an ALB and another EC2 should not.

**Describe the solution you'd like**
I would like to have an annotation for example `alb.ingress.kubernetes.io/inbound-security-groups` where I can specify a list of security groups that are allowed to send traffic to the listeners of the ALB.

**Describe alternatives you've considered**
Configure a custom security group using the `alb.ingress.kubernetes.io/security-groups` annotation. And reference the SG of the API Gateway in the SG to attach to the ALB. But this would be *extra work* since I then have to create a security group that would otherwise be managed by the load balancer controller.

runningman:
Would using the `security-groups` annotation in combination with `alb.ingress.kubernetes.io/manage-backend-security-group-rules: "true"` solve the issue? This way the load balancer controller still manages the node/pod sg's which I think you mean. This was the problem I was having and I came across this post when searching for a solution. It looks like this option was added in 2.4

vincentheet:
> Would using the `security-groups` annotation in combination with `alb.ingress.kubernetes.io/manage-backend-security-group-rules: "true"` solve the issue? This way the load balancer controller still manages the node/pod sg's which I think you mean. This was the problem I was having and I came across this post when searching for a solution. It looks like this option was added in 2.4

No it won't. When using both property's the loadbalancer controller creates a managed backend security group which is attached to the ALB. That security group is being referenced in the nodes security group as an inbound rule source. 
The security group that is configured and attached to the ALB based on the `alb.ingress.kubernetes.io/security-groups` annotation should be created and configured manually to allow either traffic from a CIDR or another security group.

[...]

PR #3829, opened 2024-09-01
Add referencing existing security groups for inbound traffic

changhyuni:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2688

### Description
To enhance security and enable more flexible management of security groups, we are adding a security group source chaining feature to inbound security groups. This is great for allowing traffic from public IP spaces, but for internal ALBs, it would be ideal to allow inbound traffic from specific security groups. For instance, allowing API Gateway traffic (via VPC Link) to an ALB without exposing the ALB to the entire subnet or VPC. It would be preferable to reference the security group of the VPC Link. Similarly, you might want to allow a specific EC2 instance (not part of the EKS cluster) to connect to an ALB while restricting access for another EC2 instance.

### Checklist
- [ ] Manually tested

### DIFF
--- a/docs/guide/ingress/annotations.md
+++ b/docs/guide/ingress/annotations.md
@@ -530,6 +530,24 @@ Access control for LoadBalancer can be controlled with following annotations:
         ```
         alb.ingress.kubernetes.io/inbound-cidrs: 10.0.0.0/24
         ```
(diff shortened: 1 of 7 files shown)


changhyuni:
@shraddhabang Hi.
Could you take a look at the PR when you have time?

toneill818 (approved):

Issue #3823, opened 2024-08-27
Make `leaseDurationSeconds` of a lease owned by AWS Load Balancer Controller configurable


gjbravi:
**Is your feature request related to a problem?**
Yes, sometimes the default lease duration time of 15s of the lease `aws-load-balancer-controller-leader` is too high.

**Describe the solution you'd like**
I would like to configure the `leaseDurationSeconds` of a lease owned by AWS Load Balancer Controller.

**Describe alternatives you've considered**
Controller-runtime already provides a way of configuring LeaseTime of a Lease in https://github.com/kubernetes-sigs/controller-runtime/blob/e6c3d139d2b6c286b1dbba6b6a95919159cfe655/pkg/manager/manager.go#L210-L213

shraddhabang:
Hello @gjbravi , I agree. This would be a good feature to implement on the controller. Thank you for bringing this to our attention. I am marking this as a good-first-issue. We welcome any community contributions for such issues.

caiofralmeida:
@gjbravi, it would be a good idea for us to contribute to the implementation, what do you think?

gjbravi:
For sure @caiofralmeida.

@shraddhabang that's awesome, we're gonna work on that.

PR #3835, opened 2024-09-03
feat: make LeaseDuration configurable via `--leader-election-lease-duration` flag

gjbravi:
### Issue

resolves https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3823

### Description

This PR adds a new flag to the controller, `--leader-election-lease-duration` will be responsible for externalizing and controlling the `LeaseDuration` of a `Lease` owned by aws-load-balancer-controller. 

`--leader-election-lease-duration` defaults to 15s if not informed, matching the default value of a `LeaseDuration` from [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/blob/38546806f2faf5973e3321a7bd5bb3afdbb5767d/pkg/manager/internal.go#L50-L52).

### Checklist
- [x] Manually tested


gjbravi:
The AWS LB controller chart version from eks-charts repo points to `1.8.2` as follows:

https://github.com/aws/eks-charts/blob/master/stable/aws-load-balancer-controller/Chart.yaml#L4

While, here is still on `1.7.1`:

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/e1d32f4f83eebc2d7bc61f600dc5e22770102e3e/helm/aws-load-balancer-controller/Chart.yaml#L1-L5

I would really appreciate some guidance on how the chart bump should be made. Ty

gjbravi:
@shraddhabang thanks for reviewing!

I've also submitted the chart bump on https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3835/commits/f8c0943c4d9824b11ac8b83bf00f89491baa1389 alongside all changes made and the PR labels were removed. Could you please re-check it? Ty

oliviassss:
@gjbravi, thanks for the contribution, I left some minor comments, overal LGTM

caiofralmeida:
> /lgtm /approve

@shraddhabang the approved label is missing (maybe should be "/approved" ?), could you help us to make this PR mergeable?

gjbravi:
@oliviassss tks for reviewing.

The current behavior should be equal to the default value brought by controller-runtime as far as I see. Implying that, if not specified in the values file, it will take the default value of 15 seconds.

During the implementation, I've also tested setting a "null" value to the const, however the [time pkg](https://pkg.go.dev/time) does not allow that, forcing me to specify a value(same as the implementation of SyncPeriod).

I can perhaps be missing something here, please let me know in case another solution is also viable here. Tks

blackdesert575:
Any updates?

PR #3847, opened 2024-09-11
Allow disabling ingress validation via helm flag

### LINKED ISSUE #2801: Option to disable ingress validation webhook

sarevalo2002:
### Issue

Fixes #2801.

### Description

Added a flag `webhookConfig.disableIngressValidation` which when set to true, will disable Ingress validation for those who don't want Ingress resources to be validated. This is a re-implementation of the changes made in #3011 due to popular demand.

#### Manual Testing Performed

First confirmed that manually editing `validatingwebhookconfigurations` via deleting the code block corresponding to validating ingress resulted in ingress validation being disabled (an invalid ingress was allowed to be applied after this update). Then confirmed that calling `helm upgrade` with the new flag `--set webhookConfig.disableIngressValidation=true` resulted in the relevant code block mentioned previously being removed.

### Checklist
- [X] Manually tested


oliviassss:
@sarevalo2002, can you please add the manual tests you've verified in the PR description?
/LGTM

sarevalo2002:
@oliviassss Done! If you'd like more detail, please let me know.

PR #3866, opened 2024-09-27
[helm chart] add default issuerRef.group for selfsigned cert

Azahorscak:
### Issue

[3867](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3867)

### Description

Certificates will fail to automatically renew when newer versions of cert-manager (tested on 1.15.2 and 1.15.3) don't have a default Issuer defined in the cluster and the certificate does not specify `issuerRef.group`. It Fails with `Unknown issuer kind: Issuer`

Since cert-manager is considered a pre-req to use the self-signed issuer/cert config in this chart, the chart should define this value by default when `enableCertManager` is set to true so that no further config is required for cert-manager to function as expected. 

This PR adds this field and sets it to `cert-manager.io`

### Checklist
- [x] Manually tested

### DIFF
--- a/helm/aws-load-balancer-controller/templates/webhook.yaml
+++ b/helm/aws-load-balancer-controller/templates/webhook.yaml
@@ -235,6 +235,7 @@ spec:
   - {{ template "aws-load-balancer-controller.webhookService" . }}.{{ .Release.Namespace }}.svc.{{ .Values.cluster.dnsDomain }}
   issuerRef:
     kind: Issuer
+    group: cert-manager.io
     name: {{ template "aws-load-balancer-controller.namePrefix" . }}-selfsigned-issuer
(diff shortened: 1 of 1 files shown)


mloiseleur:
This field is present [in the CRD since v1.0.0 of cert-manager](https://doc.crds.dev/github.com/jetstack/cert-manager/cert-manager.io/Certificate/v1@v1.0.0), so I guess this change is safe.
/lgtm

Azahorscak:
Hi a friendly reminder that this is good to go @M00nF1sh @johngmyers @kishorj

Azahorscak:
IMO this can be closed now as the origin-ca-issuer has resolved the issue, see - https://github.com/cloudflare/origin-ca-issuer/issues/149

PR #3886, opened 2024-10-10
add sagemaker-hyperpod compute type to resolve its pods via VPC ENI

amber-liu-amzn:
### Issue

[Sagemaker HyperPod](https://aws.amazon.com/blogs/aws/amazon-sagemaker-hyperpod-introduces-amazon-eks-support/) offers service-managed Kubernetes nodes accessible from customer accounts. Using [aws-load-balancer-controller](https://github.com/kubernetes-sigs/aws-load-balancer-controller) in HyperPod EKS clusters is not supported today, because nodes are in SageMaker VPC while load balancers will be in customer VPC.

**Why is it not working today for routing traffic directly to the HyperPod's pod IP?**
SageMaker HyperPod pods are in a different VPC, but LBC incorrectly maps them as EC2 pods in customer VPC, leading to incorrect ENI info retrieval and missing security group permissions.

### Description

To enable IP target mode for pods running on SageMaker HyperPod, this PR is to add sagemaker-hyperpod as a new compute type to resolve its pods via VPC ENI. 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/networking/pod_eni_info_resolver.go
+++ b/pkg/networking/pod_eni_info_resolver.go
@@ -142,20 +142,20 @@ func (r *defaultPodENIInfoResolver) saveENIInfosToCache(pods []k8s.PodInfo, eniI
 }
 
 func (r *defaultPodENIInfoResolver) resolvePodsViaCascadedLookup(ctx context.Context, pods []k8s.PodInfo) (map[types.NamespacedName]ENIInfo, error) {
-	podsOnEc2, podsOnFargate, podsOnSageMakerHyperPod, err := r.classifyPodsByComputeType(ctx, pods)
+	podsByComputeType, err := r.classifyPodsByComputeType(ctx, pods)
 	if err != nil {
 		return nil, err
 	}
 	eniInfoByPodKey := make(map[types.NamespacedName]ENIInfo)
-	if len(podsOnEc2) > 0 {
-		eniInfoByPodKeyEc2, err := r.resolveViaCascadedLookup(ctx, podsOnEc2, false)
+	if len(podsByComputeType.ec2Pods) > 0 {
+		eniInfoByPodKeyEc2, err := r.resolveViaCascadedLookup(ctx, podsByComputeType.ec2Pods, false)
 		if err != nil {
 			return nil, err
 		}
 		eniInfoByPodKey = eniInfoByPodKeyEc2
 	}
-	if len(podsOnFargate) > 0 {
-		eniInfoByPodKeyFargate, err := r.resolveViaCascadedLookup(ctx, podsOnFargate, true)
+	if len(podsByComputeType.fargatePods) > 0 {
+		eniInfoByPodKeyFargate, err := r.resolveViaCascadedLookup(ctx, podsByComputeType.fargatePods, true)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


pkg/networking/pod_eni_info_resolver.go
```diff
@@ -164,17 +165,28 @@ func (r *defaultPodENIInfoResolver) resolvePodsViaCascadedLookup(ctx context.Con
...
-func (r *defaultPodENIInfoResolver) resolveViaCascadedLookup(ctx context.Context, pods []k8s.PodInfo, isFargateNode bool) (map[types.NamespacedName]ENIInfo, error) {
+func (r *defaultPodENIInfoResolver) resolveViaCascadedLookup(ctx context.Context, pods []k8s.PodInfo, isFargateNode bool, isSageMakerHyperPodNode bool) (map[types.NamespacedName]ENIInfo, error) {
 	eniInfoByPodKey := make(map[types.NamespacedName]ENIInfo)
 	resolveFuncs := []func(ctx context.Context, pods []k8s.PodInfo) (map[types.NamespacedName]ENIInfo, error){
 		r.resolveViaPodENIAnnotation,
 		r.resolveViaNodeENIs,
 		// TODO, add support for kubenet CNI plugin(kops) by resolve via routeTable.
 	}
-	if isFargateNode {
+	if isFargateNode || isSageMakerHyperPodNode {
```
shraddhabang, line 189: Cam we change this to check for non-ec2 pod instead of having or condition for Fargate and Sagemaker pod? 
This way we will only need to send one boolean value in this function and make decision based on that.

amber-liu-amzn: yeah. makes sense. Addressed in latest commit. thanks!

pkg/networking/pod_eni_info_resolver.go
```diff
@@ -388,33 +401,39 @@ func (r *defaultPodENIInfoResolver) isPodSupportedByNodeENI(pod k8s.PodInfo, nod
 	return false
 }
 
-// classifyPodsByComputeType classifies in to ec2 and fargate groups
-func (r *defaultPodENIInfoResolver) classifyPodsByComputeType(ctx context.Context, pods []k8s.PodInfo) ([]k8s.PodInfo, []k8s.PodInfo, error) {
+// classifyPodsByComputeType classifies in to ec2, fargate and sagemaker-hyperpod groups
+func (r *defaultPodENIInfoResolver) classifyPodsByComputeType(ctx context.Context, pods []k8s.PodInfo) ([]k8s.PodInfo, []k8s.PodInfo, []k8s.PodInfo, error) {
```
zac-nixon, line 405: Can you refactor this to be more type friendly? E.g. it's non-trivial to remember arg1 is ec2, arg2 is fargate, arg3 is hyperpod.

Can you introduce a new type that basically is 

struct{
  ec2Pods []k8s.PodInfo
  fargatePods []k8s.PodInfo
  hyperpodPod []k8s.PodInfo
}

this will make the code cleaner and as we add more compute types it would be more extensible.

zac-nixon (changes requested):
Were you able to validate in an environment that has EC2, Fargate, and sagemaker-hyperpod compute to ensure everything works?

pkg/networking/pod_eni_info_resolver.go
amber-liu-amzn: yup. refactored in the last commit. thanks!

amber-liu-amzn:
> Were you able to validate in an environment that has EC2, Fargate, and sagemaker-hyperpod compute to ensure everything works?

Good question! I will try to set up such mixed cluster and verify. will get back to you on this

zac-nixon (approved):

shraddhabang (approved):

PR #3918, opened 2024-10-29
[helm chart] add resource-policy annotations to crds

aburan28:
### Issue

Guidance on upgrades for AWS LB controller includes references to uninstalling the old version before moving to the new version of the controller. In those cases it would be ideal if we could retain the CRDs during that uninstall so as to prevent downtime associated with a fresh install.

### Description
CRDs are deleted by default when the helm chart is uninstalled. Offering an option to retain those resources would be great. Addittionally it would be ideal to pass in annotations to the CRDs. 

In this PR I've added the following section which allow for configuring whether or not the helm `"helm.sh/resource-policy": keep` annotation is added to the CRDs, whether or not the CRDs are installed, and additional annotations that could be passed into the CRDs.

```yaml
crds:
  keep: false
  install: true
  annotation: {}
```

### Checklist
- [x] Manually tested

### DIFF
--- a/helm/aws-load-balancer-controller/crds/crds.yaml
+++ b/helm/aws-load-balancer-controller/crds/crds.yaml
@@ -1,8 +1,15 @@
+{{- if .Values.crds.install }}
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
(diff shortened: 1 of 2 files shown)


aburan28:
was wondering if we could have an update on this PR and what is needed to get this approved or if there are any concerns with this approach

PR #3926, opened 2024-10-31
UDP Support over IPv6 via Dualstack NLBs using SourceNAT configurations

shethyogita83:
### Issue
Until now, UDP on NLB was supported only over IPv4 address type NLBs because the NLB preserves the client-IP for IPv4 address type NLBs.
Dualstack NLBs and PrivateLink NLBs are required to do sourceNAT and thus cannot support UDP because in cases where NLB doesn’t preserve client-IP but translates (sourceNAT) it to its own IP, there is a risk of packets being mis-routed between clients.
But now we want to support UPD over IPv6 address type NLBs too since this feature allow customers (service owners) to provide UDP-based services such as media streaming, gaming, remote desktop applications, VoIP, etc., over IPv6 too.
### Description
To support UDP for NLB over IPv6 on K8 controller Service, We introduced two new optional annotations on K8 Service annotations, for the corresponding API input Parameters on CreateLoadBalancer API and SetSubnets API
Optional Annotation 1 -  [service.beta.kubernetes.io/aws-load-balancer-enable-prefix-for-ipv6-source-nat](http://service.beta.kubernetes.io/aws-load-balancer-enable-prefix-for-ipv6-source-nat): Values can be either “on” or “off”. Default value is off. Corresponding API Input property is EnablePrefixForIpv6SourceNat in CreateLoadBalancer API and SetSubnets API
Optional Annotation 2 - [service.beta.kubernetes.io/aws-load-balancer-source-nat-ipv6-prefixes](http://service.beta.kubernetes.io/aws-load-balancer-source-nat-ipv6-prefixes) : Value is a list of ipv6Prefixes or if user doesn’t want to provide particular ipv6Prefixe value, user can use “auto-assigned”.
ipv6Prefixes list should have same length as the number of subnets. Each Ipv6 prefix in the list will be applied to each corresponding subnet in the subnets annotation list in the same order. A valid ipv6 prefix ends with /80 and it should be within the CIDR range of the corresponding subnet (e.g. 2600:1f16:ed6:312:98c3::/80). Corresponding API Input property is SourceNatIpv6Prefixes in CreateLoadBalancer API and SetSubnets API
This PR is very similar to the old PR https://github.com/shethyogita83/private-aws-load-balancer-controller/pull/1, but with required modifications to suit the aws-sdk-go version update.
### Checklist
- [X] Manually tested


zac-nixon (approved):

pkg/service/model_builder_test.go
```diff
@@ -6472,6 +6499,7 @@ func Test_defaultModelBuilderTask_Build(t *testing.T) {
 				jsonString, err := d.Marshal(stack)
 				assert.Equal(t, nil, err)
 				assert.JSONEq(t, tt.wantValue, jsonString)
+				assert.JSONEq(t, tt.wantValue, jsonString)
```
shraddhabang, line 6502: Added assert twice by mistake.

PR #3950, opened 2024-11-21
Add support for capacity unit reservation for load balancers

shraddhabang:
### Description

Application Load Balancer (ALB) and Network Load Balancer (NLB) now support Load Balancer Capacity Unit (LCU) Reservation that allows you to proactively set a minimum capacity for your load balancer, complementing its existing ability to auto-scale based on your traffic pattern. Ref : https://aws.amazon.com/about-aws/whats-new/2024/11/load-balancer-capacity-unit-reservation-application-balancers/

Adding support for this feature on controller for Ingress and Services. 

### Customer Experience

End user will need to update the controller to the latest version containing this feature. They will also need to update the IAM policies and IngressClassParam CRDs after updating the controller version to use this feature. 

#### Default mode 

If the alb.ingress.kubernetes.io/minimum-load-balancer-capacity  or service.beta.kubernetes.io/aws-load-balancer-minimum-load-balancer-capacity annotations are not specified for the ingress or service in the manifest, the capacity will be set to zero on the new load balancers by default. 

####  Update capacity : 

If the end user wants to update the capacity for ALB/NLB to 1200 capacity units, they can update it by setting the new annotation as follows 

#### Sets the minimum capacity for the ALB
`alb.ingress.kubernetes.io/minimum-load-balancer-capacity: CapacityUnits=1200`

#### Sets the minimum capacity for the NLB
`service.beta.kubernetes.io/aws-load-balancer-minimum-load-balancer-capacity: CapacityUnits=1200`

#### Reset capacity: 

If the end user wants to reset the capacity back to 0, they can either remove the new annotations or they can set the capacity on the annotations to be zero as follows

#### Sets the minimum capacity for the ALB
`alb.ingress.kubernetes.io/minimum-load-balancer-capacity: CapacityUnits=0`

#### Sets the minimum capacity for the NLB
`service.beta.kubernetes.io/aws-load-balancer-minimum-load-balancer-capacity: CapacityUnits=0`

#### Turn off feature flag for  capacity reservation (New)

If the customer wants to disable this feature to avoid accidental capacity reservation, then they can turn this feature off through the controller flag by updating their controller deployment as follows

```
      containers:
      - args:
        - --cluster-name=my-cluster
        - --ingress-class=alb
        - --log-level=debug
        - --feature-gates=LBCapacityReservation=false
        image: public.ecr.aws/eks/aws-load-balancer-controller:v2.10.0
        imagePullPolicy: IfNotPresent
  ```  

#### Increase the stabilization requeue interval(New)

If the customer wants to increase the interval to monitor the LB status, then they can set through the controller flag by updating their controller deployment as follows

```
      containers:
      - args:
        - --cluster-name=my-cluster
        - --ingress-class=alb
        - --log-level=debug
        - --lb-stabilization-requeue-interval=1000s
        image: public.ecr.aws/eks/aws-load-balancer-controller:v2.10.0
        imagePullPolicy: IfNotPresent
```

### Checklist
- [x ] Manually tested


scripts/aws_sdk_model_override/setup.sh
shraddhabang, line 1: Since migration of GO SDK to V2, we dont need to generate the private SDK from model anymore. You can simple download the generated GoV2 SDK code and replace references directly.

zac-nixon (approved):

Issue #3822, opened 2024-08-26
Non-Transactional RevokeSGIngress Call Risks Full Cluster Connectivity Loss


asaf400:
**Describe the bug**
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/e1d32f4f83eebc2d7bc61f600dc5e22770102e3e/pkg/networking/security_group_reconciler.go#L104-L114

Upon reconciliation of security group ingress ports rule, specifically the function `reconcileIngressWithSGInfo`,
It's possible for the Controller to successfully revoke the cluster traffic rule `RevokeSGIngress`,
but later (within milliseconds?**), Calling `AuthorizeSGIngress` could fail if the cluster includes a manually created `TargetGroupBinding` with a misconfiguration, for example:
```
...
spec:
  networking:
    ingress:
      - from:
          - securityGroup:
              groupID: sg-defacedeface123456
...
```

Under the condition where the user specified `groupID` doesn't exists, due to the parameter roll-up (single batch for all `permissionsToGrant`) AuthorizeSGIngress would fail, thereby **the cluster is left without any ingress rule** for the backend traffic security group, and all Load Balancers fail to health checks and or reach pods (VPC CNI).

**Expected outcome**
Even when runtime configuration of the controller is not configured with `--disable-restricted-sg-rules={{ .Values.disableRestrictedSecurityGroupRules }}`, this error state should be handled with a short-circuit before a cluster networking disaster occurs.

This can be achieved by any or all of these options: 

- The controller should use DryRun parameter on the `AuthorizeSGIngress` api call, before revoking the rule,
  [API-Doc](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_AuthorizeSecurityGroupIngress.html#API_AuthorizeSecurityGroupIngress_RequestParameters)

- The entire logic flow of the `reconcileIngressWithSGInfo` function should be 'transactional',
Meaning upon a failure within the context, the controller should have a recovery method to restore the existing state before the error.
(This means that not only diff needs to be calculated, but also keep the existing 'describes' result-sets in contextual memory to be used if need be)

**Immediate resolution**
For immediate resolution of the connectivity issue, restarting the LBC pods is possible and worked in my case,
Didn't dive any deeper into the code, but from the cause and effect, seems like the controller's bootstrapping successfully restores the security group ingress rule, and thus ingress works again (temporarily), even though the invalid TGB is still present, and has not been deleted.

[...]

PR #3952, opened 2024-11-22
bug fix: try SG permission add prior to revoke

zac-nixon:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3822

### Description

As described in the issue, it's possible for a user to incorrectly configure their SG rules in such a way that the controller leaves the SG in a state that doesn't allow traffic. This PR solves the issue by reversing the order of operations such that instead allowing too few traffic we allow too much traffic (in the worse case). This way customers who mis-configure their SG rules don't risk a full outage.

I also backfilled missing test cases in the SG resolver.

Test cases tried:

- Remove permission
- Add permission
- Remove + add permission together
- Add + remove permission such that the add needs the remove to happen first to ensure we stay below the maximum SG rules limit.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/networking/security_group_manager.go
+++ b/pkg/networking/security_group_manager.go
@@ -33,7 +33,7 @@ func (opts *FetchSGInfoOptions) ApplyOptions(options ...FetchSGInfoOption) {
 
 type FetchSGInfoOption func(opts *FetchSGInfoOptions)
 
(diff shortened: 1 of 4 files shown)


shraddhabang (approved):
/lgtm

PR #3954, opened 2024-11-26
avoid using flag --aws-vpc-tag-key

### LINKED ISSUE #3889: Usage of --aws-vpc-tags with --aws-vpc-tag-key is unclear
This was introduced by @jeswinkoshyninan in https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3656

It's not clear to my why `--aws-vpc-tags` allows to pass a map with an arbitrary number of items if we only ever pass a single key-value pair (the one with the key specified in `--aws-vpc-tag-key`) as the tag filter for finding a VPC.

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/8a0147826cc45be2f824dcdbf630ff95ff85408f/pkg/aws/cloud.go#L141-L143

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/8a0147826cc45be2f824dcdbf630ff95ff85408f/pkg/aws/cloud.go#L184-L192

Shouldn't it rather just convert the map from `--aws-vpc-tags` into a list of tag filters, one for each item? This would remove the need for `--aws-vpc-tag-key` altogether.

Ref: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3656#discussion_r1799215219

jeswinkoshyninan:
### Issue

This will fix #3889

### Checklist
- [ ] Manually tested


pkg/aws/cloud.go
```diff
@@ -200,14 +201,18 @@ func inferVPCID(ec2Metadata services.EC2Metadata, ec2Service services.EC2) (stri
...
-		Filters: []ec2types.Filter{
-			{
-				Name:   aws.String("tag:" + VpcNameTagKey),
-				Values: []string{VpcNameTagValue},
-			},
-		},
+		Filters: vpcFilter,
 	})
 	if err != nil {
 		return "", fmt.Errorf("failed to fetch VPC ID with tag: %w", err)
```
alfredkrohmer, line 218: I think this should say "... with tags:" or ".. with tag(s):" now. Same for the errors below.

pinkavaj (commented):
Looks good, mey-be squash the 2 commits to one and possibly change the MR/commit title (it will not be obvious for me, when reading changelog, what this change does). My suggestion is to go with something like: "remove --aws-vpc-tag-key  flag, use values from --aws-vpc-tags to specify the tag key"

pinkavaj:
> > Looks good, mey-be squash the 2 commits to one and possibly change the MR/commit title (it will not be obvious for me, when reading changelog, what this change does). My suggestion is to go with something like: "remove --aws-vpc-tag-key flag, use values from --aws-vpc-tags to specify the tag key"
> 
> Updated the title, I think commit will be squashed before merging to master. I am not able to see the option seems due to permission, but definitely the default option will be the squash commits + merge ?

Thx for update.

Not sure if they get squashed (I would kind of expect to be manual step, and it is up to the author to keep the git history clean, but not sure what are the exact rules here).

Btw there are some checkboxes in the MR description not checked, not sure how serious are the maintainers here, but it might help to get this throught (as I'm really looking forward to it!)

jeswinkoshyninan:
|Thx for update.

Not sure if they get squashed (I would kind of expect to be manual step, and it is up to the author to keep the git history clean, but not sure what are the exact rules here).

Btw there are some checkboxes in the MR description not checked, not sure how serious are the maintainers here, but it might help to get this throught (as I'm really looking forward to it!)

@pinkavaj my last PR had 10 commits, so I assume this should be fine.

sarevalo2002:
Hi @jeswinkoshyninan, thanks for the contribution! I've verified that the `aws-vpc-tags` flag accepts both single and multiple input values, as well as handles error cases (no vpc with specified tag exists, multiple vpcs with specified tag exists) properly. I'll loop in @shraddhabang for ok-to-test and lgtm labels.

pinkavaj:
> We should not be removing any flags. Unfortunately, we are stuck with this approach unless we cut a new major version (3.x.y).

So ignoring the flag is an option right?

Issue #3743, opened 2024-06-14
Cannot set the IPv6 addresses in dualstack mode during modification


witalisoft:
**Describe the bug**

Changing the IP Address Type to `dualstack` (via annotation `service.beta.kubernetes.io/aws-load-balancer-ip-address-type`) with a provided list of IPv6 addresses (via annotation `service.beta.kubernetes.io/aws-load-balancer-ipv6-addresses`) is not being reflected (still using the randomly selected IP from IPv6 range). AWS Console gives the option to specify the IPv6 address.  Creating instead of modifying the LoadBalancer, which works as expected.

**Steps to reproduce**

Initial Kubernetes service specification:

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-name: test
    service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
    service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: '*'
    service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
    service.beta.kubernetes.io/aws-load-balancer-type: external
  name: test
spec:
  ports:
  - name: http
    port: 80
    targetPort: http
  - name: https
    port: 443
    targetPort: https
  selector:
    app.kubernetes.io/name: kubernetes-ingress
  type: LoadBalancer
```

Migration to `dualstack` mode with specified IPv6 addresses

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-ip-address-type: dualstack
    service.beta.kubernetes.io/aws-load-balancer-ipv6-addresses: <list of IPv6 addresses>
    service.beta.kubernetes.io/aws-load-balancer-name: test
    service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
    service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: '*'
    service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
    service.beta.kubernetes.io/aws-load-balancer-type: external
  name: test
spec:
  ports:
  - name: http
    port: 80
    targetPort: http
  - name: https
    port: 443
    targetPort: https
  selector:
    app.kubernetes.io/name: kubernetes-ingress
  type: LoadBalancer
```

**Expected outcome**

[...]

PR #3959, opened 2024-11-27
fix: Cannot set the IPv6 addresses in dualstack mode during modification

wweiwei-li:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3743

### Description
* The issue occurs when changing the IP address Type to `dualstack` with a provided list of IPv6 addresses. The update is not being reflected because we only check if the subnets ID has changed. 
* To support this, we should also check the IPv6 address. However, since IPv6 address can only be set once, we need to ensure it is being set for the first time. Additionally, we need to ensure this applies only to NLB

-->

### Checklist
- [x] Manually tested


zac-nixon (approved):

shraddhabang (approved):
/lgtm

witalisoft:
ping

witalisoft:
@wweiwei-li any updates

PR #3960, opened 2024-11-28
Add elasticloadbalancing:DescribeListenerAttributes action to AWSLoadBalancerControllerIAMPolicy

lga-zurich:
### Description

Following [this tutorial](https://docs.aws.amazon.com/eks/latest/userguide/lbc-helm.html) to install the AWS Load Balancer Controller on an EKS cluster, I had to allow the action in the patch in the AWSLoadBalancerControllerIAMPolicy policy or creation of loadbalancers would fail with:

```
Failed deploy model due to operation error Elastic Load Balancing v2: DescribeListenerAttributes, https response error StatusCode: 403, RequestID: XXX, api error AccessDenied: User: arn:aws:sts::XXX:assumed-role/AmazonEKSLoadBalancerControllerRole/XXX is not authorized to perform: elasticloadbalancing:DescribeListenerAttributes because no identity-based policy allows the elasticloadbalancing:DescribeListenerAttributes action
```

### DIFF
--- a/docs/install/iam_policy_iso.json
+++ b/docs/install/iam_policy_iso.json
@@ -33,6 +33,7 @@
                 "elasticloadbalancing:DescribeLoadBalancers",
                 "elasticloadbalancing:DescribeLoadBalancerAttributes",
                 "elasticloadbalancing:DescribeListeners",
+                "elasticloadbalancing:DescribeListenerAttributes",
                 "elasticloadbalancing:DescribeListenerCertificates",
                 "elasticloadbalancing:DescribeSSLPolicies",
                 "elasticloadbalancing:DescribeRules",


zac-nixon:
This isn't really needed. This API is not supported in the isolated partition. The other AWS partitions that support this API already have the correct permissions.

lga-zurich:
> This isn't really needed. This API is not supported in the isolated partition. The other AWS partitions that support this API already have the correct permissions.

ok, thanks @zac-nixon ! It took me a while to figure out what was wrong when following [the controller installation guide](https://docs.aws.amazon.com/eks/latest/userguide/lbc-helm.html#lbc-helm-iam) but it may have been my own confusion.

Issue #3946, opened 2024-11-20
Add Customization Option for Controller Name


andpozo:
**Is your feature request related to a problem?**

Yes, the lack of customization for the controller name makes it impossible to deploy multiple instances of the AWS Load Balancer Controller within the same Kubernetes cluster. This limitation is problematic for setups requiring separate public and private ingress controllers, especially in compliance-driven architectures such as those mandated by SOC2 and ISO27001.

**Describe the solution you'd like**

I'm using the AWS Load Balancer Controller in a setup where I need to register two instances of the controller: one for public ingress and one for private ingress. This setup aligns with the requirements outlined in the AWS whitepaper for building scalable and secure multi-VPC network infrastructure ([link](https://docs.aws.amazon.com/whitepapers/latest/building-scalable-secure-multi-vpc-network-infrastructure/using-network-firewall-for-centralized-ingress.html)), particularly for SOC2 and ISO27001 compliance.

Currently, the controller name is hardcoded in the pkg/ingress/class_loader.go file ([link to line](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/5c892f5467f363ce0770e49bbfb49673c3c2e6b2/pkg/ingress/class_loader.go#L21)).

I propose adding an extra configuration option to the pkg/config/ingress_config.go file ([link](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/5c892f5467f363ce0770e49bbfb49673c3c2e6b2/pkg/config/ingress_config.go)) that allows users to customize the controller name.

#### This change would enable users to:

- Register multiple instances of the AWS Load Balancer Controller.
- Differentiate between public and private ingress configurations.
#### Benefits
- Provides flexibility for advanced ingress setups.
- Supports compliance with SOC2 and ISO27001 by enabling centralized ingress management.

#### Suggested Implementation
Add a new field (e.g., controllerName) to the IngressConfig struct in pkg/config/ingress_config.go.
Modify the hardcoded controller name in pkg/ingress/class_loader.go to use the configurable value from IngressConfig.
Update the default behavior to maintain backward compatibility (e.g., default to the existing hardcoded name if the new option is not specified).

Architecture Diagram

Below is an architectural diagram that illustrates the proposed setup with two Load Balancer Controllers, one managing public ingress and another managing private ingress. This setup ensures compliance with centralized ingress requirements for secure multi-VPC environments:

[...]

Issue #3634, opened 2024-04-03
Support for ALBs in multiple AWS accounts


marcosdiez:
We have multiple EKS clusters, in multiple VPCs. That being said, we plan to have a centralized VPC just for ingress, following this pattern: https://docs.aws.amazon.com/whitepapers/latest/building-scalable-secure-multi-vpc-network-infrastructure/using-network-firewall-for-centralized-ingress.html

Since less is more, we want to keep the number of hops as low as possible, so we plan to have the following traffic setup:

```

Browser  -> PUBLIC_ALB_IN_THE_INGRESS_ACCOUNT -> IP_OF_THE_EKS_POD_IN_THE_APPLICATION_ACCOUNT

```

That means we have at least two AWS accounts: INGRESS_ACCOUNT and APPLICATION_ACCOUNT 

Our current setup is created with terraform, including the EKS cluster, the ALB, the DNS entry, the ALB rule and the ALB target group.
Which means the only thing that the `aws-load-balancer-controller` has to do is to sync the IPs of the k8s service with the target group.

For that, we use `TargetGroupBindings` ( https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.7/guide/targetgroupbinding/targetgroupbinding/ )

Just for reference, this is what a `TargetGroupBinding` look like:

```
apiVersion: elbv2.k8s.aws/v1beta1
kind: TargetGroupBinding
metadata:
  name: my-tgb
spec:
  serviceRef:
    name: awesome-service # route traffic to the awesome-service
    port: 80
  targetGroupARN: <arn-to-targetGroup>
```

So what we need is for the `aws-load-balancer-controller` to be able to interact with the target groups in different AWS accounts.

More specifically, we need this IAM permission on different AWS accounts:

```
                "elasticloadbalancing:DescribeTargetGroups",
                "elasticloadbalancing:DescribeTargetHealth",
                "elasticloadbalancing:ModifyTargetGroup",
                "elasticloadbalancing:ModifyTargetGroupAttributes",
                "elasticloadbalancing:RegisterTargets",
                "elasticloadbalancing:DeregisterTargets"
```

Ideally, we should be able to do that with AWS RAM, but AWS RAM can only share whole VPCs, which not something my security team is happy about.

[...]

PR #3691, opened 2024-05-13
TargetGroupBindings can now manipulate target groups from different aws accounts

marcosdiez:
### Description

Now, every time a `TargetGroupBinding` has the `alb.ingress.kubernetes.io/IamRoleArnToAssume` (and optionally `alb.ingress.kubernetes.io/AssumeRoleExternalId`) annotations, we assume it before interacting to it (actually, the assume role operation is cached).

In real life, I changed: 

* targetgroupbinding_validator
* targetgroupbinding_mutator
* target manager (which has `RegisterTargets` / `DeregisterTargets` / `ListTargets`

This way we can easily manipulate target groups from target group bindings for every single AWS account in the world, as long as there is a role we can assume. In order to prevent the _confused deputy problem_, `external id` is supported.

The way this was implemented was actually way simpler than initially imagined:

the `service.elbv2Client` object now supports a new method called `AssumeRole` which returns `service.elbv2Client` on the specified role. The catch is that if `AssumeRole` is called with a blank role, the object returns itself and life goes on.

In other words, every time we call some method from `elbv2Client` from something related to target groups, we also call `AssumeRole` and that's basically it.

Caching the assumed role was also trivial, a simple map that is stored on `cloud.go`

I decided to have `cloud.go` manage the caching becuase in the future it could assume role for other objects if needed and all the assume role logic would be in the same place.

The only unfortunate change that needed to be made is that `aws.Cloud` became `services.Cloud` else golang would complain about circular dependency :(

This is what the logs look like:

```

{"level":"info","ts":"2024-05-13T03:24:31Z","msg":"awsCloud","method":"GetAssumedRoleELBV2","AssumeRoleArn":"arn:aws:iam::7550XXXXXXXX:role/alb-controller-policy-to-impersonate","externalId":"XXXXXX"}
{"level":"info","ts":"2024-05-13T03:24:33Z","msg":"registering endpoints using the targetGroup's vpcID vpc-00XXXXXXXXXXXXXXX which is different from the cluster's vpcID vpc-04XXXXXXXXXXXXXXX"}
{"level":"info","ts":"2024-05-13T03:24:33Z","msg":"registering targets","arn":"arn:aws:elasticloadbalancing:us-east-1:7550XXXXXXXX:targetgroup/helloworld-8080-a8054/5741da332459b560","targets":[{"AvailabilityZone":"all","Id":"10.244.0.22","Port":8080}]}
[...]

PR #3963, opened 2024-11-30
adding a customizable value for controller class

andpozo:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3946
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3634

### Description
This PR introduces support for setting a customizable value for the controller class in the AWS Load Balancer Controller.

Key Changes

- Added a new configuration option to allow users to specify a custom value for the controller class.
- Updated the controller logic to use the provided value if specified; defaults remain intact for backward compatibility, the default value will remain the same as the previously hardcoded value (`ingress.k8s.aws/alb`).
- Adjusted the documentation to include instructions on how to configure the custom controller class.

Benefits

- Provides greater flexibility for users managing multiple controllers within the same cluster.
- Enhances compatibility with diverse use cases where unique controller class names are necessary.

### Checklist
- [ ] Manually tested


andpozo:
Hey @johngmyers,
wouldn't it be a good idea to use the controller name instead of the controller class? what do you think?
LMK I'll stay tuned

zac-nixon:
I would prefer to solve this issue using this PR that's been submitted https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3691 that seems to be a more general purpose solution.

andpozo:
> I would prefer to solve this issue using this PR that's been submitted #3691 that seems to be a more general purpose solution.

Hey @zac-nixon, probably yes, but... we should isolate the two controller classes to separately handle internal and external load balancing, thoughts? this option will be able the deploy the same controller with different classes for different purposes

andpozo:
reopened due to an incorrect closure

PR #3974, opened 2024-12-10
feat: add advertise ca for mtls listener

zac-nixon:
### Issue

ALB recently added a new flag to mtls listeners:
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/mutual-authentication.html#advertise-ca-subject

This CR adds the ability to set the advertise ca flag on an mtls listener from an ingress / service annotation.

### Description

Adds parsing logic to `alb.ingress.kubernetes.io/mutual-authentication` that parses the `advertiseTrustStoreCaNames` field and passes it onto the ELB API to set the new flag on an mtls listener.

Tested locally in a cluster with the following setups:

1/ No mTLS
2/ mTLS off
3/ mTLS pass through mode
4/ mTLS verify with advertise ca off
5/ mTLS verify with advertise ca on

### Checklist
- [x] Manually tested


pkg/deploy/elbv2/listener_manager.go
```diff
@@ -365,11 +365,17 @@ func buildSDKMutualAuthenticationConfig(modelMutualAuthenticationCfg *elbv2model
...
-	return &elbv2types.MutualAuthenticationAttributes{
+	attributes := &elbv2types.MutualAuthenticationAttributes{
 		IgnoreClientCertificateExpiry: modelMutualAuthenticationCfg.IgnoreClientCertificateExpiry,
 		Mode:                          awssdk.String(modelMutualAuthenticationCfg.Mode),
 		TrustStoreArn:                 modelMutualAuthenticationCfg.TrustStoreArn,
 	}
+
+	if modelMutualAuthenticationCfg.Mode == string(elbv2model.MutualAuthenticationVerifyMode) {
+		attributes.AdvertiseTrustStoreCaNames = translateAdvertiseCAToEnum(modelMutualAuthenticationCfg.AdvertiseTrustStoreCaNames)
+	}
```
andreybutenko, line 376: Should `IgnoreClientCertificateExpiry` be handled the same way? Or should this be handled the same way as `IgnoreClientCertificateExpiry`?

They are both verify-specific features, this code section is more difficult to understand if similar features are implemented in different ways

zac-nixon: Unfortunately, the features might be implemented the same way but the API validation is different for the two. When the default value is set for `IgnoreClientCertificateExpiry` the API accepts the input no matter what mode we're using. However, the same is not true for `AdvertiseTrustStoreCaNames`, if the a value is set for this field and the mode is not verify, then the API will fail.

andreybutenko: Thanks for the clarification! If this is consistent with the API behavior, that's okay with me :)

andreybutenko (approved):

shraddhabang (approved):
/lgtm

PR #3983, opened 2024-12-16
fix(helm): change topologySpreadConstraints default value

yann-soubeyrand:
topologySpreadConstraints is a list not a map.


zac-nixon (approved):
K8s docs explaining the field - https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#scheduling

PR #3990, opened 2024-12-20
feat: sync created Backend SG tags

### LINKED ISSUE #2838: Default tags don't get applied to previously created "backend" security group
**Describe the bug**
Tags from `--default-tags` argument don't get applied to previously created "backend" security group.

**Steps to reproduce**
1. Deploy chart with minimal config.
2. Create at least 1 ALB Ingress.
3. Update chart, providing `defaultTags` option.
4. Observe default tags only on frontend security group.

**Expected outcome**
Both frontend and backend security groups get provided tags.

**Environment**

* AWS Load Balancer controller version: v2.4.4
* EKS version: v1.23.10-eks-15b7512

**Additional Context**:
n/a

phuhung273:
### Issue

- Close #2838

### Description

- If previous created Backend SG found, sync its tags

### Checklist
- [x] Manually tested


pkg/networking/backend_sg_provider.go
```diff
@@ -253,6 +259,20 @@ func (p *defaultBackendSGProvider) allocateBackendSG(ctx context.Context, resour
...
 
+func (p *defaultBackendSGProvider) syncBackendSGTags(ctx context.Context, sgID string) error {
+	tagSpecifications := p.buildBackendSGTags(ctx)
+
+	createReq := &ec2sdk.CreateTagsInput{
+		Resources: []string{sgID},
+		Tags:      tagSpecifications[0].Tags,
+	}
+
+	if _, err := p.ec2Client.CreateTagsWithContext(ctx, createReq); err != nil {
```
zac-nixon, line 270: Can you please add a check to ensure that this create call will actually do something? Ideally, we shouldn't call the CreateTags API when there is no changes to be made.

zac-nixon (changes requested):
Appreciate the PR. This looks good aside from the comment I left.

pkg/networking/backend_sg_provider.go
phuhung273: Thanks for your review, I found this code with the same reconcilation logic https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/2ff2e59711c8e749c93b46c62dd60975598115c3/pkg/deploy/ec2/tagging_manager.go#L80

But we cannot import since circular dependency, so I copied here. Do you think we should refactor to remove duplication ?

phuhung273: Hi @zac-nixon, can you help review the new change

phuhung273: Hi @zac-nixon, can you help take a look. Thanks.

PR #4021, opened 2025-01-16
[feat] allow multiple targetgroupbindings to reference same tg arn if using multicluster mode

zac-nixon:
### Issue

n/a

### Description

Users would like to utilize the same Target Group ARN for multiple TargetGroupBindings. Previously, before Multicluster support, this was not possible as the reconcile logic would be unaware of other entities registering targets into the configured Target Group. Now, using Multicluster TargetGroupBindings, the reconcile logic is able to only operate on targets in the Target Group that belong to the specified TargetGroupBinding.

We can relax the validation logic to allow users to specify duplicate Target Group ARNs if they also specify the Multicluster flag on each binding. 

### Checklist
- [x] Manually tested

### DIFF
--- a/webhooks/elbv2/targetgroupbinding_validator.go
+++ b/webhooks/elbv2/targetgroupbinding_validator.go
@@ -3,6 +3,7 @@ package elbv2
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/types"
 	"regexp"
 	"strings"
 
@@ -90,6 +91,9 @@ func (v *targetGroupBindingValidator) ValidateUpdate(ctx context.Context, obj ru
 	if err := v.checkAssumeRoleConfig(tgb); err != nil {
 		return err
 	}
+	if err := v.checkExistingTargetGroups(tgb); err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


webhooks/elbv2/targetgroupbinding_validator.go
```diff
@@ -72,6 +73,9 @@ func (v *targetGroupBindingValidator) ValidateCreate(ctx context.Context, obj ru
 	if err := v.checkAssumeRoleConfig(tgb); err != nil {
 		return err
 	}
+	if err := v.checkExistingTargetGroups(tgb); err != nil {
```
wweiwei-li, line 76: `checkExistingTargetGroups` is being called twice, which is not need ?

zac-nixon: bad rebase on my part.

wweiwei-li (approved):

webhooks/elbv2/targetgroupbinding_validator.go
```diff
@@ -90,6 +91,9 @@ func (v *targetGroupBindingValidator) ValidateUpdate(ctx context.Context, obj ru
...
```
jaroslav-muller-jamf, line 94: This is a breaking change, that wasn't mentioned in release notes.
We have a cluster with two TGBs with duplicate target group. (Yes, it is wrong, but it worked with 2.10)
Previously, this was checked only on Create. So reconcile worked OK.
Now, this is checked on Update too, so both TGBs are invalid and no reconcile is done.

While I agree that the root problem is actually the duplicate TG, this is a breaking change that we didn't expect when upgrading.

PR #4026, opened 2025-01-21
Follow up for 3691 [TargetGroupBindings can now manipulate target groups from different aws accounts]

zac-nixon:
### Issue

Previous PR, https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3691

Issue, https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3634

### Description

The original PR (#3691) was a very good start to implementing cross account Target Group management and I used the same logic. This PR mostly fixes up bugs in the approach, as 3691 didn't work using SDKv2. Things I've added / changed:

- Use a singleton STS client.
- Use a garbage collected cache.
- Do not hold on to assumed role elbv2 clients indefinitely, these clients eventually expire and need to be recreated.
- Remove annotation based configuration for the Target Group Binding, instead rely only on the TGB spec to hold configuration data.
- Fix TG ARN resolution (the previous PR didn't apply the same logic between ARN / name specification for the TG). Also refactored the logic to remove code duplication.
- Add webhook validation to prevent specifying instance target groups + assume role. ELB API doesn't allow cross account instance registration.
- Added documentation and examples.

Test Cases:
- Same account TGB interactions work the same.
- Cross account with valid role + external id works.
- Cross account with invalid external id gracefully fails in the webhook validation.
- Long running controller (5 days) has no significant memory increase nor any errors related to the AssumeRole management.

### Checklist
- [x] Manually tested

### DIFF
--- a/docs/guide/targetgroupbinding/spec.md
+++ b/docs/guide/targetgroupbinding/spec.md
@@ -55,15 +55,6 @@ Kubernetes meta/v1.ObjectMeta
 <table>
 <tr><td><code>annotations</code></td><td>
 
-<table>
-<tr><td><code>alb.ingress.kubernetes.io/IamRoleArnToAssume</code><br><i>string</i></td>
-<td><i>(Optional)</i> In case the target group is in a differet AWS account, you put here the role that needs to be assumed in order to manipulate the target group.
-</td></tr>
-<tr><td><code>alb.ingress.kubernetes.io/AssumeRoleExternalId</code><br><i>string</i></td>
-<td><i>(Optional)</i> The external ID for the assume role operation. Optional, but recommended. It helps you to prevent the <a href="https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html" target="_blank">confused deputy problem</a>.
-</td></tr>
-</table>
-
 <tr><td colspan=2>
 Refer to the Kubernetes API documentation for the other fields of the
 <code>metadata</code> field.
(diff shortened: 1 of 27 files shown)


docs/guide/targetgroupbinding/spec.md
```diff
@@ -55,15 +55,6 @@ Kubernetes meta/v1.ObjectMeta
 <table>
 <tr><td><code>annotations</code></td><td>
 
-<table>
```
zac-nixon, line 58: Removed this. Trying to get the tool fixed to generate this properly, but it will make the CR _huge_

docs/guide/targetgroupbinding/targetgroupbinding.md
```diff
@@ -112,10 +112,63 @@ spec:
...
+            "Action": [
+                "sts:AssumeRole"
+            ],
+            "Resource": "*"
+        }
+```
+
+In the target account, which owns the Target Group the assumed role must allow for Target Group Management, at the minimum:
+
+```json
```
shraddhabang, line 159: I was wondering if we should restrict these permissions to tag based instead of allowing them to access all tg resources for another account. We can restrict their permissions to access tg based on its ResourceTags?

shraddhabang: I am thinking about updating the policies to use a condition block to ensure these register/deregister actions can only be performed on tg that have the specified cluster tag present, adding an additional layer of security control for the tgs not owned. We do this in our [current](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/install/iam_policy.json#L184) IAM policies for ELB actions to be performed on LBs which have certain cluster tags. This ensures that LBC is only interacting with the LBs which it created. I am thinking should we also add this type of pre-req to deal with tgs from different accounts to avoid any accidental updates.

shraddhabang: While I understand it will add an extra layer of config on customer side, but it may protect their resources by adding this fine grained access.

shraddhabang (approved):
/lgtm
/approve

Issue #3816, opened 2024-08-19
Sporadic Fixed Error 404''s due to Ingress Controller Using ModifyRule API


dieherna:
**Describe the bug**
Observing sporadic fixed 404 errors on an EKS cluster hosting multiple container services. The services are fronted by a single ALB using host based routing. All services have a unique FQDN, there are no overlapping domains. The services are configured on the ALB via ingress controller using ingress groups. During the deployment of a new service version, we will deploy the new service in parallel to the existing service. Shortly after (few days) we will delete the old service version. When the old service version is deleted, the ALB will respond with fixed 404 responses for a few seconds before recovering. All changes on the ALB are exclusively done via the ingress controller. Never manually.

**[ROOT CAUSE]**
The outcome troubleshooting with the AWS ALB service team is that the issue is being caused by a race condition. Have been told by the service team that the race condition can be avoided by the ingress controller using the SetRulePriorities 
API instead of recreating the functionality through a lot of ModifyRule API. To summarize the events that lead to the race condition:

**Scenario:** Customer has 30 services deployed. All with unique FQDN's. Single ALB being managed by ingress controller ingress groups. Single ALB listener with 30 rules (1 per service)

1. New service version is deployed (service-a-v2.xyz.com) and running in parallel to existing service (service-a-v1.xyz.com). service-a-v1.xyz.com has an ALB rule priority of 3.
2. service-a-v1.xyz.com is deleted. The ingress controller will send an DeleteRule API to the ALB. ALB rules remaining have a priority of 1-2 & 4-30
   1. Implicitly (under the hood behaviour) the ingress controller will send 27 ModifyAPI requests to the ALB. The ModifyAPI's are trying to keep the rules priorities sequential (1-29). The ModifyAPI will replace the FQDN of the rule above with the FQDN of the rule below. Ultimately resulting in all the FQDN's priority being shifted up by 1.
   2. The ALB implements changes in batches. When the first API is received, the ALB will wait 10 seconds to batch all possible subsequent changes before deploying the change in the data plane. This could lead to the race condition where some of the ModifyAPI's arrive after the first API's and be implemented in the next window. A random rule for a short period of time does not exist in the ALB. e.g., service-z.xyz.com responds with fixed 404's.

**Steps to reproduce**
1. Create multiple services with ingress.group e.g 3 services
2. delete a service, preferably one which has been deployed in the ALB with the highest rule priority
   1. The order will be done using lexical order by default
[...]

PR #4039, opened 2025-02-04
Update rule management to avoid sporadic 503 errors

shraddhabang:
### Issue
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3816

### Description
The root cause  for the above issue was identified as a race condition in the ALB's rule management process. The controller uses ModifyRule API sequentially to add any new rule at the existing priority which results in loss of the existing older rule at that priority. When the first API call to modify rules is received, the ALB waits 10 seconds to batch all subsequent changes before deploying them to the data plane. If some rule modification APIs arrive after the initial 10-second window, they are implemented in the next batch, leading to a brief period where a random rule may not exist, causing 404 errors. 

The proposed improvement in the rule management logic make use of SetRulePriorities API to first push the unwanted rules down in the listener instead of deleting them. This creates the gaps in the rules order to add any new/updated rule. Once the new rule is successfully added, the controller will then delete all the unwanted rules. This way we ensure that no rules will be lost before the new/updated rules are present in the controller. This will avoid any sporadic 404 errors caused due to race conditions. 

### Checklist
- [x] Manually tested


pkg/deploy/elbv2/listener_rule_synthesizer.go
```diff
@@ -62,30 +68,75 @@ func (s *listenerRuleSynthesizer) PostSynthesize(ctx context.Context) error {
 
 func (s *listenerRuleSynthesizer) synthesizeListenerRulesOnListener(ctx context.Context, lsARN string, resLRs []*elbv2model.ListenerRule) error {
 	sdkLRs, err := s.findSDKListenersRulesOnLS(ctx, lsARN)
+	var lastAvailablePriority int32 = 50000
```
zac-nixon, line 71: Looking at the implementation on how we assign rule priorities it seems we'll never reach 50k. It might be good to make a note that changing the priority assignment would mean revisiting this logic.

zac-nixon: Actually, I thought of this more and there's potential to deadlock the reconciler.

Scenario:

4 rules with priorities (1,2,3,4), We are removing rule with priority 2,3.

Reconcile 1:

1/ We move priority 2 -> priority 50,000
2/ We move priority 3 -> priority 49,999
2/ We try to move priority 4 -> priority 2 and fail (maybe network issue on call)


Reconcile 2:
1/ We try to move one of the unmatched rules to priority 50,000
  - If we try to move the old priority rule 3 to 50,000. This will fail because the old priority rule 2 already got moved to 50k.
  - If we try to move the old priority rule 2 to 50,000. API idempotency will allow the call to succeed.

It's possible to construct a complex enough rule removal that we have permutations of rules to adjust that we never make progress on the reconciliation. 

I wonder if we even need this step to move rules to the back of the priority list? The need for sequential listener rule priorities is just an artifact of the LBC implementation and not a requirement of the ELB API. I wonder if something like this would work..


```
1. Delete all unmatched ListenerRules.
2. Use SetRulePriority on all matched ListenerRules to set the desired priority for each existing rule to priority calculated by the LBC.
3. Use CreateRule to create any missing ListenerRules with the desired priority calculated by the LBC
```

zac-nixon (changes requested):

pkg/deploy/elbv2/listener_rule_synthesizer.go
zac-nixon: Shraddha was nice enough to explain that this thinking is actually incorrect as the SetRulesPriorities is atomic and we're move all rules at one time :)

pkg/deploy/elbv2/listener_rule_synthesizer.go
```diff
@@ -110,48 +139,40 @@ type resAndSDKListenerRulePair struct {
...
+			if cmp.Equal(resLRDesiredActionsAndConditionsPair.desiredActions, sdkLR.ListenerRule.Actions, elbv2equality.CompareOptionForActions()) &&
+				cmp.Equal(resLRDesiredActionsAndConditionsPair.desiredConditions, sdkLR.ListenerRule.Conditions, elbv2equality.CompareOptionForRuleConditions()) {
+				sdkLRPriority, _ := strconv.ParseInt(awssdk.ToString(sdkLR.ListenerRule.Priority), 10, 64)
+				if resLR.Spec.Priority != int32(sdkLRPriority) {
+					matchedResAndSDKLRsBySettings = append(matchedResAndSDKLRsBySettings, resAndSDKListenerRulePair{
+						resLR: resLR,
+						sdkLR: sdkLR,
+					})
+				}
+				unmatchedSDKLRs = append(unmatchedSDKLRs[:i], unmatchedSDKLRs[i+1:]...)
```
zac-nixon, line 165: nit: I am not a fan of modifying the input of a function like this. I see that we do this is in other parts of the code but overall I think it's an anti-pattern.

pkg/deploy/elbv2/listener_rule_synthesizer.go
```diff
@@ -110,48 +139,40 @@ type resAndSDKListenerRulePair struct {
...
+				i--
+				found = true
+				break
+			}
+		}
+		if !found {
```
zac-nixon, line 171: I think this will still result in 503s (please correct me if I'm wrong)

Basically, from my read of the code, when a user changes their conditions or actions we will delete the ListenerRule and re-create the rule. If we're unlucky, it's possible for the delete to get propagated and the create to be delayed a bit which causes the 503 errors.

Ideally, we can use `modify-rule` https://docs.aws.amazon.com/cli/latest/reference/elbv2/modify-rule.html to detect these rules with the changed actions / conditions and modify the rule in place, rather than delete and create the rule.

shraddhabang: Thanks for catching this. I have updated the code to avoid such scenarios. We will first push down all the unmatched SDK LRs which are not matching any priority or settings down. We will reprioritize the rules which are matching the settings to match their desired priorities and then we will either modify/create the rules at their desired priorities and in the end delete all the unwanted pushed down rules. This will we will have have rules on the listener before we get a newer/updated one at higher priority.

pkg/deploy/elbv2/listener_rule_synthesizer.go
shraddhabang: I have done this to avoid any duplicate desired res rules matching the settings to be paired to same sdk rule.

pkg/deploy/elbv2/listener_rule_synthesizer.go
```diff
@@ -61,25 +66,49 @@ func (s *listenerRuleSynthesizer) PostSynthesize(ctx context.Context) error {
...
+		lrStatus, err := s.lrManager.Create(ctx, resLR, resLRDesiredActionsAndConditionsPairs[resLR])
 		if err != nil {
 			return err
 		}
 		resLR.SetStatus(lrStatus)
 	}
-	for _, resAndSDKLR := range matchedResAndSDKLRs {
+	// Update existing listener rules on the load balancer for their tags
+	for _, resAndSDKLR := range matchedResAndSDKLRsBySettings {
 		lsStatus, err := s.lrManager.Update(ctx, resAndSDKLR.resLR, resAndSDKLR.sdkLR)
```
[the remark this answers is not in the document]
shraddhabang: Sounds good. Updated.

pkg/deploy/elbv2/listener_rule_synthesizer.go
```diff
@@ -61,25 +66,49 @@ func (s *listenerRuleSynthesizer) PostSynthesize(ctx context.Context) error {
...
+	}
+	// Find existing listener rules on the load balancer
 	sdkLRs, err := s.findSDKListenersRulesOnLS(ctx, lsARN)
 	if err != nil {
 		return err
 	}
-
-	matchedResAndSDKLRs, unmatchedResLRs, unmatchedSDKLRs := matchResAndSDKListenerRules(resLRs, sdkLRs)
+	// matchedResAndSDKLRsBySettings : A slice of matched resLR and SDKLR rule pairs that have matching settings like actions and conditions
+	// unmatchedResLRs : A slice of resLR) that do not have a corresponding match in the sdkLRs. These rules need to be created on the load balancer.
```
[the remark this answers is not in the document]
shraddhabang: fixed

docs/install/iam_policy.json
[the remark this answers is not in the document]
shraddhabang: Yes. I agree. Nice catch fixed now. :)

pkg/deploy/elbv2/listener_rule_manager.go
```diff
@@ -116,15 +125,48 @@ func (m *defaultListenerRuleManager) Delete(ctx context.Context, sdkLR ListenerR
...
+		priorityJ, _ := strconv.Atoi(awssdk.ToString(unmatchedSDKLRs[j].ListenerRule.Priority))
+		return priorityI < priorityJ
+	})
+	// Push down all the unmatched existing SDK rules on load balancer so that updated rules can take their place
+	for _, sdkLR := range slices.Backward(unmatchedSDKLRs) {
+		sdkLR.ListenerRule.Priority = awssdk.String(strconv.Itoa(int(lastAvailablePriority)))
+		sdkLRs = append(sdkLRs, sdkLR)
+		lastAvailablePriority--
+	}
+	//Reprioratize matched rules by settings
```
zac-nixon, line 144: nit: Reprioritize

pkg/deploy/elbv2/listener_rule_manager.go
```diff
@@ -116,15 +125,48 @@ func (m *defaultListenerRuleManager) Delete(ctx context.Context, sdkLR ListenerR
 	return nil
 }
 
-func (m *defaultListenerRuleManager) updateSDKListenerRuleWithSettings(ctx context.Context, resLR *elbv2model.ListenerRule, sdkLR ListenerRuleWithTags) error {
-	desiredActions, err := buildSDKActions(resLR.Spec.Actions, m.featureGates)
-	if err != nil {
+func (m *defaultListenerRuleManager) SetRulePriorities(ctx context.Context, matchedResAndSDKLRsBySettings []resAndSDKListenerRulePair, unmatchedSDKLRs []ListenerRuleWithTags) error {
+	var lastAvailablePriority int32 = 50000
```
zac-nixon, line 129: just for my own understanding, it's impossible for a user to model a rule with 50k priority in K8s. It would have to be done via an out of band approach therefore would also be eligible for removal

zac-nixon (approved):

zac-nixon (changes requested):
Sorry I missed this before, but we should have some unit tests here.

pkg/deploy/elbv2/listener_rule_manager.go
shraddhabang: Correct.

shraddhabang:
> Sorry I missed this before, but we should have some unit tests here.

Done. Please take a look.

zac-nixon (approved):

mpayne-coveo:
Could a new release be cut soon with this change? We're waiting for this bugfix to be released. Thanks

Issue #4042, opened 2025-02-05
Dangling AWS Shield Protections after Ingress are deleted


sitaramshelke:
**Bug Description**
We use AWS Load balancer controller to manage ingress. We are using `alb.ingress.kubernetes.io/shield-advanced-protection` ingress annotation to protect the ingress using AWS Shield. This part works perfectly fine. However, when we delete our ingresses, we see that the Protection resource is not deleted. 
Protection resources have below attributes.
"Name"  = "managed by aws-load-balancer-controller"
"AWS Resource" = "<name-of-the-lb>"
"Resource type" = "Application Load Balancer"
"Status" = "Resource Deleted"
"AWS WAF web ACL" = "Error"

![Image](https://github.com/user-attachments/assets/4e3ce476-fb1c-4059-bd28-53a71f735955)

**Steps to Reproduce**

- Provision an ingress with Annotation `alb.ingress.kubernetes.io/shield-advanced-protection = true`
- Verify Shield Protection resource is created
- Delete the ingress

**Expected Behavior**

Since the protection resource is managed by load balancer controller, it should be deleted by the controller.

**Actual Behavior**

Protection resource still exists in a dangling state.

**Regression**
Was the functionality working correctly in a previous version ? [Yes / No]
If yes, specify the last version where it worked as expected
Unsure about this.

**Current Workarounds**

NA

**Environment**
- AWS Load Balancer controller version: v2.4.6
- Kubernetes version: EKS 1.29
- Using EKS (yes/no), if so version?: 1.29
- Using Service or Ingress: Ingress
- AWS region: All regions
- How was the aws-load-balancer-controller installed:
  - If helm was used then please show output of `helm ls -A | grep -i aws-load-balancer-controller`
  - If helm was used then please show output of `helm -n <controllernamespace> get values <helmreleasename>`
  - If helm was not used, then copy/paste the exact command used to install the controller, including flags and options.
- Current state of the Controller configuration:
  - `kubectl -n <controllernamespace> describe deployment aws-load-balancer-controller`
- Current state of the Ingress/Service configuration:
  - `kubectl describe ingressclasses`
  - `kubectl -n <appnamespace> describe ingress <ingressname>`
  - `kubectl -n <appnamespace> describe svc <servicename>`

**Possible Solution (Optional)**

NA

**Contribution Intention (Optional)**

- [ ] Yes, I'm willing to submit a PR to fix this issue
[...]

PR #4048, opened 2025-02-12
Dangling fix AWS Shield Protections after Ingress are deleted #4042

Lahutsin:
### Issue

[](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4042)

### Description

This is my first fix contribution. I am open to any discussion to improve my PR's.

DeleteProtection Method:

- The DeleteProtection method now calls DeleteProtectionWithContext to disable the shield protection.
- After disabling the protection, it removes the protection info from the cache using m.protectionInfoByResourceARNCache.Delete(resourceARN).
- It then calls a new method verifyProtectionDeleted to ensure that the protection resource is deleted.

verifyProtectionDeleted Method:

- This method checks if the protection resource is successfully deleted by calling DescribeProtectionWithContext.
- If the protection resource is not found, it means the resource is successfully deleted.
- If the protection resource still exists, it returns an error.

### Checklist
- [+] Added tests that cover your change: in protection_manager_test.go

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/aws/aws-sdk-go v1.55.5
 	github.com/aws/aws-sdk-go-v2 v1.32.6
 	github.com/aws/aws-sdk-go-v2/config v1.27.27
+	github.com/aws/aws-sdk-go-v2/credentials v1.17.27
 	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.16.11
 	github.com/aws/aws-sdk-go-v2/service/acm v1.28.4
 	github.com/aws/aws-sdk-go-v2/service/appmesh v1.27.7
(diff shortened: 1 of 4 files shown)


zac-nixon (changes requested):
Hi. I am really sorry, but I'm trying to understand what this code change accomplishes. It looks like we are just re-reading the protection result after deletion and throwing an error if the protection still exists. Is the implication here that the delete call is not working?

Lahutsin:
> Hi. I am really sorry, but I'm trying to understand what this code change accomplishes. It looks like we are just re-reading the protection result after deletion and throwing an error if the protection still exists. Is the implication here that the delete call is not working?

Hello. Thank you for your comment! I understand your concerns and suggest discussing this in more detail. For additional context, please refer to the issue discussion [Issue 4042](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4042).

You are correct in interpreting that the code change involves re-reading the protection result after attempting to delete it and then issuing an error if the protection still exists.

It is implied that the deletion call is not necessarily completely broken, but rather that it might not have been fully executed or reflected at the time of checking the result. Re-reading and error handling are intended to detect situations where the protection should have been deleted but, for some reason (such as propagation delays, API issues, etc.), it was not deleted in a timely manner.

Thus, this code can provide greater reliability and handle edge cases where deletion takes longer than expected or does not work as planned, issuing an error in such cases for a better understanding of what went wrong.

If you have specific suggestions or questions, let's discuss them to find the best solution.

zac-nixon:
I talked with @shraddhabang about this issue. It looks like the deletion logic is incorrect as we don't tag the resource correctly.

Issue #4054, opened 2025-02-14
Update TLS to version 1.2 or higher: Update Security Policy


riita10069:
`alb.ingress.kubernetes.io/ssl-policy`'s default value is `None`
but We should Update TLS to version 1.2 or higher

We should update `alb.ingress.kubernetes.io/ssl-policy`'s default value to `ELBSecurityPolicy-TLS13-1-3-2021-06`

shraddhabang:
@riita10069 , We already have a `default-ssl-policy` controller [flag](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.11/deploy/configurations/#:~:text=Kubernetes%20cluster%20name-,default%2Dssl%2Dpolicy,-string) which has a default value as `ELBSecurityPolicy-2016-08`. Changing default now would be a breaking change for existing customer. You may use this controller flag for now to set your default policy.

PR #4055, opened 2025-02-14
Update Default Value for Chart defaultSSLPolicy because of `Update TLS to version 1.2 or higher`

riita10069:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4054

### Description

To ensure compliance with modern security standards and to mitigate vulnerabilities associated with older TLS versions, we need to configure the Application Load Balancer (ALB) to use an SSL policy that enforces TLS 1.2 or higher. This change aligns with industry best practices and AWS recommendations.

### Requirements:

Update the ALB's SSL policy to one that supports only TLS 1.2 and TLS 1.3.

Recommended policies include:
ELBSecurityPolicy-TLS13-1-2-2021-06

**Verify that the updated policy disables support for deprecated protocols such as TLS 1.0 and TLS 1.1.**

Ensure compatibility with all client systems accessing the ALB.

### Justification:
Security Compliance: Industry standards such as PCI DSS, NIST, and others mandate the use of secure protocols like TLS 1.2 or higher.

AWS Recommendations: AWS documentation recommends using the latest predefined security policies, such as ELBSecurityPolicy-TLS13-1-2-2021-06, which support TLS 1.3 and are backward-compatible with TLS

- https://docs.aws.amazon.com/elasticloadbalancing/latest/application/describe-ssl-policies.html
- https://trendmicro.com/cloudoneconformity/knowledge-base/aws/ELBv2/security-policy.html

Mitigation of Vulnerabilities: Older protocols (TLS 1.0/1.1) are considered insecure and have been deprecated by most modern browsers and operating systems.

https://trendmicro.com/cloudoneconformity/knowledge-base/aws/ELBv2/security-policy.html

### Checklist
- [ ] Manually tested

### DIFF
--- a/helm/aws-load-balancer-controller/values.yaml
+++ b/helm/aws-load-balancer-controller/values.yaml
@@ -259,7 +259,7 @@ tolerateNonExistentBackendService:
 tolerateNonExistentBackendAction:
 
 # defaultSSLPolicy specifies the default SSL policy to use for TLS/HTTPS listeners
-defaultSSLPolicy:
+defaultSSLPolicy: ELBSecurityPolicy-TLS13-1-3-2021-06
 
 # Liveness probe configuration for the controller
 livenessProbe:


helm/aws-load-balancer-controller/values.yaml
```diff
@@ -259,7 +259,7 @@ tolerateNonExistentBackendService:
 tolerateNonExistentBackendAction:
 
 # defaultSSLPolicy specifies the default SSL policy to use for TLS/HTTPS listeners
-defaultSSLPolicy:
+defaultSSLPolicy: ELBSecurityPolicy-TLS13-1-3-2021-06
```
guessi, line 262: Default of the `defaultSSLPolicy` will be set here
- https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/config/controller_config.go#L138-L139

where the `defaultSSLPolicy` definition could be found at
- https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/config/controller_config.go#L27

Changing the `values.yaml` could be a breaking change for existing users.

zac-nixon: +1. This change is particularly dangerous. If you wish to enforce SSL policy enforcement, I would suggest a feature that sets the default at the controller level using feature flags.

shraddhabang: We already have a [controller flag](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.11/deploy/configurations/#:~:text=Kubernetes%20cluster%20name-,default%2Dssl%2Dpolicy,-string) `default-ssl-policy` which customers can use to set their desired policy. The controller level default is `ELBSecurityPolicy-2016-08` which aligns with our CLI behavior mentioned earlier. So I don't think this is a valid change as this could break existing customer.

PR #4056, opened 2025-02-14
add prometheus metrics

wweiwei-li:
### Description

-  metricAPIPermissionErrorsTotal
-  metricAPILimitExceededTotal
-  metricAPIThrottledTotal
-  metricAPIValidationErrorTotal
-  MetricControllerReconcileErrors tracks the total number of controller errors by error type.
-  MetricControllerReconcileStageDuration tracks latencies of different reconcile stages.
-  MetricWebhookValidationFailure tracks the total number of validation errors by error type.
-  MetricWebhookMutationFailure tracks the total number of mutation errors by error type.
-  MetricControllerCacheObjectCount tracks the total number of object in the controller runtime cache.
-  MetricControllerTopTalker 

### Checklist
- [x] Manually tested

### DIFF
--- a/controllers/elbv2/targetgroupbinding_controller.go
+++ b/controllers/elbv2/targetgroupbinding_controller.go
@@ -19,10 +19,11 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"time"
+
 	discv1 "k8s.io/api/discovery/v1"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/pkg/errors"
@@ -31,16 +32,19 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/aws-load-balancer-controller/controllers/elbv2/eventhandlers"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/config"
+	errmetrics "sigs.k8s.io/aws-load-balancer-controller/pkg/error"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/k8s"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/runtime"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/targetgroupbinding"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 
 	"github.com/go-logr/logr"
+	lbcmetrics "sigs.k8s.io/aws-load-balancer-controller/pkg/metrics/lbc"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	elbv2api "sigs.k8s.io/aws-load-balancer-controller/apis/elbv2/v1beta1"
+	metricsutil "sigs.k8s.io/aws-load-balancer-controller/pkg/metrics/util"
 )
 
 const (
@@ -51,7 +55,7 @@ const (
 // NewTargetGroupBindingReconciler constructs new targetGroupBindingReconciler
 func NewTargetGroupBindingReconciler(k8sClient client.Client, eventRecorder record.EventRecorder, finalizerManager k8s.FinalizerManager,
 	tgbResourceManager targetgroupbinding.ResourceManager, config config.ControllerConfig, deferredTargetGroupBindingReconciler DeferredTargetGroupBindingReconciler,
-	logger logr.Logger) *targetGroupBindingReconciler {
+	logger logr.Logger, metricsCollector lbcmetrics.MetricCollector, reconcileCounters *metricsutil.ReconcileCounters) *targetGroupBindingReconciler {
 
 	return &targetGroupBindingReconciler{
 		k8sClient:                            k8sClient,
@@ -60,6 +64,8 @@ func NewTargetGroupBindingReconciler(k8sClient client.Client, eventRecorder reco
 		tgbResourceManager:                   tgbResourceManager,
 		deferredTargetGroupBindingReconciler: deferredTargetGroupBindingReconciler,
 		logger:                               logger,
+		metricsCollector:                     metricsCollector,
+		reconcileCounters:                    reconcileCounters,
 
 		maxConcurrentReconciles:    config.TargetGroupBindingMaxConcurrentReconciles,
 		maxExponentialBackoffDelay: config.TargetGroupBindingMaxExponentialBackoffDelay,
@@ -75,6 +81,8 @@ type targetGroupBindingReconciler struct {
 	tgbResourceManager                   targetgroupbinding.ResourceManager
 	deferredTargetGroupBindingReconciler DeferredTargetGroupBindingReconciler
 	logger                               logr.Logger
+	metricsCollector                     lbcmetrics.MetricCollector
+	reconcileCounters                    *metricsutil.ReconcileCounters
 
 	maxConcurrentReconciles    int
 	maxExponentialBackoffDelay time.Duration
@@ -93,13 +101,19 @@ type targetGroupBindingReconciler struct {
 // +kubebuilder:rbac:groups="discovery.k8s.io",resources=endpointslices,verbs=get;list;watch
 
 func (r *targetGroupBindingReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
+	r.reconcileCounters.IncrementTGB(req.NamespacedName)
 	r.logger.V(1).Info("Reconcile request", "name", req.Name)
 	return runtime.HandleReconcileError(r.reconcile(ctx, req), r.logger)
 }
 
 func (r *targetGroupBindingReconciler) reconcile(ctx context.Context, req reconcile.Request) error {
 	tgb := &elbv2api.TargetGroupBinding{}
-	if err := r.k8sClient.Get(ctx, req.NamespacedName, tgb); err != nil {
+	var err error
+	fetchTargetGroupBindingFn := func() {
+		err = r.k8sClient.Get(ctx, req.NamespacedName, tgb)
+	}
+	r.metricsCollector.ObserveControllerReconcileLatency(controllerName, "fetch_targetGroupBinding", fetchTargetGroupBindingFn)
+	if err != nil {
 		return client.IgnoreNotFound(err)
 	}
 
@@ -110,15 +124,23 @@ func (r *targetGroupBindingReconciler) reconcile(ctx context.Context, req reconc
 }
 
 func (r *targetGroupBindingReconciler) reconcileTargetGroupBinding(ctx context.Context, tgb *elbv2api.TargetGroupBinding) error {
-	if err := r.finalizerManager.AddFinalizers(ctx, tgb, targetGroupBindingFinalizer); err != nil {
+	var err error
+	finalizerFn := func() {
+		err = r.finalizerManager.AddFinalizers(ctx, tgb, targetGroupBindingFinalizer)
+	}
+	r.metricsCollector.ObserveControllerReconcileLatency(controllerName, "add_finalizers", finalizerFn)
+	if err != nil {
 		r.eventRecorder.Event(tgb, corev1.EventTypeWarning, k8s.TargetGroupBindingEventReasonFailedAddFinalizer, fmt.Sprintf("Failed add finalizer due to %v", err))
-		return err
+		return errmetrics.NewErrorWithMetrics(controllerName, "add_finalizers_error", err, r.metricsCollector)
 	}
 
-	deferred, err := r.tgbResourceManager.Reconcile(ctx, tgb)
-
+	var deferred bool
+	tgbResourceFn := func() {
+		deferred, err = r.tgbResourceManager.Reconcile(ctx, tgb)
+	}
+	r.metricsCollector.ObserveControllerReconcileLatency(controllerName, "reconcile_targetgroupbinding", tgbResourceFn)
 	if err != nil {
-		return err
+		return errmetrics.NewErrorWithMetrics(controllerName, "reconcile_targetgroupbinding_error", err, r.metricsCollector)
 	}
 
 	if deferred {
@@ -126,9 +148,12 @@ func (r *targetGroupBindingReconciler) reconcileTargetGroupBinding(ctx context.C
 		return nil
 	}
 
-	if err := r.updateTargetGroupBindingStatus(ctx, tgb); err != nil {
-		r.eventRecorder.Event(tgb, corev1.EventTypeWarning, k8s.TargetGroupBindingEventReasonFailedUpdateStatus, fmt.Sprintf("Failed update status due to %v", err))
-		return err
+	updateTargetGroupBindingStatusFn := func() {
+		err = r.updateTargetGroupBindingStatus(ctx, tgb)
+	}
+	r.metricsCollector.ObserveControllerReconcileLatency(controllerName, "update_status", updateTargetGroupBindingStatusFn)
+	if err != nil {
+		return errmetrics.NewErrorWithMetrics(controllerName, "update_status_error", err, r.metricsCollector)
 	}
 
 	r.eventRecorder.Event(tgb, corev1.EventTypeNormal, k8s.TargetGroupBindingEventReasonSuccessfullyReconciled, "Successfully reconciled")
--- a/controllers/ingress/group_controller.go
+++ b/controllers/ingress/group_controller.go
@@ -3,6 +3,7 @@ package ingress
 import (
 	"context"
 	"fmt"
+
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/go-logr/logr"
@@ -21,8 +22,11 @@ import (
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/deploy"
 	elbv2deploy "sigs.k8s.io/aws-load-balancer-controller/pkg/deploy/elbv2"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/deploy/tracking"
+	errmetrics "sigs.k8s.io/aws-load-balancer-controller/pkg/error"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/ingress"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/k8s"
+	lbcmetrics "sigs.k8s.io/aws-load-balancer-controller/pkg/metrics/lbc"
+	metricsutil "sigs.k8s.io/aws-load-balancer-controller/pkg/metrics/util"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/model/core"
 	elbv2model "sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2"
 	networkingpkg "sigs.k8s.io/aws-load-balancer-controller/pkg/networking"
@@ -48,7 +52,7 @@ func NewGroupReconciler(cloud services.Cloud, k8sClient client.Client, eventReco
 	finalizerManager k8s.FinalizerManager, networkingSGManager networkingpkg.SecurityGroupManager,
 	networkingSGReconciler networkingpkg.SecurityGroupReconciler, subnetsResolver networkingpkg.SubnetsResolver,
 	elbv2TaggingManager elbv2deploy.TaggingManager, controllerConfig config.ControllerConfig, backendSGProvider networkingpkg.BackendSGProvider,
-	sgResolver networkingpkg.SecurityGroupResolver, logger logr.Logger) *groupReconciler {
+	sgResolver networkingpkg.SecurityGroupResolver, logger logr.Logger, metricsCollector lbcmetrics.MetricCollector, reconcileCounters *metricsutil.ReconcileCounters) *groupReconciler {
 
 	annotationParser := annotations.NewSuffixAnnotationParser(annotations.AnnotationPrefixIngress)
 	authConfigBuilder := ingress.NewDefaultAuthConfigBuilder(annotationParser)
@@ -61,10 +65,10 @@ func NewGroupReconciler(cloud services.Cloud, k8sClient client.Client, eventReco
 		authConfigBuilder, enhancedBackendBuilder, trackingProvider, elbv2TaggingManager, controllerConfig.FeatureGates,
 		cloud.VpcID(), controllerConfig.ClusterName, controllerConfig.DefaultTags, controllerConfig.ExternalManagedTags,
 		controllerConfig.DefaultSSLPolicy, controllerConfig.DefaultTargetType, controllerConfig.DefaultLoadBalancerScheme, backendSGProvider, sgResolver,
-		controllerConfig.EnableBackendSecurityGroup, controllerConfig.DisableRestrictedSGRules, controllerConfig.IngressConfig.AllowedCertificateAuthorityARNs, controllerConfig.FeatureGates.Enabled(config.EnableIPTargetType), logger)
+		controllerConfig.EnableBackendSecurityGroup, controllerConfig.DisableRestrictedSGRules, controllerConfig.IngressConfig.AllowedCertificateAuthorityARNs, controllerConfig.FeatureGates.Enabled(config.EnableIPTargetType), logger, metricsCollector)
 	stackMarshaller := deploy.NewDefaultStackMarshaller()
 	stackDeployer := deploy.NewDefaultStackDeployer(cloud, k8sClient, networkingSGManager, networkingSGReconciler, elbv2TaggingManager,
-		controllerConfig, ingressTagPrefix, logger)
+		controllerConfig, ingressTagPrefix, logger, metricsCollector, controllerName)
 	classLoader := ingress.NewDefaultClassLoader(k8sClient, true)
 	classAnnotationMatcher := ingress.NewDefaultClassAnnotationMatcher(controllerConfig.IngressConfig.IngressClass)
 	manageIngressesWithoutIngressClass := controllerConfig.IngressConfig.IngressClass == ""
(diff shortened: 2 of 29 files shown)


controllers/elbv2/targetgroupbinding_controller.go
```diff
@@ -110,22 +114,26 @@ func (r *targetGroupBindingReconciler) reconcile(ctx context.Context, req reconc
 }
 
 func (r *targetGroupBindingReconciler) reconcileTargetGroupBinding(ctx context.Context, tgb *elbv2api.TargetGroupBinding) error {
+	defer r.metricsCollector.ObserveControllerReconcileLatency("targetGroupBinding", "add finalizers")
```
zac-nixon, line 117: I don't think this works correctly. If the idea is to get the latency of the reconcile request, we need to get a timestamp at the start of the function. Currently, looking at the implementation of `ObserveControllerReconcileLatency` it will always report 0.

zac-nixon, line 117: Just so I'm clear, the idea is to provide more granular reconcile latency metrics than what is already provided the controller-runtime? Have you looked into what constructs the controller-runtime provides so we don't need to roll our own?

main.go
```diff
@@ -107,6 +107,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	var lbcMetricsCollector *lbcmetrics.Collector
```
zac-nixon, line 110: Why did you change this? The original goal of this implementation was to to avoid having to nil check when no metrics were requested.

main.go
```diff
@@ -206,6 +212,14 @@ func main() {
 		setupLog.Error(err, "problem wait for podInfo repo sync")
 		os.Exit(1)
 	}
+
+	go func() {
+		if err := lbcMetricsCollector.StartCollectCacheSize(ctx); err != nil {
```
zac-nixon, line 217: This will currently cause a NPE if a registry is not defined.

pkg/deploy/elbv2/load_balancer_synthesizer.go
```diff
@@ -50,9 +53,11 @@ type loadBalancerSynthesizer struct {
 	controllerConfig               config.ControllerConfig
 	lbsNeedingCapacityModification []resAndSDKLoadBalancerPair
 	capacityReservationReconciler  LoadBalancerCapacityReservationReconciler
+	metricsCollector               lbcmetrics.MetricCollector
 }
 
 func (s *loadBalancerSynthesizer) Synthesize(ctx context.Context) error {
+	defer s.metricsCollector.ObserveControllerReconcileLatency("service/ingress", "synthesize load balancer")
```
zac-nixon, line 60: It might be useful to distinguish between ALB / NLB here.

pkg/ingress/model_build_load_balancer.go
```diff
@@ -41,34 +42,42 @@ func (t *defaultModelBuildTask) buildLoadBalancer(ctx context.Context, listenPor
...
 		return elbv2model.LoadBalancerSpec{}, err
 	}
 	securityGroups, err := t.buildLoadBalancerSecurityGroups(ctx, listenPortConfigByPort, ipAddressType)
 	if err != nil {
+		t.metricsCollector.ObserveControllerReconcileError("ingress", "build model error", "build security groups")
 		return elbv2model.LoadBalancerSpec{}, err
 	}
 	coIPv4Pool, err := t.buildLoadBalancerCOIPv4Pool(ctx)
 	if err != nil {
+		t.metricsCollector.ObserveControllerReconcileError("ingress", "build model error", "build customer owned IPv4 pool")
```
zac-nixon, line 65: Instead of return an error here, it might be good to define a custom error struct that embeds the error message and various metric fields into the struct. The plus side here is that
1/ It's impossible to forget add new metrics for new fields.
2/ It's hopefully a little more readable.

LMK what you think.

pkg/metrics/aws/collector.go
```diff
@@ -71,6 +72,38 @@ func WithSDKCallMetricCollector(c *Collector) func(stack *smithymiddleware.Stack
 				labelStatusCode: statusCode,
 				labelErrorCode:  errorCode,
 			}).Inc()
+
+			// https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/CommonErrors.html
+			if statusCode == "401" || statusCode == "403" {
+				c.instruments.apiCallAuthErrorsTotal.With(map[string]string{
+					labelService:    service,
```
zac-nixon, line 79: nit: you define this same string map for every case.

pkg/metrics/aws/collector.go
```diff
@@ -71,6 +72,38 @@ func WithSDKCallMetricCollector(c *Collector) func(stack *smithymiddleware.Stack
...
+					labelOperation:  operation,
+					labelStatusCode: statusCode,
+					labelErrorCode:  errorCode,
+				}).Inc()
+			} else if errorCode == "ServiceLimitExceeded" {
```
zac-nixon, line 84: Can you double check that the current metrics don't cover these cases? If they don't, it would be good to make these cases a little more generic. Right now, all the cases are tailored to ELB specific error codes.

pkg/metrics/aws/instruments.go
```diff
@@ -60,12 +70,41 @@ func newInstruments(registerer prometheus.Registerer) *instruments {
 		Help:      "Latency of an individual HTTP request to the service endpoint",
 	}, []string{labelService, labelOperation})
 
-	registerer.MustRegister(apiCallsTotal, apiCallDurationSeconds, apiCallRetries, apiRequestsTotal, apiRequestDurationSecond)
+	apiCallAuthErrorsTotal := prometheus.NewCounterVec(prometheus.CounterOpts{
+		Subsystem: metricSubSystem,
+		Name:      metricAPIAuthErrorsTotal,
+		Help:      "Number of failed AWS API calls that due to auth or authrorization failures",
```
zac-nixon, line 76: typo: remove `that`

zac-nixon: same for all these descriptions it looks like.

pkg/metrics/lbc/collector.go
```diff
@@ -1,39 +1,140 @@
...
+
+	// collect service cache size
+	svcList := &corev1.ServiceList{}
+	if err := cache.List(ctx, svcList); err != nil {
+		return fmt.Errorf("failed to list services: %w", err)
+	}
+	c.ObserveControllerCacheSize("service", len(svcList.Items))
+
+	// Collect TargetGroupBlinding cache size
+	tgbList := &elbv2api.TargetGroupBindingList{}
```
zac-nixon, line 133: can you collect the deferred target queue cache size too?

pkg/metrics/lbc/collector.go
```diff
@@ -1,39 +1,140 @@
...
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+			c.CollectCacheSize(ctx)
+		}
+	}
+}
+
+func (c *Collector) CollectCacheSize(ctx context.Context) error {
```
zac-nixon, line 115: We should make this a bit more generic as ultimately it's the same code block repeated. Basically defining a "collectable" resource map and then allowing the type to be passed in might work. Let's talk offline about how to make this more extendable.

pkg/metrics/lbc/instruments.go
```diff
@@ -12,15 +12,41 @@ const (
...
+	// MetricControllerReconcileStageDuration tracks latencies of different reconcile stages.
+	MetricControllerReconcileStageDuration = "controller_reconcile_stage_duration"
+	// MetricWebhookValidationFailure tracks the total number of validation errors by error type.
+	MetricWebhookValidationFailure = "webhook_validation_failure_total"
+	// MetricWebhookMutationFailure tracks the total number of mutation errors by error type.
+	MetricWebhookMutationFailure = "webhook_mutation_failure_total"
+	// MetricControllerCacheObjectCount tracks the total number of object in the controller runtime cache.
+	MetricControllerCacheObjectCount = "controller_cache_object_total"
+	// MetricPodReadinessGateFlipAboveX tracks readiness gate flips that are X seconds old
+	MetricPodReadinessGateFlipAbove60Seconds = "readiness_gate_above_60_seconds"
```
zac-nixon, line 26: This should be derivable from the buckets we've defined.

pkg/metrics/lbc/collector.go
wweiwei-li: Sure, will add.

pkg/metrics/lbc/collector.go
wweiwei-li: Agree, "collectable" resource map makes sense. will update it

pkg/metrics/aws/instruments.go
wweiwei-li: Good catch, will remove them

pkg/metrics/lbc/instruments.go
wweiwei-li: Are you saying the MetricPodReadinessGateFlipAbove60Seconds are redundant ? yeah,  I know we are able to derive how many of them exceed 60s and 90s using PromQL. However, I was thinking we want to some direct metrics ?

pkg/metrics/aws/collector.go
wweiwei-li: will fix it

pkg/ingress/model_build_load_balancer.go
wweiwei-li: Good point.  I agree.  will implement that

pkg/deploy/elbv2/load_balancer_synthesizer.go
wweiwei-li: yeah, agree, will update it to distinguish between ALB and NLB

controllers/elbv2/targetgroupbinding_controller.go
wweiwei-li: Yeah, the idea is to provide more granular reconcile latency metrics. I checked controller-runtime . It only provided end to end [reconcile time](https://github.com/kubernetes-sigs/controller-runtime/blob/bbc9711d9b2db31ce828714e7d3c875c00b4c1f1/pkg/internal/controller/metrics/metrics.go#L58).

wweiwei-li: Yeah, it is not working correctly. My original plan was something like this. Would this make more sense ?
```
reconcileAddFinalizerStartTime := time.Now()
....
r.metricsCollector.ObserveControllerReconcileLatency("targetGroupBinding", "add finalizers", time.Since(econcileAddFinalizerStartTime))

reconcileUpdateStatusStartTime := time.Now()
.....

r.metricsCollector.ObserveControllerReconcileLatency("targetGroupBinding", "update status", time.Since(reconcileUpdateStatusStartTime))
```

zac-nixon: Your suggestion works but we can cut down on code duplication. I would suggest defining something like this:
```
func (mc *MetricsCollector) latencyHelper(resource string, step string, fn func()) {
	start := time.Now()
	defer mc.ObserveControllerReconcileLatency(resource, step, time.Now().Sub(start))
	fn()
}
```

Then you can call it like:

```
var err error
finalizerFn := func () {
   err = r.finalizerManager.AddFinalizers(ctx, tgb, targetGroupBindingFinalizer)
} 
r.metricsCollector.latencyHelper("targetGroupBinding", "add finalizers", finalizerFn)
... handle error here ...
```

Let me know what you think.

zac-nixon: I actually forgot about the quirk with using time.Now and deferred statements.

https://stackoverflow.com/questions/72965657/measure-elapsed-time-with-a-defer-statement-in-go

it's a small change to the example.

pkg/metrics/aws/collector.go
zac-nixon: Agreed. Sounds good.

pkg/metrics/lbc/instruments.go
zac-nixon: In this case I think we're just duplicating the bucketing function that Prometheus supports out of the box. If we want these 2 specific buckets, adding them to this vector makes more sense. https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/metrics/lbc/instruments.go#L32

controllers/elbv2/targetgroupbinding_controller.go
M00nF1sh, line 117: have whitespaces in such "label" could cause issues when consuming those metrics in downstream services/code.
maybe switch things like "add finalizers" to "add_finalizers". (make it a machine readable identifier)

M00nF1sh, line 117: this `defer r.metricsCollector.ObserveControllerReconcileLatency("targetGroupBinding", "add finalizers")` won't work, it will always be 0. you need to return a anonymous function and call it in defer

controllers/elbv2/targetgroupbinding_controller.go
```diff
@@ -110,22 +114,26 @@ func (r *targetGroupBindingReconciler) reconcile(ctx context.Context, req reconc
...
 	if err := r.finalizerManager.AddFinalizers(ctx, tgb, targetGroupBindingFinalizer); err != nil {
 		r.eventRecorder.Event(tgb, corev1.EventTypeWarning, k8s.TargetGroupBindingEventReasonFailedAddFinalizer, fmt.Sprintf("Failed add finalizer due to %v", err))
 		return err
 	}
 
+	defer r.metricsCollector.ObserveControllerReconcileLatency("targetGroupBinding", "reconcile")
```
M00nF1sh, line 123: this seems not work due to how defer works in golang. 
ideally we should collect latency metrics spend for each task, instead of "latency spend since X stage".
There is a difference between
* latency spent on adding finalizers
* latency since adding finalizer to finish reconcile.
The first one is clearly the better option and allows us to monitor on

main.go
wweiwei-li: I think that's because I saw `awsMetricsCollector ` had the nil check, and I didn't see `lbcmetrics.NewCollector()` handle nil metrics.Registry internally.  So have a feeling we need to apply the same nil check. Let me know if I am wrong

main.go
wweiwei-li: Good catch, I will add check metrics.Registry 
```
if metrics.Registry != nil {
		go func() {
			if err := lbcMetricsCollector.StartCollectCacheSize(ctx); err != nil {
				setupLog.Error(err, "problem periodically collect cache size")
				os.Exit(1)
			}
		}()
	}
```

controllers/elbv2/targetgroupbinding_controller.go
wweiwei-li: Good point, will fix it

wweiwei-li: Yeah, like Zac suggested above,  

I will have 
```
	var err error
	finalizerFn := func() {
		err = r.finalizerManager.AddFinalizers(ctx, tgb, targetGroupBindingFinalizer)
	}
	r.metricsCollector.ObserveControllerReconcileLatency("targetGroupBinding", "add_finalizers", finalizerFn)
	if err != nil {
		return err
	}
```
```
func (c *Collector) ObserveControllerReconcileLatency(controller string, stage string, fn func()) {
	start := time.Now()
	defer func() {
		c.instruments.controllerReconcileLatency.With(prometheus.Labels{
			labelController:     controller,
			labelReconcileStage: stage,
		}).Observe(time.Since(start).Seconds())
	}()
	fn()
}
````

File: main.go (hunk 2 above)
Comment (zac-nixon) [reply main.go:110]: We should continue setting `lbcMetricsCollector`. The factory knows whether or not the metric registry is nil. If it's nil, we return a no-op collector. If the metric registry is not nil, we return an actual collector. The users of the metrics collector don't need to know either way.

File: pkg/metrics/aws/instruments.go [hunk 13]
```diff
@@ -13,6 +13,11 @@ const (
 
 	metricAPIRequestsTotal          = "api_requests_total"
 	metricAPIRequestDurationSeconds = "api_request_duration_seconds"
+
+	metricAPIAuthErrorsTotal      = "api_call_auth_errors_total"
+	metricAPILimitExceededTotal   = "api_call_limit_exceeded_total"
+	metricAPIThrottledTotal       = "api_call_throttled_total"
+	metricAPIValidationErrorTotal = "api_call_validation_error_total"
```
oliviassss, line 20: nit: maybe to be consistent on using either `Error` or `Errors` for the metrics naming.

pkg/metrics/aws/collector.go
```diff
@@ -71,6 +72,38 @@ func WithSDKCallMetricCollector(c *Collector) func(stack *smithymiddleware.Stack
 				labelStatusCode: statusCode,
 				labelErrorCode:  errorCode,
 			}).Inc()
+
+			// https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/CommonErrors.html
```
oliviassss, line 76: do we only consider ELBv2 errors in this PR? how about other like EC2 API errors?

wweiwei-li: Yeah, this was only considered for ELB. I think EC2 API also matters. I can add some common errors based on https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html

controllers/elbv2/targetgroupbinding_controller.go
```diff
@@ -110,25 +119,36 @@ func (r *targetGroupBindingReconciler) reconcile(ctx context.Context, req reconc
...
-	deferred, err := r.tgbResourceManager.Reconcile(ctx, tgb)
-
+	var deferred bool
+	tgbResourceFn := func() {
+		deferred, err = r.tgbResourceManager.Reconcile(ctx, tgb)
+	}
+	r.metricsCollector.ObserveControllerReconcileLatency("targetGroupBinding", "reconcile_targetgroupblinding", tgbResourceFn)
 	if err != nil {
-		return err
+		return errmetrics.NewErrorWithMetrics("targetGroupBinding", "reconcile_targetgroupblinding_error", err, r.metricsCollector)
```
zac-nixon, line 138: typo : `blinding` -> `binding`

controllers/elbv2/targetgroupbinding_controller.go
```diff
@@ -93,6 +101,7 @@ type targetGroupBindingReconciler struct {
 // +kubebuilder:rbac:groups="discovery.k8s.io",resources=endpointslices,verbs=get;list;watch
 
 func (r *targetGroupBindingReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
+	r.reconcileCounters.IncrementTGB(req.NamespacedName)
```
zac-nixon, line 104: nice!

controllers/elbv2/targetgroupbinding_controller.go
```diff
@@ -110,25 +119,36 @@ func (r *targetGroupBindingReconciler) reconcile(ctx context.Context, req reconc
...
 		return nil
 	}
 
-	if err := r.updateTargetGroupBindingStatus(ctx, tgb); err != nil {
-		r.eventRecorder.Event(tgb, corev1.EventTypeWarning, k8s.TargetGroupBindingEventReasonFailedUpdateStatus, fmt.Sprintf("Failed update status due to %v", err))
-		return err
+	updateTargetGroupBindingStatusFn := func() {
+		err = r.updateTargetGroupBindingStatus(ctx, tgb)
+	}
+	r.metricsCollector.ObserveControllerReconcileLatency("targetGroupBinding", "update_status", updateTargetGroupBindingStatusFn)
```
zac-nixon, line 149: Can you please define the resource as a const, it's re-used for every metric collection step. I think the step names are unique and don't need be consts.

controllers/ingress/group_controller.go
```diff
@@ -116,40 +126,57 @@ type groupReconciler struct {
...
-		if err != nil {
-			return err
-		}
-		if err := r.updateIngressGroupStatus(ctx, ingGroup, lbDNS); err != nil {
-			r.recordIngressGroupEvent(ctx, ingGroup, corev1.EventTypeWarning, k8s.IngressEventReasonFailedUpdateStatus, fmt.Sprintf("Failed update status due to %v", err))
-			return err
+		dnsResolveAndUpdateStatus := func() {
+			lbDNS, err := lb.DNSName().Resolve(ctx)
+			if err != nil {
+				return
```
zac-nixon, line 163: I believe you still need to propagate this error to the parent function. Right now, it will silently fail.

controllers/ingress/group_controller.go
```diff
@@ -116,40 +126,57 @@ type groupReconciler struct {
...
+			}
+			if err := r.updateIngressGroupStatus(ctx, ingGroup, lbDNS); err != nil {
```
zac-nixon, line 165: same comment, when using this nested function approach we need to propagate the error to the parent function.

controllers/ingress/group_controller.go
```diff
@@ -116,40 +126,57 @@ type groupReconciler struct {
...
 		}
+		r.metricsCollector.ObserveControllerReconcileLatency("ingress", "dns_resolve_and_update_status", dnsResolveAndUpdateStatus)
 	}
 
 	if len(ingGroup.InactiveMembers) > 0 {
-		if err := r.groupFinalizerManager.RemoveGroupFinalizer(ctx, ingGroupID, ingGroup.InactiveMembers); err != nil {
+		removeGroupFinalizerFn := func() {
+			err = r.groupFinalizerManager.RemoveGroupFinalizer(ctx, ingGroupID, ingGroup.InactiveMembers)
+		}
+		r.metricsCollector.ObserveControllerReconcileLatency("ingress", "remove_group_finalizer", removeGroupFinalizerFn)
```
zac-nixon, line 176: please define `ingress` as a constant.

controllers/service/service_controller.go
```diff
@@ -139,28 +167,47 @@ func (r *serviceReconciler) deployModel(ctx context.Context, svc *corev1.Service
...
+	}
+
+	deployModelFn := func() {
+		err = r.deployModel(ctx, svc, stack)
 	}
-	err := r.deployModel(ctx, svc, stack)
+	r.metricsCollector.ObserveControllerReconcileLatency("service", "deploy_model", deployModelFn)
 	if err != nil {
-		return err
+		return errmetrics.NewErrorWithMetrics("service", "deploy_model_error", err, r.metricsCollector)
```
zac-nixon, line 186: same, please define `service` as a const.

main.go
```diff
@@ -206,6 +210,25 @@ func main() {
 		setupLog.Error(err, "problem wait for podInfo repo sync")
 		os.Exit(1)
 	}
+
+	if metrics.Registry != nil {
```
zac-nixon, line 214: I'd prefer to just use the no-ops collector instead nil checking the registry. It makes the code easier to reason about.

pkg/error/error_with_metrics.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+func NewErrorWithMetrics(resourceType string, errorCategory string, err error, metricCollector lbcmetrics.MetricCollector) *ErrorWithMetrics {
+	reconcileErr := &ErrorWithMetrics{
+		ResourceType:  resourceType,
+		ErrorCategory: errorCategory,
+		Err:           err,
+	}
+
+	if metricCollector != nil {
```
zac-nixon, line 20: same here, let's just utilize the no-op implementation so we don't need nil checks.

main.go
```diff
@@ -206,6 +210,25 @@ func main() {
...
+		go func() {
+			setupLog.Info("starting collect cache size")
+			if err := lbcMetricsCollector.StartCollectCacheSize(ctx); err != nil {
+				setupLog.Error(err, "problem periodically collect cache size")
+				os.Exit(1)
```
zac-nixon, line 219: What startup error(s) do we expect here?

docs/guide/metrics/prometheus/index.md
```diff
@@ -0,0 +1,81 @@
+# Monitoring Controller Metrics with Prometheus
```
zac-nixon, line 1: Thanks for documenting the process.

helm/aws-load-balancer-controller/crds/crds.yaml
```diff
@@ -54,6 +54,12 @@ spec:
           spec:
             description: IngressClassParamsSpec defines the desired state of IngressClassParams
             properties:
+              PrefixListsIDs:
```
zac-nixon, line 57: Is this a bad rebase?

pkg/metrics/lbc/collector.go
```diff
@@ -37,3 +78,107 @@ func (c *collector) ObservePodReadinessGateReady(namespace string, tgbName strin
...
+func (c *collector) StartCollectCacheSize(ctx context.Context) {
+	ticker := time.NewTicker(30 * time.Second)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			c.CollectCacheSize(ctx)
```
zac-nixon, line 137: nit: might be good to log that an error was returned here (although that shouldnt stop the collection loop)

pkg/metrics/util/reconcile_counter.go
```diff
@@ -0,0 +1,62 @@
+package util
+
+import (
+	"sort"
+
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type ReconcileCounters struct {
+	serviceReconciles map[types.NamespacedName]int
```
zac-nixon, line 10: What will garbage collect these maps?

pkg/metrics/util/reconcile_counter.go
```diff
@@ -0,0 +1,62 @@
...
+
+func NewReconcileCounters() *ReconcileCounters {
+	return &ReconcileCounters{
+		serviceReconciles: make(map[types.NamespacedName]int),
+		ingressReconciles: make(map[types.NamespacedName]int),
+		tgbReconciles:     make(map[types.NamespacedName]int),
+	}
+}
+
+func (c *ReconcileCounters) IncrementService(namespaceName types.NamespacedName) {
```
zac-nixon, line 28: These maps probably need to be guarded with a mutex.

pkg/metrics/util/reconcile_counter.go
```diff
@@ -0,0 +1,81 @@
...
+
+	topReconciles["service"] = getTopN(c.serviceReconciles)
+	topReconciles["ingress"] = getTopN(c.ingressReconciles)
+	topReconciles["targetgroupbinding"] = getTopN(c.tgbReconciles)
+
+	return topReconciles
+}
+
+func (c *ReconcileCounters) ResetCounter() {
+	c.mutex.Lock()
```
zac-nixon, line 76: You're going to drop potential requests between calling `GetTopReconciles` and `ResetCounter`. But I think that's ok, as they are just metrics.

Issue #3877, opened 2024-10-02
Disable proxy protocol for TCP/UDP ports in NLB


sknmi:
I'm using nginx-ingress and want to expose gitlab-shell service on TCP port 22. 
```
defaultBackend:
  enabled: false

tcp:
  22: gitlab/gitlab-gitlab-shell:22

  config:
    server-tokens: false
    use-proxy-protocol: true
  service:
    type: "LoadBalancer"
    annotations:
      service.beta.kubernetes.io/aws-load-balancer-type: "external"
      service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: "ip"
      service.beta.kubernetes.io/aws-load-balancer-scheme: "internal"
      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: "tcp"
      service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: true
      service.beta.kubernetes.io/aws-load-balancer-target-group-attributes: "proxy_protocol_v2.enabled=true,preserve_client_ip.enabled=true,deregistration_delay.timeout_seconds=180,deregistration_delay.connection_termination.enabled=false"
```

So for 443/80 ports I need proxy protocol and it works as expected, but for port 22 I don't need it. If I disabled it in target group by hands aws-load-balancer-controler sets proxy protocol for that tcp 22 port back after some time.

I know it is also possible to set something like this fornginx-ingress:
```
  22: gitlab/gitlab-gitlab-shell:22:PROXY:PROXY
  22: gitlab/gitlab-gitlab-shell:22:PROXY
  22: gitlab/gitlab-gitlab-shell:22::PROXY
```
But none of them works.

**Describe the solution you'd like**
Ability to disable proxy-protocol for specific ports.

**Describe alternatives you've considered**
Using https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.9/guide/targetgroupbinding/targetgroupbinding/ this maybe.

wweiwei-li:
We don't have the ability to disable it for specific ports.  We add it as feature. Any community contributes are welcome

aburan28:
@wweiwei-li What would be the ideal way to implement this? Would a target group override annotation for the service port be reasonable? for example below this would disable `proxy_protocol_v2` for port 443.


```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
[...]

PR #4079, opened 2025-03-06
Added support for setting Proxy protocol per target group based on ServicePort

pthak94:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3877

### Description

Introduced a new annotation "service.beta.kubernetes.io/aws-load-balancer-proxy-protocol-per-target-group" with which, proxy protocol setting can be configured for each target group individually. Example usage is "80, 443", which will enable proxy protocols for the target groups associated to port 80 and 443 listeners. And disable for the rest, for example port 22.

To achieve this, when target group is created, it's service port is compared to the value of this annotation, if it's found, it's value is set to true, otherwise set to false.

"service.beta.kubernetes.io/aws-load-balancer-proxy-protocol" annotation still overrides this new annotation.

Added test cases to check if the new annotation works, and that it is still overriden by "service.beta.kubernetes.io/aws-load-balancer-proxy-protocol" annotation.

Finally, added documentation regarding this new annotation.

This version of controller is tested in my EKS testing environment.

-->

### Checklist
- [x] Manually tested


docs/guide/service/annotations.md
```diff
@@ -23,6 +23,7 @@
...
 | [service.beta.kubernetes.io/aws-load-balancer-proxy-protocol](#proxy-protocol-v2)                | string                  |                           | Set to `"*"` to enable                                                                                                                                                                                                              |
+| [service.beta.kubernetes.io/aws-load-balancer-proxy-protocol-per-target-group](#proxy-protocol-v2)                | string                  |                           | If specified,configures proxy protocol per targrt group based on ServicePort. For example `"80, true, 443, true, 22, false"` will disable proxy protocol for port 22 only. This annotation is overriden by `"service.beta.kubernetes.io/aws-load-balancer-proxy-protocol"`                                                                                                                                                                                                            |
```
zac-nixon, line 26: I don't think this is the correct way to model this information. How about just using a single list `80,443` to denote that those ports should have proxy protocol enabled.

zac-nixon (commented):
Thank you for the contribution. Please consider my comment.

pkg/service/model_build_target_group.go
```diff
@@ -204,41 +204,66 @@ func (t *defaultModelBuildTask) buildTargetGroupName(_ context.Context, svcPort
...
+                }
+                enabledPorts[trimmedPort] = struct{}{}
+            }
+        }
+
+        currentPortStr := strconv.FormatInt(int64(port.Port), 10)
+        if _, enabled := enabledPorts[currentPortStr]; enabled {
+            rawAttributes[tgAttrsProxyProtocolV2Enabled] = "true"
+        } else {
+            rawAttributes[tgAttrsProxyProtocolV2Enabled] = "false"
```
zac-nixon, line 237: to keep existing behavior, can you remove this line? the default for this attribute is false anyways.

zac-nixon:
overall looks pretty good. please consider my comment.

i appreciate the contribution!

pkg/service/model_build_target_group.go
zac-nixon: Thanks for the explanation. I was just thinking that it would be nice to have the `false` value set when we didn't need it. But I'm happy with the implementation as-is and I appreciate the work you did here :)

Issue #3953, opened 2024-11-25
Ambiguity when checking for existence of Backend SGs


visit1985:
When searching for the existence of backend SGs, the controller searches by tags and does not consider the case where multiple SGs with that tags are returned.

To avoid ambiguity, the below function should either search by SG name (which is already present as an input parameter), or return an error when multiple SGs with the given tags are found.

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/8ba34e27224d1f7a56f0e416086c43b6877c7e65/pkg/networking/backend_sg_provider.go#L284-L310

shraddhabang:
Hey @visit1985 , While configuring the backend sg, the LBC uses a shared backend SG for all the resources which it creates once and applies these tags. So the customer should avoid to use the same tag for other sgs. 
If you want to configure your own backend sg, you can disable this automanagement by setting flag ``--enable-backend-security-group=false``. Then in that case the LBC will configure the SG that you provide. 

More info on SG management : https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/deploy/security_groups/#backend-security-groups

visit1985:
Yes, right. But even the  it’s better to fail instead if using a random SG via `sgs[0]`.

M00nF1sh:
@visit1985 
We should avoid dependencies on the `name`(which allows us to change naming patterns)
I'm curious why you would run into this issue? since the backendSG is already scoped by the clusterTag & resourceTag and vpcID, there shouldn't be more than one such backendSG exists. Did you created a SG with same set of tags out of the controller?

visit1985:
Yes, someone created it during preparation to migrate to command line flag  `--backend-security-group`.  For some deployments this resulted in a scenario where both SGs are in use.

Adding a raise condition for `len(sgs) > 1` should be backwards compatible, does not create a dependency on `name` and would inform the customer that he should avoid to use the same tag for other SGs.

PR #4091, opened 2025-03-16
raise an error when multiple backendSGs exist

visit1985:
### Issue

#3953

### Description

Raise an error when multiple backendSGs with same vpc-id and tags exist instead of using the first one found.

### Checklist
- [X] Manually tested

### DIFF
--- a/pkg/networking/backend_sg_provider.go
+++ b/pkg/networking/backend_sg_provider.go
@@ -298,10 +298,14 @@ func (p *defaultBackendSGProvider) getBackendSGFromEC2(ctx context.Context, sgNa
 			},
 		},
 	}
-	p.logger.V(1).Info("Queriying existing SG", "vpc-id", vpcID, "name", sgName)
+	tags := fmt.Sprintf("%v=%v, %v=%v", tagKeyK8sCluster, p.clusterName, tagKeyResource, tagValueBackend)
+	p.logger.V(1).Info("Querying existing SG", "vpc-id", vpcID, "tags", tags)
 	sgs, err := p.ec2Client.DescribeSecurityGroupsAsList(ctx, req)
 	if err != nil && !isEC2SecurityGroupNotFoundError(err) {
 		return "", err
+	}
+	if len(sgs) > 1 {
+		return "", errors.Errorf("Found multiple SGs with vpc-id %v and tags %v", vpcID, tags)
 	}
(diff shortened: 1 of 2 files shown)


pkg/networking/backend_sg_provider.go
```diff
@@ -298,11 +298,15 @@ func (p *defaultBackendSGProvider) getBackendSGFromEC2(ctx context.Context, sgNa
...
 	}
-	p.logger.V(1).Info("Queriying existing SG", "vpc-id", vpcID, "name", sgName)
+	tags := fmt.Sprintf("%v=%v, %v=%v", tagKeyK8sCluster, p.clusterName, tagKeyResource, tagValueBackend)
+	p.logger.V(1).Info("Querying existing SG", "vpc-id", vpcID, "tags", tags)
 	sgs, err := p.ec2Client.DescribeSecurityGroupsAsList(ctx, req)
 	if err != nil && !isEC2SecurityGroupNotFoundError(err) {
 		return "", err
 	}
+	if len(sgs) > 1 {
+		return "", errors.Errorf("Found multiple SGs with vpc-id %v and tags %v", vpcID, tags)
```
zac-nixon, line 308: Breaking changes like this should be feature flagged.

zac-nixon (changes requested):

pkg/networking/backend_sg_provider.go
visit1985: How do I do that?

zac-nixon: example of adding a feature flag
https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/c78d2ca84daba60ae2cad2b2043a0a34a263503f

visit1985: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3953#issuecomment-2521711703
> the customer should avoid to use the same tag for other sgs.

Imho. this is a bugfix, not a feature. It may be marked as a breaking change, yes. But, since it‘s not in the interest of the project to have customers keep the old behavior, I don’t see why we should introduce a feature flag for it.

zac-nixon: We are not merging a change that results in a breaking change.

PR #4093, opened 2025-03-17
Feat(performance): Use protobuf as default type instead of json

RiskyAdventure:
### Issue

Currently the controller only accepts the JSON responses from kube-apiserver. Accepting the more efficient Protobuf representation responses will improve performance at scale. This affects majorly the Pod cache ListWatch that is set [here](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/k8s/pod_info_repo.go#L37) and can act as a bottleneck in clusters with large amount of pods.

By default, Kubernetes returns objects serialized to JSON with content type application/json. This is the default serialization format for the API. However, clients may request the more efficient [Protobuf representation](https://kubernetes.io/docs/reference/using-api/api-concepts/#protobuf-encoding) of these objects for better performance at scale.

Discussion on issue 3749 requested protobuf as the default vs. using flags to enable.
https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3749

### Checklist
- [ ] Manually tested
- [ x] Made sure the title of the PR is a good description that can go into the release notes

### DIFF
--- a/pkg/config/runtime_config.go
+++ b/pkg/config/runtime_config.go
@@ -109,7 +109,6 @@ func BuildRestConfig(rtCfg RuntimeConfig) (*rest.Config, error) {
 	}
 
 	restCFG.AcceptContentTypes = "application/vnd.kubernetes.protobuf,application/json"
-	restCFG.ContentType = "application/vnd.kubernetes.protobuf"
 	restCFG.QPS = defaultQPS
 	restCFG.Burst = defaultBurst
 	return restCFG, nil
--- a/pkg/config/runtime_config.go
+++ b/pkg/config/runtime_config.go
@@ -108,6 +108,8 @@ func BuildRestConfig(rtCfg RuntimeConfig) (*rest.Config, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


zac-nixon:
I think it's a good change but curious why the feature flag approach wasn't taken. The linked issue doesn't have much details.

RiskyAdventure:
> I think it's a good change but curious why the feature flag approach wasn't taken? The linked issue doesn't have much details.

I should've covered that, thank you for asking. 

- Other projects such as aws-ebs-csi-driver do this by default. It would be a confusing customer experience to have to change a flag on this project, but not other common AWS drivers.
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2138/commits/75421fae6cc06cc9feadaa96a6c3176317b47a9d

- The serialization impact is difficult to troubleshoot as you need to pull the annotations on the audit log to understand if you're having this problem. Since the performance impact can be quite substantial, having a default set, would potentially prevent customer issues from happening in the first place.

zac-nixon:
This doesn't work because we need some more work to support CRDs using Protobuf:
https://anhduongviet.medium.com/working-with-kubernetes-crd-model-as-protobuf-ce5c8031cea0

Issue #4089, opened 2025-03-13
v2.12.0 Cross Account TG Binding enforcing Availability Zone = All


listellm:
**Bug Description**
As documented in the [AWS EKS Best Practises Guide](URL) under the "Sharing VPC across multiple accounts" section, we are using AWS RAM to share our cluster private and cluster public subnets with other accounts in our organisation.  It allows us to reduce hops between our cluster account and our customer account.

When using the new cross-account TG binding feature in v2.12.0, the AWS LB controller forces the AZ to be set to `all` as shown in your code [here](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/879e7159d345a73dbabc09fe9986505bddef3c85/pkg/targetgroupbinding/resource_manager.go#L555).  

```
	if tgb.Spec.IamRoleArnToAssume != "" {
		// If we're interacting with another account, then we should always be setting "all" AZ to allow this
		// target to get registered by the ELB API.
		overrideAzFn = func(_ netip.Addr) bool {
			return true
		}
	}
```

This is in-turn then denied by the AWS API when using shared subnets as the pod IP is on the same VPC CIDR as the LB subnets.

```
operation error Elastic Load Balancing v2: RegisterTargets, https response error StatusCode: 400, RequestID: fxxxx-xxxx-xxxx-xxxx--xxxxx, api error ValidationError: The IP address '10.80.45.136' is within the VPC, and cannot have its Availability Zone overridden to 'all' from 'eu-central-1b'
```

This prevents using cross-account TG binding with shared subnets.

**Steps to Reproduce**
In the following examples, we will refer to Cluster Owner (CO) and Target Group Owner (TGO) accounts.

1/Ensure that your CO VPC has shared subnets
2/ Create your LB in the TGO account on a shared subnet from CO account
3/ Create a TG Binding utilising the cross-account LB feature per documented instructions
4/ Observe the LB controller logs which will state the above AWS API error

**Expected Behavior**
The LB controller detects the AZ that the target pod IP is in and sets this on the TG binding request for the AZ instead of forcing `all`.

**Actual Behavior**
AWS API denies the request as the source POD IP is within the same VPC CIDR that the LB is in but the LB controller tries to set `all` which should only be set if the private IP address is outside of the VPC scope

**Regression**
Was the functionality working correctly in a previous version? N/A.  its a newly launched feature
If yes, specify the last version where it worked as expected

**Current Workarounds**
[...]

PR #4095, opened 2025-03-17
[bug fix] handle ram shared VPCs for cross account tgb

zac-nixon:
### Issue
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4089

### Description

The initial implementation didn't account for RAM shared VPCs, targets in target groups that use RAM shared VPCs can't set the AZ = 'all' as the VPC exists in the account that owns the target group.

This CR adds support to handle either peered VPCs (which will return not found when we resolve the VPC as we use the cluster accounts' credentials to resolve the VPC) and RAM shared VPCs (which will get resolved from the resolve VPC call).

I wsa able to test this out in my own cluster using a peered VPC and RAM shared VPC. Furthermore, I backfilled some missing tests and reiterated the fix details in the code.

### Checklist
- [x] Manually tested

### DIFF
--- a/helm/aws-load-balancer-controller/crds/crds.yaml
+++ b/helm/aws-load-balancer-controller/crds/crds.yaml
@@ -54,6 +54,12 @@ spec:
           spec:
             description: IngressClassParamsSpec defines the desired state of IngressClassParams
             properties:
+              PrefixListsIDs:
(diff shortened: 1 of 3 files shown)


shraddhabang (approved):
/lgtm

PR #4112, opened 2025-03-28
Docs: mention ACM in Certificate Discovery

sdarwin:
Hi,

This is both an Issue and an attempted Pull Request to solve the issue.

On the page "Certificate Discovery" https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/ingress/cert_discovery/  , I was reading and wondered about where the discovered TLS certificates come from. Consider that the main kubernetes documentation about Ingress https://kubernetes.io/docs/concepts/services-networking/ingress/ includes a nearly identical spec:

```
spec:
  tls:
  - hosts:
      - https-example.foo.com
```

Where do those TLS certs on the official page come from?  
They are k8s Secrets, which are resources in the cluster, similar to ConfigMaps.      
If this controller is doing something similar to the official Ingress spec, but "discovering" tls, perhaps that also means from k8s secrets.    
"ACM" "AWS Certificate Manager" is referred to 0 times on the "Certificate Discovery" page.    
All of this might lead one to think about "secrets" and not "ACM". Which is it? And so, the text update in this pull request.   
Let me know, if it should to be edited/modified.


sdarwin:
There is another topic that could be mentioned on this page also. 
1. Often, certificates are associated with multiple domains.  
2. One ingress might have multiple different entries. with multiple domains.

text -

1.  "If an ACM certificate has many domains and/or wildcards, it's sufficient for the discovery process to find one of those domains, and that will count as a match."
2. show an example?

zac-nixon:
I agree with your two points, that would be helpful to mention! We can work on updating those examples, or would you prefer to work off this PR?

sdarwin:
Hi @zac-nixon ,

If you will work on it, that's great.   
For me, it doesn't matter if that's through this PR, or creating another one.    This PR has "Allow edits by maintainers".   But a new one is fine also.    
"I agree with your two points".  Yes. There is an issue I often find in documentation: let's say an --options-flag or a config setting can take multiple entries.   The docs might say that "You can add multiple domains".    The reader is left wondering "Would those be comma-separated, or completely separate flags, or in a formatted json array, or..."    Any time something can be multiplied it's worthwhile to actually show examples. I was trying to think of such cases here.  "What if a cert has multiple domains".  "What if an ingress takes multiple certs". etc.

sdarwin:
/remove-lifecycle stale

All the discussions in this thread are extra/optional. The pull request could be merged as-is.   It's a small change. Simply merge.

sdarwin:
/remove-lifecycle stale

The existing pull request is a small text modification, and could be merged as-is.

PR #4137, opened 2025-04-09
Subnet Discovery for Gateway API

zac-nixon:
### Description

This PR introduces subnet discovery for the Gateway API. It sets up some of the basic model builders to resolve the information needed to construct the ELB Subnets, such as the LB scheme and IP Address Type.

The goal here is to make re-usable logic for both NLB and ALB subnet creation; essentially adding, merging, and refactoring the existing subnet logic from the service and ingress packages.

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/service/model_build_load_balancer.go#L323

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/ingress/model_build_load_balancer.go#L200

The goal I have in mind is to prevent one big builder object like the Ingress and Service models currently have. Instead, I'd like to strive more granular pieces of the model builder. 

### Checklist
- [x] Manually tested


pkg/gateway/model/model_build_subnet.go
```diff
@@ -0,0 +1,221 @@
...
+		if (subnetConfig.PrivateIPv4Allocation != nil) != privateIPv4AllocationSpecified {
+			return false, errors.Errorf("Either specify all private ipv4 allocations or none.")
+		}
+
+		if (subnetConfig.SourceNatIPv6Prefix != nil) != sourceNATSpecified {
+			return false, errors.Errorf("Either specify all source nat prefixes or none.")
+		}
+	}
+
+	if eipAllocationSpecified {
```
shraddhabang, line 130: Nit : Lets move these checks above before we iterate through subnetConfigs to return early.

pkg/gateway/model/subnet/eip_mutator.go
```diff
@@ -0,0 +1,34 @@
...
+	if elbSubnets == nil || len(subnetConfig) == 0 || subnetConfig[0].EIPAllocation == nil {
+		return nil
+	}
+
+	if len(elbSubnets) != len(subnetConfig) {
+		return errors.Errorf("Unable to assign EIP allocations we have %+v subnets and %v EIP allocations", len(elbSubnets), len(subnetConfig))
+	}
+
+	for i, elbSubnet := range elbSubnets {
+		fmt.Println("HERE!")
```
shraddhabang, line 30: Nit: leftover log

pkg/gateway/model/subnet/ipv6_mutator.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (mutator *ipv6Mutator) Mutate(elbSubnets []*elbv2model.SubnetMapping, ec2Subnets []ec2types.Subnet, subnetConfig []elbv2gw.SubnetConfiguration) error {
+
+	// We've already validated it's already all or nothing for ipv6 allocations.
+	if elbSubnets == nil || len(subnetConfig) == 0 || subnetConfig[0].IPv6Allocation == nil {
+		return nil
+	}
+
+	if len(elbSubnets) != len(subnetConfig) {
```
shraddhabang, line 35: As this is common error among all the mutators, lets check this once before starting the mutators for each config to avoid duplicate code.

zac-nixon: Nice catch. I did this on purpose because. i wanted the mutators to be agnostic of each other. This was to future proof these mutators where we might run a subset of them in the future.

pkg/gateway/model/subnet/eip_mutator.go
zac-nixon: Will fix

shraddhabang (approved):
/lgtm

PR #4142, opened 2025-04-14
[feat: gw api] Security Group discovery and management

zac-nixon:
### Description

Adds logic to manage security groups, either via static IDs / names, or by controller managed SG. This logic combines the ingress and service paths to work for either nlb / alb.

### Checklist
- [x] Manually tested


pkg/gateway/model/model_build_loadbalancer.go
```diff
@@ -0,0 +1,120 @@
...
+		Scheme:                      scheme,
+		IPAddressType:               ipAddressType,
+		SubnetMappings:              subnets.subnets,
+		SecurityGroups:              securityGroupTokens,
+		LoadBalancerAttributes:      lbModelBuilder.buildLoadBalancerAttributes(lbConf),
+		MinimumLoadBalancerCapacity: lbModelBuilder.buildLoadBalancerMinimumCapacity(lbConf),
+		Tags:                        tags,
+	}
+
+	if lbModelBuilder.loadBalancerType == elbv2model.LoadBalancerTypeNetwork {
```
shraddhabang, line 63: Can we have dedicated functions to build these LB type specific params? instead of big if blocks for better readability?

pkg/gateway/model/model_build_loadbalancer.go
```diff
@@ -0,0 +1,120 @@
...
+	if lbModelBuilder.loadBalancerType == elbv2model.LoadBalancerTypeApplication {
+		spec.CustomerOwnedIPv4Pool = lbConf.Spec.CustomerOwnedIpv4Pool
+		spec.IPv4IPAMPool = lbConf.Spec.IPv4IPAMPoolId
+	}
+
+	return spec, nil
+}
+
+func (lbModelBuilder *loadBalancerBuilderImpl) translateSourcePrefixEnabled(b bool) elbv2model.EnablePrefixForIpv6SourceNat {
+	if b {
```
shraddhabang, line 80: NIT: Lets add proper variable name for this

pkg/gateway/model/model_build_loadbalancer.go
```diff
@@ -0,0 +1,120 @@
...
+		return elbv2model.EnablePrefixForIpv6SourceNatOn
+	}
+	return elbv2model.EnablePrefixForIpv6SourceNatOff
+}
+
+func (lbModelBuilder *loadBalancerBuilderImpl) buildLoadBalancerName(lbConf *elbv2gw.LoadBalancerConfiguration, gw *gwv1.Gateway, scheme elbv2model.LoadBalancerScheme) (string, error) {
+	if lbConf.Spec.LoadBalancerName != nil {
+		name := *lbConf.Spec.LoadBalancerName
+		// The name of the loadbalancer can only have up to 32 characters
```
shraddhabang, line 89: Already added as part of crd validation. https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/apis/gateway/v1beta1/loadbalancerconfig_types.go#L173

pkg/gateway/model/model_build_loadbalancer.go
zac-nixon: done

pkg/gateway/model/model_build_loadbalancer.go
zac-nixon: done

pkg/gateway/model/model_build_loadbalancer.go
zac-nixon: i was just being defensive but yeah fair enough the whole point of using the CRD was to avoid having duplicate validations

shraddhabang (approved):
/lgtm /approve

pkg/gateway/model/model_build_security_group.go
```diff
@@ -0,0 +1,305 @@
...
+	includeIPv6 := isIPv6Supported(ipAddressType)
+
+	// Port Loop
+	for port, cfg := range routes {
+		// Protocol Loop
+		for _, protocol := range generateProtocolListFromRoutes(cfg) {
+			// CIDR Loop
+			for _, cidr := range sourceRanges {
+				isIPv6 := isIPv6CIDR(cidr)
+
```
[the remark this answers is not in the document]
zac-nixon: IMO, I don't think it's more readable. Also, it looks like there is a bug as it would always enable icmpv4 if icmp was not enabled.

PR #4156, opened 2025-04-25
[feat:gw-api] add support for ServiceExternalTrafficPolicyLocal

shuqz:
### Description
add support for ServiceExternalTrafficPolicyLocal when ExternalTrafficPolicy is defined in Service Spec for NLB. To have default value for Local case, it needs to satisfy: 
1. targetType = Instance 
2. externalTrafficPolicy = Local
3. load balancer type = NLB

### Checklist
- [✅] Added tests that cover your change (if possible)
- [] Added/modified documentation as required (such as the `README.md`, or the `docs` directory)
- [✅] Manually tested (changed in kind:Service YAML file)
    - test case 1: set `externalTrafficPolicy: Cluster` -> no difference, this is the default case 
    - test case 2: set `externalTrafficPolicy: Local` and not specify healthCheckNodePort -> random number(range 30000-32767) is chosen for health check node port and other default values are taken
    - test case 3: set `externalTrafficPolicy: Local` and specify healthCheckNodePort -> healthCheckNodePort is chosen for health check node port and other default values are taken
    - test case 4: set `externalTrafficPolicy: Local` and specify healthCheckNodePort and under Type:NodePort -> get error " spec.healthCheckNodePort: Invalid value: 30001: may only be set when `type` is 'LoadBalancer' and `externalTrafficPolicy` is 'Local'"
- [✅] Made sure the title of the PR is a good description that can go into the release notes

### DIFF
--- a/pkg/gateway/model/model_build_target_group.go
+++ b/pkg/gateway/model/model_build_target_group.go
@@ -53,6 +53,14 @@ type targetGroupBuilderImpl struct {
 	defaultHealthyThresholdCount              int32
 	defaultHealthCheckTimeout                 int32
 	defaultHealthCheckInterval                int32
+
+	// Default health check settings for NLB instance mode with spec.ExternalTrafficPolicy set to Local
+	defaultHealthCheckProtocolForInstanceModeLocal           elbv2model.Protocol
+	defaultHealthCheckPathForInstanceModeLocal               string
+	defaultHealthCheckIntervalForInstanceModeLocal           int32
+	defaultHealthCheckTimeoutForInstanceModeLocal            int32
+	defaultHealthCheckHealthyThresholdForInstanceModeLocal   int32
+	defaultHealthCheckUnhealthyThresholdForInstanceModeLocal int32
 }
 
 func newTargetGroupBuilder(clusterName string, vpcId string, tagHelper tagHelper, loadBalancerType elbv2model.LoadBalancerType, disableRestrictedSGRules bool, defaultTargetType string) targetGroupBuilder {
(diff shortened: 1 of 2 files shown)


pkg/gateway/model/model_build_target_group.go
```diff
@@ -515,6 +542,12 @@ func (builder *targetGroupBuilderImpl) buildTargetGroupHealthCheckPath(targetGro
 		return &builder.defaultHealthCheckPathGRPC
 	}
 
+	if targetGroupProps == nil || targetGroupProps.HealthCheckConfig == nil || targetGroupProps.HealthCheckConfig.HealthCheckPath == nil {
+		if builder.loadBalancerType == elbv2model.LoadBalancerTypeNetwork && isServiceExternalTrafficPolicyTypeLocal {
```
zac-nixon, line 546: nit: isServiceExternalTrafficPolicyTypeLocal will never be true for application lb, so you can drop the nlb check.

pkg/gateway/model/model_build_target_group.go
```diff
@@ -544,30 +577,42 @@ func (builder *targetGroupBuilderImpl) buildTargetGroupHealthCheckMatcher(target
 	}
 }
 
-func (builder *targetGroupBuilderImpl) buildTargetGroupHealthCheckIntervalSeconds(targetGroupProps *elbv2gw.TargetGroupProps) int32 {
+func (builder *targetGroupBuilderImpl) buildTargetGroupHealthCheckIntervalSeconds(targetGroupProps *elbv2gw.TargetGroupProps, isServiceExternalTrafficPolicyTypeLocal bool) int32 {
 	if targetGroupProps == nil || targetGroupProps.HealthCheckConfig == nil || targetGroupProps.HealthCheckConfig.HealthCheckInterval == nil {
-		return builder.defaultHealthCheckInterval
+		return map[bool]int32{
```
zac-nixon, line 582: interesting pattern :)

zac-nixon (approved):
I was expecting this to as a new property in the TG config to fit the pattern of defining TG attributes in that CRD. But I think this is ok, and it's not a one-way door. We can add the TG prop at a later time.

PR #4162, opened 2025-04-30
[feat:gw-api] add support for capacity reservation

shuqz:
[feat:gw-api] add support for capacity reservation 

### Description
1. add support for capacity reservation 
2. for testing purpose, also add simple logic to process loadbalancerconfiguration, it was passed in as empty object before
3. pre-requirement for testing: submit service quota request to increase reserve LCU quota (default is 0)

### Checklist
- [x] Manually tested
        - test with following YAML file, was able to see 100LCU requested from console and modify it to 0 will reset the request. verified from both AWS console and log message. 
```
metadata:
  name: test-gw-alb-lb-config
  namespace: gateway-alb
spec:
  loadBalancerName: "customized-alb-lb-name"
  minimumLoadBalancerCapacity:
    capacityUnits: 100
---
kind: Gateway
metadata:
  name: test-gw-alb
  namespace: gateway-alb
spec:
  gatewayClassName: alb-gateway
  infrastructure:
    parametersRef:
      name: test-gw-alb-lb-config
      group: gateway.k8s.aws
  listeners:
  - protocol: HTTP
    port: 80
    name: test-server
    allowedRoutes:
      namespaces:
        from: Same
```

- [x] Backfilled missing tests for code in same general area :tada: -> added test file for model_build_load_balancer.go


pkg/gateway/model/model_build_loadbalancer.go
```diff
@@ -117,7 +117,11 @@ func (lbModelBuilder *loadBalancerBuilderImpl) buildLoadBalancerAttributes(lbCon
 	return attributes
 }
 
-// TODO -- Fill this in at a later time.
 func (lbModelBuilder *loadBalancerBuilderImpl) buildLoadBalancerMinimumCapacity(lbConf *elbv2gw.LoadBalancerConfiguration) *elbv2model.MinimumLoadBalancerCapacity {
-	return nil
+	if lbConf.Spec.MinimumLoadBalancerCapacity == nil {
```
shraddhabang, line 121: check the nil for lbconf too. There may be a case where we dont specify LB config for either gwclass or gateway.

zac-nixon: Good call out. I found this to be annoying, so I've refactored the model builders to not use a ptr for the lb conf. I suggest adding the nil check here and I'll rebase to remove the nil check whenever https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4163 is merged.

apis/gateway/v1beta1/loadbalancerconfig_types.go
```diff
@@ -279,6 +283,17 @@ type LoadBalancerConfigurationList struct {
 	Items           []LoadBalancerConfiguration `json:"items"`
 }
 
+// CapacityUnits Unit for setting the capacity on load balancer
+const (
+	CapacityUnits string = "CapacityUnits"
```
zac-nixon, line 288: What is this used for?

shuqz: not used, i forgot to remove it

pkg/gateway/model/model_build_loadbalancer.go
shuqz: i noticed that change in your code too, and was wondering why made that change, now i know lol. i will update to check nil here so you can rebase all later

PR #4163, opened 2025-05-01
[feat gw api] Add Gateway Class Reconciler + Config merging logic.

zac-nixon:
### Description

Adds a reconciler to process Gateway Class changes, including changes to LoadBalancer configurations attached to a Gateway Class.

Adds basic logic to ensure that Gateway reconciler will only progress Gateways that have an accepted Gateway Class.

Adds a configuration merger that implements merging logic to handle GatewayClass and Gateway LoadBalancer configs. [TODO: Add documentation]

Refactored to drop pointer references to the the lb config in the gateway model builders, this removes the need for nil checks in all of the model builders.

### Checklist
- [] Added/modified documentation as required (such as the `README.md`, or the `docs` directory)
- [x] Manually tested


pkg/config/controller_config.go
```diff
@@ -149,6 +153,8 @@ func (cfg *ControllerConfig) BindFlags(fs *pflag.FlagSet) {
 		"Maximum number of concurrently running reconcile loops for service")
 	fs.IntVar(&cfg.TargetGroupBindingMaxConcurrentReconciles, flagTargetGroupBindingMaxConcurrentReconciles, defaultMaxConcurrentReconciles,
 		"Maximum number of concurrently running reconcile loops for targetGroupBinding")
+	fs.IntVar(&cfg.GatewayClassMaxConcurrentReconciles, flagGatewayClassMaxConcurrentReconciles, 1,
```
shraddhabang, line 156: should we use `defaultMaxConcurrentReconciles` instead of 1?

pkg/gateway/config_merger.go
```diff
@@ -0,0 +1,224 @@
...
+		mergedAttributes := make([]elbv2gw.LoadBalancerAttribute, 0, len(baseAttributesMap))
+
+		for k, v := range baseAttributesMap {
+			mergedAttributes = append(mergedAttributes, elbv2gw.LoadBalancerAttribute{
+				Key:   k,
+				Value: v,
+			})
+		}
+
+		sort.Slice(mergedAttributes, func(i, j int) bool {
```
shraddhabang, line 81: Any reason for sorting?

pkg/gateway/config_merger.go
```diff
@@ -0,0 +1,224 @@
...
+		merged.VpcId = lowPriority.Spec.VpcId
+	}
+
+	if highPriority.Spec.EnableICMP != nil {
+		merged.EnableICMP = highPriority.Spec.EnableICMP
+	} else {
+		merged.EnableICMP = lowPriority.Spec.EnableICMP
+	}
+
+	if highPriority.Spec.ManageBackendSecurityGroupRules != nil {
```
shraddhabang, line 219: I wish Go had ternary testing operation :(

controllers/gateway/gateway_controller.go
```diff
@@ -454,3 +463,63 @@ func (r *gatewayReconciler) setupNLBGatewayControllerWatches(ctrl controller.Con
...
+	}
+
+	if gatewayClassLBConfig != nil {
+		storedVersion := getStoredProcessedConfig(gwClass)
+		if storedVersion == nil || *storedVersion != gatewayClassLBConfig.ResourceVersion {
+			var safeVersion string
+			if storedVersion != nil {
+				safeVersion = *storedVersion
+			}
+			return elbv2gw.LoadBalancerConfiguration{}, errors.Errorf("GatewayClass hasn't processed latest loadbalancer config. Processed version %s, Latest version %s", safeVersion, gatewayClassLBConfig.ResourceVersion)
```
shraddhabang, line 489: Same as above. Lets add gwclass names.

pkg/config/controller_config.go
zac-nixon: That is 3, which is probably overkill. But I made the change.

pkg/gateway/config_merger.go
zac-nixon: It makes the output predictable which means the we can just use:
```
assert.Equal(t, tc.expected, result)
```
in our unit tests. Seeing as the attributes list is going to be relatively small it's a small performance hit.

config/crd/gateway/gateway-crds.yaml
```diff
@@ -230,6 +230,14 @@ spec:
                   specifies whether you want the controller to configure security group rules on Node/Pod for traffic access
                   when you specify securityGroups
                 type: boolean
+              mergingMode:
+                description: |-
+                  mergingMode defines the merge behavior when both the Gateway and GatewayClass have a defined LoadBalancerConfiguration.
+                  This field is only honored for the configuration attached to the GatewayClass.
+                enum:
+                - prefer-gateway
```
shuqz, line 238: i am think should we be more specific in naming, cuz the name `mergingMode` does not imply it is for lbConfig and enum does not imply it too. Users can read doc for sure but maybe we can try to make name more explicit

controllers/gateway/eventhandlers/gatewayclass/load_balancer_configuration_events.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (h *enqueueRequestsForLoadBalancerConfigurationEvent) Update(ctx context.Context, e event.TypedUpdateEvent[*elbv2gw.LoadBalancerConfiguration], queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	lbconfigNew := e.ObjectNew
+	h.logger.V(1).Info("enqueue loadbalancerconfiguration update event", "loadbalancerconfiguration", lbconfigNew.Name)
+	h.enqueueImpactedGatewayClass(ctx, lbconfigNew, queue)
+}
+
+func (h *enqueueRequestsForLoadBalancerConfigurationEvent) Delete(ctx context.Context, e event.TypedDeleteEvent[*elbv2gw.LoadBalancerConfiguration], queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	lbconfig := e.Object
+	h.logger.V(1).Info("enqueue loadbalancerconfiguration delete event", "loadbalancerconfiguration", lbconfig.Name)
```
shuqz, line 54: do we also wanna logging namespace? (if name can be same for different namespace)

zac-nixon: yeah good idea

pkg/gateway/config_merger.go
```diff
@@ -0,0 +1,224 @@
...
+	if lowPriority.Spec.Tags != nil {
+		baseTags = algorithm.MergeStringMap(baseTags, *lowPriority.Spec.Tags)
+	}
+
+	if len(baseTags) > 0 {
+		merged.Tags = &baseTags
+	}
+}
+
+func (merger *configMergerImpl) performTakeOneMerges(merged *elbv2gw.LoadBalancerConfigurationSpec, highPriority elbv2gw.LoadBalancerConfiguration, lowPriority elbv2gw.LoadBalancerConfiguration) {
```
shuqz, line 140: it looks like in this function, it is assume if a property is not specified in highPriority spec, it will be specified in low priority spec, but what if `lowPriority.Spec.LoadBalancerName` is also nil (unless it is never gonna be the case), and if we originally have a default value for LoadBalancerName, then after this function, the default value will be overriden and it is gonna be nil

controllers/gateway/gateway_controller.go
```diff
@@ -454,3 +463,63 @@ func (r *gatewayReconciler) setupNLBGatewayControllerWatches(ctrl controller.Con
 	return nil
 
 }
+
+func (r *gatewayReconciler) getLoadBalancerConfigForGateway(ctx context.Context, k8sClient client.Client, gw *gwv1.Gateway, gwClass *gwv1.GatewayClass) (elbv2gw.LoadBalancerConfiguration, error) {
+
+	// If the Gateway Class isn't accepted, we shouldn't try to reconcile this Gateway.
```
shraddhabang, line 469: Can we move this check outside of this function in the main reconciler so that this function sticks to just getting the lb config.

controllers/gateway/gateway_controller.go
```diff
@@ -454,3 +463,63 @@ func (r *gatewayReconciler) setupNLBGatewayControllerWatches(ctrl controller.Con
...
+	}
+
+	if gatewayClassLBConfig != nil {
+		storedVersion := getStoredProcessedConfig(gwClass)
+		if storedVersion == nil || *storedVersion != gatewayClassLBConfig.ResourceVersion {
+			var safeVersion string
+			if storedVersion != nil {
+				safeVersion = *storedVersion
+			}
+			return elbv2gw.LoadBalancerConfiguration{}, errors.Errorf("GatewayClass [%s] hasn't processed latest loadbalancer config. Processed version %s, Latest version %s", gwClass.Name, safeVersion, gatewayClassLBConfig.ResourceVersion)
```
shraddhabang, line 489: Instead of throwing error, do we want to retry/wait for gwclass to complete its processing its cfg?

controllers/gateway/eventhandlers/load_balancer_configuration_events.go
```diff
@@ -62,7 +65,7 @@ func (h *enqueueRequestsForLoadBalancerConfigurationEvent) Generic(ctx context.C
 }
 
 func (h *enqueueRequestsForLoadBalancerConfigurationEvent) enqueueImpactedService(ctx context.Context, lbconfig *elbv2gw.LoadBalancerConfiguration, queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
-	gwClasses := GetImpactedGatewayClassesFromLbConfig(ctx, h.k8sClient, lbconfig, h.gwController)
+	gwClasses := GetImpactedGatewayClassesFromLbConfig(ctx, h.k8sClient, lbconfig, h.gwControllerSet)
```
shraddhabang, line 68: Since we have separate controller for gateway class? Do we even need to enqueue gwclass for LB config changes in gateway controller?

controllers/gateway/eventhandlers/gatewayclass/load_balancer_configuration_events.go
```diff
@@ -0,0 +1,74 @@
...
+type enqueueRequestsForLoadBalancerConfigurationEvent struct {
+	gwClassEventChan chan<- event.TypedGenericEvent[*gatewayv1.GatewayClass]
+	k8sClient        client.Client
+	eventRecorder    record.EventRecorder
+	gwControllers    sets.Set[string]
+	logger           logr.Logger
+}
+
+func (h *enqueueRequestsForLoadBalancerConfigurationEvent) Create(ctx context.Context, e event.TypedCreateEvent[*elbv2gw.LoadBalancerConfiguration], queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	// We don't need to respond to create events.
```
shraddhabang, line 43: What if the gwclass is created first referencing a non-existing cfg, user realises his mistake as they see not found status on the gwclass and decides to create the lbcfg now? We need to listen to this event and enqueue the gwclass referencing it to update its status.

controllers/gateway/gatway_class_controller.go
```diff
@@ -0,0 +1,140 @@
...
+	updateLastProcessedConfigFn func(ctx context.Context, k8sClient client.Client, gwClass *gwv1.GatewayClass, lbConf *elbv2gw.LoadBalancerConfiguration) error
+	configResolverFn            func(ctx context.Context, k8sClient client.Client, reference *gwv1.ParametersReference) (*elbv2gw.LoadBalancerConfiguration, error)
+}
+
+func (r *gatewayClassReconciler) SetupWatches(_ context.Context, ctrl controller.Controller, mgr ctrl.Manager) error {
+
+	gwClassEventChan := make(chan event.TypedGenericEvent[*gwv1.GatewayClass])
+	lbEventHandler := gatewayclasseventhandlers.NewEnqueueRequestsForLoadBalancerConfigurationEvent(gwClassEventChan, r.k8sClient, r.eventRecorder, r.enabledControllers, r.logger)
+
+	if err := ctrl.Watch(source.Kind(mgr.GetCache(), &gwv1.GatewayClass{}, &handler.TypedEnqueueRequestForObject[*gwv1.GatewayClass]{})); err != nil {
```
shraddhabang, line 63: Do we not need to enqueue the gwclass to update the status of gwclass on create/update/delete events?

pkg/gateway/config_merger.go
zac-nixon: I don't follow -- The merged spec starts off w/ a blank spec, so it's all zero values. We will always use the high priority value (if it exists) otherwise we use the low priority value, which could be a real value or a zero value. If it's a zero value, the merged spec still has a zero value which is correct.

I don't think it's possible we ever start with a default value for the load balancer name.

controllers/gateway/gateway_controller.go
zac-nixon: sure

controllers/gateway/gateway_controller.go
zac-nixon: Throwing an error is the same thing as wait + retry.

controllers/gateway/eventhandlers/load_balancer_configuration_events.go
zac-nixon: Yeah.

They serve two different purposes.

In the GatewayClass reconciler, it's meant to validate the configuration is valid and move the Gateway Class status to Accepted.

In the Gateway reconciler, it's meant to materialize the ELB resources according to the Gateway config.

controllers/gateway/eventhandlers/gatewayclass/load_balancer_configuration_events.go
zac-nixon: This should already been handled, because we have the event queued for the gwclass creation. Since it failed to find the config, it will keep reconciling. Eventually when the config is created, everything should just work.

controllers/gateway/gatway_class_controller.go
zac-nixon: I don't follow. Those CRUD actions are the only actions we care about. Can you elaborate on the concern?

pkg/gateway/config_merger.go
shuqz: i see, thanks for the explanation

controllers/gateway/gateway_controller.go
shraddhabang: Ok got it.

controllers/gateway/gatway_class_controller.go
shraddhabang: Never mind.

PR #4169, opened 2025-05-02
Bump version to 2.13

zac-nixon:
Bumps version to 2.13.0. Also fixes some fresh install errors found for the Gateway API.


Makefile
```diff
@@ -2,7 +2,7 @@
 MAKEFILE_PATH = $(dir $(realpath -s $(firstword $(MAKEFILE_LIST))))
 
 # Image URL to use all building/pushing image targets
-IMG ?= public.ecr.aws/eks/aws-load-balancer-controller:v2.12.0
```
shraddhabang, line 5: Can you also update the [docs-publish](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/Makefile#L175) target to point the latest docs to 2.13.

config/rbac/role.yaml
```diff
@@ -179,6 +179,7 @@ rules:
   verbs:
   - get
   - list
```
shraddhabang, line 181: We also need to update the helm roles for helm installation right?
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/helm/aws-load-balancer-controller/templates/rbac.yaml

zac-nixon: thanks fixed.

shraddhabang (approved):
/lgtm
/approved

PR #4174, opened 2025-05-08
[feat gw-api] add hostname handling logic

shuqz:
### Issue

Add hostname handling logic for Route

### Description
This PR added logic to handle hostname and covers 
1. hostname format validation
2. handle scenario when both listener hostname and route hostname are defined -> checks hostname compatibility 
3. added precedence rule, priority oder as follow: exact > wildcard, longer character > shorter character

it does not cover 
- handle of status update for Route

### Checklist
- [x] Manually tested 

### DIFF
--- a/pkg/gateway/model/model_build_listener.go
+++ b/pkg/gateway/model/model_build_listener.go
@@ -165,6 +165,10 @@ func (l listenerBuilderImpl) buildL4ListenerSpec(ctx context.Context, stack core
 }
 
 func (l listenerBuilderImpl) buildListenerRules(stack core.Stack, ls *elbv2model.Listener, lb *elbv2model.LoadBalancer, securityGroups securityGroupOutput, gw *gwv1.Gateway, port int32, lbCfg elbv2gw.LoadBalancerConfiguration, routes map[int32][]routeutils.RouteDescriptor) error {
+
+	// add hostname handling (sort by precedence order)
+	sortRoutesByHostnamePrecedence(routes[port])
+
 	// TODO for L7 Gateway Implementation
 	// This is throw away code
 	// This is temporary implementation for supporting basic multiple HTTPRoute for simple backend refs. We will create default forward action for all the backend refs for all HTTPRoutes for this listener
@@ -185,6 +189,21 @@ func (l listenerBuilderImpl) buildListenerRules(stack core.Stack, ls *elbv2model
 						},
 					},
 					}
+
+					// add host header condition
+					if hostnames := descriptor.GetHostnames(); len(hostnames) > 0 {
+						hostnamesStringList := make([]string, len(descriptor.GetHostnames()))
+						for i, j := range descriptor.GetHostnames() {
+							hostnamesStringList[i] = string(j)
+						}
+						conditions = append(conditions, elbv2model.RuleCondition{
+							Field: elbv2model.RuleConditionFieldHostHeader,
+							HostHeaderConfig: &elbv2model.HostHeaderConditionConfig{
+								Values: hostnamesStringList,
+							},
+						})
+					}
+
 					actions := buildL4ListenerDefaultActions(targetGroup)
 					tags, tagsErr := l.tagHelper.getGatewayTags(lbCfg)
 					if tagsErr != nil {
--- a/pkg/gateway/model/utilities.go
+++ b/pkg/gateway/model/utilities.go
@@ -1,7 +1,9 @@
 package model
 
 import (
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/gateway/routeutils"
 	elbv2model "sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2"
+	"sort"
 	"strings"
 )
 
@@ -17,4 +19,26 @@ func isIPv6Supported(ipAddressType elbv2model.IPAddressType) bool {
 // TODO - Refactor?
 func isIPv6CIDR(cidr string) bool {
 	return strings.Contains(cidr, ":")
+}
+
+func sortRoutesByHostnamePrecedence(routes []routeutils.RouteDescriptor) {
+	sort.SliceStable(routes, func(i, j int) bool {
+		hostnameOne := routes[i].GetHostnames()
+		hostnameTwo := routes[j].GetHostnames()
+
+		if len(hostnameOne) == 0 && len(hostnameTwo) == 0 {
+			return false
+		}
+		if len(hostnameOne) == 0 {
+			return false
+		}
+		if len(hostnameTwo) == 0 {
+			return true
+		}
+		precedence := routeutils.GetHostnamePrecedenceOrder(string(hostnameOne[0]), string(hostnameTwo[0]))
+		if precedence != 0 {
+			return precedence < 0 // -1 means higher precedence
+		}
+		return false
+	})
 }
--- a/pkg/gateway/routeutils/listener_attachment_helper.go
+++ b/pkg/gateway/routeutils/listener_attachment_helper.go
@@ -32,6 +32,7 @@ func newListenerAttachmentHelper(k8sClient client.Client, logger logr.Logger) li
 // listenerAllowsAttachment utility method to determine if a listener will allow a route to connect using
 // Gateway API rules to determine compatibility between lister and route.
 func (attachmentHelper *listenerAttachmentHelperImpl) listenerAllowsAttachment(ctx context.Context, gw gwv1.Gateway, listener gwv1.Listener, route preLoadRouteDescriptor) (bool, error) {
+	// check namespace
 	namespaceOK, err := attachmentHelper.namespaceCheck(ctx, gw, listener, route)
 	if err != nil {
 		return false, err
@@ -39,7 +40,25 @@ func (attachmentHelper *listenerAttachmentHelperImpl) listenerAllowsAttachment(c
 	if !namespaceOK {
 		return false, nil
 	}
-	return attachmentHelper.kindCheck(listener, route), nil
+
+	// check kind
+	kindOK := attachmentHelper.kindCheck(listener, route)
+	if !kindOK {
+		return false, nil
+	}
+
+	// check hostname
+	if (route.GetRouteKind() == HTTPRouteKind || route.GetRouteKind() == GRPCRouteKind || route.GetRouteKind() == TLSRouteKind) && route.GetHostnames() != nil {
+		hostnameOK, err := attachmentHelper.hostnameCheck(listener, route)
+		if err != nil {
+			return false, err
+		}
+		if !hostnameOK {
+			return false, nil
+		}
+	}
+
+	return true, nil
 }
 
 // namespaceCheck namespace check implements the Gateway API spec for namespace matching between listener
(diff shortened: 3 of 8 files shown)


pkg/gateway/routeutils/utils.go
```diff
@@ -87,3 +90,83 @@ func isServiceReferredByRoute(route preLoadRouteDescriptor, svcID types.Namespac
...
+// 1. no IP allowed
+// 2. wildcard is only allowed as leftmost character
+// Allowed Characters: Hostname labels must only contain lowercase ASCII letters (a-z), digits (0-9), and hyphens (-).
+// Starting with a Digit: RFC 1123 allows labels to begin with a digit, which is a departure from the previous RFC 952 restriction.
+// Length: Each label in a hostname can be between 1 and 63 characters long.
+// Overall Hostname Length: The entire hostname, including the periods separating labels, cannot exceed 253 characters.
+// Case: Hostnames are case-insensitive.
+// Underscore: Underscores are not permitted in hostnames.
+// Other Symbols: No other symbols, punctuation, or whitespace is allowed in hostnames
+func IsHostNameInValidFormat(hostName string) (bool, error) {
```
shraddhabang, line 104: We dont need this since the gateways and routes CRDS both perform hostname validation. Reference : https://gateway-api.sigs.k8s.io/reference/spec/#hostname

pkg/gateway/model/model_build_listener.go
```diff
@@ -185,6 +189,30 @@ func (l listenerBuilderImpl) buildListenerRules(stack core.Stack, ls *elbv2model
 						},
 					},
 					}
+
+					// get valid hostnames
+					var validHostnames []string
+					if hostnames := descriptor.GetHostnames(); len(hostnames) > 0 {
+						for _, hostname := range hostnames {
+							isListenerHostnameValid, err := routeutils.IsHostNameInValidFormat(string(hostname))
```
shraddhabang, line 197: We have already performed validation for this while checking the compatibility with listeners. We can assume these are valid.

pkg/gateway/routeutils/listener_attachment_helper.go
```diff
@@ -32,14 +32,33 @@ func newListenerAttachmentHelper(k8sClient client.Client, logger logr.Logger) li
...
-	return attachmentHelper.kindCheck(listener, route), nil
+
+	// check kind
+	kindOK := attachmentHelper.kindCheck(listener, route)
+	if !kindOK {
+		return false, nil
+	}
+
+	// check hostname
+	if (route.GetRouteKind() == HTTPRouteKind || route.GetRouteKind() == GRPCRouteKind) && route.GetHostnames() != nil {
```
zac-nixon, line 51: TLS route too?

pkg/gateway/routeutils/utils.go
```diff
@@ -87,3 +90,83 @@ func isServiceReferredByRoute(route preLoadRouteDescriptor, svcID types.Namespac
...
+// isHostnameCompatible checks if given two hostnames are compatible with each other
+// this function is used to check if listener hostname and Route hostname match
+func isHostnameCompatible(hostnameOne, hostnameTwo string) bool {
+	// exact match
+	if hostnameOne == hostnameTwo {
+		return true
+	}
+
+	// suffix match - hostnameOne is a wildcard
+	if strings.HasPrefix(hostnameOne, "*.") {
```
zac-nixon, line 137: nit: 
```
if strings.HasPrefix(hostnameOne, "*.") && strings.HasSuffix(hostnameTwo, hostnameOne[1:]){
return true
}
```

to be more concise.

pkg/gateway/routeutils/utils.go
```diff
@@ -87,3 +90,83 @@ func isServiceReferredByRoute(route preLoadRouteDescriptor, svcID types.Namespac
...
+	// suffix match - hostnameTwo is a wildcard
+	if strings.HasPrefix(hostnameTwo, "*.") {
+		if strings.HasSuffix(hostnameOne, hostnameTwo[1:]) {
+			return true
+		}
+	}
+	return false
+}
+
+// GetHostnamePrecedenceOrder Hostname precedence ordering rule:
```
zac-nixon, line 151: thanks for documenting the function

pkg/gateway/routeutils/utils.go
zac-nixon: That regex will allow IP addresses through, I think we need this additional validation.

pkg/gateway/routeutils/utils.go
```diff
@@ -87,3 +90,83 @@ func isServiceReferredByRoute(route preLoadRouteDescriptor, svcID types.Namespac
...
+
+	labels := strings.Split(hostName, ".")
+	if strings.HasPrefix(hostName, "*.") {
+		labels = labels[1:]
+	}
+	for _, label := range labels {
+		if len(label) < 1 || len(label) > 63 {
+			return false, fmt.Errorf("invalid hostname label length, length must between 1 and 63")
+		}
+		if !regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9-]*[a-zA-Z0-9]$`).MatchString(label) {
```
zac-nixon, line 121: It's expensive to compile this regex. Can you do it once at the top?

pkg/gateway/routeutils/utils_test.go
```diff
@@ -434,3 +435,308 @@ func Test_IsServiceReferredByRoute(t *testing.T) {
...
+			wantErr:  "",
+		},
+		{
+			name:     "Valid wildcard domain",
+			hostname: "*.example.com",
+			want:     true,
+			wantErr:  "",
+		},
+		{
+			name:     "Invalid wildcard domain",
```
shraddhabang, line 466: This is blocked by CRD validations. No need to check for this.

pkg/gateway/routeutils/utils_test.go
```diff
@@ -434,3 +435,308 @@ func Test_IsServiceReferredByRoute(t *testing.T) {
...
+			wantErr:  "invalid hostname length",
+		},
+		{
+			name:     "Label too long",
+			hostname: strings.Repeat("a", 64) + ".com",
+			want:     false,
+			wantErr:  "invalid hostname label length",
+		},
+		{
+			name:     "Invalid characters",
```
shraddhabang, line 496: Same : `The Gateway "test-alb-1" is invalid: spec.listeners[0].hostname: Invalid value: "test!.com": spec.listeners[0].hostname in body should match '^(\*\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'`

pkg/gateway/routeutils/listener_attachment_helper.go
```diff
@@ -104,3 +123,35 @@ func (attachmentHelper *listenerAttachmentHelperImpl) kindCheck(listener gwv1.Li
...
+		return true, nil
+	}
+
+	// validate listener hostname, return if listener hostname is not valid
+	isListenerHostnameValid, err := IsHostNameInValidFormat(string(*listener.Hostname))
+	if err != nil {
+		attachmentHelper.logger.Error(err, "listener hostname is not valid", "listener", listener.Name, "hostname", *listener.Hostname)
+		return false, err
+	}
+	if !isListenerHostnameValid {
```
shraddhabang, line 139: No need to check for this. We should always throw error if the host name is invalid.

pkg/gateway/routeutils/listener_attachment_helper.go
```diff
@@ -104,3 +123,35 @@ func (attachmentHelper *listenerAttachmentHelperImpl) kindCheck(listener gwv1.Li
...
+		return false, nil
+	}
+
+	for _, hostname := range route.GetHostnames() {
+		// validate route hostname, skip invalid hostname
+		isHostnameValid, err := IsHostNameInValidFormat(string(hostname))
+		if err != nil || !isHostnameValid {
```
shraddhabang, line 146: Same as above.

pkg/gateway/model/model_build_listener.go
shuqz: that's true, if it is not valid, it won't be added to the list, i will remove this part

pkg/gateway/routeutils/listener_attachment_helper.go
shuqz: yes, will add it

pkg/gateway/routeutils/utils.go
zac-nixon: Yeah. I see the confusion. It's weird that some validation is handled by the CRD while others are not.

shuqz: i thought about this check in CRD, 1. it is missing IP check 2. it does not check if each label is within 1-63 length. i added other checks just to double check, but since it is expensive to do it, in this hostname validation, we can verify IP and label length, other checks will be done by CRD synthax check.

pkg/gateway/routeutils/utils.go
shuqz: please see my reply above

pkg/gateway/routeutils/utils_test.go
shuqz: please see my reply above, same answer applies to all hostname validation related comments

pkg/gateway/routeutils/utils_test.go
shuqz: same as above

pkg/gateway/routeutils/utils_test.go
```diff
@@ -434,3 +435,308 @@ func Test_IsServiceReferredByRoute(t *testing.T) {
...
+			wantErr:  "invalid hostname character",
+		},
+		{
+			name:     "Label starting with number",
+			hostname: "1test.com",
+			want:     false,
+			wantErr:  "invalid hostname character",
+		},
+		{
+			name:     "Label ending with hyphen",
```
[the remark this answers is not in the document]
shuqz: same as above

pkg/gateway/routeutils/utils.go
zac-nixon: Sorry, I don't follow. You can specify it at the top of the file:
```
var (
	validateHostNameRegex = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9-]*[a-zA-Z0-9]$`)
)
```

and then re-use the regex:
```
if !validateHostNameRegexMatchString(label) {
...
```

pkg/gateway/routeutils/listener_attachment_helper.go
shuqz: see comment below

pkg/gateway/routeutils/listener_attachment_helper.go
shuqz: see comment below

pkg/gateway/routeutils/utils.go
shuqz: we do not really need to do regex check here, CRD takes care of it. so this function will only check if it is IP and label length. i will remove all other logics and let CRD handles it.

zac-nixon (approved):

pkg/gateway/routeutils/listener_attachment_helper.go
shraddhabang: Sorry I mean `IsHostNameInValidFormat` is always returning `false, err` in case of invalid hostname. So there wont be any case where its throwing `false, nil` and it should not. So only err check is sufficient to return early, we dont need this block
`if !isListenerHostnameValid {
		return false, nil
}`

pkg/gateway/model/utilities.go
```diff
@@ -18,3 +20,25 @@ func isIPv6Supported(ipAddressType elbv2model.IPAddressType) bool {
...
+		if len(hostnameOne) == 0 && len(hostnameTwo) == 0 {
+			return false
+		}
+		if len(hostnameOne) == 0 {
+			return false
+		}
+		if len(hostnameTwo) == 0 {
+			return true
+		}
+		precedence := routeutils.GetHostnamePrecedenceOrder(string(hostnameOne[0]), string(hostnameTwo[0]))
```
shraddhabang, line 38: Just for my understanding, Why are we only considering first hostname in sorting decision?

shuqz: synced offline

PR #4190, opened 2025-05-20
[feat gw-api] add route status update

shuqz:
### Description
- add a route reconciler that handles route status update
- route will be enqueued within gateway_controller.go when 1. route is accepted 2. route is not accepted due to hostname not compatible or route does not allow attachment to listener etc during mapping 

### Checklist
- [x] Manually tested - see output in task

- [x] Refactored something and made the world a better place :star2: - typo in gateway_class_controller.go file, it was gatway_class_controller.go, so renamed this file


pkg/gateway/routeutils/route_reconciler.go
```diff
@@ -0,0 +1,420 @@
...
+	}
+}
+
+func (d *routeReconcilerImpl) Enqueue(routeStatusInfo RouteStatusInfo, routeMetadataDescriptor preLoadRouteDescriptor) {
+	routeData := RouteData{
+		routeStatusInfo:         routeStatusInfo,
+		routeMetadataDescriptor: routeMetadataDescriptor,
+	}
+	d.enqueue(routeData)
+	d.logger.Info("enqueued new Route for status update: ", "route", routeMetadataDescriptor.GetRouteNamespacedName().Name)
```
shraddhabang, line 63: NIT: Can you use verbose logger here instead of standard? We may not want to flood the prod logs for each status event for routes?

controllers/gateway/gateway_controller.go
```diff
@@ -60,6 +68,12 @@ func newGatewayReconciler(controllerName string, lbType elbv2model.LoadBalancerT
 
 	cfgResolver := newGatewayConfigResolver()
 
+	// create deferred route reconciler
+	delayingQueue := workqueue.NewDelayingQueueWithConfig(workqueue.DelayingQueueConfig{
```
zac-nixon, line 72: This will create one reconciler per LB type. I think it would be better to create one reconciler in main.go and use it for both ALB and NLB.

pkg/gateway/routeutils/route_listener_mapper.go
```diff
@@ -40,6 +40,17 @@ func (ltr *listenerToRouteMapperImpl) mapGatewayAndRoutes(ctx context.Context, g
 		ltr.logger.V(1).Info("Route is eligible for attachment", "route", route.GetRouteNamespacedName(), "allowed attachment", allowsAttachment)
 		if allowsAttachment {
 			routesForGateway = append(routesForGateway, route)
+		} else {
+			// for route not allowed to attach by gateway, need to update route status
+			deferredRouteReconciler.Enqueue(
```
zac-nixon, line 45: Is there a part of the Gateway API that mandates this? I think it might be noisy to have this status update.

pkg/gateway/routeutils/loader.go
```diff
@@ -87,11 +87,26 @@ func (l *loaderImpl) LoadRoutesForGateway(ctx context.Context, gw gwv1.Gateway,
...
 	if err != nil {
 		return nil, err
 	}
 
+	// update status for accepted routes
+	for _, routeList := range mappedRoutes {
+		for _, route := range routeList {
+			deferredRouteReconciler.Enqueue(
+				RouteStatusInfo{
+					Accepted:     true,
```
zac-nixon, line 100: It would be good to mention what Gateway is accepting the route.

pkg/gateway/routeutils/route_reconciler.go
```diff
@@ -0,0 +1,420 @@
...
+
+func (d *routeReconcilerImpl) handleRouteStatusUpdate(routeData RouteData) error {
+	d.logger.Info("Handling route status update", "route", routeData.routeMetadataDescriptor.GetRouteNamespacedName().Name)
+	routeName := routeData.routeMetadataDescriptor.GetRouteNamespacedName().Name
+	routeNamespace := routeData.routeMetadataDescriptor.GetRouteNamespacedName().Namespace
+	routeKind := routeData.routeMetadataDescriptor.GetRouteKind()
+	// double check if route exist and get route
+	var route client.Object
+	switch routeKind {
+	case "HTTPRoute":
```
shraddhabang, line 111: Need to include L4 routes.

pkg/gateway/routeutils/route_reconciler.go
```diff
@@ -0,0 +1,420 @@
...
+func (d *routeReconcilerImpl) updateRouteStatus(route client.Object, routeStatusInfo RouteStatusInfo) error {
+	// initial route status if not exist already
+	var ParentRefs []gwv1.ParentReference
+	switch r := route.(type) {
+	case *gwv1.HTTPRoute:
+		if r.Status.Parents == nil {
+			r.Status.Parents = []gwv1.RouteParentStatus{}
+		}
+		ParentRefs = r.Spec.ParentRefs
+	case *gwv1.GRPCRoute:
```
shraddhabang, line 147: Same L4 routes

pkg/gateway/routeutils/route_reconciler.go
```diff
@@ -0,0 +1,420 @@
...
+	default:
+		return fmt.Errorf("%s: %s", unKnownRouteKind, r)
+	}
+
+	// set conditions
+	var newRouteStatus []gwv1.RouteParentStatus
+	for _, parentRef := range ParentRefs {
+		newRouteParentStatus := gwv1.RouteParentStatus{
+			ParentRef: parentRef,
+			// TODO: add support for NLB
```
shraddhabang, line 161: NVM I see this is TODO.

pkg/gateway/routeutils/route_listener_mapper.go
zac-nixon: Checking this:
```
	// Note that parent references that cannot be resolved by an implementation
	// of this API will not be added to this list. Implementations of this API
	// can only populate Route status for the Gateways/parent resources they are
	// responsible for.
```

This status update should not be here.

pkg/gateway/routeutils/route_listener_mapper.go
```diff
@@ -51,10 +62,20 @@ func (ltr *listenerToRouteMapperImpl) mapGatewayAndRoutes(ctx context.Context, g
 			// for connection viability.
 			if !ltr.routeAttachmentHelper.routeAllowsAttachmentToListener(listener, route) {
 				ltr.logger.V(1).Info("Route doesnt allow attachment")
+				// for route not allowed to attach, need to update route status
+				deferredRouteReconciler.Enqueue(
```
zac-nixon, line 66: same comment as above, we should not update status here.

pkg/gateway/routeutils/route_reconciler.go
```diff
@@ -0,0 +1,420 @@
+package routeutils
```
zac-nixon, line 1: This doesn't "feel" like a good place for this reconciler. I would prefer to put it in `controllers/gateway`.

pkg/gateway/routeutils/route_reconciler.go
shuqz: yea, i can update it

pkg/gateway/routeutils/route_reconciler.go
```diff
@@ -0,0 +1,420 @@
...
+	switch routeKind {
+	case "HTTPRoute":
+		route = &gwv1.HTTPRoute{}
+	case "GRPCRoute":
+		route = &gwv1.GRPCRoute{}
+	default:
+		return fmt.Errorf("%s: %s", unKnownRouteKind, routeKind)
+	}
+
+	if err := d.k8sClient.Get(context.Background(), types.NamespacedName{Namespace: routeNamespace, Name: routeName}, route); err != nil {
```
zac-nixon, line 119: IMO, we should skip this GET, and instead include the route in the queued item.

controllers/gateway/gateway_controller.go
```diff
@@ -326,6 +342,7 @@ func (r *gatewayReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manag
 	if err != nil {
 		return nil, err
 	}
+	go r.routeReconciler.Run()
```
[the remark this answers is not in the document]
shuqz: i will move it and the reconciler to main.go

pkg/gateway/routeutils/route_reconciler.go
```diff
@@ -0,0 +1,420 @@
...
+		r.Status.Parents = newRouteStatus
+	case *gwv1.GRPCRoute:
+		r.Status.Parents = newRouteStatus
+	default:
+		return fmt.Errorf("%s: %s", unKnownRouteKind, r)
+	}
+	return nil
+}
+
+func (d *routeReconcilerImpl) resolveRefGateway(parentRef gwv1.ParentReference, namespace string) (*gwv1.Gateway, error) {
```
zac-nixon, line 188: We already know this at enqueue time, can you please add it to the queue item to prevent re-querying.

pkg/gateway/routeutils/route_reconciler.go
```diff
@@ -0,0 +1,420 @@
...
+	}
+}
+
+func (d *routeReconcilerImpl) setConditionsBasedOnResolveRefGateway(route client.Object, parentStatus *gwv1.RouteParentStatus, resolveErr error) {
+	timeNow := metav1.NewTime(time.Now())
+	parentStatus.Conditions = []metav1.Condition{
+		{
+			Type:               string(gwv1.RouteConditionAccepted),
+			Status:             metav1.ConditionFalse,
+			Reason:             "ParentRefDoesNotExist",
```
zac-nixon, line 276: nit: please use a const

pkg/gateway/routeutils/route_listener_mapper.go
shuqz: ah, that's what it means. yea i can move this status update.

pkg/gateway/routeutils/loader.go
shuqz: i put it under message, so customer will see it , do you want to add it any where else? 
`Message:      "Route is accepted by Gateway",`

pkg/gateway/routeutils/route_reconciler.go
shuqz: yap it is under TODO, currently we only have HTTPRoute and GPRCRoute under "sigs.k8s.io/gateway-api/apis/v1"

pkg/gateway/routeutils/route_listener_mapper.go
shuqz: will remove it

pkg/gateway/routeutils/route_reconciler.go
shuqz: yea i originally put it under controllers/gateway, but then saw a lot issues with cyclic importing because of the cross directory refer, i will see if i can move it once i put reconciler to main.go

zac-nixon: Some more comments:
I wonder if this can be implemented in such a way where we can de-dupe items. Right now, a route might have multiple statuses to process. E.g

t0 - publish route failure
t1 - publish route failure
t2 - publish route failure
t3 - the reconciler publishes t0 event
t4 - the reconciler publishes t1 event
t5 - the reconciler publishes t2 event

It would be better if we could just skip to the publishing the t2 event.

pkg/gateway/routeutils/route_reconciler.go
shuqz: i think the main goal for this get is more about re-checking if the route still exists after enqueue and before processing, let me know what you think

pkg/gateway/routeutils/route_reconciler.go
```diff
@@ -0,0 +1,420 @@
+package routeutils
+
```
zac-nixon, line 2: Sorry.. even more comments. I don't see where we handle this:

route has 2 parent refs - gw1 and gw2.

t0 - reconcile gw1 and route
t1 - reconcile gw2 and route
t2 - update route status w/ gw1
t3 - update route status w/ gw2


Now, I remove one of those parent refs, for gw1.
t4 - remove gw1 from route parent refs

The problem is that we don't have any workflow that will clean up the route status for gw1. This might be a bigger change, so let's chat and figure out if we want to tackle the problem at a different time.

pkg/gateway/routeutils/route_reconciler.go
```diff
@@ -0,0 +1,420 @@
...
+		d.handleItemError(routeData, err, "Failed to update route status")
+		return
+	}
+}
+
+func (d *routeReconcilerImpl) handleItemError(routeData RouteData, err error, msg string) {
+	err = client.IgnoreNotFound(err)
+	if err != nil {
+		d.logger.Error(err, msg)
+		if !strings.Contains(err.Error(), unKnownRouteKind) {
```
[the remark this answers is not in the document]
shuqz: we do not want to re-queue it back if this is a route kind that's not supported

pkg/gateway/routeutils/route_reconciler.go
shuqz: sure

pkg/gateway/routeutils/route_reconciler.go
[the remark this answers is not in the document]
zac-nixon: How would that happen? Programmer error?

pkg/gateway/routeutils/loader.go
zac-nixon: This should go under the parent status that has the gateway name/ns.

zac-nixon: We talked about this. We can just use the parent ref field in the status.

pkg/gateway/routeutils/route_reconciler.go
zac-nixon: We chatted about this, we think the queue item is stable enough to actually de-dupe correctly.

zac-nixon: AI: Come back and validate later.

pkg/gateway/routeutils/route_reconciler.go
shraddhabang: This is currently blocking all the L4 route testing though.

shuqz: going to added in new revision

pkg/gateway/routeutils/route_reconciler.go
shuqz: i updated routeData, with the current info passed in, the dedup feature should work

pkg/gateway/routeutils/loader.go
```diff
@@ -87,11 +87,20 @@ func (l *loaderImpl) LoadRoutesForGateway(ctx context.Context, gw gwv1.Gateway,
...
+	for _, routeList := range mappedRoutes {
+		for _, route := range routeList {
+			deferredRouteReconciler.Enqueue(
+				GenerateRouteData(true, true, string(gwv1.RouteConditionAccepted), RouteStatusInfoAcceptedMessage, route, gw),
+			)
+		}
+	}
+
 	// 3. Load the underlying resource(s) for each route that is configured.
 	return l.loadChildResources(ctx, mappedRoutes)
```
zac-nixon, line 105: Talked with Shelley that by doing the route before this load call, we are missing out on implementing a bunch of `RouteConditionResolvedRefs` statuses. We'll tackle that in a different commit.

zac-nixon (approved):

zac-nixon (approved):

PR #4192, opened 2025-05-21
[feat: gw api] support gw deletion

zac-nixon:
### Description

Originally, we missed adding the deletion logic to Gateways, meaning that the AWS resources would not get cleaned up after Gateway deletion.

- Added deletion logic
- Refactored gateway controller to use gateway specific events
- Removed redundant LB config resolver function

Testing:
- Deleted an NLB and ALB gateway. 

### Checklist
- [x] Manually tested


shraddhabang (approved):
/lgtm
/approved

PR #4198, opened 2025-05-23
[feat: gw api][bug fix] generate sg rules using listener protocols

shraddhabang:
### Description

1. This PR fixes the generation of sg rules using listener protocols instead  route protocols and its corresponding tests
2. I have removed the the check of spec and deletion updates from gwclass event handler since we want to reconcile the gateways impacted by gatewayclass if we change the lb configuration attached to gateway. We update the last updated version on gw class when we change the lb conf referrered by gwclass. Hence as soon as it changes, we wanna enqueue all the impacted gateways. 
3. Small logger update. 

### Checklist
- [x] Manually tested

### DIFF
--- a/controllers/gateway/eventhandlers/gateway_class_events.go
+++ b/controllers/gateway/eventhandlers/gateway_class_events.go
@@ -3,7 +3,6 @@ package eventhandlers
 import (
 	"context"
 	"github.com/go-logr/logr"
-	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/k8s"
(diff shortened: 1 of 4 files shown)


controllers/gateway/eventhandlers/gateway_class_events.go
```diff
@@ -42,17 +41,8 @@ func (h *enqueueRequestsForGatewayClassEvent) Create(ctx context.Context, e even
...
-	// we only care below update event:
-	//  1. GatewayClass spec updates
-	//  3. GatewayClass deletions
-	if equality.Semantic.DeepEqual(gwClassOld.Spec, gwClassNew.Spec) &&
-		equality.Semantic.DeepEqual(gwClassOld.DeletionTimestamp.IsZero(), gwClassNew.DeletionTimestamp.IsZero()) {
-		return
-	}
-
 	h.logger.V(1).Info("enqueue gatewayclass update event", "gatewayclass", gwClassNew.Name)
 	h.enqueueImpactedGateways(ctx, gwClassNew, queue)
```
zac-nixon, line 47: This is going to re-queue all gateways every sync period. Is the idea that we want this to happen in order to support lb conf changes?

pkg/gateway/model/model_build_security_group.go
```diff
@@ -200,97 +199,88 @@ func (builder *securityGroupBuilderImpl) buildManagedSecurityGroupIngressPermiss
...
-			break
-		default:
-			// Ignore? Throw error?
-		}
+func getSgRuleProtocol(protocol gwv1.ProtocolType) ec2types.Protocol {
+	switch protocol {
+	case gwv1.HTTPProtocolType, gwv1.HTTPSProtocolType, gwv1.TCPProtocolType, gwv1.TLSProtocolType:
+		return ec2types.ProtocolTcp
+	case gwv1.UDPProtocolType:
+		return ec2types.ProtocolUdp
```
zac-nixon:
```
  if protocol == gwv1.UDPProtocolType {
          return ec2types.ProtocolUdp
        }
        return  ec2types.ProtocolTcp
```
might be more concise

controllers/gateway/eventhandlers/gateway_class_events.go
shraddhabang: Yes we want to enqueue gateways when lb conf changes happens.

pkg/gateway/model/model_build_security_group.go
shraddhabang: Updated

zac-nixon (approved):

PR #4217, opened 2025-06-04
[feat gw-api]implement finalizer

shuqz:
### Description
- implement finalizer for target group configuration and load balancer configuration 
- LBConfig finalizer will be added during reconcileUpdate and removed during reconcileDelete, it will check both gateway and gatewayClass 
- TargetGroupConfig will be added in backend loader and removed when service is deleted

### Checklist
- [ ] Added tests that cover your change (if possible)
- [ ] Added/modified documentation as required (such as the `README.md`, or the `docs` directory)
- [x] Manually tested
========== test targetgroupconfiguration =======
targetgroupconfiguration will be removed when service is deleted 
Test process: 
1. targetgroupconfiguration refers to a service
2. http route added this service to trigger reconcile
2. delete service -> finalizer should be deleted
Result 
```
sic-test-alb % k describe targetgroupconfiguration targetgroupconfiguration-sample -n gateway-alb
Name:         targetgroupconfiguration-sample
Namespace:    gateway-alb
Labels:       <none>
Annotations:  <none>
API Version:  gateway.k8s.aws/v1beta1
Kind:         TargetGroupConfiguration
Metadata:
  Creation Timestamp:  2025-05-30T01:11:34Z
  Finalizers:
    gateway.k8s.aws/targetgroupconfigurations
  Generation:        5
  Resource Version:  21594197
  UID:               582df46b-2d0a-4a41-b23f-ffc42e07bdd0
Spec:
  Default Configuration:
    Health Check Config:
      Health Check Interval:      30
      Health Check Path:          /health
      Health Check Protocol:      http
      Health Check Timeout:       5
      Healthy Threshold Count:    2
      Unhealthy Threshold Count:  2
    Protocol:                     HTTP
    Protocol Version:             HTTP1
    Target Type:                  ip
  Target Reference:
    Group:
    Kind:   Service
    Name:   echoserver2
Events:     <none>
```
Load balancer Configuration result 
```
k describe loadbalancerconfiguration test-gw-alb-lb-config-1 -n gateway-alb
Name:         test-gw-alb-lb-config-1
Namespace:    gateway-alb
Labels:       <none>
Annotations:  <none>
API Version:  gateway.k8s.aws/v1beta1
Kind:         LoadBalancerConfiguration
Metadata:
  Creation Timestamp:  2025-06-03T20:48:36Z
  Finalizers:
    gateway.k8s.aws/loadbalancerconfigurations
  Generation:        1
  Resource Version:  21606856
[...]

PR #4225, opened 2025-06-06
[feat: gw api] Fix edge cases for lb config finalizers

shraddhabang:
### Description

1. Fixed few edge cases as mentioned in [here](https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4217#discussion_r2127518231). 

- Edge case: If the gwClass C with lbconfig C has two gateways A and B.  Finalizer is added on lbcfg C.  If both the gateways are deleted, the finalizer will not be removed from lbcfg sinc it is still being referenced by gwclass C. So when we delete gwclass C, it should also trigger the removal of finalizer on lbcfg C if its not in use by any other gateway/gwclas. 
- Edge case: If the gateway A is created with lbconfig A, the finalizer is added on lbcfg A. Now I change the gateway A to reference another lbcfg B. This does not remove finalizer on initial lbcfg A. So when the deletion of lbcfg gets blocked. Hence, if we get a deletion event for lbcfg, we should check whether its being used by any gw/gwclasses currenlty. If not, finalizer should be removed from it.

2. Moved the shared utils for gateway and lb cfg  in their own utils files to avoid cyclic imports.
3. Added missing rbac permissions to patch resources with finalizers
4. refactored the code for `IsLBConfigInUse` to make it modular based on gw and gwclass. 

### Checklist
- [x] Manually tested

### DIFF
--- a/controllers/gateway/eventhandlers/gateway_class_events.go
+++ b/controllers/gateway/eventhandlers/gateway_class_events.go
@@ -3,7 +3,6 @@ package eventhandlers
 import (
 	"context"
 	"github.com/go-logr/logr"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/gateway/gatewayutils"
@@ -53,13 +52,6 @@ func (h *enqueueRequestsForGatewayClassEvent) Update(ctx context.Context, e even
 
 // Delete is not implemented for this handler as GatewayClass deletion should be finalized and is prevented while referenced by Gateways
 func (h *enqueueRequestsForGatewayClassEvent) Delete(ctx context.Context, e event.TypedDeleteEvent[*gatewayv1.GatewayClass], queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
-	gwClass := e.Object
-	h.logger.V(1).Info("enqueue gatewayclass delete event", "gatewayclass", gwClass.Name)
-	// remove the load balancer configuration finalizer if there are any referred by this gwclass
-	if err := gatewayutils.RemoveLoadBalancerConfigurationFinalizers(ctx, nil, gwClass, h.k8sClient, h.finalizerManager, sets.New(h.gwController)); err != nil {
-		h.logger.V(1).Info("failed to remove finalizers on load balancer configuration for ", "gateway class", gwClass.Name)
-		return
-	}
 }
 
 func (h *enqueueRequestsForGatewayClassEvent) Generic(ctx context.Context, e event.TypedGenericEvent[*gatewayv1.GatewayClass], queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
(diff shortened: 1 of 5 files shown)


controllers/gateway/config_resolver.go
```diff
@@ -52,17 +53,7 @@ func (resolver *gatewayConfigResolverImpl) getLoadBalancerConfigForGateway(ctx c
...
-		// Convert local param ref -> namespaced param ref
-		ns := gwv1.Namespace(gw.Namespace)
-		gwParametersRef = &gwv1.ParametersReference{
-			Group:     gw.Spec.Infrastructure.ParametersRef.Group,
-			Kind:      gw.Spec.Infrastructure.ParametersRef.Kind,
-			Name:      gw.Spec.Infrastructure.ParametersRef.Name,
-			Namespace: &ns,
-		}
-	}
+	var gwParametersRef = gatewayutils.GetNamespacedParamRefForGateway(gw)
```
zac-nixon, line 64:
```suggestion
	gwParametersRef := gatewayutils.GetNamespacedParamRefForGateway(gw)
```

controllers/gateway/eventhandlers/gatewayclass/load_balancer_configuration_events.go
```diff
@@ -46,6 +49,14 @@ func (h *enqueueRequestsForLoadBalancerConfigurationEvent) Create(ctx context.Co
 func (h *enqueueRequestsForLoadBalancerConfigurationEvent) Update(ctx context.Context, e event.TypedUpdateEvent[*elbv2gw.LoadBalancerConfiguration], queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
```
zac-nixon, line 51: Why is this going under the Update handler? What should trigger it?

controllers/gateway/gateway_class_controller.go
zac-nixon, line 1: Why don't we have an add / remove finalizer call in the gw class controller?

controllers/gateway/eventhandlers/gateway_class_events.go
```diff
@@ -49,6 +53,13 @@ func (h *enqueueRequestsForGatewayClassEvent) Update(ctx context.Context, e even
 
 // Delete is not implemented for this handler as GatewayClass deletion should be finalized and is prevented while referenced by Gateways
 func (h *enqueueRequestsForGatewayClassEvent) Delete(ctx context.Context, e event.TypedDeleteEvent[*gatewayv1.GatewayClass], queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	gwClass := e.Object
+	h.logger.V(1).Info("enqueue gatewayclass delete event", "gatewayclass", gwClass.Name)
+	// remove the load balancer configuration finalizer if there are any referred by this gwclass
```
zac-nixon: this isn't the right place to perform the finalizer deletion. If it fails, it won't get retried.

shraddhabang: Even if its fails, when the actual deletion would occur for lbcfg, we will handle the removal of finalizers in lbcfg events as well which will succed right so they will be able to delete it eventually.

controllers/gateway/eventhandlers/gatewayclass/load_balancer_configuration_events.go
shraddhabang: When the the lbcfg with finalizers are attempted to delete, it triggers the update event as only the deletiontimestamp changes. hence added here so that if any residual lbfgs with finalizers gets delete attempt we try to clean these for them.

controllers/gateway/gateway_class_controller.go
shraddhabang: The finalizers are only added if there are any gateways referenced by gwclass referencing the lbcfg which we do in gateway controller logic. There is no need to add finalizers on lbcfg with just gatewayClass and no gateways.

PR #4230, opened 2025-06-12
tg / lb conf finalizers

zac-nixon:
### Description

Adds finalizers to TargetGroupConfiguration, LoadbalancerConfiguration, and GatewayClass (only if the controller manages it).

- TargetGroupConfiguration: IMO, we should not allow deletion when it's actively attached to a Gateway. Users can potentially delete their TG config and revert their targets back to the defaults which has a high potential of causing impact. Previously, the finalizer was preventing deletion until the referring service was deleted which is a bit clunky, what if I no longer want to use ELB but want to keep my service around?
- LoadbalancerConfiguration: Finalizer prevents deletion until no Gateway / GatewayClass refers to the configuration. This will ensure that users are able to have correctly working Gateways. Same logic as TG config, we should not allow customers to unknowingly reset their configuration by deleting the LBConfig. If they want to reset to defaults they can either modify the LBConfig or explicitly detach it from the gw / gwc.
- GatewayClass: Finalizer prevents deletion until no Gateways reference it. This makes sure that customers aren't able to break their existing gateways.

### Checklist
- [x] Manually tested


controllers/gateway/eventhandlers/gateway_class_events.go
```diff
@@ -60,13 +58,16 @@ func (h *enqueueRequestsForGatewayClassEvent) Generic(ctx context.Context, e eve
 }
 
 func (h *enqueueRequestsForGatewayClassEvent) enqueueImpactedGateways(ctx context.Context, gwClass *gatewayv1.GatewayClass, queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
-	gwList := gatewayutils.GetGatewaysManagedByGatewayClass(ctx, h.k8sClient, gwClass, h.gwController)
+	gwList, err := gatewayutils.GetGatewaysManagedByLBController(ctx, h.k8sClient, h.gwController)
```
shraddhabang: We should only enqueue the Gateways which are managed by GatewayClass rather than controller for gwclass events. Use `GetGatewaysManagedByGatewayClass`

controllers/gateway/eventhandlers/gateway_class_events.go
```diff
@@ -60,13 +58,16 @@ func (h *enqueueRequestsForGatewayClassEvent) Generic(ctx context.Context, e eve
...
+	if err != nil {
+		h.logger.Error(err, "failed to get gateways managed by controller")
```
shraddhabang: Also please add the `gwClass` for which this has failed to get the idea of which gwclass is at fault.

controllers/gateway/gateway_class_controller.go
```diff
@@ -100,6 +103,21 @@ func (r *gatewayClassReconciler) reconcile(ctx context.Context, req reconcile.Re
...
 
+	if gwClass.DeletionTimestamp == nil || gwClass.DeletionTimestamp.IsZero() {
+		return r.handleUpdate(ctx, gwClass)
+	}
+
+	return r.handleDelete(ctx, gwClass)
+}
+
+func (r *gatewayClassReconciler) handleUpdate(ctx context.Context, gwClass *gwv1.GatewayClass) error {
+	if k8s.HasFinalizer(gwClass, shared_constants.GatewayClassFinalizer) {
```
shraddhabang: NIT : `!k8s.HasFinalizer(gwClass, shared_constants.GatewayClassFinalizer)`. If it does not have a finalizer, then add one.

controllers/gateway/gateway_controller.go
```diff
@@ -165,7 +167,11 @@ type gatewayReconciler struct {
 func (r *gatewayReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
 	r.reconcileTracker(req.NamespacedName)
 	err := r.reconcileHelper(ctx, req)
-	return runtime.HandleReconcileError(err, r.logger)
+	res, reconcileErr := runtime.HandleReconcileError(err, r.logger)
```
shraddhabang: Why cant we simply return `runtime.HandleReconcileError(err, r.logger)`?

controllers/gateway/gateway_controller.go
```diff
@@ -224,17 +230,24 @@ func (r *gatewayReconciler) reconcileHelper(ctx context.Context, req reconcile.R
...
 			r.logger.Error(err, "Failed to process gateway delete")
+			return err
 		}
+		r.serviceReferenceCounter.UpdateRelations([]types.NamespacedName{}, k8s.NamespacedName(gw), true)
+		return nil
+	}
+	err = r.reconcileUpdate(ctx, gw, gwClass, stack, lb, backendSGRequired)
+	r.serviceReferenceCounter.UpdateRelations(getServicesFromRoutes(allRoutes), k8s.NamespacedName(gw), false)
+	if err != nil {
+		r.logger.Error(err, "Failed to process gateway update")
```
shraddhabang: NIT : Please add the failing `gateway` in the error log.

controllers/gateway/gateway_controller.go
```diff
@@ -224,17 +230,24 @@ func (r *gatewayReconciler) reconcileHelper(ctx context.Context, req reconcile.R
...
+		err = r.reconcileDelete(ctx, gw, stack, allRoutes)
 		if err != nil {
 			r.logger.Error(err, "Failed to process gateway delete")
+			return err
 		}
+		r.serviceReferenceCounter.UpdateRelations([]types.NamespacedName{}, k8s.NamespacedName(gw), true)
+		return nil
+	}
+	err = r.reconcileUpdate(ctx, gw, gwClass, stack, lb, backendSGRequired)
+	r.serviceReferenceCounter.UpdateRelations(getServicesFromRoutes(allRoutes), k8s.NamespacedName(gw), false)
```
shraddhabang, line 237: If the gateway update fails, should we still update the serviceReferenceCounter?

controllers/gateway/gateway_class_controller.go
zac-nixon: i think this a bug not a nit :))

controllers/gateway/gateway_controller.go
zac-nixon: initially had logging here, will change back

controllers/gateway/gateway_controller.go
zac-nixon: Yes, I think so. The gateway still relies on the Service / TG config at that time, we so want to prevent deletion of those resources

controllers/gateway/loadbalancer_configuration_controller.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	inUse, err := gatewayutils.IsLBConfigInUse(context.Background(), lbConf, r.k8sClient, constants.FullGatewayControllerSet)
+
+	if err != nil {
+		return err
+	}
+	// if the loadbalancer configuration is still in use, we should not delete it
+	if inUse {
+		return fmt.Errorf("loadbalancer configuration [%+v] is still in use", k8s.NamespacedName(lbConf))
```
shraddhabang, line 91: Does it have to be an error? Can we simply not log it as info?

zac-nixon: Yeah it's an error because the user is doing something they shouldn't be doing.

PR #4241, opened 2025-06-23
[feat gw-api]implement httproute matching

shuqz:
### Description
- added httproute matching logic 
- added precedence order logic to assign priority
- sourceIp is not cover in this PR 
- some unit testing and annotations are missing, i will update in another commit 
- i am giving pathType priority as exact > regex > pathPrefix, but this can be changed

### Checklist
- [x] Manually tested
```
Provided YAML :
spec:
  parentRefs:
    - name: sample-gw-alb
      port: 80
  hostnames:
#    - "test.example.com"
    - "*.example.com"
  rules: # no matches specified
    - backendRefs:
        - name: echoserver
          port: 80
      matches:
        - path:
            type: Exact
            value: /exactPath
        - path:
            type: PathPrefix
            value: /pathPrefixMatch
          headers:
            - name: "firstMatch"
              value: "someValue"
#          queryParams:
#            - name: "user"
#              value: "john"
#          method: GET
        - path: # Regex path match
            type: RegularExpression
            value: "/v+?/users"
    - backendRefs:
        - name: echoserver # this can be another backendRef
          port: 80
      matches:
        - path: # not specify type, should be treated as pathPrefix
            value: "/pathPrefixMatch"
          headers:
            - name: "version"
              type: RegularExpression
              value: "*[0-9]a-Z*"
            - name: "secondHeader"
              value: "someValue"
```
Screenshot 2025-06-23 at 1 41 05 PM


pkg/gateway/model/model_build_listener.go
```diff
@@ -164,63 +164,50 @@ func (l listenerBuilderImpl) buildL4ListenerSpec(ctx context.Context, stack core
 	return listenerSpec, nil
 }
 
+// this is only for L7 ALB
 func (l listenerBuilderImpl) buildListenerRules(stack core.Stack, ls *elbv2model.Listener, lb *elbv2model.LoadBalancer, securityGroups securityGroupOutput, gw *gwv1.Gateway, port int32, lbCfg elbv2gw.LoadBalancerConfiguration, routes map[int32][]routeutils.RouteDescriptor) error {
```
zac-nixon, line 167: There are conflicting comments here, you say only l7 here but then later on call an l4 function.

pkg/gateway/model/model_build_listener.go
```diff
@@ -164,63 +164,50 @@ func (l listenerBuilderImpl) buildL4ListenerSpec(ctx context.Context, stack core
...
+		if tagsErr != nil {
+			return tagsErr
+		}
+		for _, backend := range rule.GetBackends() {
+			targetGroup, tgErr := l.tgBuilder.buildTargetGroup(stack, gw, lbCfg, lb.Spec.IPAddressType, route, backend, securityGroups.backendSecurityGroupToken)
+			if tgErr != nil {
+				return tgErr
 			}
+			actions := buildL4ListenerDefaultActions(targetGroup) // TODO: update this and weighted target group support
+			albRules = append(albRules, ingress.Rule{
```
zac-nixon: It would be best to not mingle the `ingress` and `gateway` packages. Would it be possible to take some time to refactor what we need out of the ingress package into a common package.

pkg/gateway/model/utilities.go
```diff
@@ -62,3 +63,128 @@ func sortRoutesByHostnamePrecedence(routes []routeutils.RouteDescriptor) {
 		return false
 	})
 }
+
+type RulePrecedence struct {
```
zac-nixon: nit: consider putting this into it's own file, as to not bloat the general utilities file.

pkg/gateway/model/utilities.go
```diff
@@ -62,3 +63,128 @@ func sortRoutesByHostnamePrecedence(routes []routeutils.RouteDescriptor) {
...
+	RouteNamespacedName       string
+}
+
+func sortAllRulesByPrecedence(routes []routeutils.RouteDescriptor) []RulePrecedence {
+	var allRoutes []RulePrecedence
+
+	for _, route := range routes {
+		routeNamespacedName := route.GetRouteNamespacedName().String()
+		routeCreateTimestamp := getRouteCreateTimestamp(route)
+		highestPrecedenceHostname := getHighestPrecedenceHostname(route.GetHostnames())
```
zac-nixon: This is interesting. Were you following a specific specification here?

zac-nixon: Nvm, found it here. https://gateway-api.sigs.k8s.io/reference/spec/#httproute

zac-nixon: I guess I need some clarification, because I only see these specific rules under the GRPC route.
https://gateway-api.sigs.k8s.io/reference/spec/#grpcroutespec

pkg/gateway/routeutils/route_rule_condition.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	if match.Method != nil {
+		methodCondition := buildHttpMethodCondition(match.Method)
+		conditions = append(conditions, methodCondition...)
+	}
+	return conditions
+}
+
+func buildHttpPathCondition(path *gwv1.HTTPPathMatch) []elbv2model.RuleCondition {
+	// +kubebuilder:default=PathPrefix
```
zac-nixon: This looks out of place.

pkg/gateway/model/model_build_listener.go
shuqz: yea i already modified it to in my local, it was from throw away code.

pkg/gateway/model/model_build_listener.go
shuqz: i will take a look

pkg/gateway/routeutils/route_rule_condition.go
shuqz: forgot to delete, it was for me. will remove it

pkg/gateway/routeutils/route_rule_condition.go
```diff
@@ -0,0 +1,118 @@
...
+
+func buildHttpPathCondition(path *gwv1.HTTPPathMatch) []elbv2model.RuleCondition {
+	// Path Type will never be nil, default is prefixPath
+	// Path Value will never be nil, default is "/"
+	pathType := *path.Type
+	pathValue := *path.Value
+	if pathType == gwv1.PathMatchPathPrefix {
+		// the paths `/abc`, `/abc/`, and `/abc/def` would all match the prefix `/abc`, but the path `/abcd` would not.
+		if pathValue == "/" {
+			pathValue = "/*"
```
zac-nixon: Can you explain why this is needed? I was checking against the Ingress implementation and we translate `/` -> `/` without adding the wildcard.

pkg/gateway/routeutils/route_rule_precedence.go
```diff
@@ -0,0 +1,199 @@
...
+
+// sortHostnameListByPrecedence Given a hostname list, sort it by precedence order
+func sortHostnameListByPrecedence(hostnames []string) {
+	// sort hostnames based on their precedence
+	sort.Slice(hostnames, func(i, j int) bool {
+		return getHostnamePrecedenceOrder(hostnames[i], hostnames[j]) < 0
+	})
+}
+
+func getHostnameListPrecedenceOrder(hostnameOne, hostnameTwo []string) int {
```
zac-nixon: nit: adding a comment stating that the behavior is to NOT tiebreak when lists are not equal length but have the same elements, e.g
```
routeA: 
hostname: 
- "test.example.com" -> highest
- "test.a.com"
- "*.a.com"

routeB: 
hostname: 
- "test.example.com"
```

pkg/gateway/routeutils/route_rule_condition.go
shuqz: i was thinking, when pathType and value is not specified, it will be automatically assigned as prefix and /. and in this case, we should pass /* to ALB to present that this allows all path. 
```
	// +optional
	// +kubebuilder:default={type: "PathPrefix", value: "/"}
```

pkg/gateway/model/utilities.go
shuqz: synced offline regarding this, added the missing piece in new commit

pkg/gateway/routeutils/route_rule_precedence.go
shuqz: ah i have a comment regarding this before return 0 at the end, but i can add a general comment for this function

pkg/gateway/routeutils/route_rule_condition.go
zac-nixon: Is this what you're referring to?
```
	// If no matches are specified, the default is a prefix
	// path match on "/", which has the effect of matching every
	// HTTP request.
```

zac-nixon (approved):

pkg/gateway/routeutils/route_rule_precedence.go
```diff
@@ -0,0 +1,199 @@
...
+	RouteCreateTimestamp time.Time
+	RouteNamespacedName  string
+}
+
+func SortAllRulesByPrecedence(routes []RouteDescriptor) []RulePrecedence {
+	var allRoutes []RulePrecedence
+
+	for _, route := range routes {
+		routeNamespacedName := route.GetRouteNamespacedName().String()
+		routeCreateTimestamp := getRouteCreateTimestamp(route)
```
shraddhabang: NIT : Add this `getRouteCreateTimestamp` as part of `routeMetadataDescriptor` interface as its part of metadata of the routes. 
```
type routeMetadataDescriptor interface {
    GetRouteNamespacedName() types.NamespacedName
    GetRouteKind() RouteKind
    GetHostnames() []gwv1.Hostname
    GetParentRefs() []gwv1.ParentReference
    GetRawRoute() interface{}
    GetBackendRefs() []gwv1.BackendRef
    GetRouteGeneration() int64
    GetRouteCreateTimestamp() time.Time  // Add this method
}
```

pkg/gateway/model/model_build_listener.go
```diff
@@ -164,63 +163,57 @@ func (l listenerBuilderImpl) buildL4ListenerSpec(ctx context.Context, stack core
...
-					rules = append(rules, ingress.Rule{
-						Conditions: conditions,
-						Actions:    actions,
-						Tags:       tags,
-					})
-				}
+		var conditionsList []elbv2model.RuleCondition
+		switch route.GetRouteKind() {
+		case routeutils.HTTPRouteKind:
+			conditionsList = routeutils.BuildHttpRuleConditions(ruleWithPrecedence.HTTPMatch, hostnamesStringList)
```
shraddhabang: we should add validation for total number conditions while building a rule since we only allow 5 max conditions per rule. We should not depend on the API for this error.
```
{"level":"debug","ts":"2025-06-25T22:58:38Z","logger":"events","msg":"Failed deploy model due to operation error Elastic Load Balancing v2: ModifyRule, https response error StatusCode: 400, RequestID: d6c1ce56-6abe-437d-ac42-7073b4c0a0e9, api error ValidationError: You can specify up to 5 conditions","type":"Warning","object":{"kind":"Gateway","namespace":"shrabang-ns","name":"test-alb-1","uid":"bce97194-5f60-492a-ab31-a3f2236663b8","apiVersion":"gateway.networking.k8s.io/v1","resourceVersion":"31935251"},"reason":"FailedDeployModel"}
```

pkg/gateway/routeutils/route_rule_precedence.go
```diff
@@ -0,0 +1,199 @@
...
+	Rule             RouteRule
+	RuleIndexInRoute int // index of the rule in the route
+	MatchIndexInRule int // index of the match in the rule
+
+	HTTPMatch *v1.HTTPRouteMatch
+	GRPCMatch *v1.GRPCRouteMatch
+
+	// factors determining precedence
+	Hostnames            []string // raw hostnames from route, unsorted
+	PathType             int      // 3=exact, 2=regex, 1=prefix
```
shraddhabang, line 21: Can we use enums for these instead of these int values?

pkg/gateway/routeutils/route_rule_precedence.go
shraddhabang: NIT : `func getHostnameListPrecedenceOrder(hostnameListOne, hostnameListTwo []string)`

pkg/gateway/model/model_build_listener.go
```diff
@@ -164,63 +163,57 @@ func (l listenerBuilderImpl) buildL4ListenerSpec(ctx context.Context, stack core
 	return listenerSpec, nil
 }
 
+// this is only for L7 ALB
 func (l listenerBuilderImpl) buildListenerRules(stack core.Stack, ls *elbv2model.Listener, lb *elbv2model.LoadBalancer, securityGroups securityGroupOutput, gw *gwv1.Gateway, port int32, lbCfg elbv2gw.LoadBalancerConfiguration, routes map[int32][]routeutils.RouteDescriptor) error {
+	// sort all rules based on precedence
+	routesWithPrecedenceOrder := routeutils.SortAllRulesByPrecedence(routes[port])
```
shraddhabang: NIT:  rename`routesWithPrecedenceOrder` to `rulesWithPrecedenceOrder`

pkg/gateway/model/model_build_listener.go
```diff
@@ -164,63 +163,57 @@ func (l listenerBuilderImpl) buildL4ListenerSpec(ctx context.Context, stack core
...
+
+	var albRules []elbv2model.Rule
+	for _, ruleWithPrecedence := range routesWithPrecedenceOrder {
+		route := ruleWithPrecedence.RouteDescriptor
+		rule := ruleWithPrecedence.Rule
+		var hostnamesStringList []string
+		// add hostname to condition
+		if hostnames := route.GetHostnames(); len(hostnames) > 0 {
```
shraddhabang: We have this info already in `ruleWithPrecedence`. No need to create hostnamesStringList again from route. 
Replace this with `hostnamesStringList :=ruleWithPrecedence.Hostnames`

pkg/gateway/routeutils/route_rule_condition.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func buildHttpHeaderCondition(headers []gwv1.HTTPHeaderMatch) []elbv2model.RuleCondition {
+	var conditions []elbv2model.RuleCondition
+	for _, header := range headers {
+		headerCondition := []elbv2model.RuleCondition{
+			{
+				Field: elbv2model.RuleConditionFieldHTTPHeader,
+				HTTPHeaderConfig: &elbv2model.HTTPHeaderConditionConfig{
+					HTTPHeaderName: string(header.Name),
```
shraddhabang, line 99: Add validations that header name can be only 40 char long. 
Refer Console Validations: `Valid characters are a-z, A-Z, 0-9 and special characters. HTTP header name must be 1-40 characters.`

pkg/gateway/routeutils/route_rule_condition.go
```diff
@@ -0,0 +1,118 @@
...
+					Values:         []string{header.Value},
```
shraddhabang, line 100: Add validations that header value can be only 128 char long.
refer console validations: `Valid characters are a-z, A-Z, 0-9 and special characters. HTTP header value must be 1-128 characters.`

pkg/gateway/routeutils/route_rule_condition.go
```diff
@@ -0,0 +1,118 @@
...
+				},
+			},
+		}
+		conditions = append(conditions, headerCondition...)
+	}
+	return conditions
+}
+
+func buildHttpQueryParamCondition(queryParams []gwv1.HTTPQueryParamMatch) []elbv2model.RuleCondition {
+	var conditions []elbv2model.RuleCondition
```
shraddhabang, line 110: Same add validations if required

pkg/gateway/routeutils/route_rule_precedence.go
shuqz: ENUM is good for understand but it will be it slower to compare, i am using value to compare directly in `comparePrecedence`

pkg/gateway/routeutils/route_rule_precedence.go
shuqz: good idea

pkg/gateway/routeutils/route_rule_condition.go
shuqz: synced offline

shuqz:
synced with [shraddhabang](https://github.com/shraddhabang) we will defer validation check to API directly, resolving all validation related comments.

zac-nixon (approved):

PR #4264, opened 2025-07-14
[feat gw-api] backfill E2E tests

shuqz:
### Description
- backfil E2E test for httproute match and filter feature, mostly added verifier for rule in E2E test
- added some unit tests
- added exception for other header modification since we only support redirect 
- added function for building target stickiness (this will be used later)
- fixed some minor issues i saw during implementation

### Checklist
- [x] Manually tested
```
ginkgo -v -r test/e2e/gateway --  --kubeconfig=$KUBECONFIG --cluster-name=xxx --aws-region=xx --aws-vpc-id=vpc-xx --enable-gateway-tests -ginkgo.focus="test k8s alb gateway using instance targets reconciled by the aws load balancer controller|test k8s alb gateway using ip targets reconciled by the aws load balancer controller"

with result: 
------------------------------
SS

Ran 6 of 14 Specs in 1825.423 seconds
SUCCESS! -- 6 Passed | 0 Failed | 0 Pending | 8 Skipped
PASS
```

### DIFF
--- a/pkg/gateway/model/model_build_listener.go
+++ b/pkg/gateway/model/model_build_listener.go
@@ -14,7 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	elbv2gw "sigs.k8s.io/aws-load-balancer-controller/apis/gateway/v1beta1"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/aws/services"
-	certs "sigs.k8s.io/aws-load-balancer-controller/pkg/certs"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/certs"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/gateway/routeutils"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/k8s"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/model/core"
@@ -204,7 +204,7 @@ func (l listenerBuilderImpl) buildListenerRules(stack core.Stack, ls *elbv2model
 				Weight:         &weight,
 			})
 		}
-		actions := buildL7ListenerActions(targetGroupTuples)
+		actions := buildL7ListenerForwardActions(targetGroupTuples, nil)
 
 		// configure actions based on filters
 		switch route.GetRouteKind() {
@@ -344,13 +344,24 @@ func buildL4ListenerDefaultActions(targetGroup *elbv2model.TargetGroup) []elbv2m
 	}
 }
 
-func buildL7ListenerActions(targetGroupTuple []elbv2model.TargetGroupTuple) []elbv2model.Action {
+func buildL7ListenerForwardActions(targetGroupTuple []elbv2model.TargetGroupTuple, duration *int32) []elbv2model.Action {
+
+	forwardConfig := &elbv2model.ForwardActionConfig{
+		TargetGroups: targetGroupTuple,
+	}
+
+	// Configure target group stickiness if a duration is specified
+	if duration != nil {
+		forwardConfig.TargetGroupStickinessConfig = &elbv2model.TargetGroupStickinessConfig{
+			Enabled:         awssdk.Bool(true),
+			DurationSeconds: awssdk.Int32(*duration),
+		}
+	}
+
 	return []elbv2model.Action{
 		{
-			Type: elbv2model.ActionTypeForward,
-			ForwardConfig: &elbv2model.ForwardActionConfig{
-				TargetGroups: targetGroupTuple,
-			},
+			Type:          elbv2model.ActionTypeForward,
+			ForwardConfig: forwardConfig,
 		},
 	}
 }
--- a/pkg/gateway/routeutils/route_rule_condition.go
+++ b/pkg/gateway/routeutils/route_rule_condition.go
@@ -98,7 +98,8 @@ func buildHttpHeaderCondition(headers []gwv1.HTTPHeaderMatch) []elbv2model.RuleC
 				Field: elbv2model.RuleConditionFieldHTTPHeader,
 				HTTPHeaderConfig: &elbv2model.HTTPHeaderConditionConfig{
 					HTTPHeaderName: string(header.Name),
-					Values:         []string{header.Value},
+					// for a given HTTPHeaderName, ALB rule can accept a list of values. However, gateway route headers only accept one value per name, and name cannot duplicate.
+					Values: []string{header.Value},
 				},
 			},
 		}
@@ -146,15 +147,14 @@ func buildGrpcRouteRuleConditions(matches RouteRule) ([][]elbv2model.RuleConditi
 	return conditions, nil
 }
 
+// BuildHttpRuleActionsBasedOnFilter only request redirect is supported, header modification is limited due to ALB support level.
 func BuildHttpRuleActionsBasedOnFilter(filters []gwv1.HTTPRouteFilter) ([]elbv2model.Action, error) {
 	for _, filter := range filters {
 		switch filter.Type {
-		case gwv1.HTTPRouteFilterRequestHeaderModifier:
-			// TODO: decide behavior for request header modifier
 		case gwv1.HTTPRouteFilterRequestRedirect:
 			return buildHttpRedirectAction(filter.RequestRedirect)
-		case gwv1.HTTPRouteFilterResponseHeaderModifier:
-			// TODO: decide behavior for response header modifier
+		default:
+			return nil, errors.Errorf("Unsupported filter type: %v. Only request redirect is supported. To specify header modification, please configure it through LoadBalancerConfiguration.", filter.Type)
 		}
 	}
 	return nil, nil
--- a/test/e2e/gateway/alb_instance_target_test.go
+++ b/test/e2e/gateway/alb_instance_target_test.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"fmt"
 	awssdk "github.com/aws/aws-sdk-go-v2/aws"
+	elbv2types "github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2/types"
+	"github.com/gavv/httpexpect/v2"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	elbv2gw "sigs.k8s.io/aws-load-balancer-controller/apis/gateway/v1beta1"
+	elbv2model "sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2"
 	"sigs.k8s.io/aws-load-balancer-controller/test/framework/http"
 	"sigs.k8s.io/aws-load-balancer-controller/test/framework/utils"
 	"sigs.k8s.io/aws-load-balancer-controller/test/framework/verifier"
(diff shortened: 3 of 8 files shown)


pkg/gateway/model/model_build_listener.go
```diff
@@ -344,7 +344,24 @@ func buildL4ListenerDefaultActions(targetGroup *elbv2model.TargetGroup) []elbv2m
 	}
 }
 
-func buildL7ListenerActions(targetGroupTuple []elbv2model.TargetGroupTuple) []elbv2model.Action {
+func buildL7ListenerForwardActions(targetGroupTuple []elbv2model.TargetGroupTuple, duration *int) []elbv2model.Action {
```
shraddhabang: It would easier to define stickiness duration as `*int32` to avoid unnecessary type conversion (`int32(*duration)`).

pkg/gateway/model/model_build_listener.go
```diff
@@ -344,7 +344,24 @@ func buildL4ListenerDefaultActions(targetGroup *elbv2model.TargetGroup) []elbv2m
...
+	// forward action with stickiness
+	if duration != nil {
```
shraddhabang, line 354: We can avoid code duplication by ceating the ForwardConfig once and conditionally adding stickiness. Something like, 

```	
forwardConfig := &elbv2model.ForwardActionConfig{
       TargetGroups: targetGroupTuple,
}
	
// Configure target group stickiness if a duration is specified
if duration != nil {
forwardConfig.TargetGroupStickinessConfig = &elbv2model.TargetGroupStickinessConfig{
	Enabled:         awssdk.Bool(true),
	DurationSeconds: awssdk.Int32(int32(*duration)),
}
}
	
return []elbv2model.Action{
{
	Type:          elbv2model.ActionTypeForward,
	ForwardConfig: forwardConfig,
},
}
```

test/e2e/gateway/shared_resource_definitions.go
```diff
@@ -346,3 +346,214 @@ func buildHTTPRoute(hostnames []string) *gwv1.HTTPRoute {
...
+								{
+									Name:  testQueryStringKeyOne,
+									Value: "queryValueOne",
+								},
+								{
+									Name:  testQueryStringKeyTwo,
+									Value: "queryValueTwo",
+								},
+							},
+							//Headers: []gwv1.HTTPHeaderMatch{
```
shuqz: i will remove this

test/e2e/gateway/shared_resource_definitions.go
```diff
@@ -346,3 +346,214 @@ func buildHTTPRoute(hostnames []string) *gwv1.HTTPRoute {
...
+			Rules: []gwv1.HTTPRouteRule{
+				// rule 1
+				{
+					Matches: []gwv1.HTTPRouteMatch{
+						{
+							Path: &gwv1.HTTPPathMatch{
+								Type:  &[]gwv1.PathMatchType{gwv1.PathMatchExact}[0],
+								Value: awssdk.String(testPathString),
+							},
+							//QueryParams: []gwv1.HTTPQueryParamMatch{
```
shuqz: same here, forgot to cleanup

test/e2e/gateway/alb_instance_target_test.go
```diff
@@ -107,6 +121,269 @@ var _ = Describe("test k8s alb gateway using instance targets reconciled by the
...
+			}
+			tgSpec := elbv2gw.TargetGroupConfigurationSpec{}
+			gwListeners := []gwv1.Listener{
+				{
+					Name:     "test-listener",
+					Port:     80,
+					Protocol: gwv1.HTTPProtocolType,
+				},
+			}
+			httpr := buildHTTPRouteWithMatchesAndTargetGroupWeights()
```
shraddhabang: Instead of creating a dedicated func to build HTTP Routes with Matches and TG weights, Can you update the existing `buildHTTPRoute` function and send the required Filters and Matches for the rules. That way we can extend the same func to to process different testcases in future. I modified the same func to accept different hostnames we can extend it to accept filters and matches too.

test/framework/verifier/aws_resource_verifier.go
```diff
@@ -371,3 +378,205 @@ func VerifyListenerMutualAuthentication(lsMutualAuth *elbv2types.MutualAuthentic
...
+	if len(actual) != len(expected) {
+		return errors.Errorf("expected %d listener rule conditions, got %d", len(expected), len(actual))
+	}
+
+	// Create map of expected key-value pairs for lookup
+	expectedMap := make(map[string]string)
+	expectedMapForHeaders := make(map[string]string)
+	for _, expectedCondition := range expected {
+		expectedField := awssdk.ToString(expectedCondition.Field)
+		if expectedField == "query-string" {
```
shraddhabang: Can you use constants for expected fields as they are getting referenced in switch below too? 
Reference: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/ingress/model_build_listener_rules.go#L139

shuqz: those seems to be from ingress, i think we want to separate out them. but i can use what's defined in elbv2model, `elbv2model.RuleConditionFieldQueryString`, i was using those in other files but forgot to change them here. nice catch

test/e2e/gateway/alb_instance_target_test.go
shuqz: my intentional was to keep them independent since those will not be used in basic httproute tests. especially there are cases that when request redirect cannot be together with backendRefs, instead of handling all those in one function, i felt rather to keep them clean

shraddhabang: We could customize the httpRoutes with our without backend refs too if needed later. I think rather than creating dedicated func for each of them it will be easier to keep one buildHTTPRoute func so that we can customize it as much as we want. Then we only need to think about the possible combinations of Filters, Matches to build test cases easily.

test/e2e/gateway/alb_instance_target_test.go
```diff
@@ -107,6 +121,269 @@ var _ = Describe("test k8s alb gateway using instance targets reconciled by the
...
+				err = verifier.WaitUntilTargetsAreHealthy(ctx, tf, lbARN, len(nodeList))
+				Expect(err).NotTo(HaveOccurred())
+			})
+			By("waiting until DNS name is available", func() {
+				err := utils.WaitUntilDNSNameAvailable(ctx, dnsName)
+				Expect(err).NotTo(HaveOccurred())
+			})
+			By("sending http request to the lb", func() {
+				url := fmt.Sprintf("http://%v%s", dnsName, testPathString)
+				err := tf.HTTPVerifier.VerifyURL(url, http.ResponseCodeMatches(200))
```
shraddhabang, line 321: Do we also want to check the modified response headers in the response body to test e2e?

test/e2e/gateway/alb_ip_target_test.go
```diff
@@ -103,6 +116,279 @@ var _ = Describe("test k8s alb gateway using ip targets reconciled by the aws lo
...
+				Expect(lbARN).ToNot(BeEmpty())
+			})
+
+			By("waiting until DNS name is available", func() {
+				err := utils.WaitUntilDNSNameAvailable(ctx, dnsName)
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			By("testing redirect with ReplaceFullPath", func() {
+				httpExp := httpexpect.New(tf.LoggerReporter, fmt.Sprintf("http://%v", dnsName))
```
shraddhabang, line 371: This is great. Thank you for going above and beyond on this one.

pkg/gateway/model/model_build_listener.go
```diff
@@ -344,13 +344,24 @@ func buildL4ListenerDefaultActions(targetGroup *elbv2model.TargetGroup) []elbv2m
...
-func buildL7ListenerActions(targetGroupTuple []elbv2model.TargetGroupTuple) []elbv2model.Action {
+func buildL7ListenerForwardActions(targetGroupTuple []elbv2model.TargetGroupTuple, duration *int32) []elbv2model.Action {
+
+	forwardConfig := &elbv2model.ForwardActionConfig{
+		TargetGroups: targetGroupTuple,
+	}
+
+	// Configure target group stickiness if a duration is specified
+	if duration != nil {
+		forwardConfig.TargetGroupStickinessConfig = &elbv2model.TargetGroupStickinessConfig{
```
zac-nixon, line 355: We should be validating that the user is setting stickiness via Cookie. ALB doesn't support stickiness via Header.

pkg/gateway/routeutils/route_rule_condition.go
```diff
@@ -98,7 +98,8 @@ func buildHttpHeaderCondition(headers []gwv1.HTTPHeaderMatch) []elbv2model.RuleC
 				Field: elbv2model.RuleConditionFieldHTTPHeader,
 				HTTPHeaderConfig: &elbv2model.HTTPHeaderConditionConfig{
 					HTTPHeaderName: string(header.Name),
-					Values:         []string{header.Value},
+					// for a given HTTPHeaderName, ALB rule can accept a list of values. However, gateway route headers only accept one value per name, and name cannot duplicate.
```
zac-nixon, line 101: Interesting. Are we going to use the rule CRD to enhance this?

pkg/gateway/routeutils/route_rule_condition.go
```diff
@@ -146,15 +147,14 @@ func buildGrpcRouteRuleConditions(matches RouteRule) ([][]elbv2model.RuleConditi
...
 	for _, filter := range filters {
 		switch filter.Type {
-		case gwv1.HTTPRouteFilterRequestHeaderModifier:
-			// TODO: decide behavior for request header modifier
 		case gwv1.HTTPRouteFilterRequestRedirect:
 			return buildHttpRedirectAction(filter.RequestRedirect)
-		case gwv1.HTTPRouteFilterResponseHeaderModifier:
-			// TODO: decide behavior for response header modifier
+		default:
+			return nil, errors.Errorf("Unsupported filter type: %v. Only request redirect is supported. To specify header modification, please configure it through LoadBalancerConfiguration.", filter.Type)
```
zac-nixon, line 157: Can you please add a task to track this in documentation? I suspect we'll get a couple confused users for this.

pkg/gateway/model/model_build_listener.go
shuqz: those will be done when i actually do the target group stickiness implementation. i had this change and then realize we are going to do it in Rule CRD, but felt at least this part of code does not need to be discarded.

pkg/gateway/routeutils/route_rule_condition.go
shuqz: sure i will include it in documentation task so we do not forget

pkg/gateway/routeutils/route_rule_condition.go
shuqz: actually, that's not discussed yet. i am ok to add it later when customer needs it.

test/e2e/gateway/alb_instance_target_test.go
shuqz: i merged response headers modification test into the basic test with lbConfig. felt should be enough
```
			By("verifying listener attributes header modification is applied", func() {
				lsARN := verifier.GetLoadBalancerListenerARN(ctx, tf, lbARN, "80")
				err := verifier.VerifyListenerAttributes(ctx, tf, lsARN, map[string]string{
					headerModificationServerEnabled: "true",
					headerModificationMaxAge:        headerModificationMaxAgeValue,
				})
				Expect(err).NotTo(HaveOccurred())
			})
```

test/e2e/gateway/alb_instance_target_test.go
shuqz: ok i can make this change if it is preferred by you. i am fine with both

pkg/gateway/routeutils/route_rule_condition.go
shuqz: let me know if you want me to add it in rule-CRD, let's discuss it tomorrow

PR #4265, opened 2025-07-14
[feat: gw api] Add Rules CRD for customizing Gateway API L7 routes

shraddhabang:
# Add ListenerRuleConfiguration CRD for customizing Gateway API L7 routes with ELB-specific features

## Summary

This PR introduces a new custom resource definition (CRD) called `ListenerRuleConfiguration` that extends Gateway API's HTTPRoute capabilities with AWS Application Load Balancer (ALB) specific features. This implementation bridges the functionality gap between Gateway API's current feature set and AWS ALB's advanced capabilities.

## Problem Statement

While Gateway API provides a standardized way to manage ingress traffic across Kubernetes clusters, it currently lacks native support for several AWS ALB-specific features that customers rely on, such as:

- Authentication (OIDC and Cognito)
- Fixed response actions
- Target group stickiness
- Source IP conditions
- Advanced redirect capabilities (query parameter manipulation)

## Solution

We've implemented a new `RuleConfiguration` CRD that can be attached to HTTPRoutes using an `ExtensionRef` filter. This approach:

1. Preserves Gateway API compatibility while extending functionality
2. Focuses exclusively on features not natively supported by Gateway API
3. Maintains a clean separation between standard and AWS-specific configurations
4. Follows the same pattern established by our existing LoadBalancerConfiguration and TargetGroupConfiguration CRDs

The CRD allows users to configure:

- Authentication actions (OIDC and Cognito)
- Fixed response actions with custom status codes, content types and bodies
- Forward actions with target group stickiness
- Redirect actions with query parameter support
- Source IP conditions for rule matching

## Design Considerations

- We've limited the scope to features that don't overlap with Gateway API's native capabilities
- The implementation follows the existing pattern of our other Configuration CRDs for consistency
- The CRD includes appropriate validation rules to ensure configurations meet AWS requirements
- Documentation clearly indicates which fields are supported and which require native Gateway API configuration

## Usage

Users can reference the ListenerRuleConfiguration in their HTTPRoutes:

```yaml
kind: HTTPRoute
metadata:
  name: auth-oidc-http-route
  namespace: example-ns
spec:
  parentRefs:
    - name: test-alb-1
      sectionName: http-1
  hostnames:
    - example.com
  rules:
    - backendRefs:
        - name: echoserver
          port: 80
      filters:
        - type: ExtensionRef
          extensionRef:
            group: gateway.k8s.aws
            kind: RuleConfiguration
            name: cognito-auth-example
```

With a corresponding RuleConfiguration:

```yaml
metadata:
  name: cognito-auth-example
  namespace: example-ns
spec:
  actions:
    - type: authenticate-cognito
      authenticateCognitoConfig:
        userPoolID: "us-east-1_USUcEHSmh"
        userPoolClientID: "1q4ba1iri8o1va2tlsq5cj7nc6"
        userPoolDomain: "us-east-1usucehsmh"
        scope: "openid"
        sessionCookieName: "AWSELBAuthSessionCookie"
        sessionTimeout: 604800
        onUnauthenticatedRequest: "authenticate"
  tags:
    Environment: production
    Service: echoserver
```

## Documentation

- Added CRD reference documentation

### Checklist
- [x] Manually tested


apis/gateway/v1beta1/ruleconfig_types.go
```diff
@@ -0,0 +1,325 @@
+package v1beta1
```
shuqz: i notice we've been calling it between `RuleConfiguration` and `ListenerRuleConfiguration` in different places (descriptions/code), this can be confusing. let's make it fixed, i vote for `ListenerRuleConfiguration`

apis/gateway/v1beta1/ruleconfig_types.go
```diff
@@ -0,0 +1,325 @@
...
+	AuthenticateCognitoActionConditionalBehaviorEnumAllow        AuthenticateCognitoActionConditionalBehaviorEnum = "allow"
+	AuthenticateCognitoActionConditionalBehaviorEnumAuthenticate AuthenticateCognitoActionConditionalBehaviorEnum = "authenticate"
+)
+
+// AuthenticateOidcActionConditionalBehaviorEnum defines the behavior when a user is not authenticated
+// +kubebuilder:validation:Enum=deny;allow;authenticate
+type AuthenticateOidcActionConditionalBehaviorEnum string
+
+// Enum values for AuthenticateOidcActionConditionalBehaviorEnum
+const (
```
shuqz: they seem to be same as line 20-24, why not just use `AuthenticateActionConditionalBehaviorEnumDeny` ?

apis/gateway/v1beta1/ruleconfig_types.go
```diff
@@ -0,0 +1,325 @@
...
+	// Information for an authenticate-cognito action
+	// +optional
+	AuthenticateCognitoConfig *AuthenticateCognitoActionConfig `json:"authenticateCognitoConfig,omitempty"`
+
+	// Information for an authenticate-oidc action
+	// +optional
+	AuthenticateOIDCConfig *AuthenticateOidcActionConfig `json:"authenticateOIDCConfig,omitempty"`
+}
+
+// RuleSpec defines the desired state of RuleConfiguration
```
shuqz: if we decide to go with ListenerRuleConfiguration, we should update it in comments too

apis/gateway/v1beta1/ruleconfig_types.go
```diff
@@ -0,0 +1,325 @@
...
+	// Actions defines the set of actions to be performed when conditions match.
+	// This CRD implementation currently supports only  authenticate-oidc, authenticate-cognito, and fixed-response action types fully and forward and redirect actions partially
+	//
+	// For other fields in forward and redirect actions, please use the standard Gateway API HTTPRoute or other route resources, which provide
+	// native support for those conditions through the Gateway API specification.
+	//
+	// At most one authentication action can be specified (either authenticate-oidc or authenticate-cognito).
+	// +optional
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=2
```
shuqz: why is maxItem=2? i thought it is 4? they can specify fixed-response, forward, redirect and one of the auth together?

apis/gateway/v1beta1/ruleconfig_types.go
```diff
@@ -0,0 +1,325 @@
...
+	// Information for a source IP condition
+	// +optional
+	SourceIPConfig *SourceIPConditionConfig `json:"sourceIPConfig,omitempty"`
+}
+
+// ActionType defines the type of action for the rule
+// +kubebuilder:validation:Enum=forward;fixed-response;redirect;authenticate-cognito;authenticate-oidc
+type ActionType string
+
+const (
```
shuqz: i do not see any of those being used

shuqz:
you might wanna fix unit test

apis/gateway/v1beta1/ruleconfig_types.go
shraddhabang: I also wondered about this. But I kept them separate to match enums in ELB SDK Model. Just to avoid any confusion in future if we want to add new values for either of them.

apis/gateway/v1beta1/ruleconfig_types.go
shraddhabang: We have Type field in Action struct. https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4265/files#diff-349a0bf5ed031528f2ce597c1951b5c665d2bc4a9a4f571dc4be8bcb6a3eb063R239

apis/gateway/v1beta1/ruleconfig_types.go
```diff
@@ -0,0 +1,325 @@
...
+	ActionTypeForward             ActionType = "forward"
+	ActionTypeFixedResponse       ActionType = "fixed-response"
+	ActionTypeEnumRedirect        ActionType = "redirect"
+	ActionTypeAuthenticateCognito ActionType = "authenticate-cognito"
+	ActionTypeAuthenticateOIDC    ActionType = "authenticate-oidc"
+)
+
+// Information about the target group stickiness for a rule.
```
zac-nixon: This is already defined in https://gateway-api.sigs.k8s.io/reference/spec/#sessionpersistence. We should use it rather than defining another stickiness parameter.

apis/gateway/v1beta1/ruleconfig_types.go
shraddhabang: There can be atmost 2 actions for a rule. One of auth and other should be one of fixed-response, forward, redirect. Hence this validation. You cannot specify multiple non-auth actions.

apis/gateway/v1beta1/ruleconfig_types.go
zac-nixon: We are handling redirect and forward natively within the Gateway API. I think the only actions we need to define here are:
1. Fixed Response
2. Cognito
3. OIDC

apis/gateway/v1beta1/ruleconfig_types.go
```diff
@@ -0,0 +1,325 @@
...
+type ForwardActionConfig struct {
+
+	// The target group stickiness for the rule.
+	// Note: ForwardActionConfig only supports target group stickiness configuration through CRD.
+	// All other forward action fields must be set through the Gateway API native way.
+	TargetGroupStickinessConfig *TargetGroupStickinessConfig `json:"targetGroupStickinessConfig,omitempty"`
+}
+
+// Information about a redirect action.
+type RedirectActionConfig struct {
```
zac-nixon: This is supported by https://gateway-api.sigs.k8s.io/guides/http-redirect-rewrite/?h=redirect#redirects. Is there a reason to re-define it?

apis/gateway/v1beta1/ruleconfig_types.go
shraddhabang: Will update

apis/gateway/v1beta1/ruleconfig_types.go
shraddhabang: The gateway does not let your customize the query customiztion in url while ALB does suppport it. So to bridge that gap we are only allowing customization of query in redirect, the rest of the cofigurations must be done in natively through gateway api for redirect action.

apis/gateway/v1beta1/ruleconfig_types.go
shraddhabang: Shelly did try it out but the session persistence is still in experimental mode in L7 routes and has not graduated. Since we are only supporting standard features for the L7 routes we decided that we will not support session persistence until it graduates.

apis/gateway/v1beta1/listenerruleconfig_types.go
```diff
@@ -0,0 +1,325 @@
...
+	// Information for a source IP condition
+	// +optional
+	SourceIPConfig *SourceIPConditionConfig `json:"sourceIPConfig,omitempty"`
+}
+
+// ActionType defines the type of action for the rule
+// +kubebuilder:validation:Enum=forward;fixed-response;redirect;authenticate-cognito;authenticate-oidc
+type ActionType string
+
+const (
```
zac-nixon, line 59: I don't think we need to define Forward or Redirect actions as Gateway API defines those already.

apis/gateway/v1beta1/listenerruleconfig_types.go
```diff
@@ -0,0 +1,325 @@
...
+	ActionTypeForward             ActionType = "forward"
+	ActionTypeFixedResponse       ActionType = "fixed-response"
+	ActionTypeEnumRedirect        ActionType = "redirect"
+	ActionTypeAuthenticateCognito ActionType = "authenticate-cognito"
+	ActionTypeAuthenticateOIDC    ActionType = "authenticate-oidc"
+)
+
+// Information about the target group stickiness for a listener rule.
+type TargetGroupStickinessConfig struct {
```
zac-nixon, line 68: This is defined within the Gateway API so we don't need it here. 
https://gateway-api.sigs.k8s.io/reference/spec/#sessionpersistence

apis/gateway/v1beta1/listenerruleconfig_types.go
```diff
@@ -0,0 +1,325 @@
...
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=604800
+	DurationSeconds *int32 `json:"durationSeconds,omitempty"`
+
+	// Indicates whether target group stickiness is enabled.
+	// +kubebuilder:default=false
+	Enabled *bool `json:"enabled,omitempty"`
+}
+
+// Information about a forward action.
```
zac-nixon, line 82: I don't understand why we need this.

zac-nixon (changes requested):

zac-nixon:
I added comments originally, but I see the intent now to specifically support Auth actions.

zac-nixon (approved):

PR #4267, opened 2025-07-16
[gw api] Reference Grant support to allow cross namespace service access

zac-nixon:
### Description

This is a long one because it involves,
1/ Setting up a new event handler for Reference Grant events
2/ Refactoring the existing logic in the listener builder to support filtering out listeners that have no backends.
3/ Refactoring the route util logic to not fail on various situations [Invalid Svc, Invalid Ref Grant]
4/ Refactoring the E2E tests to allow for reference grant tests, and to allow more in-depth validations for the Target Groups created.

I have broken down the changes into a couple commits:

[reference grant event handler](https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/cd44402cfe6e196aea4d63ec603bbacbb63177e4)

Adds the event handler to respond to ReferenceGrant changes. The handler will send route events to any routes that are listed in the ReferenceGrant in order to trigger a reconcile. To support route deletions, we calculate routes that exist in the new reference grant OR in the old reference grant.

TODO: This should probably be done conditionally, if the user installed the ReferenceGrant CRD. But this PR is big enough as-is.

[refactor build listener to support ref grants](https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/1b8586cdd07d094acb557004e4d2640a18e4e32d)

This commit refactors the listener builder to handle listeners with 0 backends. In the case of ALB, this means throwing out any rules that have 0 actions and using the default 404 action. In the case of NLB, this means not provisioning the listener at all.

[refactor route utils to support ref grant](https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/859ebd096b9cbce6a007df08159ae8523a7053bc)

This commit refactors the route util to have loading errors (when we load the corresponding service for the backend ref) to be warning or fatal errors. Warning errors will lead to the route being updated with the updated status, but it won't cause the reconcile to fail. To the model builder, it will be like that the route doesn't exist. Fatal errors are produced during k8s api outage and will cause the reconcile cycle to fail.

The warning / fatal logic is used particularly for reference grant revocation, as we need to materialize the new LB state to remove the backend when a reference grant is revoked. I also expanded this logic to non-existent services to make the reconcile logic more robust).

[reference grant e2e tests](https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/618fccaa676dd21a4cddd4e4b601e4223e6c979e)

This commit fixes some bugs found in the ALB path for Reference Grants. This commit introduces e2e test logic for alb ip / instance tests and nlb ip / instance tests.

E2E test results:

```
Ran 12 of 12 Specs in 6967.095 seconds
SUCCESS! -- 12 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```

### Checklist
- [x] Manually tested


shuqz:
i still have e2e commit to go, but for other 3 commits, left comments in each commit

test/e2e/gateway/alb_instance_target_test.go
```diff
@@ -119,9 +125,58 @@ var _ = Describe("test k8s alb gateway using instance targets reconciled by the
...
+				}
+
+				err = verifier.VerifyAWSLoadBalancerResources(ctx, tf, lbARN, verifier.LoadBalancerExpectation{
+					Type:         "application",
+					Scheme:       "internet-facing",
+					Listeners:    stack.albResourceStack.getListenersPortMap(),
+					TargetGroups: expectedTargetGroups,
+				})
+				Expect(err).NotTo(HaveOccurred())
+			})
```
shraddhabang, line 166: Can we verifyURL for `http://%v:5000/any-path` before removing ref grant?

pkg/gateway/model/model_build_listener.go
```diff
@@ -153,7 +160,7 @@ func (l listenerBuilderImpl) buildL4ListenerSpec(ctx context.Context, stack core
 	}
 	routeDescriptor := routes[0]
 	if routeDescriptor.GetAttachedRules()[0].GetBackends() == nil || len(routeDescriptor.GetAttachedRules()[0].GetBackends()) == 0 {
-		return &elbv2model.ListenerSpec{}, errors.Errorf("no backend refs found for route %v for gateway %v, one backend ref must be specified", routeDescriptor.GetRouteNamespacedName(), k8s.NamespacedName(gw))
+		return nil, nil
```
shraddhabang, line 164: Lets add a log here that we are skipping this route and listener creation since we did not find any backend refs. Also need to change this doc limitation if we are skipping creation in such cases rather than failing. 
[listenerSpec, err = l.buildL7ListenerSpec(ctx, lb, subnets, gw, lbCfg, port, gwLsCfg, lbLsCfg)](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.13/guide/gateway/l4gateway/#single-backend-reference-per-l4-route)

pkg/gateway/model/model_build_listener.go
```diff
@@ -220,11 +232,13 @@ func (l listenerBuilderImpl) buildListenerRules(stack core.Stack, ls *elbv2model
 			// TODO: add case for GRPC
 		}
 
-		albRules = append(albRules, elbv2model.Rule{
-			Conditions: conditionsList,
-			Actions:    actions,
-			Tags:       tags,
-		})
+		if len(actions) > 0 {
+			albRules = append(albRules, elbv2model.Rule{
```
shraddhabang: Same here a log message would help that we are skipping adding rules if we find no actions.

controllers/gateway/eventhandlers/reference_grant_events.go
```diff
@@ -0,0 +1,174 @@
...
+			routes, err := routeutils.ListTCPRoutes(ctx, h.k8sClient, &client.ListOptions{Namespace: string(impactedFrom.Namespace)})
+			if err == nil {
+				for _, route := range routes {
+					h.tcpRouteEventChan <- event.TypedGenericEvent[*gwalpha2.TCPRoute]{
+						Object: route.GetRawRoute().(*gwalpha2.TCPRoute),
+					}
+				}
+
+			} else {
+				h.logger.Error(err, "Unable to list impacted grpc routes for reference grant event handler")
```
shraddhabang: Wrong route type in log message for all L4 routes.

test/e2e/gateway/alb_instance_target_test.go
```diff
@@ -119,9 +125,58 @@ var _ = Describe("test k8s alb gateway using instance targets reconciled by the
...
+			By("removing ref grant", func() {
+				err := auxiliaryStack.DeleteReferenceGrants(ctx, tf)
+				Expect(err).NotTo(HaveOccurred())
+				// Give some time to have the reference grant to be deleted
```
shraddhabang, line 175: Why do we need to wait for 2 minutes for deletion?

controllers/gateway/gateway_controller.go
```diff
@@ -501,6 +507,8 @@ func (r *gatewayReconciler) setupNLBGatewayControllerWatches(ctrl controller.Con
 		loggerPrefix.WithName("TLSRoute"))
 	svcEventHandler := eventhandlers.NewEnqueueRequestsForServiceEvent(nil, nil, tcpRouteEventChan, udpRouteEventChan, tlsRouteEventChan, r.k8sClient, r.eventRecorder,
 		loggerPrefix.WithName("Service"), constants.NLBGatewayController)
+	refGrantHandler := eventhandlers.NewEnqueueRequestsForReferenceGrantEvent(nil, nil, tcpRouteEventChan, udpRouteEventChan, tlsRouteEventChan, r.k8sClient, r.eventRecorder,
```
shraddhabang, line 512: Do we need to add RBAC perm for Ref grant resources?

test/e2e/gateway/alb_instance_target_test.go
zac-nixon: done

pkg/gateway/model/model_build_listener.go
zac-nixon: done

pkg/gateway/model/model_build_listener.go
zac-nixon: done

controllers/gateway/eventhandlers/reference_grant_events.go
zac-nixon: fixed

test/e2e/gateway/alb_instance_target_test.go
zac-nixon: this gives time for the ELB CP to propagate to the ELB DP.

controllers/gateway/gateway_controller.go
zac-nixon: yup. let me fix.

pkg/gateway/routeutils/utils.go
```diff
@@ -120,6 +121,37 @@ func IsHostNameInValidFormat(hostName string) (bool, error) {
 	return true, nil
 }
 
+func attachedRulesAccumulator[RuleType any](ctx context.Context, k8sClient client.Client, route preLoadRouteDescriptor, rules []RuleType, backendRefIterator func(RuleType) []gwv1.BackendRef, ruleConverter func(*RuleType, []Backend) RouteRule) ([]RouteRule, []routeLoadError) {
```
shuqz: are we using this function anywhere?

pkg/gateway/model/model_build_listener.go
```diff
@@ -220,11 +233,17 @@ func (l listenerBuilderImpl) buildListenerRules(stack core.Stack, ls *elbv2model
 			// TODO: add case for GRPC
 		}
 
+		if len(actions) == 0 {
+			l.logger.Info("Filling in no backend actions with fixed 503")
```
shuqz, line 237: can we add some info related to which rule?

pkg/gateway/routeutils/utils.go
zac-nixon: ah no, this was my first iteration of the accumulator. will remove

pkg/gateway/routeutils/backend.go
```diff
@@ -32,24 +33,72 @@ type Backend struct {
...
 	if backendRef.Kind != nil && *backendRef.Kind != "Service" {
 		initialErrorMessage := "Backend Ref must be of kind 'Service'"
 		wrappedGatewayErrorMessage := generateInvalidMessageWithRouteDetails(initialErrorMessage, routeKind, routeIdentifier)
-		return nil, wrapError(errors.Errorf("%s", initialErrorMessage), gwv1.GatewayReasonListenersNotValid, gwv1.RouteReasonInvalidKind, &wrappedGatewayErrorMessage, nil)
+		return nil, wrapError(errors.Errorf("%s", initialErrorMessage), gwv1.GatewayReasonListenersNotValid, gwv1.RouteReasonInvalidKind, &wrappedGatewayErrorMessage, nil), nil
 	}
 
 	if backendRef.Weight != nil && *backendRef.Weight == 0 {
-		return nil, nil
+		return nil, nil, nil
```
shraddhabang, line 95: Sorry I missed this earlier. This looks like we are skipping the backends with zero weights. However ALB allows the target groups with weight 0 in weighted scenario. Do we wanna still keep these backends with zero weights to match that behavior?

zac-nixon: Weight 0 means do not forward to this backend in the Gateway API
```
	// If only one backend is specified and it has a weight greater than 0, 100%
	// of the traffic is forwarded to that backend. If weight is set to 0, no
	// traffic should be forwarded for this entry. If unspecified, weight
	// defaults to 1.
```

zac-nixon: While diving for that blurb I did realize there is a mismatch between GW API and ALB. GW API allows weights 1-100,000

while ALB only allows up to 999.

I can add that validation in an upcoming PR.

shraddhabang: Good catch on the validation. But as per our offline discussion, Lets allow these zero weighted backend refs for on-fly traffic switching and avoiding the delay due to tg creation and registration of targets. Also, we will handle edge case where if we find single tgs with zero weight on forward action, we will simply consider as no backend ref found and add 503 fixed response action. This due to the fact that there is mismatch between behavior of gw and ALB for such single tgs with zero weight on forward action. Gateway recommends that no traffic should be sent to such tgs and ALB send 100% traffic.

pkg/gateway/routeutils/backend.go
```diff
@@ -32,24 +33,72 @@ type Backend struct {
 	Weight                int
 }
 
+type attachedRuleAccumulator[RuleType any] interface {
```
shraddhabang, line 36: This is a great refactor. Pure genius. :-)

test/e2e/gateway/alb_instance_target_test.go
shraddhabang: Got it.

PR #4270, opened 2025-07-21
[feat gw-api] implement grpcroute match

shuqz:
### Description
- implement grpcroute match 
- refactor code in rule precedence file, the big changes are 
   - grpcroute and httproute has independent compare precedence logic 
   - the sort logic will take all routes from same port but it will sort grpcroute and httproute independently, and append them together. based on current logic, rules from httproute will be appended first since they tend to be more specific, but we might plan to have it configurable by user later. This is to handle the case when both  grpcroute and httproute attached to same port, and for old logic, if will try to compare a rule from grpcroute and a rule from httproute, which does not make sense since they are different. Also, because the only time both can be attached is when hostname does not overlap, so they will never have same routing rule. we just need to keep the relative priority within each route to be correct.

- change priority of prefix and regex for easy of comparison, before it is exact > regex > prefix, now it is exact > prefix > regex (align with ingress)
- did not add e2e test yet, did manual testing and run some existing e2e test 

### Checklist
- [x] Manually tested
tested with 2 yaml files below
```
  rules:
    - matches:
        - method:
            # provide only method
            method: User
        - headers:
            # no method provided, will be treated as /*
            - name: test-header-4
              value: test-value-4
      backendRefs:
        - name: echoserver3
          port: 80
    - matches:
        - method:
            # provide both service and method
            service: com.example
            method: Login
          headers:
            - name: test-header-1
              value: test-value-1
            - name: test-header-2
              value: test-value-2
        - method:
            # provide only service
            service: second.com
          headers:
            - name: test-header-3
              value: test-value-3
-----------------------------------
  rules:
    - matches:
        - method:
            # provide only method
            type: RegularExpression
            method: User*
          headers:
            - name: test-header-5
              value: test-value-5
        - method:
            # provide only service
            type: RegularExpression
            service: service?
          headers:
            - name: test-header-6
              value: test-value-6
```
result:
Screenshot 2025-07-24 at 11 04 26 AM

### DIFF
--- a/pkg/gateway/model/model_build_listener.go
+++ b/pkg/gateway/model/model_build_listener.go
@@ -188,18 +188,19 @@ func (l listenerBuilderImpl) buildListenerRules(stack core.Stack, ls *elbv2model
 
 	var albRules []elbv2model.Rule
 	for _, ruleWithPrecedence := range rulesWithPrecedenceOrder {
-		route := ruleWithPrecedence.RouteDescriptor
-		rule := ruleWithPrecedence.Rule
+		route := ruleWithPrecedence.CommonRulePrecedence.RouteDescriptor
+		rule := ruleWithPrecedence.CommonRulePrecedence.Rule
 
 		var conditionsList []elbv2model.RuleCondition
 		var err error
 		switch route.GetRouteKind() {
 		case routeutils.HTTPRouteKind:
 			conditionsList, err = routeutils.BuildHttpRuleConditions(ruleWithPrecedence)
-			if err != nil {
-				return err
-			}
-			// TODO: add case for GRPC
+		case routeutils.GRPCRouteKind:
+			conditionsList, err = routeutils.BuildGrpcRuleConditions(ruleWithPrecedence)
+		}
+		if err != nil {
+			return err
 		}
 		tags, tagsErr := l.tagHelper.getGatewayTags(lbCfg)
 		if tagsErr != nil {
--- a/pkg/gateway/routeutils/route_rule_condition.go
+++ b/pkg/gateway/routeutils/route_rule_condition.go
@@ -11,7 +11,7 @@ import (
 // BuildHttpRuleConditions each match will be mapped to a ruleCondition, conditions within same match will be ANDed
 func BuildHttpRuleConditions(rule RulePrecedence) ([]elbv2model.RuleCondition, error) {
 	match := rule.HTTPMatch
-	hostnamesStringList := rule.Hostnames
+	hostnamesStringList := rule.CommonRulePrecedence.Hostnames
 	var conditions []elbv2model.RuleCondition
 	if hostnamesStringList != nil && len(hostnamesStringList) > 0 {
 		conditions = append(conditions, elbv2model.RuleCondition{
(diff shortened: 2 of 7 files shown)


pkg/gateway/routeutils/route_rule_condition.go
```diff
@@ -141,12 +141,106 @@ func buildHttpMethodCondition(method *gwv1.HTTPMethod) []elbv2model.RuleConditio
...
+			},
+		}
+		conditions = append(conditions, headerCondition...)
+	}
+	return conditions
+}
+
+func buildGrpcMethodCondition(method *gwv1.GRPCMethodMatch) ([]elbv2model.RuleCondition, error) {
+	var pathValue string
+	if *method.Type == gwv1.GRPCMethodMatchExact {
```
shraddhabang: Can we avoid the code duplication in if else statement as both are identical? Also can we add early validation for invalid match type? Something like this 
```
func buildGrpcMethodCondition(method *gwv1.GRPCMethodMatch) ([]elbv2model.RuleCondition, error) {

	if *method.Type != gwv1.GRPCMethodMatchExact && *method.Type != gwv1.GRPCMethodMatchRegularExpression {
		return nil, errors.Errorf("Invalid grpc method match type: %v", method.Type)
	}

	var pathValue string
        if method.Service != nil && method.Method != nil {
	        pathValue = fmt.Sprintf("/%s/%s", *method.Service, *method.Method)
        } else if method.Service != nil {
	        pathValue = fmt.Sprintf("/%s/*", *method.Service)
        } else if method.Method != nil {
	        pathValue = fmt.Sprintf("/*/%s", *method.Method)
        } else {
	        return nil, errors.Errorf("Invalid grpc method match: %v", method)
        }

	return []elbv2model.RuleCondition{
		{
			Field: elbv2model.RuleConditionFieldPathPattern,
			PathPatternConfig: &elbv2model.PathPatternConditionConfig{
				Values: []string{pathPattern},
			},
		},
	}, nil
}
```

pkg/gateway/model/model_build_listener.go
```diff
@@ -185,7 +185,11 @@ func (l listenerBuilderImpl) buildListenerRules(stack core.Stack, ls *elbv2model
 			if err != nil {
 				return err
 			}
-			// TODO: add case for GRPC
+		case routeutils.GRPCRouteKind:
+			conditionsList, err = routeutils.BuildGrpcRuleConditions(ruleWithPrecedence)
```
zac-nixon, line 200: nit you can move the err check outside the case statement to remove a branch:
```
		case routeutils.HTTPRouteKind:
			conditionsList, err = routeutils.BuildHttpRuleConditions(ruleWithPrecedence)
			if err != nil {
				return err
			}
		case routeutils.GRPCRouteKind:
			conditionsList, err = routeutils.BuildGrpcRuleConditions(ruleWithPrecedence)
			if err != nil {
				return err
			}
		}
```
-->
```
		case routeutils.HTTPRouteKind:
			conditionsList, err = routeutils.BuildHttpRuleConditions(ruleWithPrecedence)
		case routeutils.GRPCRouteKind:
			conditionsList, err = routeutils.BuildGrpcRuleConditions(ruleWithPrecedence)
		}
	        if err != nil {
		  return err
	        }
```

pkg/gateway/routeutils/route_rule_condition.go
```diff
@@ -141,12 +141,106 @@ func buildHttpMethodCondition(method *gwv1.HTTPMethod) []elbv2model.RuleConditio
...
 
+func buildGrpcHeaderCondition(headers []gwv1.GRPCHeaderMatch) []elbv2model.RuleCondition {
+	var conditions []elbv2model.RuleCondition
+	for _, header := range headers {
+		headerCondition := []elbv2model.RuleCondition{
+			{
+				Field: elbv2model.RuleConditionFieldHTTPHeader,
+				HTTPHeaderConfig: &elbv2model.HTTPHeaderConditionConfig{
+					HTTPHeaderName: string(header.Name),
+					// for a given HTTPHeaderName, ALB rule can accept a list of values. However, gateway route headers only accept one value per name, and name cannot duplicate.
```
zac-nixon, line 203: This will be fixed by Shraddha's CRD, right?

pkg/gateway/routeutils/route_rule_condition.go
zac-nixon: plus 100!

pkg/gateway/routeutils/route_rule_condition_test.go
```diff
@@ -440,3 +440,210 @@ func Test_BuildHttpRuleActionsBasedOnFilter(t *testing.T) {
...
+	testKey := "testKey"
+	testValue := "testValue"
+	testKeyTwo := "testKeyTwo"
+	testValueTwo := "testValueTwo"
+	tests := []struct {
+		name        string
+		headerMatch []gwv1.GRPCHeaderMatch
+		want        []elbv2model.RuleCondition
+	}{
+		{
```
zac-nixon, line 682: how about testing no match?

pkg/gateway/routeutils/route_rule_condition.go
shuqz: yes, i will work on this after CRD is done

pkg/gateway/routeutils/route_rule_condition.go
shuqz: actually, no need to check type validation, it won't pass validation when they apply in yaml

pkg/gateway/routeutils/route_rule_condition_test.go
shuqz: it won't enter buildGrpcHeaderCondition or buildGrpcMethodCondition if match is nil or match.Method == nil, see `BuildGrpcRuleConditions` 
```
	if match == nil || match.Method == nil {
		// If Method field is not specified, all services and methods will match.
		conditions = append(conditions, elbv2model.RuleCondition{
			Field: elbv2model.RuleConditionFieldPathPattern,
			PathPatternConfig: &elbv2model.PathPatternConditionConfig{
				Values: []string{"/*"},
			},
		})
		return conditions, nil
	}
```
but i can add unit test for BuildGrpcRuleConditions

PR #4273, opened 2025-07-22
refactor how no backends are modeled in NLB / ALB Gateways

zac-nixon:
### Description

follow up from: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4267

Key changes:
1/ No longer filter out weight = 0 backends in the route builder. Instead, we let the model builder handle the backend depending on which LB type we are working with.

NLB -> Only 1 TG allowed in a Listener, don't provision listener when the TG has weight = 0
ALB -> Propagate all weight = 0 target groups. We special case all target groups having weight = 0, we will instead insert a 503 fixed response rule. HOWEVER, the target groups and targets are still provisioned meaning that customers don't need to re-provision these resources later.

2/ When an NLB / ALB Listener has no backends, we follow the same rules as outlined above.

I was able to manually test NLB and ALB. Further, I've updated the E2E tests to also test that the behavior I have described above is actually implemented.

I noticed that the ListenerRules builder had no UT, so I've gone ahead and created two basic ones for my own scenarios I have added.

### Checklist
- [x] Manually tested


pkg/gateway/model/model_build_listener.go
```diff
@@ -176,10 +182,11 @@ func (l listenerBuilderImpl) buildL4ListenerSpec(ctx context.Context, stack core
 }
 
 // this is only for L7 ALB
-func (l listenerBuilderImpl) buildListenerRules(stack core.Stack, ls *elbv2model.Listener, lb *elbv2model.LoadBalancer, securityGroups securityGroupOutput, gw *gwv1.Gateway, port int32, lbCfg elbv2gw.LoadBalancerConfiguration, routes map[int32][]routeutils.RouteDescriptor) error {
+func (l listenerBuilderImpl) buildListenerRules(stack core.Stack, ls *elbv2model.Listener, ipAddressType elbv2model.IPAddressType, securityGroups securityGroupOutput, gw *gwv1.Gateway, port int32, lbCfg elbv2gw.LoadBalancerConfiguration, routes map[int32][]routeutils.RouteDescriptor) error {
 	// sort all rules based on precedence
 	rulesWithPrecedenceOrder := routeutils.SortAllRulesByPrecedence(routes[port])
 
+	fmt.Printf("Got these rules %d (initial %d)\n", len(rulesWithPrecedenceOrder), len(routes[port]))
```
shraddhabang: leftover?

pkg/gateway/routeutils/route_rule_precedence.go
```diff
@@ -57,6 +58,8 @@ func SortAllRulesByPrecedence(routes []RouteDescriptor) []RulePrecedence {
 					getHttpMatchPrecedenceInfo(&httpMatch, &match)
 					allRoutes = append(allRoutes, match)
 				}
+			default:
+				fmt.Printf("GOT THIS KIND!! %s\n", r)
```
shraddhabang: leftover?

pkg/networking/security_group_reconciler.go
```diff
@@ -114,6 +114,7 @@ func (r *defaultSecurityGroupReconciler) reconcileIngressWithSGInfo(ctx context.
 
 	if len(permissionsToGrant) > 0 {
 		if err := r.sgManager.AuthorizeSGIngress(ctx, sgInfo.SecurityGroupID, permissionsToGrant); err != nil {
+			r.logger.Error(err, "GOT THIS ERROR!")
```
shraddhabang: leftover?

pkg/gateway/model/model_build_listener.go
```diff
@@ -167,6 +167,12 @@ func (l listenerBuilderImpl) buildL4ListenerSpec(ctx context.Context, stack core
 		return &elbv2model.ListenerSpec{}, errors.Errorf("multiple backend refs found for route %v for listener on port:protocol %v:%v for gateway %v , only one must be specified", routeDescriptor.GetRouteNamespacedName(), port, listenerSpec.Protocol, k8s.NamespacedName(gw))
 	}
 	backend := routeDescriptor.GetAttachedRules()[0].GetBackends()[0]
+
+	if backend.Weight == 0 {
```
shuqz, line 171: do we wanna check backend.Weight != nil here?

pkg/gateway/model/model_build_listener.go
```diff
@@ -545,3 +551,16 @@ func newListenerBuilder(ctx context.Context, loadBalancerType elbv2model.LoadBal
 		logger:           logger,
 	}
 }
+
+// shouldProvisionActions -- determine if the given target groups are acceptable for ELB Actions.
+// The criteria -
+// 1/ One or more target groups are present.
+// 2/ At least one target group has a weight greater than zero.
+func shouldProvisionActions(targetGroups []elbv2model.TargetGroupTuple) bool {
```
shuqz, line 559: i might missed it, but where is this part being implemented? 
"ALB - We special case all target groups having weight = 0, we will instead insert a 503 fixed response rule."

pkg/gateway/model/model_build_listener.go
zac-nixon: This can't be nil. The route loader forces all `nil` values to be 1 per Gateway API spec:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/gateway/routeutils/backend.go#L205-L219

pkg/gateway/model/model_build_listener.go
zac-nixon: actions are built here (IIF there are going to be target groups to send to)
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/ebefb4f0946e86c9b4c4ee522470c016ef9c0fd7/pkg/gateway/model/model_build_listener.go#L224-L226

when we generate no forward actions: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/ebefb4f0946e86c9b4c4ee522470c016ef9c0fd7/pkg/gateway/model/model_build_listener.go#L242-L245

PR #4274, opened 2025-07-22
[feat: gw api] Add event handling for Rules CRD

shraddhabang:
### Description

1. Added event handlers for Rules CRD
2. Fixed few minor typos in logs 
3. Added unit tests for all the utility functions

### Checklist
- [x] Manually tested


controllers/gateway/eventhandlers/listener_rule_configuration_events.go
```diff
@@ -0,0 +1,97 @@
...
+
+// NewEnqueueRequestsForListenerRuleConfigurationEvent creates handler for ListenerRuleConfiguration resources
+func NewEnqueueRequestsForListenerRuleConfigurationEvent(httpRouteEventChan chan<- event.TypedGenericEvent[*gatewayv1.HTTPRoute],
+	grpcRouteEventChan chan<- event.TypedGenericEvent[*gatewayv1.GRPCRoute], k8sClient client.Client, eventRecorder record.EventRecorder, gwController string, logger logr.Logger) handler.TypedEventHandler[*elbv2gw.ListenerRuleConfiguration, reconcile.Request] {
+	return &enqueueRequestsForListenerRuleConfigurationEvent{
+		httpRouteEventChan: httpRouteEventChan,
+		grpcRouteEventChan: grpcRouteEventChan,
+		k8sClient:          k8sClient,
+		eventRecorder:      eventRecorder,
+		gwController:       gwController,
```
zac-nixon: is this needed?

controllers/gateway/eventhandlers/listener_rule_configuration_events.go
```diff
@@ -0,0 +1,97 @@
...
+		gwControllerSet:    sets.New(gwController),
```
zac-nixon: is this needed?

pkg/gateway/routeutils/descriptor.go
```diff
@@ -18,6 +18,7 @@ type routeMetadataDescriptor interface {
 	GetParentRefs() []gwv1.ParentReference
 	GetRawRoute() interface{}
 	GetBackendRefs() []gwv1.BackendRef
+	GetListenerRuleConfigs() []gwv1.LocalObjectReference
```
zac-nixon, line 21: Do you want to return the local object reference, or is it better to load the actual CRDs and then attach them to the route descriptor? The idea of this route descriptor is that it holds data needed to make a complete gateway object, so we would need to have the CRD content loaded for the model builder.

pkg/gateway/routeutils/tcp.go
```diff
@@ -102,6 +102,10 @@ func (tcpRoute *tcpRouteDescription) GetBackendRefs() []gwv1.BackendRef {
 	return backendRefs
 }
 
+func (tcpRoute *tcpRouteDescription) GetListenerRuleConfigs() []gwv1.LocalObjectReference {
+	return nil
```
zac-nixon: nit: so we don't have to worry about nil checking, IMO this should just return an empty array.

config/rbac/role.yaml
```diff
@@ -128,6 +128,30 @@ rules:
...
+  - listenerruleconfigurations
+  verbs:
+  - get
+  - list
+  - patch
+  - watch
+- apiGroups:
+  - gateway.k8s.aws
+  resources:
+  - listenerruleconfigurations/finalizers
```
zac-nixon, line 143: Same plan as we have for LB / TG configs? If so, it might be beneficial to just knock the logic out in this PR.

pkg/gateway/routeutils/descriptor.go
shraddhabang: I prefer to keep the local object reference refs here for eventhandling since we may not need to load and attach that crd to the route if its not impacting any gateway. I plan to load the actual CRDs and attach those to `RouteRule` when we load attached rules since its more connected to rule than routes. https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/gateway/routeutils/loader.go#L73

config/rbac/role.yaml
shraddhabang: I wanna keep them separate so its easier to review with small changes. :)

pkg/gateway/routeutils/tcp.go
shraddhabang: Good point. Fixed it.

controllers/gateway/eventhandlers/listener_rule_configuration_events.go
shraddhabang: No fixed it.

PR #4275, opened 2025-07-24
[feat: gw api] Add finalizers  for Rules CRD

shraddhabang:
### Description

Following the same strategy as of tgcfg and lbcfg to add the finalizers for rules CRD.
We will only check if the rules CRD is being referred by any L7 routes. If not, we remove the finalizers on the CRD. 

### Checklist
- [ ] Manually tested


pkg/gateway/constants/controller_constants.go
```diff
@@ -63,4 +63,7 @@ const (
 
 	//TargetGroupConfigurationController the controller that reconciles TargetGroupConfiguration changes
 	TargetGroupConfigurationController = "aws-lbc-targetgroupconfiguration-controller"
+
+	//ListenerRuleConfigurationController the controller that reconciles TargetGroupConfiguration changes
```
shuqz: nit: ListenerRuleConfigurationController the controller that reconciles ListenerRuleConfiguration changes

pkg/gateway/routeutils/utils.go
```diff
@@ -33,7 +32,7 @@ func ListL4Routes(ctx context.Context, k8sClient client.Client) ([]preLoadRouteD
 	}
 	l4Routes = append(l4Routes, tlsRoutes...)
 	if len(failedRoutes) > 0 {
-		err = errors.Errorf("failed to list L4 routes, %s", failedRoutes)
+		err = fmt.Errorf("failed to list L4 routes, %v", failedRoutes)
```
shuqz, line 33: why do we make this change?

pkg/gateway/routeutils/listener_rule_config_utils.go
```diff
@@ -0,0 +1,17 @@
+package routeutils
+
+import (
+	"context"
+	elbv2gw "sigs.k8s.io/aws-load-balancer-controller/apis/gateway/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func IsListenerRuleConfigInUse(ctx context.Context, listenerRuleConfig *elbv2gw.ListenerRuleConfiguration, k8sClient client.Client) (bool, error) {
```
shuqz, line 11: i felt this one does not need a separate utils file especially when it is also using FilterRoutesByListenerRuleCfg from utils.go and it is just one function. i've rather keep it in utils.go (a centralized place to look for when needed) so it is easier to know this function exists.

pkg/gateway/routeutils/utils.go
shraddhabang: Ohh it was printing our very verbose error which was very difficult to read. This simplifies it.

pkg/gateway/routeutils/listener_rule_config_utils.go
shraddhabang: Actually the route utils was getting bigger. So I now have a dedicated utils for listener rule cfgs. It will be easier to add more if needed for listener rule cfgs. I also have moved the tests for better management.

Issue #4262, opened 2025-07-11
WAF/Shield support for Gateway API implementation


timothy-spencer:
**Describe the feature you are requesting**
I would like a way to configure a Gateway that provisions an ALB to have a WAF ACL associated with it, and so that Shield can be enabled as well.

**Motivation**
Ingresses can have annotations applied to them like
```
  annotations:
    alb.ingress.kubernetes.io/wafv2-acl-arn: arn:aws:wafv2:us-west-2:xxx:regional/webacl/xxx/xxx
    alb.ingress.kubernetes.io/shield-advanced-protection: true
```
The gateway stuff should be able to as well.  This will move the Gateway implementation towards full feature parity with Ingress.

**Describe the proposed solution you'd like**
I don't mind having annotations on the Gateway like they are on the Ingress, but I'm open to anything.

Ideally, there would be documentation too, so that this wouldn't be a hidden mystery feature.  :-)

**Describe alternatives you've considered**
There seems to be no workaround to this, though perhaps I could plug the WAF into the ALB by hand.  I haven't tried, because we want to have everything be done with code.

**Contribution Intention (Optional)**

-[ ] Yes, I am willing to contribute a PR to implement this feature
-[ ] No, I cannot work on a PR at this time
-[X] Maybe, I don't know the codebase very well, so if I can figure it out, I will.

sarevalo2002:
Thank you for reaching out to us! We'll put this on our roadmap for the Gateway API implementation.

zac-nixon:
Released this in v2.14.0

PR #4277, opened 2025-07-29
[Gateway API] Add Addon Support

zac-nixon:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4262

### Description

Adds the same add-on support to the Gateway API implementation that currently exists within the Ingress implementation. The idea is to add the addon configuration at the LB configuration level. Currently, we will support WAFv2 and Shield addons however the implementation is flexible enough to extend to other addons in the future. I didn't port over classic WAF as it's being deprecated ~9/30.

The implementation is rather simple, we track the enabled addons in the Gateway annotations, doing so let's us accurately detach addon resources from the LoadBalancer  when the addon is disabled. I also added some logic to detach addons from the LB as the first step in the deletion process. Some addons like Shield do not automatically detach themselves from deleted resources.

Logic break down:

1/ Calculate previous addon configuration from annotations.
2/ Calculate current addon configuration from LB CFG.
3/ In the model builder we determine if any addons were added / removed using the results from 1 & 2.
4/ For any new addons, write an annotation for that new addon. That tracks that we are enabling the addon.
5/ Materialize the new addon configuration in the deployer.
6/ For any addons that were enabled but were removed in step 5, remove that addon from the annotations.

### Checklist
- [x] Manually tested


controllers/gateway/addons.go
```diff
@@ -0,0 +1,114 @@
...
+// generateAddOnKey translates an addon into the respective annotation key value.
+func generateAddOnKey(a addon.Addon) string {
+	return fmt.Sprintf("%s%s", constants.GatewayLBPrefixEnabledAddon, strings.ToLower(string(a)))
+}
+
+// parseAddOnEnabledValue parses an annotation key value into a boolean, assuming false if the value is malformed.
+func parseAddOnEnabledValue(e string) bool {
+	b, err := strconv.ParseBool(e)
+	if err != nil {
+		// Assume corrupted value is false (todo - maybe error log?)
```
shraddhabang: A debug log would suffice here.

apis/gateway/v1beta1/loadbalancerconfig_types.go
```diff
@@ -140,6 +140,18 @@ type MutualAuthenticationAttributes struct {
...
+// ShieldConfiguration configuration parameters used to configure Shield
+type ShieldConfiguration struct {
+	// Enabled whether Shield Advanced should be configured with the Gateway
+	Enabled bool `json:"enabled,omitempty"`
+}
+
+// WAFv2Configuration configuration parameters used to configure WAFv2
+type WAFv2Configuration struct {
+	// ACL The WebACL to configure with the Gateway
+	ACL string `json:"webACL"`
```
shraddhabang, line 152: NIT: `WebACLArn string `json:"webACLARN`?
In future we may accept name if needed?

pkg/gateway/model/model_build_addon.go
```diff
@@ -0,0 +1,108 @@
...
+	return result, nil
+}
+
+func (aob *addOnBuilderImpl) buildWAFv2(stack core.Stack, lbARN core.StringToken, lbCfg elbv2gw.LoadBalancerConfiguration, previousAddonConfig []addon.Addon) (bool, error) {
+	var webACLARN string
+	if lbCfg.Spec.WAFv2 != nil {
+		webACLARN = lbCfg.Spec.WAFv2.ACL
+	}
+
+	aob.logger.Info("WAF ACL", "ACL", webACLARN)
```
shraddhabang: NIT: More descriptive log?

apis/gateway/v1beta1/loadbalancerconfig_types.go
zac-nixon: I think it would go under a different field name to reduce ambiguity.

pkg/gateway/model/model_build_addon.go
zac-nixon: just removed this log. not sure it was a good one to add lol

config/crd/gateway/gateway.k8s.aws_listenerruleconfigurations.yaml
```diff
@@ -1,383 +0,0 @@
----
```
shuqz, line 1: are those files being empty now? this question applies to lbConfig crd and tgConfig crd yaml files too

config/crd/gateway/gateway-crds.yaml
```diff
@@ -282,6 +282,15 @@ spec:
                 items:
                   type: string
                 type: array
+              shieldConfiguration:
+                description: ShieldAdvanced define the AWS Shield settings for a Gateway
+                  [Application Load Balancer]
+                properties:
+                  enabled:
```
shuqz, line 289: ~~if shieldConfiguration is given but no `enabled` is provided, what is our behavior? (because enabled is not required based on definition)~~ nvm, found it

controllers/gateway/addons.go
```diff
@@ -0,0 +1,116 @@
...
+func generateAddOnKey(a addon.Addon) string {
+	return fmt.Sprintf("%s%s", constants.GatewayLBPrefixEnabledAddon, strings.ToLower(string(a)))
+}
+
+// parseAddOnEnabledValue parses an annotation key value into a boolean, assuming false if the value is malformed.
+func parseAddOnEnabledValue(e string, logger logr.Logger) bool {
+	b, err := strconv.ParseBool(e)
+	if err != nil {
+		// Assume corrupted value is false (todo - maybe error log?)
+		logger.V(1).Info("Unknown boolean value", "val", e)
```
shuqz: maybe "Unknown boolean value, default it to false"

controllers/gateway/gateway_controller.go
```diff
@@ -223,12 +224,37 @@ func (r *gatewayReconciler) reconcileHelper(ctx context.Context, req reconcile.R
...
 	if err != nil {
 		return err
 	}
 
+	r.logger.V(1).Info("Got this addon config", "current", currentAddOns, "new addon", newAddOnConfig)
+
+	// To accurately track the set of enabled addons, we need to figure out if any addons were added / removed during this run.
+	addOnAdditions, addOnRemovals := diffAddOns(currentAddOns, newAddOnConfig)
+
+	// To handle addon additions, we need to write the latest add on config back to the Gateway before actually materializing the addon.
```
shuqz, line 248: what happens when annotation attached successfully, but final materialize failed for some reason, do we remove the attached annotation? same thing applies to remove addon annotation, is there any way we can make it in a pattern that succeed or fail together? e.g. add annotation + materialization -> if any fails, all fails, what's done will be undo

pkg/addon/addon.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+var (
+	AllAddons = []Addon{WAFv2, Shield}
+)
+
+type AddonMetadata struct {
+	Name    Addon
+	Enabled bool
+	// Potentially other metadata? //
```
shuqz: this can be removed

pkg/gateway/constants/controller_constants.go
```diff
@@ -66,4 +66,7 @@ const (
 
 	//ListenerRuleConfigurationController the controller that reconciles ListenerRuleConfiguration changes
 	ListenerRuleConfigurationController = "aws-lbc-listenerruleconfiguration-controller"
+
+	// GatewayLBPrefixEnabledAddon The prefix tracking addons within the Gateway.
+	GatewayLBPrefixEnabledAddon = "gateway.k8s.aws.addon."
```
shuqz, line 71: extra dot at the end?

pkg/gateway/model/base_model_builder.go
```diff
@@ -109,64 +111,82 @@ type baseModelBuilder struct {
...
-		return nil, nil, false, err
+		return nil, nil, nil, false, err
+	}
+
+	addOnCfg := lbConf
+	if isPreDelete {
+		addOnCfg = elbv2gw.LoadBalancerConfiguration{}
+	}
+
+	currentAddonConfig, err := baseBuilder.addOnBuilder.buildAddons(stack, lb.LoadBalancerARN(), addOnCfg, previousAddonConfig)
```
shuqz: do you think we can be consistent on naming? cuz i see we have "currentAddon, newAddon" and here we are using "addon, previousAddOn"

config/crd/gateway/gateway.k8s.aws_listenerruleconfigurations.yaml
zac-nixon: I am not sure if these should be included or not? The past couple commits I have done they get added / remove. Should they be committed to the repository?

controllers/gateway/gateway_controller.go
zac-nixon: if you know how to transactionally have two different systems commit changes I am open to hearing about it (You might also have a billion dollar+ company idea :)) 

The general idea here is also used in the MultiCluster support. We have two different systems that need to know about the state of the data. Consider if we DO NOT write the annotation first:

1. Materialize addon
2. Initial annotation write fails.
3. User removes addon
<We have now orphaned the add on>

Consider what happens when we write first:

1. Initial addon annotation update fails
2. User removes addon
<add on not orphaned>

Same logic applies the other way around too, for removal of addon data.

pkg/addon/addon.go
zac-nixon: ok

pkg/gateway/constants/controller_constants.go
zac-nixon: not extra. it's the prefix, see how it gets used later on

pkg/gateway/model/base_model_builder.go
zac-nixon: sure i can make the naming consistent

shraddhabang (approved):
/lgtm

controllers/gateway/gateway_controller.go
shuqz: lol i am over-simplifying things, somehow i was thinking about other designs i saw that will populate message to SQS in similar case and make all events fail or success. but that requires human intervention and only guarantee eventual consistency

Issue #4280, opened 2025-07-31
Allow targetGroupName instead of targetGroupARN in forward action ingress annotations


pascal-hofmann:
**Describe the feature you are requesting**
Allow specifying a `targetGroupName` (instead of `targetGroupARN`) in forward action ingress annotations.

**Motivation**
Currently we need to set the correct target group ARN depending on which of our environments we are deploying to. If we could just specify the target group name instead, we would not have to differentiate per env, as the target groups have the same name in all our envs.

**Describe the proposed solution you'd like**
`targetGroupName` should be added to the `TargetGroupTuple` struct, and be used to determine the `targetGroupARN`, in case it is not set. This is the same behavior as for TargetGroupBindings: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/aefed36aa30528e35a73281a5dc131ed26caa0f4/webhooks/elbv2/targetgroupbinding_mutator.go#L68-L74

**Describe alternatives you've considered**

--

**Contribution Intention (Optional)**

- [x] Yes, I am willing to contribute a PR to implement this feature
- [ ] No, I cannot work on a PR at this time

PR #4281, opened 2025-07-31
feat: allow targetGroupName instead of targetGroupARN in forward action ingress annotations

pascal-hofmann:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4280

### Description

Allow targetGroupName instead of targetGroupARN in forward action ingress annotations

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/ingress/model_build_actions.go
+++ b/pkg/ingress/model_build_actions.go
@@ -4,12 +4,9 @@ import (
 	"context"
 	"fmt"
 	awssdk "github.com/aws/aws-sdk-go-v2/aws"
-	elbv2sdk "github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2"
-	elbv2types "github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2/types"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/aws-load-balancer-controller/pkg/aws/services"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/model/core"
 	elbv2model "sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2"
 	"strings"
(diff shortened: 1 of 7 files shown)


pascal-hofmann:
Hi @M00nF1sh, hi @oliviassss, hi @wweiwei-li,
it would be great to get this reviewed / merged soon. It would save us quite some manual fiddling with different ARNs in different environments.

Thanks in advance,
Pascal

pkg/ingress/model_build_actions.go
```diff
@@ -105,6 +108,12 @@ func (t *defaultModelBuildTask) buildForwardAction(ctx context.Context, ing Clas
 		var tgARN core.StringToken
 		if tgt.TargetGroupARN != nil {
 			tgARN = core.LiteralStringToken(*tgt.TargetGroupARN)
+		} else if tgt.TargetGroupName != nil {
```
zac-nixon, line 108: This will be pretty expensive to run, every reconcile cycle would have to resolve the arn from the tg name. Is your use case that you're deleting / re-creating the TG constantly?


I think you'll want to attach some TG name -> TG ARN cache so we're not continuously resolving the TG ARN.

zac-nixon:
Thanks for the contribution! I left a comment around some caching which I would consider very important. The LBC already makes a lot of ELB API calls, it would be great to make sure this PR doesn't balloon the call volume.

pkg/ingress/model_build_actions.go
pascal-hofmann: Yeah, that makes sense. I’ll have a look on how caching is done elsewhere in the controller and implement a similar cache for the target group name/ARN resolution.

pascal-hofmann: Hi @zac-nixon,
I added caching in a separate commit. Give me a shout whether this is ok, or you imagined a different, simpler solution (like `tgByResID`?).

Cheers
Pascal

PS: In my use-case the target group is not deleted/created constantly. We have multiple target group with given names in multiple AWS accounts and want to be able to reference them just by name for simplicity.

zac-nixon: Looks great. Thanks for the making the change.

zac-nixon (approved):
/approved
/lgtm

Thank you! This looks like a great addition.

pascal-hofmann:
Thanks for the quick feedback and merge. This was a very pleasant pull request experience. 🚀

PR #4282, opened 2025-08-01
Add TG protocol into TGB

zac-nixon:
### Description

Adds TG protocol in TGB for an upcoming project. Also refactors the webhook logic to only look up TG once per invocation, previously we would re-fetch the TG from ELB for every validation.

Scenarios tested:

Deploy new LBC, don't update TGB CRD.

- E2E tests still work.
- TGB operations still work.

Deploy old LBC, update TGB CRD.

- E2E tests still work
- TGB operations still work (protocol is just nil)

Deploy new LBC, update TGB CRD.

- E2E tests work.
- Tested without inserting protocol, it gets auto populated.
- Ingress / Service / Gateway API all auto populate protocol.
- Manually created TGB rejects unknown protocol types.
- Manually created TGB rejects protocol types when it doesn't matched the protocol stored within ELB.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/deploy/elbv2/target_group_binding_manager.go
+++ b/pkg/deploy/elbv2/target_group_binding_manager.go
@@ -15,7 +15,7 @@ import (
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/annotations"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/deploy/tracking"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/k8s"
-	elbv2model "sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2/k8s"
+	elbv2modelk8s "sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2/k8s"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"time"
 )
@@ -29,9 +29,9 @@ const (
 
 // TargetGroupBindingManager is responsible for create/update/delete TargetGroupBinding resources.
 type TargetGroupBindingManager interface {
-	Create(ctx context.Context, resTGB *elbv2model.TargetGroupBindingResource) (elbv2model.TargetGroupBindingResourceStatus, error)
+	Create(ctx context.Context, resTGB *elbv2modelk8s.TargetGroupBindingResource) (elbv2modelk8s.TargetGroupBindingResourceStatus, error)
 
-	Update(ctx context.Context, resTGB *elbv2model.TargetGroupBindingResource, k8sTGB *elbv2api.TargetGroupBinding) (elbv2model.TargetGroupBindingResourceStatus, error)
+	Update(ctx context.Context, resTGB *elbv2modelk8s.TargetGroupBindingResource, k8sTGB *elbv2api.TargetGroupBinding) (elbv2modelk8s.TargetGroupBindingResourceStatus, error)
 
 	Delete(ctx context.Context, k8sTGB *elbv2api.TargetGroupBinding) error
 }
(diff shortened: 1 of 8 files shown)


pkg/gateway/model/mock_tg_builder.go
```diff
@@ -6,6 +6,7 @@ import (
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/gateway/routeutils"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/model/core"
 	elbv2model "sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2/k8s"
```
shuqz, line 9: let's make this consistent: `sigs.k8s.io/aws-load-balancer-controller/pkg/model/elbv2/k8s` i saw it was used as `elbv2model`, `k8s2` and `k8s`

webhooks/elbv2/targetgroupbinding_mutator.go
```diff
@@ -152,8 +199,10 @@ func (m *targetGroupBindingMutator) getTargetGroupIPAddressTypeFromAWS(ctx conte
 	return ipAddressType, nil
 }
 
-func (m *targetGroupBindingMutator) getVpcIDFromAWS(ctx context.Context, tgb *elbv2api.TargetGroupBinding) (string, error) {
-	targetGroup, err := getTargetGroupFromAWS(ctx, m.elbv2Client, tgb)
+func (m *targetGroupBindingMutator) getVpcIDFromAWS(tgCache func() tgCacheObject) (string, error) {
+	obj := tgCache()
```
shuqz, line 203: looks like for all getXXXFromAWS functions, we are doing 
```
obj := tgCache()
targetGroup := obj.tg
err := obj.error
if err != nil {
		return "", err
	}
```
then maybe instead of pass in `tgCache func() tgCacheObject`, can we check error right after 
```
	targetGroupCache := sync.OnceValue(func() tgCacheObject {
		targetGroup, err := getTargetGroupFromAWS(ctx, m.elbv2Client, tgb)
		return tgCacheObject{
			tg:    targetGroup,
			error: err,
		}
	})
```
and then pass targerGroup: `tgCache().tg`

pkg/gateway/model/mock_tg_builder.go
zac-nixon: sounds good. i was getting confused about the various names assigned. :)

webhooks/elbv2/targetgroupbinding_mutator.go
zac-nixon: the idea was that we only want to call at most one time. in the current setup we only call `getTargetGroupFromAWS` when we need to, which means we need to have each invocation be prepared to handle an error. it sounds like your suggestion would be to just call `getTargetGroupFromAWS` and pass the value around? 

lmk if i misunderstood.

shuqz: for functions getVpcIDFromAWS, getTargetGroupIPAddressTypeFromAWS, obtainSDKTargetGroupProtocolFromAWS etc, this part of code is in all of them 
```
obj := tgCache()
targetGroup := obj.tg
err := obj.error
if err != nil {
		return "", err
	}
```
because we passed targetGroupCache in `m.defaultingIPAddressType(tgb, targetGroupCache)`, i am thinking if we can check cache error before calling all those m.defaultXXXX functions and just pass `m.defaultingIPAddressType(tgb, targetGroupCache.tg)` 
in this way, we only need to check err once and remove duplicated codes, let me know if this applies

zac-nixon: ah ok. i did understand what you were saying. I don't think we can do that, if we want to maintain the minimum amount of calls.

We only want to fetch the Target Group if we need to, so in this case we need each function to handle the error. For example, we might not need to fetch the TG prior to validating the TG VPC, if that call fails we need to have error handling.

shuqz: oh i see your point now, thanks

Issue #2311, opened 2021-10-21
Replace ALB annotations with IngressClassParams


stevehipwell:
**Is your feature request related to a problem?**
I'd like to be able to replace more of the `alb.ingress.kubernetes.io` annotations with the `IngressClassParams` to abstract away the ALB specification from the `Ingress` resources using it. I mistakenly thought that #2190 was doing this, but after re-reading I think that's just replacing the `alb.ingress.kubernetes.io/load-balancer-attributes` annotation.

**Describe the solution you'd like**
I'd like to be able to specify the following annotations in the `IngressClassParams` directly.

- alb.ingress.kubernetes.io/target-type
- alb.ingress.kubernetes.io/subnets
- alb.ingress.kubernetes.io/listen-ports
- alb.ingress.kubernetes.io/ssl-redirect
- alb.ingress.kubernetes.io/inbound-cidrs
- alb.ingress.kubernetes.io/certificate-arn

**Describe alternatives you've considered**
n/a

nikskiz:
It would be nice to have `spec.params.targetGroupAttributes`

Nuru:
Really, I think any of the 14 annotations that are "Exclusive" (must only be set once per Group) should be configurable via IngressClass (possibly via IngressClassParams). Then we can move that configuration out of the individual services so we do not have to worry about them fighting over control of them. 

As of version 2.4, the following annotations are marked "Exclusive", but only a few of them can be specified in IngressClass and most (all?) of the rest cannot be specified anywhere but via annotations. This is a poor separation of concerns.

1. [alb.ingress.kubernetes.io/load-balancer-name](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/annotations/#load-balancer-name)
2. [alb.ingress.kubernetes.io/ip-address-type](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/annotations/#ip-address-type)
3. [alb.ingress.kubernetes.io/scheme](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/annotations/#scheme)
4. [alb.ingress.kubernetes.io/subnets](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/annotations/#subnets)
[...]

Issue #2600, opened 2022-04-05
Provide more defaults via IngressClassParams


visit1985:
**Is your feature request related to a problem?**
Even after introducing `IngressClass`/`IngressClassParams` we have a lot of redundant annotations in our Ingress definitions for the same Ingress group:
- `alb.ingress.kubernetes.io/actions.ssl-redirect`
- `alb.ingress.kubernetes.io/certificate-arn`
- `alb.ingress.kubernetes.io/listen-ports`
- ~~`alb.ingress.kubernetes.io/ssl-policy`~~

**Describe the solution you'd like**
We would like to specify the defaults for these annotations as `IngressClassParams`.

zmully:
I'd add to this:
- `alb.ingress.kubernetes.io/wafv2-acl-arn`

Additionally, updating `IngressClassParams` appears to only affect _new_ ALBs, not existing ALBs. While I can understand the reasoning behind this, having changes to the `IngressClassParams` optionally trigger updates to all ALBs managed by the controller would be helpful. Especially for users of ALBs using `group.name` tagging, trying to update ALB shared attributes via application ingress annotations is fraught and unpredictable.

kishorj:
We will put it in the roadmap, this requires further design to incorporate the attributes in the ingress class params.

nikskiz:
Similar issue: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2311

visit1985:
`ssl-policy` can be set now: https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.5/guide/ingress/ingress_class/#specsslpolicy

omri-shilton:
is there an update on this issue? specifically for waf purposes

franklouwers:
Note that some of the annotations listed above, don't even make sense on an individual ingress level.

Eg: If one of the ingresses in an IngressGroup set `alb.ingress.kubernetes.io/ssl-redirect`, then it's impossible to have another Ingress to disable the redirect for that individual Ingress. It applies on the group level, not on the individual level.

By lack of an IngressGroup object, the logical sense to configure this would be on the IngressClassParams.

Besides that, I agree: we should be able to configure defaults for any annotation which can be set on an Ingress, in an IngressClassParams level.

jglick:
Has this been obsoleted by the Gateway API?

1ms-ms:
[...]

PR #4290, opened 2025-08-04
feature: add load balancer name to IngressClassParams

1ms-ms:
### Issue

Entry point to https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2311. If accepted, we will need to add `WAF arn/name` to `IngressClassParams` spec as well.
### Description
Currently, to provision load balancer with a hard coded name (that is a name which isnt generated from `group.name`), we need to use [alb.ingress.kubernetes.io/load-balancer-name](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/ingress/annotations/#load-balancer-name) annotation, added to at least one ingress resource. When I create an `IngressClass`, I wish I could specify the load balancer name as part of `IngressClassParams` spec and not care about the lifecycle of the `load-balancer-name` annotation.

### Checklist
- [ ] Manually tested

### DIFF
--- a/apis/elbv2/v1beta1/ingressclassparams_types.go
+++ b/apis/elbv2/v1beta1/ingressclassparams_types.go
@@ -116,6 +116,10 @@ type IPAMConfiguration struct {
 
 // IngressClassParamsSpec defines the desired state of IngressClassParams
 type IngressClassParamsSpec struct {
+	// LoadBalancerName defines the name of the load balancer that will be created with this IngressClassParams.
+	// +optional
+	LoadBalancerName string `json:"loadBalancerName,omitempty"`
+
 	// CertificateArn specifies the ARN of the certificates for all Ingresses that belong to IngressClass with this IngressClassParams.
 	// +optional
 	CertificateArn []string `json:"certificateArn,omitempty"`
--- a/config/crd/bases/elbv2.k8s.aws_ingressclassparams.yaml
+++ b/config/crd/bases/elbv2.k8s.aws_ingressclassparams.yaml
@@ -148,6 +148,10 @@ spec:
                   - value
                   type: object
                 type: array
+              loadBalancerName:
+                description: LoadBalancerName defines the name of the load balancer
+                  that will be created with this IngressClassParams.
+                type: string
               minimumLoadBalancerCapacity:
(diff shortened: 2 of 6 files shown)


zac-nixon:
Hey, thanks for the contribution. I don't quite follow what problem this is solving though. A load balancer name is unique within a region / account, meaning that that the ingress class param would be a 1:1 mapping to ingress resources. I don't think it makes sense to add the LB name to the ingress class params.

1ms-ms:
Hi @zac-nixon, thanks for the quick reply. This PR is motivated by the fact that I couldn't find a satisfactory way to provision an ALB with a specific name using this controller.

> A load balancer name is unique within a region/account, meaning that the IngressClassParam would be a 1:1 mapping to ingress resources.

This is exactly the reason for this PR. I manage 30 ingresses and want to logically group and bind them to a single ALB. The best option currently is using `IngressClassParams` since there is no `IngressGroup` or similar CRD. Other than that, out of all the ideas I had, only two made it to a somewhat standard approach:

- Create an `init Ingress`, which will act as a placeholder for common settings like the ALB name, WAF ARN, etc. With this solution, I wouldn’t need to specify annotations in the other ingresses.

- Use the same `load-balancer-name annotation` for each ingress I create, which leverages the same ALB.

I don’t like either of these options. I have reviewed other issues like https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2600 and the one I linked earlier (https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2311), and the approach I propose seems to be mentioned by other users as well.

In the end, I believe this PR wont change anything for people who use the annotation for ALB name, but introduces flexibility for people like me, who try to group ingresses and have cleaner setup with `IngressClassParams`/`IngressClass`.

EDIT: doesnt [spec.group](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/ingress/ingress_class/#specgroup) imply a singular ALB as well?

zac-nixon:
hmm, great point about the existing ingress group name logic. I honestly can't think of a use-case to specify that, but since we have a precedence already set and this will help your use-case, I am happy to approve it :D

PR #4298, opened 2025-08-08
[feat:gw api] Add tg stickiness and fixed response

shuqz:
### Description
This PR split into 2 commits, built-on previous commit regarding CRD loader
#### New feature implemented: 
- target group stickiness support 
- fixed response action support 
- specify query in listenerRuleConfiguration and add it to redirect action
#### Refactor: 
- added gateway.k8s.aws_listenerruleconfigurations.yaml to Kustomization and it is now included in gateway-crd yaml
- move rule action related functions to a new file (also moved unit tests)
- corrected some typo
#### TO-DO: 
Need to backfil E2E tests but since this PR is large enough, going to add it later. Manual tests are performed. See below for details.
### Checklist
- [x] Manually tested
#### Fixed response 
```
---
metadata:
  name: custom-rule-config-fixed-response
  namespace: sample-gateway-alb
spec:
  actions:
    - type: fixed-response
      fixedResponseConfig:
        statusCode: 200
        contentType:  -> did not specify contentType, it will take default value
        messageBody: "fixed-response-crd-test-body"
```
#### tg stickiness (enabled)
```
---
metadata:
  name: custom-rule-config
  namespace: sample-gateway-alb
spec:
  actions:
    - type: forward
      forwardConfig:
        targetGroupStickinessConfig:
          enabled: true 
          -> did not specify durationSeconds, will take default value 3600
````
#### tg stickiness (empty)
```
---
metadata:
  name: custom-rule-config
  namespace: sample-gateway-alb
spec:
  actions:
    - type: forward
      forwardConfig:
        targetGroupStickinessConfig:
           -> did not specify enabled, will take default value: false
````
#### Redirect
```
---
metadata:
  name: custom-rule-config-redirect
  namespace: sample-gateway-alb
spec:
  actions:
    - type: redirect
      redirectConfig:
        query:

---
metadata:
  name: custom-rule-config-redirect-specified
  namespace: sample-gateway-alb
spec:
  actions:
    - type: redirect
      redirectConfig:
        query: "redirect-test-query"
---
kind: HTTPRoute
metadata:
  name: http-app-redirect-query
  namespace: sample-gateway-alb
spec:
  rules:
    - filters:
        - type: ExtensionRef
          extensionRef:
            group: "gateway.k8s.aws"
            kind: "ListenerRuleConfiguration"
            name: "custom-rule-config-redirect"
        - type: RequestRedirect # request redirect cannot be together with backendRefs
          requestRedirect:
            path:
              type: ReplaceFullPath
              replaceFullPath: "/newExactPath"
            statusCode: 301
            scheme: https
---
kind: HTTPRoute
metadata:
  name: http-app-redirect-query-2
  namespace: sample-gateway-alb
spec:
  rules:
    - filters:
        - type: ExtensionRef
          extensionRef:
            group: "gateway.k8s.aws"
            kind: "ListenerRuleConfiguration"
            name: "custom-rule-config-redirect-specified"
        - type: RequestRedirect # request redirect cannot be together with backendRefs
          requestRedirect:
            path:
              type: ReplaceFullPath
              replaceFullPath: "/newExactPath2"
            statusCode: 301
            scheme: https
```
#### Result 
Screenshot 2025-08-08 at 4 13 30 PM

### CODE REVIEW

File: apis/gateway/v1beta1/listenerruleconfig_types.go [hunk 1]
```diff
@@ -81,7 +81,7 @@ type TargetGroupStickinessConfig struct {
 
 // Information about a forward action.
 type ForwardActionConfig struct {
-
+	// +kubebuilder:default={}
```
shuqz, line 84: ref: If you have a pointer to a struct, and you want to ensure that the nested fields within that struct are defaulted even if the outer pointer is nil, you can apply +kubebuilder:default={} to the pointer field itself. This will cause the pointer to be initialized to an empty instance of the struct, allowing the nested fields' defaults to then be applied.

this is needed for enabled and durationSecond to have default value for case 
```
metadata:
  name: custom-rule-config
  namespace: sample-gateway-alb
spec:
  actions:
    - type: forward
      forwardConfig:
        targetGroupStickinessConfig:
```

zac-nixon: How does our source code know if the user didn't specify the ForwardActionConfig (e.g. if it's nil)

pkg/gateway/model/model_build_listener.go
```diff
@@ -202,9 +202,12 @@ func (l listenerBuilderImpl) buildListenerRules(stack core.Stack, ls *elbv2model
...
 		}
-		tags, tagsErr := l.tagHelper.getGatewayTags(lbCfg)
-		if tagsErr != nil {
-			return tagsErr
+
+		// set up for building routing actions
+		var actions []elbv2model.Action
+		var routingAction *elbv2gw.Action
+		if rule.GetListenerRuleConfig() != nil {
+			routingAction = getRoutingAction(rule.GetListenerRuleConfig())
```
zac-nixon, line 210: Looks like we can append a nil value here?

zac-nixon: To clarify `getRoutingAction` can return nil if the action is mis-configured.

pkg/gateway/routeutils/route_rule_action.go
```diff
@@ -0,0 +1,222 @@
...
+		}
+		if redirectActions != nil {
+			actions = redirectActions
+		}
+	}
+	return actions, nil
+}
+
+func buildFixedResponseRoutingActions(fixedResponseConfig *elbv2gw.FixedResponseActionConfig) ([]elbv2model.Action, error) {
+	if fixedResponseConfig != nil {
```
shraddhabang: `fixedResponseConfig` can never be `nil` if the the action type is `FixedResponse`. We don't allow this from CRD.

pkg/gateway/routeutils/route_rule_action.go
```diff
@@ -0,0 +1,222 @@
...
+	if routingAction != nil && routingAction.Type == elbv2gw.ActionTypeFixedResponse {
+		fixedResponseActions, err := buildFixedResponseRoutingActions(routingAction.FixedResponseConfig)
+		if err != nil {
+			return nil, err
+		}
+		if fixedResponseActions != nil {
+			actions = fixedResponseActions
+		}
+	} else {
+		// Build Rule Routing Actions - Forward
```
shraddhabang, line 28: Dont we need to check if the routing action is  of forward type before building the ForwardRoutingActions? What is the routing action is of redirect type?

pkg/gateway/routeutils/http_test.go
```diff
@@ -167,9 +205,13 @@ func Test_HTTP_LoadAttachedRules(t *testing.T) {
 	assert.Equal(t, 2, len(convertedRules[0].GetBackends()))
 	assert.Equal(t, 4, len(convertedRules[1].GetBackends()))
 	assert.Equal(t, 0, len(convertedRules[2].GetBackends()))
+
+	assert.NotNil(t, convertedRules[0].GetListenerRuleConfig())
```
zac-nixon: nit: it would be better to assert what you expect the config to be rather if it's nil or not (i have this as a nit because im pretty sure i also did this same thing :D)

pkg/gateway/routeutils/route_rule_action.go
```diff
@@ -0,0 +1,222 @@
...
+		forwardActions, err := buildForwardRoutingActions(routingAction, targetGroupTuples)
```
zac-nixon, line 29: The expectation is that the CRD makes sure we only have one of these defined. It would be good to note that here.

apis/gateway/v1beta1/listenerruleconfig_types.go
shuqz: when type is forward, forwardConfig must be provided, otherwise it won't pass kubebuilder validation
```
// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'forward' ? has(self.forwardConfig) : !has(self.forwardConfig)",message="forwardConfig must be specified only when type is 'forward'"
```

pkg/gateway/model/model_build_listener.go
shuqz: yes, and in this case in BuildRuleRoutingActions it will try to create actions based on whats' defined it route (forward or redirect) and if none is specified , so it will fall back to create 503 action at the end

pkg/gateway/routeutils/route_rule_action.go
shuqz: you mean they define some forward action in route matches and then a redirect Type in rule CRD? redirect action requires statusCode to be specified, that has to be specified through route filter. and if filter-redirect, no backends can be added (it will fail kube builder validation) so there won't be such case that they define some forward action in route matches and then a redirect Type in rule CRD

pkg/gateway/model/model_build_listener.go
zac-nixon: But don't we need to worry about nil pointers? IMO, we shouldn't add nil to the list.

pkg/gateway/routeutils/route_rule_action.go
shraddhabang: Got it.

pkg/gateway/model/model_build_listener.go
shuqz: which list are you referring to? this routingAction is used in BuildRuleRoutingActions and why would it add nil to list?

shuqz: synced offline

pkg/gateway/routeutils/route_rule_action.go
shuqz: updated

pkg/gateway/routeutils/http_test.go
shuqz: updated test

PR #4309, opened 2025-08-14
[gw api] Fix overwrite of route status

zac-nixon:
### Description

There is a possibility for the Gateway reconciler to enter an infinite loop when there's a partially working route. The flow is that a route gets it's status updated to accepted = false, however the previous implementation would always set all route statuses that got loaded to accepted = true. This behavior would cause the status to get overwritten and cause another gateway reconcile loop to get kicked off.

I approached the fix by refactoring the code to do route status updates in one place, and removed the route reconciler injection in other places and centralized the reconciler enqueue in the route mapper. By centralizing the enqueue logic, I was able to implement dedupe logic to prevent one reconcile cycle from writing a route status more than once.

### Checklist
- [x] Manually tested

### DIFF
--- a/config/crd/bases/elbv2.k8s.aws_ingressclassparams.yaml
+++ b/config/crd/bases/elbv2.k8s.aws_ingressclassparams.yaml
@@ -11,7 +11,7 @@ spec:
     kind: IngressClassParams
     listKind: IngressClassParamsList
     plural: ingressclassparams
-    singular: ingressclassparams
+    singular: ingressclassparam
   scope: Cluster
   versions:
   - additionalPrinterColumns:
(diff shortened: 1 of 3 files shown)


config/crd/bases/elbv2.k8s.aws_ingressclassparams.yaml
```diff
@@ -11,7 +11,7 @@ spec:
     kind: IngressClassParams
     listKind: IngressClassParamsList
     plural: ingressclassparams
-    singular: ingressclassparam
+    singular: ingressclassparams
```
shuqz: why make singular same as plural? i thought there was a PR several days ago to fix it: https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/3174f96d3867d1adbbd35028a08dd98a84ffeb5c

PR #4312, opened 2025-08-19
feature: get waf arn from name

1ms-ms:
### Description
Adding a possibility to specify WAFv2 with name annotation/ingressclassparam as another possiblity to build the WAFv2 association

I would like to provide WAFv2 name instead of ARN and have a function `getArnByName` to retrieve the ARN for me. In my implementation `web acl name` annotation/ingressclassparam takes priority over `web acl arn` annotation/ingressclassparam (although you should carefully review the logic for that as I may have missed some edge case not covered by tests I added or there may be a better way to write these conditions)
### Checklist
- [ ] Manually tested

### DIFF
--- a/docs/guide/ingress/ingress_class.md
+++ b/docs/guide/ingress/ingress_class.md
@@ -400,4 +400,4 @@ When this param is absent or empty, the controller will keep LoadBalancer WAFv2 
 Cluster administrators can use the optional `wafv2AclName` field to specify name of the Amazon WAFv2 web ACL.
 Only Regional WAFv2 is supported.
 When this param is absent or empty, the controller will keep LoadBalancer WAFv2 settings unchanged. To disable WAFv2, explicitly set the param value to 'none'.
-You cannot specify 2 different web acl names in 2 different ingresses belonging to the same `IngressGroup`
+If the field is specified, LBC will ignore the 'alb.ingress.kubernetes.io/wafv2-acl-name' annotation.
--- a/pkg/ingress/model_build_load_balancer_addons.go
+++ b/pkg/ingress/model_build_load_balancer_addons.go
@@ -35,7 +35,7 @@ func (t *defaultModelBuildTask) buildLoadBalancerAddOns(ctx context.Context, lbA
 func (t *defaultModelBuildTask) buildWAFv2WebACLAssociation(ctx context.Context, lbARN core.StringToken) (*wafv2model.WebACLAssociation, error) {
 	explicitWebACLARNs := sets.NewString()
 	explicitWebACLNames := sets.NewString()
-	webACLARN := ""
+
 	for _, member := range t.ingGroup.Members {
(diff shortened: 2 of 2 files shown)


docs/guide/ingress/ingress_class.md
```diff
@@ -384,4 +393,11 @@ Cluster administrators can use `Listeners` field to specify the [Listener Attrib
...
 Only Regional WAFv2 is supported.
-When this annotation is absent or empty, the controller will keep LoadBalancer WAFv2 settings unchanged. To disable WAFv2, explicitly set the annotation value to 'none'.
+When this param is absent or empty, the controller will keep LoadBalancer WAFv2 settings unchanged. To disable WAFv2, explicitly set the annotation value to 'none'.
+
+#### spec.wafv2AclName
+
+Cluster administrators can use the optional `wafv2AclName` field to specify name of the Amazon WAFv2 web ACL.
+Only Regional WAFv2 is supported.
+When this param is absent or empty, the controller will keep LoadBalancer WAFv2 settings unchanged. To disable WAFv2, explicitly set the param value to 'none'.
+You cannot specify 2 different web acl names in 2 different ingresses belonging to the same `IngressGroup`
```
shraddhabang: We always use IngressClassParams to resolve conflict between the ingresses within same group. So instead of saying `You cannot specify 2 different web acl names in 2 different ingresses belonging to the same `IngressGroup` we should say that the `If the field is specified, LBC will ignore the `alb.ingress.kubernetes.io/wafv2-acl-name` annotation.` This aligns with our other fields.

pkg/ingress/model_build_load_balancer_addons.go
```diff
@@ -32,26 +32,59 @@ func (t *defaultModelBuildTask) buildLoadBalancerAddOns(ctx context.Context, lbA
...
+			explicitWebACLNames.Insert(rawWebACLName)
 		}
 	}
-	if len(explicitWebACLARNs) == 0 {
-		return nil, nil
+
+	rawWebACLARN := ""
+	if len(explicitWebACLNames) == 0 {
+		for _, member := range t.ingGroup.Members {
+			_ = t.annotationParser.ParseStringAnnotation(annotations.IngressSuffixWAFv2ACLARN, &rawWebACLARN, member.Ing.Annotations)
```
shraddhabang: Sorry I know this is not your change, but this behavior of web acl arn parsing conflicts with our our other spec params. Usually, we the spec params values should be used instead of annotations values if they are present. Can you change please this behavior to match web acl names behavior? If not, I will make another PR for it.

1ms-ms: @shraddhabang I have fixed it to align with behavior of other ingressclassparams/annotations.

shraddhabang (approved):
/lgtm

pkg/ingress/model_build_load_balancer_addons.go
shraddhabang: Thank you. :)

PR #4316, opened 2025-08-21
fix http / grpc route rule generation

zac-nixon:
### Description

When no matches are given to for a HTTP / GRPC Route Rule, the controller doesn't generate any ALB rules. This behavior is incorrect as per the Gateway API documentation:

For GRPC routes -
```
	// If no matches are specified, the implementation MUST match every gRPC request.
```

For HTTP routes -
```
	// If no matches are specified, the default is a prefix
	// path match on "/", which has the effect of matching every
	// HTTP request.
```

This PR implements the logic as specified ^^^. I also added unit tests to validate the change worked. I was able to validate my GRPC set up now works with this change.

### Checklist
- [x] Manually tested


pkg/gateway/routeutils/route_rule_precedence.go
```diff
@@ -69,7 +70,22 @@ func SortAllRulesByPrecedence(routes []RouteDescriptor) []RulePrecedence {
...
+					common := routeInfo
+					common.Rule = rule
+					common.RuleIndexInRoute = ruleIndex
+					common.MatchIndexInRule = math.MaxInt
+					match := RulePrecedence{
+						HTTPMatch:                        &v1.HTTPRouteMatch{},
+						HttpSpecificRulePrecedenceFactor: &HttpSpecificRulePrecedenceFactor{},
+						CommonRulePrecedence:             common,
+					}
+					httpRoutes = append(httpRoutes, match)
```
shuqz: i think `getHttpMatchPrecedenceInfo(&httpMatch, &match)` is missing when route does not match. it is ok for most of them (headers/query params) because if those does not get specified,they will have default value 0 or false. but at least for pathType, we want to treat them like prefix /*
```
func getHttpMatchPrecedenceInfo(httpMatch *v1.HTTPRouteMatch, matchPrecedence *RulePrecedence) {
	matchPrecedence.HttpSpecificRulePrecedenceFactor.PathType = getHttpRoutePathType(httpMatch.Path)
	// httpMatch.Path.Value won't be nil, default is /
	matchPrecedence.HttpSpecificRulePrecedenceFactor.PathLength = len(*httpMatch.Path.Value)
	matchPrecedence.HttpSpecificRulePrecedenceFactor.HasMethod = httpMatch.Method != nil
	matchPrecedence.HttpSpecificRulePrecedenceFactor.HeaderCount = len(httpMatch.Headers)
	matchPrecedence.HttpSpecificRulePrecedenceFactor.QueryParamCount = len(httpMatch.QueryParams)

}
```

PR #4317, opened 2025-08-22
feat(nlb): add eip auto selection support for nlb

aschaber1:
### Issue

This pull request introduces a new annotation, `service.beta.kubernetes.io/aws-load-balancer-eip-tags`, allowing users to dynamically assign Elastic IPs (EIPs) to an internet-facing Network Load Balancer (NLB) based on their AWS tags.

We don't want to hardcode certain IP allocations, and unfortunately the NLB does not support selecting IPs directly from an IPAMPool.

### Description

This change adds a new annotation, `service.beta.kubernetes.io/aws-load-balancer-eip-tags`, that enables selecting unassociated EIPs for an internet-facing NLB by their tags instead of their specific allocation IDs. This simplifies the process for users who manage many EIPs, as they no longer need to manually find and specify individual allocation IDs.

The annotation's value is a comma-separated list of `key=value` pairs, which the controller uses to filter available EIPs. It finds all unassociated EIPs that match all specified tags. The controller then assigns these EIPs to the NLB, ensuring there are enough matching EIPs for the number of subnets the NLB is deployed in. This new annotation provides a more flexible way to manage and assign static IPs to NLBs.

This functionality complements the existing `service.beta.kubernetes.io/aws-load-balancer-eip-allocations` annotation, offering an alternative for EIP management.

### Checklist
- [X] Manually tested

### DIFF
--- a/pkg/service/model_build_load_balancer.go
+++ b/pkg/service/model_build_load_balancer.go
@@ -10,6 +10,8 @@ import (
 	"sort"
 	"strconv"
 
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+
 	awssdk "github.com/aws/aws-sdk-go-v2/aws"
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


pkg/service/model_build_load_balancer.go
```diff
@@ -335,13 +337,60 @@ func (t *defaultModelBuildTask) buildLoadBalancerTags(ctx context.Context) (map[
...
+			filters = append(filters, ec2types.Filter{
+				Name:   awssdk.String(fmt.Sprintf("tag:%s", k)),
+				Values: []string{v},
+			})
+		}
+		filters = append(filters, ec2types.Filter{
+			Name:   awssdk.String("domain"),
+			Values: []string{"vpc"},
+		})
+		addresses, err := t.ec2Client.DescribeAddressesAsList(context.Background(), &ec2.DescribeAddressesInput{
```
zac-nixon, line 373: This would be good to have a cache in front of it as I don't expect this result to change often.
example: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/644c4b3497ea66706ecb9ace1bad8021f1421c9d/pkg/ingress/model_builder.go#L551-L556

aschaber1: Hello, thank you for looking into this, unfortunately I am not sure if I am familiar with Go and/or the codebase to implement caching. The code you've referenced is just part of the WebACLName Caching bits and pieces, if I understand the code correctly.
Any chances of helping me out a bit more? Thank you.

PR #4321, opened 2025-08-26
[feat gw-api]support multiple header value in condition

shuqz:
### Description
route match headers does not allow multiple http header values for same header name, but ALB supports it. we now allow customer to parsing string with comma so we can separate values by comma. customer can also choose to escape it if needs to be literal. 

### Checklist
- [x] Manually tested
```
      matches:
        - headers:
            - name: "version"
              type: Exact
              value: "bar\\,bat"
            - name: "headerOne"
              type: Exact
              value: "value1,value2"
```
gives 
Screenshot 2025-08-26 at 1 52 12 PM

### DIFF
--- a/pkg/gateway/routeutils/route_rule_condition.go
+++ b/pkg/gateway/routeutils/route_rule_condition.go
@@ -2,10 +2,11 @@ package routeutils
 
 import (
 	"fmt"
+	"strings"
+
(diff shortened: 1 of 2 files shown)


zac-nixon (approved):
/lgtm
/approved

Issue #4307, opened 2025-08-13
IMDS is still used even though aws-region and vpc-id are specified


dataviruset:
**Bug Description**
After upgrading [terraform-aws-eks](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3333) to version 21, the default `http_put_response_hop_limit` changed from `2` to `1` which seems more secure but it prevents the AWS load balancer controller from working even though I specified `--aws-region` and `--aws-vpc-id` manually as described in the [AWS LBC documentation](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.13/deploy/installation/#using-the-amazon-ec2-instance-metadata-server-version-2-imdsv2) and [AWS documentation](https://docs.aws.amazon.com/eks/latest/userguide/lbc-manifest.html#lbc-install).

It seems that it's still using IMDS for something even though I configured IRSA.

```
{"level":"error","ts":"2025-08-13T12:43:00Z","msg":"Reconciler error","controller":"targetGroupBinding","controllerGroup":"elbv2.k8s.aws","controllerKind":"TargetGroupBinding","TargetGroupBinding":{"name":"k8s-default-mypodname-1ff9343f87","namespace":"default"},"namespace":"default","name":"k8s-default-mypodname-1ff9343f87","reconcileID":"96fb8462-30af-4090-8ceb-328621e829a3","error":"operation error Elastic Load Balancing v2: DescribeTargetHealth, get identity: get credentials: failed to refresh cached credentials, no EC2 IMDS role found, operation error ec2imds: GetMetadata, canceled, context deadline exceeded"}
```

The issue was also reported here https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3955#issuecomment-2876405451 but that guy gave up and set `http_put_response_hop_limit` to `3`.

**Steps to Reproduce**
```
      containers:
        - args:
            - '--cluster-name=my-cluster-name'
            - '--ingress-class=alb'
            - '--aws-region=ap-southeast-1'
            - '--aws-vpc-id=vpc-abcdefg'
            - '--enable-shield=false'
            - '--enable-waf=false'
            - '--enable-wafv2=false'
          image: public.ecr.aws/eks/aws-load-balancer-controller:v2.13.4
```

Using the latest IAM policy from https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/install/iam_policy.json

**Expected Behavior**
[...]

PR #4327, opened 2025-08-29
doc updates for scaling, IMDS usage

zac-nixon:
issue for imds verbage: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4307

generally, when speaking with users they have a hard time understanding how to scale the LBC for massive clusters. I wrote a quick document that explains how to scale the LBC.

### DIFF
--- a/docs/deploy/scaling.md
+++ b/docs/deploy/scaling.md
@@ -1,0 +1,110 @@
+# Scaling your controller deployment
+
+The AWS Load Balancer Controller (LBC) implements a standard Kubernetes controller. The controller reads changes from the cluster
+API server, calculates an intermediate representation (IR) of your AWS resources, then ensures the actual AWS resources match
+the IR state. The controller can perform CRUD operations to ensure the Kubernetes and AWS resources stay in-sync. This page is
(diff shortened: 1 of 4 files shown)


docs/deploy/scaling.md
```diff
@@ -0,0 +1,104 @@
...
+- 3 threads for Ingress management (ALB)
+- 3 threads for Service management (NLB)
+- 3 threads for ALB Gateway management (IF ENABLED)
+- 3 threads for NLB Gateway management (IF ENABLED)
+- 3 threads for TargetGroupBinding management (Target Registration for ALB / NLB)
+
+For 99.9% of use-cases, these values are enough. When managing a large number of resources, the threads should be tuned in turn the
+memory and CPU resources should be tuned. Here's a general formula:
+
+** This formula is just a suggestion, and many workloads might perform differently. It's important to load test your exact scenario **
```
shraddhabang, line 41: Should we also mention that if they increase the thread counts, there is a possibility of throttling and should take a look at our throttling section to understand that so that they know the side-effect of increasing the worker counts especially for TGBs?

zac-nixon: good idea. will add

PR #4334, opened 2025-09-08
[feat gw api] Add authenticate oidc action support for L7 gateway

shraddhabang:
### Description

This PR adds support for OIDC authentication actions in ALB Gateway API through the new `ListenerRuleConfiguration` CRD, enabling users to configure OpenID Connect authentication at the load balancer level.

#### __Targeted Secret Management & Validation__

- __Real-time secret validation__: LRC controller validates referenced secrets exist and are accessible
- __Status reporting__: `ListenerRuleConfiguration.Status.Accepted` indicates whether all required secrets are available
- __Targeted secret monitoring__: Only monitors secrets actively referenced by accepted LRCs, avoiding cluster-wide secret watching overhead
- __Security enhanced__: Minimal RBAC footprint - only requires access to secrets referenced in LRCs
- __Event-driven lifecycle__: Automatically starts monitoring when LRC becomes accepted, stops when LRC is deleted

#### __Gateway Integration__

- __Route loading validation__: Gateway controller validates LRC acceptance before building models
- __Fail-fast behavior__: Routes with unaccepted LRCs are rejected during route loading
- __Clear error reporting__: Gateway status provides detailed feedback about LRC readiness

### Checklist
- [x] Manually tested


pkg/gateway/routeutils/listener_rule_config_utils.go
```diff
@@ -43,3 +44,59 @@ func isListenerRuleConfigReferredByRoute(route preLoadRouteDescriptor, ruleConfi
...
+
+// FilterListenerRuleConfigBySecret filters listenerruleconfigs based on secret reference.
+// Returns a list of ListenerRuleConfiguration objects that reference the given secret.
+func FilterListenerRuleConfigBySecret(ctx context.Context, k8sClient client.Client, secret *corev1.Secret) ([]*elbv2gw.ListenerRuleConfiguration, error) {
+	if secret == nil {
+		return nil, nil
+	}
+
+	// List all ListenerRuleConfiguration objects
+	listenerRuleCfgList := &elbv2gw.ListenerRuleConfigurationList{}
```
zac-nixon, line 56: We're not allowing cross-namespace LRC -> Secret right? Eventually, we might want to using a ReferenceGrant, but for now since we're not we should scope this list call down to only the namespace of the secret.

controllers/gateway/gateway_controller.go
```diff
@@ -279,7 +281,7 @@ func (r *gatewayReconciler) reconcileHelper(ctx context.Context, req reconcile.R
 	return nil
 }
 
-func (r *gatewayReconciler) reconcileDelete(ctx context.Context, gw *gwv1.Gateway, stack core.Stack, routes map[int32][]routeutils.RouteDescriptor) error {
+func (r *gatewayReconciler) reconcileDelete(ctx context.Context, gw *gwv1.Gateway, stack core.Stack, routes map[int32][]routeutils.RouteDescriptor, secrets []types.NamespacedName) error {
```
zac-nixon: Why does the reconcile delete function need the list of secrets? IIRC, deletion should only care about removing any in-memory relation of the secret <-> gateway.

controllers/gateway/listenerrule_configuration_controller.go
```diff
@@ -100,3 +139,120 @@ func (r *listenerRuleConfigurationReconciler) SetupWithManager(_ context.Context
...
+			return []types.NamespacedName{secretKey}, nil
+		}
+	}
+	return nil, nil
+}
+
+// updateStatus updates the ListenerRuleConfiguration status with the current validation state
+func (r *listenerRuleConfigurationReconciler) updateStatus(ctx context.Context, listenerRuleConf *elbv2gw.ListenerRuleConfiguration, accepted bool, message string) error {
+	// Check if status actually needs updating
+	if listenerRuleConf.Status.Accepted != nil &&
```
zac-nixon, line 192: Can we re-use this logic? https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/controllers/gateway/route_reconciler.go#L421

pkg/gateway/routeutils/route_rule_action.go
```diff
@@ -84,12 +89,53 @@ func buildAuthenticateCognitoAction(authCognitoActionConfig *elbv2gw.Authenticat
...
+	namespace := route.GetRouteNamespacedName().Namespace
+	if authenticateOIDCActionConfig.Secret.Namespace != nil {
+		namespace = awssdk.ToString(authenticateOIDCActionConfig.Secret.Namespace)
+	}
+	secretKey := types.NamespacedName{
+		Namespace: namespace,
+		Name:      authenticateOIDCActionConfig.Secret.Name,
+	}
+	secret := &corev1.Secret{}
+	if err := k8sClient.Get(ctx, secretKey, secret); err != nil {
```
zac-nixon: Can you maintain a cache here to prevent loading the same secret multiple times? It can be in the context of this reconcile run.

pkg/gateway/routeutils/route_rule_action.go
```diff
@@ -84,12 +89,53 @@ func buildAuthenticateCognitoAction(authCognitoActionConfig *elbv2gw.Authenticat
...
+		return nil, nil, err
+	}
+	rawClientID, ok := secret.Data["clientID"]
```
zac-nixon: nit: let's use named constants here.

test/e2e/gateway/alb_instance_target_test.go
```diff
@@ -844,6 +844,213 @@ var _ = Describe("test k8s alb gateway using instance targets reconciled by the
...
+					{
+						Type: elbv2gw.ActionTypeAuthenticateOIDC,
+						AuthenticateOIDCConfig: &elbv2gw.AuthenticateOidcActionConfig{
+							Issuer:                testOidcIssuer,
+							AuthorizationEndpoint: testOidcAuthorizationEndpoint,
+							TokenEndpoint:         testOidcTokenEndpoint,
+							UserInfoEndpoint:      testOidcUserInfoEndpoint,
+							Secret: &elbv2gw.Secret{
+								Name: oidcSecretName,
+								// Namespace will default to same as ListenerRuleConfiguration
```
zac-nixon, line 901: Hmm, maybe my earlier comment was wrong? Do we support cross namespace referencing?

test/e2e/gateway/alb_ip_target_test.go
```diff
@@ -884,6 +884,215 @@ var _ = Describe("test k8s alb gateway using ip targets reconciled by the aws lo
 			})
 		})
 	})
+	Context("with ALB ip target configuration with secure HTTPRoute and authenticate oidc action", func() {
```
zac-nixon, line 887: Thanks for adding these comprehensive tests.

pkg/gateway/routeutils/listener_rule_config_utils.go
shraddhabang: Actually we are allowing to access to cross-namespace secret access  right? The LRC need not be in same namespace as secrets. So why do we want to scope it to secret namespace?

controllers/gateway/gateway_controller.go
shraddhabang: Yes correct, This will always be nil. Will update it

pkg/gateway/routeutils/listener_rule_config_utils.go
zac-nixon: If we're allowing cross-namespace access we _must_ implement ReferenceGrants: https://gateway-api.sigs.k8s.io/api-types/referencegrant.

Can a route reference an LRC outside of the current namespace? Basically, anytime we cross the namespace boundary there must be a ReferenceGrant that allows the access otherwise users can perform an escalation of privilege.

test/e2e/gateway/alb_instance_target_test.go
shraddhabang: Yes we do.

controllers/gateway/listenerrule_configuration_controller.go
shraddhabang: Actually the types differ for status for routes and LRC. Hence I did not use it. LRC status is much simpler with only accepted and message.

pkg/gateway/routeutils/listener_rule_config_utils.go
shraddhabang: Sure. Will update. Thanks.

apis/gateway/v1beta1/listenerruleconfig_types.go
```diff
@@ -116,12 +116,10 @@ type FixedResponseActionConfig struct {
 	MessageBody *string `json:"messageBody,omitempty"`
 }
 
-// Secret holds OAuth 2.0 clientID and clientSecret. You need to create this secret and provide its name and namespace
+// Secret holds OAuth 2.0 clientID and clientSecret. You need to create this secret and provide its name
```
zac-nixon, line 119: Please ensure to note that the namespace of the LRC is used here.

PR #4335, opened 2025-09-08
[feat gw-api]support source ip and handle grpc filter

shuqz:
### Description
- emit exception for grpc route filter (we do not support any type except ExtensionsRef)
- support source ip with listener rule configuration CRD, below are expected behaviors 
   - when indexes are not specified, it will be added to all rule conditions under same grpcRouteRule/httpRouteRule -> this means customer can also try to have separate route and rule crd for each source ip if they think it is more clear for them to manage 
   - when indexes are provided, source ip config will be added to the specified indexes (but we recommend customer to maintain the order of rule because it can apply to wrong rule if order changes. to keep minimal effort, they can always put those rules need source ip at beginning of manifest) 

### Checklist
- [x] Manually tested

**Test case 1: not specify indexes - applies to all**
```
---
metadata:
  name: custom-rule-config-source-ip
  namespace: sample-gateway-alb
spec:
  conditions:
    - field: source-ip
      sourceIPConfig:
        values:
          - 10.0.0.0/6
---
kind: HTTPRoute
metadata:
  name: http-app-source-ip
  namespace: sample-gateway-alb
spec:
  parentRefs:
    - name: sample-gw-alb
      port: 90
  rules:
    - backendRefs:
        - name: echoserver
          port: 80
      filters:
        - type: ExtensionRef
          extensionRef:
            group: "gateway.k8s.aws"
            kind: "ListenerRuleConfiguration"
            name: "custom-rule-config-source-ip"
      matches:
        - path: # Path Pattern
            type: Exact
            value: /pathExactMatch
          queryParams: # Query String
            - name: "user"
              value: "john"
          method: GET # HTTP Request Method
        - path: # Regex path match
            type: RegularExpression
            value: "/firstRule/some?/users"
    - backendRefs:
        - name: echoserver
          port: 80
      filters:
        - type: ExtensionRef
          extensionRef:
            group: "gateway.k8s.aws"
            kind: "ListenerRuleConfiguration"
            name: "custom-rule-config-source-ip"
      matches:
        - path: # Path Pattern
            type: Exact
            value: /secondRulePath
          method: POST # HTTP Request Method
        - path: # Regex path match
            type: RegularExpression
            value: "/secondRule/some?/users"
        - path:
            type: "PathPrefix"
            value: "/secondRule"
```
Screenshot 2025-09-08 at 10 02 52 AM

**Test case 2: specify multiple indexes**
```
---
metadata:
  name: custom-rule-config-source-ip
  namespace: sample-gateway-alb
spec:
  conditions:
    - field: source-ip
      indexes:
        - ruleIndex: 1
          matchIndex: 0
        - ruleIndex: 1
          matchIndex: 1
      sourceIPConfig:
        values:
          - 10.0.0.0/6
```
Screenshot 2025-09-08 at 10 05 15 AM


apis/gateway/v1beta1/listenerruleconfig_types.go
```diff
@@ -50,6 +50,11 @@ type ListenerRuleCondition struct {
 	// Information for a source IP condition
 	// +optional
 	SourceIPConfig *SourceIPConditionConfig `json:"sourceIPConfig,omitempty"`
+
+	// Indexes of Match in a rule to apply source ip to
+	// If MatchIndexes is not provided, SourceIpConfig will be applied to all conditions where ListenerRuleConfiguration is used
+	// +optional
+	MatchIndexes *[]int `json:"matchIndexes,omitempty"`
```
shraddhabang, line 57: `MatchIndexes []int `json:"matchIndexes,omitempty"`` would be simpler to avoid nil checks.

pkg/gateway/routeutils/route_rule_condition.go
```diff
@@ -234,6 +235,54 @@ func buildGrpcMethodCondition(method *gwv1.GRPCMethodMatch) ([]elbv2model.RuleCo
...
+	}
+	return nil
+}
+
+// BuildSourceIpInCondition : takes source ip configuration from listener rule configuration CRD, then AND it to condition list
+func BuildSourceIpInCondition(ruleWithPrecedence RulePrecedence, conditionsList []elbv2model.RuleCondition) []elbv2model.RuleCondition {
+	rule := ruleWithPrecedence.CommonRulePrecedence.Rule
+	matchIndex := ruleWithPrecedence.CommonRulePrecedence.MatchIndexInRule
+	if rule.GetListenerRuleConfig() != nil {
+		conditionsFromRuleConfig := getRoutingConditions(rule.GetListenerRuleConfig())
```
shraddhabang: Can we not simply access the conditions from cfg directly?
``conditionsFromRuleConfig := rule.GetListenerRuleConfig().Spec.Conditions``. No need to copy the conditions into another array in `getRoutingConditions`

zac-nixon:
can you add an e2e test that makes sure the rule is correctly materialized 🙏

apis/gateway/v1beta1/listenerruleconfig_types.go
shuqz: this is optional param, they do not have to speicify MatchIndexes. when not specified, it will apply to all

pkg/gateway/routeutils/route_rule_condition.go
shuqz: yea i can update it

apis/gateway/v1beta1/listenerruleconfig_types.go
shraddhabang: No if they dont specify it. it will be empty only. So you dont need to have a pointer for it.

shuqz: synced offline

Issue #4315, opened 2025-08-21
Reconcile errors on EKS cluster with Hybrid nodes


sanbyk:
**Bug Description**
Logs and TargetGroupBinding objects are flooded with errors when target pods are deployed on hybrid nodes.

**Steps to Reproduce**

- Step-by-step guide to reproduce the bug:
  - Deploy EKS cluster with hybrid nodes as per: https://docs.aws.amazon.com/eks/latest/userguide/hybrid-nodes-overview.html
  - Configure CNI and Ingress
- Manifests applied while reproducing the issue:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      nodeSelector:
        eks.amazonaws.com/compute-type: hybrid
      containers:
      - name: nginx
        image: nginx:latest
        ports:
        - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: nginx
spec:
  selector:
    app: nginx
  ports:
  - protocol: TCP
    port: 80
    targetPort: 80
  type: ClusterIP
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: nginx
  labels:
    name: nginx
  annotations:
    alb.ingress.kubernetes.io/scheme: internet-facing
    alb.ingress.kubernetes.io/target-type: ip
spec:
  ingressClassName: alb
  rules:
  - http:
      paths:
      - pathType: Prefix
        path: "/"
        backend:
          service:
            name: nginx
            port:
              number: 80
---
```
- Controller logs/error messages while reproducing the issue:
```
{"level":"error","ts":"2025-08-21T13:17:46Z","msg":"Requesting network requeue due to error from ReconcileForPodEndpoints","tgb":{"name":"k8s-default-nginx-1234567890","namespace":"default"},"error":"providerID eks-hybrid:///ap-northeast-1/eks-test/mi-01c49bb1234567890 is invalid for EC2 instances, node: mi-01c49bb1234567890"}
{"level":"info","ts":"2025-08-21T13:17:46Z","msg":"Requeue for networking requeue","tgb":{"name":"k8s-default-nginx-1234567890","namespace":"default"}}
```

**Expected Behavior**

[...]

PR #4336, opened 2025-09-09
feat: add EKS Hybrid Nodes support

sanbyk:
### Issue
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4315

### Description
This PR adds support for pods running on EKS Hybrid Nodes (on-premise). These nodes have no EC2 instance or VPC ENI and previously caused reconciliation failures when resolving pod endpoints and node instance IDs. The controller now detects hybrid nodes/pods, skips EC2-/ENI-specific logic, and reconciles TargetGroupBindings for hybrid endpoints.

### Changes
Hybrid nodes detection
- Update ExtractNodeInstanceID to recognize `eks-hybrid` ProviderID
- Return a clearly non-EC2 instance identifier (special prefix) to:
  - Avoid accidental DescribeInstances calls
  - Keep a stable, non-empty key for logging/maps/metrics

Pod ENI resolution
- Extend pod classification to include hybrid using `eks.amazonaws.com/compute-type=hybrid`
- Skip ENI lookups entirely for hybrid pods; return placeholder ENI info to mark “no ENI”

Networking reconciliation
- Skip backend Security Group resolution/management for hybrid endpoints, since there are no ENIs/SGs to manage

Node instance fetch
- Exclude hybrid nodes from EC2 DescribeInstances fan-out in NodeInfoProvider

Tests
- Add hybrid nodes focused unit tests for classification and resolution
- Extend node instance ID extraction tests to cover eks-hybrid providerIDs

### Checklist
- [X] Manually tested

### DIFF
--- a/pkg/networking/networking_manager.go
+++ b/pkg/networking/networking_manager.go
@@ -162,7 +162,7 @@ func (m *defaultNetworkingManager) computeIngressPermissionsPerSGWithPodEndpoint
 	podsBySG := make(map[string][]k8s.PodInfo)
 	for podKey, eniInfo := range eniInfoByPodKey {
 		// Handle hybrid pods specially - they don't have security groups
-		if eniInfo.NetworkInterfaceID == "hybrid-no-eni" {
+		if eniInfo.NetworkInterfaceID == hybridNetworkInterfaceID {
 			continue
 		}
 		sgID, err := m.resolveEndpointSGForENI(ctx, eniInfo)
--- a/pkg/networking/pod_eni_info_resolver.go
+++ b/pkg/networking/pod_eni_info_resolver.go
@@ -28,6 +28,7 @@ const (
 
 	labelEKSComputeType       = "eks.amazonaws.com/compute-type"
 	labelSageMakerComputeType = "sagemaker.amazonaws.com/compute-type"
+	hybridNetworkInterfaceID  = "hybrid-no-eni"
 )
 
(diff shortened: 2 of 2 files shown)


malodie:
Bump.  We're waiting on this feature to move forward with hybrid nodes, as the aws-load-balancer-controller is the documented way (by AWS) to do ingress.  It appears to work when we provision pods first on EC2, but then falls apart when the pods return to the hybrid state.  I am also happy to test this in my working lab cluster once it releases (or before, if necessary).

pkg/networking/networking_manager.go
```diff
@@ -161,6 +161,10 @@ func (m *defaultNetworkingManager) computeIngressPermissionsPerSGWithPodEndpoint
 
 	podsBySG := make(map[string][]k8s.PodInfo)
 	for podKey, eniInfo := range eniInfoByPodKey {
+		// Handle hybrid pods specially - they don't have security groups
+		if eniInfo.NetworkInterfaceID == "hybrid-no-eni" {
```
zac-nixon: Can you please use a const to represent this special value?

sanbyk: @zac-nixon Thank you for your review! 
I've declared a const in `pod_eni_info_resolver.go` where it's being initially set, and since this is the same package, it's accessible in `networking_manager.go`

zac-nixon: Perfect, thanks!

Issue #4319, opened 2025-08-25
Mutual TLS remains enabled after deletion of alb.ingress.kubernetes.io/mutual-authentication annotation


justfalter:
**Bug Description**
Deleting an `alb.ingress.kubernetes.io/mutual-authentication` annotation on an mTLS-enabled ingress does not result in the disabling of mTLS for a listener.

This behavior was introduced by a fix #3715 (pr #3717). That issue was dealing with the fact that a lack of `alb.ingress.kubernetes.io/mutual-authentication` annotation was treated as "explicitly disable mutual TLS", which would fail for Outputs and Local Zones, where mTLS is not supported.  In fixing #3715, the decision was made to require that mutual-authentication annotations be explicitly set to `mode: off` for mtls to be disabled. This is [covered in the docs](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/ingress/annotations/#mutual-authentication): `Once the Mutual Authentication is set, to turn it off, you will have to explicitly pass in this annotation with mode : "off".`

The problem is that this is antithetical to the Kubernetes principle of a resource configuration representing the desired state; two resources with identical configurations should have the same state. Unfortunately, there's no way of knowing whether an ingress lacking an `alb.ingress.kubernetes.io/mutual-authentication` annotation does or does not have mutual TLS enabled.

It is a struggle to use the `alb.ingress.kubernetes.io/mutual-authentication` with common tools for managing Kubernetes deployments (helm, terraform, pulumi), as they all rely on resources adhering to the desired state principle of Kubernetes resources.  They often have two options: create or destroy, and lack sufficient patterns for "on destroy, add this annotation to a resource so that the aws-load-balanacer-controller knows to remove mutual TLS auth, wait for that to be destroyed, and then delete the resource." Having to manually patch the annotation to disable mTLS is cumbersome. 

**Steps to Reproduce**

- Have an AWS TrustStore.
- Have an Ingress resource without the `alb.ingress.kubernetes.io/mutual-authentication` annotation.
- Add annotation to ingress `alb.ingress.kubernetes.io/mutual-authentication: [{"port":12345,"mode":"verify","trustStore":"[the trust store's arn]"}]`.
- Wait for event `SuccessfullyReconciled` for ingress.
- Observe that the ingress AWS load balanacer listener configuration now has mTLS enabled, and is associated with the referenced Trust Store.
[...]

PR #4344, opened 2025-09-19
Add EnhancedDefaulting flag to do sensible defaulting when possible

zac-nixon:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4319

### Description

Adds a controller flag, that will remove mTLS configuration when the annotation is removed from the annotations. This flag could potentially be moved to other APIs, like Provisioned Capacity, but those require a bigger lift. This change preserves the initial LBC logic, so no breaking API changes. See the code for the various cases that I considered.

With Gateway API, we were handling removal of mTLS configuration much better. With this change, I am going ahead and enabling EnhancedDefault behavior for Gateway API as I believe this implementation is correct and better than ignoring missing annotations. Because the deployer is smarter now, we can remove the Local Zone / Outpost check from the Gateway API model builder.

Summary of additions:

- Defaulting behavior for ALPN, ProvisionedCapacity, and mTLS in Gateway API.
- Same defaulting behavior for ALPN and mTLS offered to Ingress API users, behind a feature flag.
- Refactored Gateway API addons to support ProvisionedCapacity, that way we can use the same defaulting logic currently in-use by Shield and WAF

### Checklist
- [x] Manually tested

### DIFF
--- a/apis/gateway/v1beta1/loadbalancerconfig_types.go
+++ b/apis/gateway/v1beta1/loadbalancerconfig_types.go
@@ -175,7 +175,6 @@ type ListenerConfiguration struct {
 
 	// mutualAuthentication defines the mutual authentication configuration information.
 	// +optional
-	// +kubebuilder:default={"mode": "off"}
 	MutualAuthentication *MutualAuthenticationAttributes `json:"mutualAuthentication,omitempty"`
 
 	// listenerAttributes defines the attributes for the listener
--- a/config/crd/gateway/gateway-crds.yaml
+++ b/config/crd/gateway/gateway-crds.yaml
@@ -515,8 +515,6 @@ spec:
                         type: object
                       type: array
                     mutualAuthentication:
-                      default:
-                        mode: "off"
                       description: mutualAuthentication defines the mutual authentication
                         configuration information.
                       properties:
--- a/config/crd/gateway/gateway.k8s.aws_loadbalancerconfigurations.yaml
+++ b/config/crd/gateway/gateway.k8s.aws_loadbalancerconfigurations.yaml
@@ -117,8 +117,6 @@ spec:
                         type: object
                       type: array
                     mutualAuthentication:
-                      default:
-                        mode: "off"
                       description: mutualAuthentication defines the mutual authentication
                         configuration information.
                       properties:
--- a/helm/aws-load-balancer-controller/crds/gateway-crds.yaml
+++ b/helm/aws-load-balancer-controller/crds/gateway-crds.yaml
@@ -515,8 +515,6 @@ spec:
                         type: object
                       type: array
                     mutualAuthentication:
-                      default:
-                        mode: "off"
                       description: mutualAuthentication defines the mutual authentication
                         configuration information.
                       properties:


pkg/gateway/model/model_build_listener.go
```diff
@@ -410,29 +408,12 @@ func buildL4ListenerDefaultActions(targetGroup *elbv2model.TargetGroup) []elbv2m
...
-	// mTLS is not supported for Local Zone or Outposts AZs
-	// We only need to check one of the subnets at this point since we have already resolved subnets for this LoadBalancer and we don't allow mix AZs
-	isSubnetInLocalZoneOrOutpost, err := subnetsResolver.IsSubnetInLocalZoneOrOutpost(ctx, subnets.subnets[0].SubnetID)
-	if err != nil {
-		return nil, err
-	}
-	if isSubnetInLocalZoneOrOutpost {
-		// Skip mTLS configuration for Local Zone or Outpost subnets
-		l.logger.V(1).Info("skipping mutual authentication configuration as it is not supported in local zone or outpost")
+	if !isSecureProtocol(gwLsCfg.protocol) || lbLsCfg == nil || lbLsCfg.MutualAuthentication == nil {
```
shraddhabang, line 413: Why are we removing the outpost/lz checks from here?
This will fail in creating listeners in outpost/lz as our CRD default is OFF (even if customer does not specify it). the drift logic in deployer only run for updates not for create. So for new listeners we will have desired mtls setting as non-nil with OFF setting, and we will fail in creating listener. we still need to check outpost/lz if we are building listener object from CRD for such cases. Correct me if I am wrong.

pkg/gateway/model/addons/model_build_addon.go
```diff
@@ -0,0 +1,121 @@
...
+			break
+		case addon.Shield:
+			shieldEnabled, shieldPrestack := aob.buildShield(lbCfg, previousAddonConfig)
+			metadata = append(metadata, addon.AddonMetadata{
+				Name:    addon.Shield,
+				Enabled: shieldEnabled,
+			})
+			prestack = append(prestack, shieldPrestack)
+			break
+		case addon.ProvisionedCapacity:
```
shraddhabang, line 52: I really like this. Thanks for unifying these

pkg/gateway/model/model_build_listener.go
zac-nixon: It's on the lines below:
```

	// Default to "off" mode when no explicit mTLS configuration exists
	if lbLsCfg == nil || lbLsCfg.MutualAuthentication == nil {
		return &elbv2model.MutualAuthenticationAttributes{
			Mode: string(elbv2gw.MutualAuthenticationOffMode),
		}, nil
	}
```

I removed this block, as the new defaulting checks within the deployer package make it so we don't need to inject the off mode.

pkg/deploy/elbv2/listener_manager.go
```diff
@@ -288,14 +305,81 @@ func isSDKListenerSettingsDrifted(lsSpec elbv2model.ListenerSpec, sdkLS Listener
...
+		// 2. desired = nil, sdk = None, Result = no drift
+		// 3. desired = nil, sdk = Other value, Result = drift.
+		// 4. desired = None, sdk = nil. Result = no drift.
+		// 5. desired = None, sdk = None. Result = no drift
+		// 6. desired = None, sdk = Other value. Result = drift
+		// 7. desired = Other value, sdk = Nil. Result = drift
+		// 8. desired = Other value, sdk = None. Result = drift
+		// 9. desired = Other value, sdk = Other value. Result = drift depending on equality.
+
+		desiredIsNilOrNone := lsSpec.ALPNPolicy == nil || len(lsSpec.ALPNPolicy) == 0 || lsSpec.ALPNPolicy[0] == string(elbv2model.ALPNPolicyNone)
```
shuqz, line 321: can we check nil here? is it possible lsSpec.ALPNPolicy=nil so lsSpec.ALPNPolicy[0] will panic. same for sdkLS.Listener, and then sdkLS.Listener.AlpnPolicy will panic

pkg/gateway/model/model_build_loadbalancer.go
```diff
@@ -48,15 +44,14 @@ func (lbModelBuilder *loadBalancerBuilderImpl) buildLoadBalancerSpec(scheme elbv
...
 
 	spec := elbv2model.LoadBalancerSpec{
-		Name:                        name,
-		Type:                        lbModelBuilder.loadBalancerType,
-		Scheme:                      scheme,
-		IPAddressType:               ipAddressType,
-		SubnetMappings:              subnets.subnets,
-		SecurityGroups:              securityGroupTokens,
-		LoadBalancerAttributes:      lbModelBuilder.buildLoadBalancerAttributes(lbConf),
-		MinimumLoadBalancerCapacity: lbModelBuilder.buildLoadBalancerMinimumCapacity(lbConf),
```
shuqz, line 58: why are we not taking MinimumLoadBalancerCapacity from load balancer spec anymore?

pkg/gateway/model/model_build_listener.go
shraddhabang: Thank you for changing the default behavior here and remove it from CRD. Going forward, we will not support the defaults for such bi-model defaults on CRD. Rather we will handle them in business logic.

PR #4348, opened 2025-09-22
feat: Introduce annotation for setting explicit egress CIDR ranges on Service LB frontend security group

mtweten:
### Description

This PR implements a new annotation (`service.beta.kubernetes.io/aws-load-balancer-outbound-cidrs`) for Service load balancers to enable explicitly specifying a list of CIDR ranges to be added as egress rules to managed frontend security groups. 

Currently, `aws-load-balancer-controller` doesn't set any explicit egress rules during SG creation, relying instead on AWS to create a default all-protocol `0.0.0.0/0` rule when no explicit egress rules are specified during SG creation. This egress rule is necessary for health checks on target groups, etc.

However, some organizations may have security scans that trigger on this `0.0.0.0/0` rule, and it would be desirable to be able to scope that down to a specific set of CIDR ranges instead (e.g. the VPC CIDR, for example). There is currently no mechanism in `aws-load-balancer-controller` to allow setting this, so that is what this is intended to provide.

Note that this only implements this for the Service load balancer (e.g. nlb) workflow. I wanted to keep the PR more narrowly scoped, but it would probably relatively simple to implement for other workflows as well. There is also the `backend` SG where a command flag would need to be introduced, but it's easy enough just to use a statically created SG for the backend SG using the existing mechanisms instead.

### Checklist
- [x] Manually tested


jwhitcraft:
hey-a @M00nF1sh or @wweiwei-li is there any movement on this?  Our security team is asking for this specific thing.

PR #4350, opened 2025-09-23
Fix weighted target group comparison for forward action

shraddhabang:
### Issue

LBC was incorrectly handling listener rules with weighted target groups, causing unnecessary rule modifications and creations instead of efficient priority reordering.

### Description

The `CompareOptionForTargetGroupTuples` function was not sorting target groups by ARN, causing equivalent configurations with different orders to be considered different. This led to:

- Unnecessary ModifyRule/CreateRule API calls instead of SetRulePriorities
- Temporary rule deletions causing 4XX errors during updates

Fix: 
1. Add ARN-based sorting in target group comparison transformer
2. Maintained backward compatibility for single target group scenarios
3. Added Comprehensive unit tests for complex scenarios

## Local Testing Verification

Using an ingress with two rules having weighted target group actions:

### Before Fix:

- Changing priority of weighted target group rules → __2 ModifyRule API calls__

### After Fix:

- Changing priority of weighted target group rules → __1 SetRulePriorities API call__

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/equality/elbv2/compare_option_for_actions_test.go
+++ b/pkg/equality/elbv2/compare_option_for_actions_test.go
@@ -554,10 +554,706 @@ func TestCompareOptionForAction(t *testing.T) {
 			},
 			want: true,
 		},
+		{
+			name: "actions with weighted target groups equal regardless of order",
+			args: args{
+				lhs: elbv2types.Action{
+					Type: elbv2types.ActionTypeEnum("forward"),
+					ForwardConfig: &elbv2types.ForwardActionConfig{
+						TargetGroups: []elbv2types.TargetGroupTuple{
+							{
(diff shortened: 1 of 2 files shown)


pkg/equality/elbv2/compare_option_for_actions_test.go
```diff
@@ -554,10 +554,626 @@ func TestCompareOptionForAction(t *testing.T) {
 			},
 			want: true,
 		},
+		{
+			name: "actions with weighted target groups equal regardless of order",
```
zac-nixon, line 558: can you add a test case where lhs has 3 target groups and rhs has 2 and then the inverse

shraddhabang: Good idea. Done. Thanks.

Issue #4324, opened 2025-08-29
awslbc_reconcile_errors_total metric increases without any error logs


daangn-rainn:
**Bug Description**

After updating from version 2.12.0 to 2.13.4, I tried to use a new metric to create alerts.
I then observed the awslbc_reconcile_errors_total{error_category="reconcile_targetgroupbinding_error"} metric increasing without any corresponding error logs.

<img width="1850" height="608" alt="Image" src="https://github.com/user-attachments/assets/a4431047-f3de-4353-83af-33e2c444b924" />

```
{"level":"info","ts":"2025-08-29T07:22:24Z","msg":"deRegistering targets","arn":"arn:aws:elasticloadbalancing:ap-northeast-2:000000000000:targetgroup/k8s-mockit-mockit-0c140c049e/fd85181e107a0bc0","targets":[{"Id":"10.254.9.153","AvailabilityZone":"ap-northeast-2a","Port":3000}]}
{"level":"info","ts":"2025-08-29T07:22:24Z","msg":"deRegistering targets","arn":"arn:aws:elasticloadbalancing:ap-northeast-2:000000000000:targetgroup/k8s-test-worker-cf9c8d458b/097de3019e482847","targets":[{"Id":"10.254.25.209","AvailabilityZone":"ap-northeast-2a","Port":8080}]}
{"level":"info","ts":"2025-08-29T07:22:24Z","msg":"deRegistering targets","arn":"arn:aws:elasticloadbalancing:ap-northeast-2:000000000000:targetgroup/k8s-mockit-mockit-013420b522/6aac63d2d6dc0cb1","targets":[{"Id":"10.254.9.153","AvailabilityZone":"ap-northeast-2a","Port":5000}]}
{"level":"info","ts":"2025-08-29T07:22:24Z","msg":"deRegistered targets","arn":"arn:aws:elasticloadbalancing:ap-northeast-2:000000000000:targetgroup/k8s-mockit-mockit-0c140c049e/fd85181e107a0bc0","targets":[{"Id":"10.254.9.153","AvailabilityZone":"ap-northeast-2a","Port":3000}]}
[...]

PR #4356, opened 2025-09-25
fix metric population by requeueneeded error

zac-nixon:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4324

### Description

Our metric collector didn't differentiate between "real" and "expected" errors. This changes add the ability to detect a RequeueNeeded* error, and when the Requeue error is detected we *do not* emit a metric pertaining to the error.

Note, I had to move the RequeueNeeded errors out of the runtime package to prevent an import cycle.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/error/error_with_metrics_test.go
+++ b/pkg/error/error_with_metrics_test.go
@@ -1,0 +1,44 @@
+package ctrlerrors
+
+import (
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	lbcmetrics "sigs.k8s.io/aws-load-balancer-controller/pkg/metrics/lbc"
+	"testing"
+)
+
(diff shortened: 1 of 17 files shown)


pkg/error/error_with_metrics_test.go
```diff
@@ -0,0 +1,44 @@
...
+			err:                 errors.New("bad thing"),
+			expectedInvocations: 1,
+		},
+		{
+			name: "requeue needed error",
+			err:  &RequeueNeeded{},
+		},
+		{
+			name: "requeue needed after error",
+			err:  &RequeueNeeded{},
```
shuqz: non-blocking: can we add a test case for RequeueNeededAfter?

shraddhabang: +1

PR #4357, opened 2025-09-25
[feat gateway-api]implement gateway listener status

shuqz:
### Description
- add listener status to gateway status 
- there are several ListenerConditionTypes: Conflicted, Accepted, ResolvedRefs, Programmed, OverlappingTLSConfig. and each type gets its ListenerConditionReason. i did not get all of them implemented since some of them are not applied.
- in short, the current behavior is: in loader phase, before mapGatewayAndRoutes, we will validate listeners configurations first. listener status should only related to listener configurations, it is different from what we are doing in mapGatewayAndRoutes, which checks listener <-> route. and this result is stored (won't break reconcile immediately) then during gateway status update, we include listener status based on validation results

### Checklist
- [x] Manually tested


controllers/gateway/gateway_controller.go
```diff
@@ -394,8 +401,18 @@ func (r *gatewayReconciler) updateGatewayStatusSuccess(ctx context.Context, lbSt
...
+	if err != nil {
+		r.logger.Info("failed to build listeners status: %v", k8s.NamespacedName(gw))
+	} else if !isListenerStatusIdentical(gw.Status.Listeners, ListenerStatuses) {
+		gw.Status.Listeners = ListenerStatuses
+		needPatch = true
+	}
+
 	if needPatch {
 		if err := r.k8sClient.Status().Patch(ctx, gw, client.MergeFrom(gwOld)); err != nil {
+			fmt.Printf("failed to update status: %s\n", err.Error())
```
shuqz: will remove it

controllers/gateway/gateway_controller.go
```diff
@@ -407,13 +424,27 @@ func (r *gatewayReconciler) updateGatewayStatusSuccess(ctx context.Context, lbSt
...
+			r.logger.Info("failed to build listeners status: %v", k8s.NamespacedName(gw))
+		} else if !isListenerStatusIdentical(gw.Status.Listeners, ListenerStatuses) {
+			gw.Status.Listeners = ListenerStatuses
+			needPatch = true
+		}
+	}
+
 	if needPatch {
 		if err := r.k8sClient.Status().Patch(ctx, gw, client.MergeFrom(gwOld)); err != nil {
+			fmt.Printf("failed to update status: %s\n", err.Error())
```
shuqz: will remove it

controllers/gateway/gateway_controller.go
```diff
@@ -237,10 +248,6 @@ func (r *gatewayReconciler) reconcileHelper(ctx context.Context, req reconcile.R
 
 	stack, lb, newAddOnConfig, backendSGRequired, secrets, err := r.buildModel(ctx, gw, mergedLbConfig, allRoutes, currentAddOns)
 
-	if err != nil {
```
zac-nixon, line 240: Seems this got removed, then didn't get re-added at a later place.

controllers/gateway/gateway_controller.go
```diff
@@ -394,8 +401,18 @@ func (r *gatewayReconciler) updateGatewayStatusSuccess(ctx context.Context, lbSt
 		needPatch = true
 	}
 
+	// update listeners status
+	ListenerStatuses, err := buildListenerStatus(r.controllerName, *gw, attachedRoutesMap, nil)
```
zac-nixon: It seems this not possible to fail, can you remove returning an error?

pkg/gateway/routeutils/listener_utils.go
```diff
@@ -0,0 +1,175 @@
...
+			result.IsValid = false
+			result.Reason = gwv1.ListenerReasonInvalidRouteKinds
+			result.Message = fmt.Sprintf("Invalid route kind for listener %s", listener.Name)
+			results.HasErrors = true
+		} else if listener.Port < 1 || listener.Port > 65535 {
+			result.IsValid = false
+			result.Reason = gwv1.ListenerReasonPortUnavailable
+			result.Message = fmt.Sprintf("Port %d is not available (listener name %s)", listener.Port, listener.Name)
+			results.HasErrors = true
+		} else if controllerName == gateway_constants.ALBGatewayController &&
```
zac-nixon, line 62: I don't think this will ever happen. As the loader will only load specific routes according to the gateway type:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/f1566e2d18d9d27cb99e3eff0388c57670f374ec/pkg/gateway/routeutils/loader.go#L103-L113

I don't know if we'll want this check, because it would require users to install the NLB CRDs (even if they don't intend to use them)

zac-nixon: Ah, I see. The user can specify routes manually, so we'll need this. Good catch.

pkg/gateway/routeutils/listener_utils.go
```diff
@@ -0,0 +1,175 @@
...
+					result.IsValid = false
+					result.Reason = gwv1.ListenerReasonHostnameConflict
+					result.Message = fmt.Sprintf("Hostname conflict for port %d with hostname %s", listener.Port, hostname)
+					results.HasErrors = true
+				} else {
+					portHostnameMap[listener.Port][hostname] = true
+				}
+			}
+
+			// Check cross-namespace references
```
zac-nixon: This looks duplicated, and potentially incorrect - https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/f1566e2d18d9d27cb99e3eff0388c57670f374ec/pkg/gateway/routeutils/backend.go#L277

pkg/gateway/routeutils/listener_utils.go
```diff
@@ -0,0 +1,175 @@
...
+						result.Message = fmt.Sprintf("Protocol conflict for port %d", listener.Port)
+						results.HasErrors = true
+					}
+				}
+			} else {
+				portProtocolMap[listener.Port] = listener.Protocol
+			}
+
+			// Check hostname conflicts - only when hostname is specified
+			if listener.Hostname != nil {
```
zac-nixon, line 91: Is this check duplicated too?

shuqz: it is different from hostname check. the existing hostname check checks if hostname in Route and hostname in listeners are compatible. but this hostname check here checks if hostnames in listeners have conflict.

pkg/gateway/routeutils/listener_utils.go
shuqz: correct me if i am wrong, i felt it is a bit different? the referenceGrantCheck we had is checking From and To. but this one here is checking if another ns is specified here, there should be a referenceGrant for it
```
      allowedRoutes:
        namespaces:
          from: Selector
          selector:
            matchLabels:
              name: xx-random-ns
```

pkg/gateway/routeutils/listener_utils.go
zac-nixon: Ok, I read more so the actual conflict would be two listeners both using `foo.com` or `bar.com`. Do you know what configuration we produce when you have something like
```
  - allowedRoutes:
      namespaces:
        from: Same
    hostname: foo.com
    name: https
    port: 443
    protocol: HTTPS
  - allowedRoutes:
      namespaces:
        from: Same
    hostname: foo.com
    name: https-2
    port: 443
    protocol: HTTPS
```
?

pkg/gateway/routeutils/listener_utils.go
zac-nixon: Right, I think you're doing a subset of the checks already performed at the ReferenceGrant check that I linked. In the linked check, we are
1: Resolving the namespace (using the selector if given)
2. Looking up the RefenceGrants in the namespace: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/f1566e2d18d9d27cb99e3eff0388c57670f374ec/pkg/gateway/routeutils/backend.go#L135-L136
3. Ensuring that to and from are correct.

The check that you added is just doing (2)

controllers/gateway/route_reconciler.go
```diff
@@ -261,7 +262,7 @@ func (d *routeReconcilerImpl) setConditionsWithRouteStatusInfo(route client.Obje
 			{
 				Type:               string(gwv1.RouteConditionResolvedRefs),
 				Status:             metav1.ConditionTrue,
-				Reason:             string(gwv1.RouteReasonAccepted),
+				Reason:             string(gwv1.RouteReasonResolvedRefs),
```
zac-nixon, line 265: nice catch

controllers/gateway/utils.go
```diff
@@ -12,19 +18,15 @@ import (
...
 )
 
 const (
 	gatewayClassAnnotationLastProcessedConfig          = "elbv2.k8s.aws/last-processed-config"
 	gatewayClassAnnotationLastProcessedConfigTimestamp = gatewayClassAnnotationLastProcessedConfig + "-timestamp"
 
 	// The max message that can be stored in a condition
-	maxMessageLength = 32700
+	maxMessageLength            = 32700
+	gatewayAcceptedFalseMessage = "Gateway is not accepted because there is invalid listener."
```
zac-nixon: Gateway is not accepted because there is *an* invalid listener

pkg/gateway/routeutils/listener_utils.go
shuqz: i do not think this is allowed, so the combination of hostname+port+protocol needs to be unique. it is gonna fail at validation during apply

pkg/gateway/routeutils/listener_utils.go
shuqz: in this case, if i just get result from referebceGrant check, then it is doing more checks than needed for RefNotPermit error (it does not need step 1 and 3) should i keep my change in validateListner and just remove step2 from ReferenceGrant check?

zac-nixon: Well, preferably we don't fetch the same reference grant twice. Is there some way to incorporate the existing reference grant check into the listener status?

pkg/gateway/routeutils/listener_utils.go
shuqz: synced offline, we are good here

pkg/gateway/routeutils/listener_utils.go
shuqz: synced offline, gonna remove this part. refNotpermitted is related to tls specifically

pkg/gateway/routeutils/listener_utils.go
shuqz:
```
The Gateway "shuqz-test-gw-listener-status" is invalid: spec.listeners: Invalid value: "array": Combination of port, protocol and hostname must be unique for each listener
```

```
* spec.listeners: Invalid value: "array": Listener name must be unique within the Gateway
```

Issue #4360, opened 2025-09-27
Limit the number of targets being added to an ELB instance


mmiller-sh:
**Describe the feature you are requesting**

While load testing our workloads, we've run into a failure scenario that is particularly catastrophic if unanticipated: Exceeding the ["Targets per Target Group"](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-limits.html#target-groups-quotas) quota for a given ELB instance will result in failed reconciliation until the number of targets drops below the quota. The consequences of this is that all targets in the ELB can become stale over time, resulting in complete outages. While it makes obvious sense to request quota increases, or update your networking architecture to avoid hitting these limits, a mechanism to limit the max number of targets being added to a group could avoid an _unanticipated_ scaling event from blocking the reconciliation loop.

I intend to contribute this feature if it will be accepted.

**Motivation**

**Describe the proposed solution you'd like**
An optional controller flag that sets the maximum number of targets that can be added to a TargetGroup, taking the first X number of targets from the list of all endpoints to apply to the ELB. The default would remain the current behavior.

**Describe alternatives you've considered**

**Contribution Intention (Optional)**

-[x] Yes, I am willing to contribute a PR to implement this feature
-[ ] No, I cannot work on a PR at this time

PR #4361, opened 2025-09-28
feat: Add flag to limit the max number of targets added to an ELB

mmiller-sh:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4360 

### Description

This change adds an optional flag to limit the maximum number of targets that are attempted to be added to a particular load balancer instance. The primary use-case is to prevent "Max targets" quotas from being hit and blocking the reconciliation loop.

The existing behavior is preserved by default (no limits). 

### Checklist
- [x] Manually tested


pkg/targetgroupbinding/resource_manager.go
```diff
@@ -240,6 +243,9 @@ func (m *defaultResourceManager) reconcileWithIPTargetType(ctx context.Context,
 			return "", "", false, ctrlerrors.NewErrorWithMetrics(controllerName, "update_tracked_ip_targets_error", err, m.metricsCollector)
 		}
 
+		eligibleTargetsCount := m.getMaxNewTargets(len(unmatchedEndpoints), len(targets), tgbScopedLogger)
```
zac-nixon: instead of always doing the endpoint slicing, how about only doing it if `m.maxTargetsPerInstance != 0`

docs/deploy/configurations.md
```diff
@@ -117,6 +117,7 @@ Currently, you can set only 1 namespace to watch in this flag. See [this Kuberne
...
 | alb-gateway-max-concurrent-reconciles                                           | int                       | 3                                          | Maximum number of concurrently running reconcile loops for ALB gateways, if enabled                                                                                           |
 | nlb-gateway-max-concurrent-reconciles                                           | int                       | 3                                          | Maximum number of concurrently running reconcile loops for NLB gateways, if enabled                                                                                           |
+| max-targets-per-instance                                                        | int                       | 0                                          | Maximum number of targets that will be added to a given ELB instance. The default value of zero will leave the number of targets unlimited                                    |
```
zac-nixon: name is a little confusing, per your issue link you ran into the max number of targets in a target group limitation. how about `max-targets-per-target-group`?

to expand, `instance` is pretty general, and it's not clear if you're referring to targets per load balancer, targets per target group, how many pods can get packed into a node, etc. There is also a separate limits: 

https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-limits.html `Targets per Application Load Balancer`

https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-limits.html `Targets per Network Load Balancer`


In this case, it looks like you're enforcing the targets per target group quota.

mmiller-sh: Thanks @zac-nixon. I agree `max-targets-per-target-group` makes more sense for this. I've updated the PR to reflect this, as well as addressed your other comment regarding limiting when we perform the endpoint slice.

zac-nixon:
/lgtm
/approved

thanks for the contribution!

wweiwei-li (approved):

demikl:
Hello,  
Thank you very much for this PR — our team was really excited about this new parameter, especially for large clusters with several hundred nodes.

Our use case:  
We have several Ingress resources that **do not use node targeting via `alb.ingress.kubernetes.io/target-node-labels`**, which results in Target Groups ending up with *all* the nodes from the cluster. The introduction of `max-targets-per-target-group` looked like an ideal way to avoid this situation.

However, we noticed an unexpected behavior:  
We initially assumed that this limit would apply **only to Ingresses configured with `alb.ingress.kubernetes.io/target-type: instance`**, since this mode relies on node-level targeting.  
Conversely, we expected that Ingresses running in **IP mode**, which directly register pods as targets, **would not be affected**.

But in practice, the limit also applies to IP mode.

As a result:  
For one of our deployments with more than 10 replicas, exposed through an IP-mode Ingress, setting

    max-targets-per-target-group=10

causes **only a subset of the pods** to be registered in the Target Group, and therefore only that subset receives traffic.

This behavior seems potentially problematic, because a limit intended for instance mode Target Groups can introduce significant side effects in IP mode.

We would appreciate your feedback on this:

*   Is this the expected behavior?
*   Would it be possible for the limit to apply *only* to instance mode?
*   Or for an option to differentiate the two modes?

Thanks again for your work on the controller and for your help!

mmiller-sh:
> Hello, Thank you very much for this PR — our team was really excited about this new parameter, especially for large clusters with several hundred nodes.
> 
> Our use case: We have several Ingress resources that **do not use node targeting via `alb.ingress.kubernetes.io/target-node-labels`**, which results in Target Groups ending up with _all_ the nodes from the cluster. The introduction of `max-targets-per-target-group` looked like an ideal way to avoid this situation.
> 
> However, we noticed an unexpected behavior: We initially assumed that this limit would apply **only to Ingresses configured with `alb.ingress.kubernetes.io/target-type: instance`**, since this mode relies on node-level targeting. Conversely, we expected that Ingresses running in **IP mode**, which directly register pods as targets, **would not be affected**.
> 
> But in practice, the limit also applies to IP mode.
> 
> As a result: For one of our deployments with more than 10 replicas, exposed through an IP-mode Ingress, setting
> 
> ```
> max-targets-per-target-group=10
> ```
> 
> causes **only a subset of the pods** to be registered in the Target Group, and therefore only that subset receives traffic.
> 
> This behavior seems potentially problematic, because a limit intended for instance mode Target Groups can introduce significant side effects in IP mode.
> 
> We would appreciate your feedback on this:
> 
> * Is this the expected behavior?
> * Would it be possible for the limit to apply _only_ to instance mode?
> * Or for an option to differentiate the two modes?
> 
> Thanks again for your work on the controller and for your help!

Using this in IP mode was actually the exact case we were testing against when opening the PR. Yes, it is true that only a subset of the Pods will receive traffic in this mode, however, this was still considered a large improvement compared to blocking controller reconciliation when ALB target group quotas were being hit (our use-case).

Issue #4030, opened 2025-01-22
Allow K8s Service to override controller default tags by annotation


jtackaberry:
**Is your feature request related to a problem?**

We define default tags at the AWS LB controller level, with the intention that these tags should be used as, well, defaults for the environment if the LoadBalancer type K8s Service doesn't specify `service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags`.

Unfortunately that annotation doesn't override default tags, which I understand is clearly documented but doesn't actually provide the behavior we need, which is that default tags are _defaults_ unless they are overridden by more granular resources.

**Describe the solution you'd like**

Even though IMO this violates the Principle of Least Astonishment, I accept  that the behavior of `service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags` is by now well established and probably can't change without disruption, so I propose a new Service annotation `service.beta.kubernetes.io/aws-load-balancer-resource-tags-overrides` or some such, where the defaults defined at the controller level are used but _can_ be overridden by any tags specified in this new annotation.

**Describe alternatives you've considered**

We'll have to drop default tags entirely from the AWS LB controller and instead vigilantly police all K8s Service objects for compliance instead.  This isn't ideal.

andreybutenko:
Hi! Thank you for the issue

`service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags` is an annotation that enables *additional* tags. Today, there is no annotation to override tags at the service level

Could you tell us more about your usecase, why you'd like to override tags at the service level? Thanks!

jtackaberry:
Hi @andreybutenko 

> service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags is an annotation that enables additional tags. Today, there is no annotation to override tags at the service level

Yep, understood, which is why I suggested a new annotation, one which can introduce new semantics without affecting use of the existing annotation.

> Could you tell us more about your usecase, why you'd like to override tags at the service level?

Sure.

At my organization, we often have EKS clusters which host multiple services that have different billing attributes and/or different support teams.  It's an internal requirement that all cloud resources contain specific tags to indicate contact information and service-level information (an identifier we call a service record).  The more granular the tag values the better -- contact details for specific teams rather than departmental (or platform team) contacts, specific services rather than higher level service families.

[...]

PR #4384, opened 2025-10-09
feat: add feature toggle to allow controller default tags to be overridden by annotation supplied tags

mwhittington21:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4030

### Description

Adds an option to override the existing behaviour where tags added to the controller via the `--default-tags` argument will override any annotation-specified tags. When the feature is enabled, annotation-specified tags will now override default tags. This had to be added as a toggle-able feature because there are likely many customers that are relying on the current behaviour.

The flag for the new behaviour has been added as a feature gate. It could also be an argument, but the main reason I chose to use a feature gate was to minimise refactoring involved. Feature gates are already passed through the entire codebase, so the amount of scaffolding changes were minimised. Tests were easier to write, and existing tests needed minimal changes.

The feature gate is called `EnableDefaultTagsLowPriority`, indicating that when enabled it will change the default tags to be of lower priority than other tags.

The majority of changes are in tests. Every other change is just an extra conditional statement in all the areas of code that merge two dicts of tags. I chose to add these conditional statements rather than change the tag merge logic and add an argument as it would only touch the parts of the code related to tags. Anything else using that function remains untouched.

### Checklist
- [x] Manually tested


zac-nixon:
Thanks for the contribution, it makes sense to me. This PR made me realized how much code duplication we have in our service and ingress implementations 👎.

/lgtm
/approved

mwhittington21:
> Thanks for the contribution, it makes sense to me. This PR made me realized how much code duplication we have in our service and ingress implementations 👎.
> 
> /lgtm /approved

@zac-nixon yeah, I thought about having a go at refactoring it but I didn't want to have a forked PR that would need consistent rebasing for this. Hopefully it can get addressed at some future point 😅. One tagging subsystem would have made it a lot easier. It looks like this started in the gateway API package, but hasn't made it back to the other packages.

Issue #4388, opened 2025-10-13
Use of wafv2-acl-name annotation causes panic


kh34:
**Bug Description**

After upgrading the AWS Load Balancer Controller to v2.14.0, use of wafv2-acl-name annotation causes panic. The controller keeps emitting:

```
{
    "level": "error",
    "ts": "2025-10-13T17:18:01Z",
    "msg": "Observed a panic",
    "controller": "ingress",
    "object": {
        "name": "my-ingress",
        "namespace": "my-namespace"
    },
    "namespace": "my-namespace",
    "name": "my-ingress",
    "reconcileID": "f15a4de7-d07c-4415-86e3-6c9ce40d4e4f",
    "panic": "runtime error: invalid memory address or nil pointer dereference",
    "panicGoValue": "\"invalid memory address or nil pointer dereference\"",
    "stacktrace": "goroutine 459 [running[]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x7f4fae00b778, 0xc000fc7aa0}, {0x7f4fada8cac0, 0x7f4faf835470})
	/go/pkg/mod/k8s.io/apimachinery@v0.33.3/pkg/util/runtime/runtime.go:132 +0xbc
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...[]).Reconcile.func1()
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:108 +0x112
panic({0x7f4fada8cac0?, 0x7f4faf835470?})
	/usr/local/go/src/runtime/panic.go:792 +0x132
sigs.k8s.io/aws-load-balancer-controller/pkg/ingress.(*webACLNameToArnMapper).getArnByName(0x0, {0x7f4fae00b778, 0xc000fc7aa0}, {0xc002a023c0, 0x17})
	/workspace/pkg/ingress/model_builder.go:606 +0x5d
sigs.k8s.io/aws-load-balancer-controller/pkg/ingress.(*defaultModelBuildTask).buildWAFv2WebACLAssociation(0xc000d33488, {0x7f4fae00b778, 0xc000fc7aa0}, {0x7f4fadfcae40, 0xc00327f890})
	/workspace/pkg/ingress/model_build_load_balancer_addons.go:86 +0x405
[...]

PR #4390, opened 2025-10-14
fix waf name resolution

zac-nixon:
### Issue

[](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4388)

### Description

The initial implementation of grabbing the WAFv2 ACL via it's name was _incorrect_ :(. The Get API requires both the name and ID. To implement the `Get By Name` API, we need to utilize the ListWebACLs API to grab a list and filter it server side, unfortunately WAF doesn't provide a server side filtering option. Also, we were missing the scope setting, so I've set to be `Regional` which captures the ALB use case. Also, we forgot to set the name -> arn mapper which caused null pointers.

Manual test cases:

- Set a name
- Change a name (foo1 -> foo2)
- Set name to none, to validate it's removed.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/aws/services/wafv2.go
+++ b/pkg/aws/services/wafv2.go
@@ -10,7 +10,7 @@ type WAFv2 interface {
 	AssociateWebACLWithContext(context.Context, *wafv2.AssociateWebACLInput) (*wafv2.AssociateWebACLOutput, error)
 	DisassociateWebACLWithContext(ctx context.Context, req *wafv2.DisassociateWebACLInput) (*wafv2.DisassociateWebACLOutput, error)
 	GetWebACLForResourceWithContext(ctx context.Context, req *wafv2.GetWebACLForResourceInput) (*wafv2.GetWebACLForResourceOutput, error)
-	ListWebACLs(ctx context.Context, req *wafv2.ListWebACLsInput) (*wafv2.ListWebACLsOutput, error)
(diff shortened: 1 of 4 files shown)


pkg/aws/services/wafv2.go
```diff
@@ -10,7 +10,7 @@ type WAFv2 interface {
 	AssociateWebACLWithContext(context.Context, *wafv2.AssociateWebACLInput) (*wafv2.AssociateWebACLOutput, error)
 	DisassociateWebACLWithContext(ctx context.Context, req *wafv2.DisassociateWebACLInput) (*wafv2.DisassociateWebACLOutput, error)
 	GetWebACLForResourceWithContext(ctx context.Context, req *wafv2.GetWebACLForResourceInput) (*wafv2.GetWebACLForResourceOutput, error)
-	GetWebACLWithContext(ctx context.Context, req *wafv2.GetWebACLInput) (*wafv2.GetWebACLOutput, error)
+	ListWebACLs(ctx context.Context, req *wafv2.ListWebACLsInput) (*wafv2.ListWebACLsOutput, error)
```
wweiwei-li: For consistency, should we name it `ListWebACLsContext`. Other than this , looks good.

shraddhabang: +1

Issue #4373, opened 2025-10-02
ALB rule limit is reached and the controller can no longer delete or edit any ingress


lc-jeff-lemos:
**Bug Description**

I'm using the ALB controller along with the `group.name` annotation. When the ALB reaches the rules limit, the alb controller fails to reconcile  even if I delete an ingress to remove rules.

When it happens, I try to delete an ingress but it will never work. It only works if I manually delete the rules from the ALB.

How it usually happens:
1. I add a new ingress and it never reconciles, never get the ALB address.
2. I check the group.name, check the ALB and it has reached the limit
    1. In my case, we've  extended it to 200 rules per ALB 
3. I try to delete the recent created or any other ingress but it never happens.
4. I go to the aws console and manually remove the rules.

### My theory

I'm not completely sure how's the reconciling process but looks like when the ingress is deleted the alb controller  move its rule orders (higher than 49900), re-create the left rules with the new priority and then delete the ingress rules. As the ALB has no more space left, it fails.

**Steps to Reproduce**

- Step-by-step guide to reproduce the bug:
- Manifests applied while reproducing the issue:
- Controller logs/error messages while reproducing the issue:

1. Considering the default ALB limit, which is 100 rules:
    1. Create 20 ingress with 5 host + path based rules each
    2. All within the same `group.name`
    3. It's better to apply 18 or less at once and then one by one apply the other until it reaches the limit
4. Wait for it to reach the rules limit the alb controller reconciliation 
5. Get any rule from the middle of the rules list, identify its ingress
    1. It will fail too if a rule (via alb controller when deleting the the ingress) from the beginning  of the list is deleted
    2. It WILL NOT fail if a rule from the end of the list is deleted (via alb controller when deleting the the ingress) .
7. Delete the ingress
8. The alb controller will start to log the error

Example of a manifest:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    alb.ingress.kubernetes.io/certificate-arn: arn:aws:acm:us-west-2:xxxxx:certificate/xxxxxx5d
    alb.ingress.kubernetes.io/group.name: alb-ingress
    alb.ingress.kubernetes.io/healthcheck-interval-seconds: "15"
    alb.ingress.kubernetes.io/healthcheck-path: /health-check
    alb.ingress.kubernetes.io/healthcheck-port: traffic-port
    alb.ingress.kubernetes.io/healthcheck-protocol: HTTP
[...]

PR #4393, opened 2025-10-16
[bug fix] handle rules exceeded error in listener rule synthesizer

zac-nixon:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4373

### Description

Fixes situation described in #4373. When the listener rule synthesizer moves rules into the last priorities (50,000 - 49900) and the listener is at all the rule limit, then the synthesizer gets stuck. This is because we always attempt to create rules prior to deleting, in order to prevent requests from not getting 404 errors due to missing listener rules. At first, I thought that the modify rule path should handle this, but closer inspection told me that we only call modify rule if the priority of the new rule matched the priority of a rule that we were deleting. This means that the only time you can successfully replace a rule is when you're at the maximum rules for a listener is to modify the rule at the exact priority (and perform no other modifications!). 

The solution I came up with is to create a greedy algorithm that attempts to create all desired rules, up until we detect the listener is it's max rule limit. At this point of time, we start flip-flopping between delete and create in order to minimize time between rules missing on the listener. Finally, once all missing rules are created, we will delete all remaining undesired rules.

Some other solutions that I played around with..

1/ Always call modify rule. This meant determining which unmatched SDK and result LRs to map together to modify. The downside to this approach is that we can't guarantee that the customer provided priority ordering would match on the listener. This is because the Modify api doesn't allow priority modification. I was afraid of misrouting scenarios because of corrupted priority ordering.
2/ Something along the lines of again trying to map unmatched sdk and result LRs and perform the create / delete in that order. It was pretty hard to come up with a correct algorithm to do this matching.. 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/deploy/elbv2/listener_rule_synthesizer.go
+++ b/pkg/deploy/elbv2/listener_rule_synthesizer.go
@@ -93,21 +93,6 @@ func (s *listenerRuleSynthesizer) synthesizeListenerRulesOnListener(ctx context.
 		return err
 	}
 
-	s.logger.Info("unmatchedResLRs size", "size", len(unmatchedResLRs))
-	for _, resLR := range unmatchedResLRs {
-		s.logger.Info("Unmatched", "res lr", *resLR)
-	}
-
-	s.logger.Info("unmatchedSDKLRs size", "size", len(unmatchedSDKLRs))
-	for _, sdkLr := range unmatchedSDKLRs {
-		s.logger.Info("Unmatched", "res lr", *sdkLr.ListenerRule)
-	}
-
-	s.logger.Info("match priority size", "size", len(matchedResAndSDKLRsBySettings))
-	for k, v := range matchedResAndSDKLRsBySettings {
-		s.logger.Info("matched priority", "priority", k, "value", *v.sdkLR.ListenerRule)
-	}
-
 	// Re-prioritize matched listener rules.
 	if len(matchedResAndSDKLRsBySettings) > 0 {
 		err := s.lrManager.SetRulePriorities(ctx, matchedResAndSDKLRsBySettings, unmatchedSDKLRs)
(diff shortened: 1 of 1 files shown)


pkg/deploy/elbv2/listener_rule_synthesizer.go
```diff
@@ -89,6 +92,22 @@ func (s *listenerRuleSynthesizer) synthesizeListenerRulesOnListener(ctx context.
 	if err != nil {
 		return err
 	}
+
+	s.logger.Info("unmatchedResLRs size", "size", len(unmatchedResLRs))
+	for _, resLR := range unmatchedResLRs {
+		s.logger.Info("Unmatched", "res lr", *resLR)
```
shraddhabang: Lets add these to debug logs.

pkg/deploy/elbv2/listener_rule_synthesizer.go
```diff
@@ -210,6 +221,72 @@ func (s *listenerRuleSynthesizer) matchResAndSDKListenerRules(resLRs []*elbv2mod
...
+		if len(unmatchedResLRs) > resLRIndex && ruleCounter != maxRules {
+			// We should prioritize the creation of rules. If we've reached the maximum number of rules on the listener,
+			// we need to perform a deletion instead.
+			resLR := unmatchedResLRs[resLRIndex]
+			lrStatus, err := s.lrManager.Create(ctx, resLR, resLRDesiredActionsAndConditionsPairs[resLR])
+			if err != nil {
+				// Detect too many rules error, we want this exception to only trigger once.  triggering multiple times
+				// either means outside forces are changing listener rules OR we have a bug.
+				if isTooManyRulesErr(err) && maxRules == -1 {
+					maxRules = ruleCounter
```
shraddhabang, line 248: Can we add a debug log here that we have started flip-flop logic now for troubleshooting purposes?

PR #4404, opened 2025-10-20
[gw api] add disable security group flag to lb config for gateway users

zac-nixon:
### Description

This ports over a behavior from the Service API which allows users to provision NLBs with no security groups attached. This only applies to NLB gateways, as ALB most always have an SG attached.

### Checklist
- [x] Manually tested

### DIFF
--- a/apis/gateway/v1beta1/loadbalancerconfig_types.go
+++ b/apis/gateway/v1beta1/loadbalancerconfig_types.go
@@ -233,6 +233,11 @@ type LoadBalancerConfigurationSpec struct {
 	// +optional
 	ListenerConfigurations *[]ListenerConfiguration `json:"listenerConfigurations,omitempty"`
 
+	// disableSecurityGroup provisions a load balancer with no security groups.
+	// Allows an NLB to be provisioned with no security groups.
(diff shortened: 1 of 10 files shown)


shraddhabang:
NIT: Can you also add a follow up e2e test for this?

apis/gateway/v1beta1/loadbalancerconfig_types.go
```diff
@@ -233,6 +233,11 @@ type LoadBalancerConfigurationSpec struct {
 	// +optional
 	ListenerConfigurations *[]ListenerConfiguration `json:"listenerConfigurations,omitempty"`
 
+	// disableSecurityGroup provisions a load balancer with no security groups.
+	// Allows an NLB to be provisioned with no security groups.
+	// [Network Load Balancer]
+	DisableSecurityGroup *bool `json:"disableSecurityGroup,omitempty"`
```
shuqz, line 239: what's the default value for DisableSecurityGroup? do we wanna enforce it here/mention it in description?

zac-nixon: It's false.

Issue #4391, opened 2025-10-14
Gateway API: certificate discovery for HTTPRoute hostnames


artem-collectai:
**Describe the feature you are requesting**
Extend certificate discovery to the hostnames defined on HTTPRoute level.

**Motivation**
Certificate discovery currently only works for the hostnames defined on the Gateway level, in the listeners configurations. This makes it inconvenient to manage an ALB that serves multiple domains: ideally, we would have one HTTPRoute per domain, whereas the central listener configuration would remain constant.

**Describe the proposed solution you'd like**
In `model_build_listener.go`, pass `routes` array down to `buildL7ListenerSpec -> buildListenerSpec -> buildCertificates` and include all their hostnames into the list for which certificate ARNs are inferred.

**Describe alternatives you've considered**
Support for ListenerSets would also do the trick (https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4320) - however, ListenerSets are at the moment in experimental status.

**Contribution Intention (Optional)**

- [ ] Yes, I am willing to contribute a PR to implement this feature
- [x] No, I cannot work on a PR at this time

zac-nixon:
Yeah, this was an oversight in our implementation of cert detection. Thanks for bringing it up and providing a solution. We'll make sure to get this included ASAP.

We'd like to avoid experimental features as much as possible, as we don't want to experience too much churn which comes with experimental apis.

zac-nixon:
This was included in v2.14.1

PR #4407, opened 2025-10-22
[gw api] discover certs for route hostnames

zac-nixon:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4391

### Description

The Ingress API will discover certs for all host names, on the listener and each rule. We missed that for the Gateway API, so I am back porting the same logic.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/gateway/model/model_build_listener.go
+++ b/pkg/gateway/model/model_build_listener.go
@@ -27,7 +27,7 @@ import (
 // TODO: Add more relevant info like TLS settings and hostnames later wherever applicable
 type gwListenerConfig struct {
 	protocol  elbv2model.Protocol
-	hostnames []string
+	hostnames sets.Set[string]
 }
 
 type listenerBuilder interface {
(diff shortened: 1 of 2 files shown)


pkg/gateway/model/model_build_listener.go
```diff
@@ -26,8 +26,8 @@ import (
 
 // TODO: Add more relevant info like TLS settings and hostnames later wherever applicable
 type gwListenerConfig struct {
-	protocol  elbv2model.Protocol
-	hostnames []string
+	protocol     elbv2model.Protocol
+	hostnamesFoo sets.Set[string]
```
shraddhabang: Can we change the name of this variable to `hostnames`?

zac-nixon: oops, i changed this because i wanted to see all uses of this field 😮‍💨

PR #4417, opened 2025-10-28
[feat gw-api]handle hostname intersection

shuqz:
### Description
we were not properly handle case when listener has hostname but route does not, therefore failing conformance test - HTTPRouteListenerHostnameMatching
This implementation covers this gap by 
- setting compatible hostname per port for route
- use compatible hostnames in listener builder phase and gw listener configuration (for certs discovery on hostnames)

**Case 1 (route attaches to several listeners with different ports):**

Port 80 rules use hostnames: ["api.example.com"]

Port 443 rules use hostnames: ["*.example.com"]

Final result: {80: ["api.example.com"], 443: ["*.example.com"]} 

**Case 2 (route attaches to several listeners with Same Port):**

Port 80 rules use hostnames: ["api.example.com", "web.example.com"]

Both hostnames are available for precedence sorting. Final result: {80: ["api.example.com", "web.example.com"]} -> this will be treated as ALB listener rule with host header=api.example.com OR web.example.com

### Checklist
- [ ] Manually tested

re-run HTTPRouteListenerHostnameMatching and got 
```
    --- PASS: TestConformance/HTTPRouteListenerHostnameMatching (245.50s)
        --- PASS: TestConformance/HTTPRouteListenerHostnameMatching/6_request_to_'foo.com/'_should_receive_one_of_[] (0.13s)
        --- PASS: TestConformance/HTTPRouteListenerHostnameMatching/2_request_to_'baz.bar.com/'_should_go_to_infra-backend-v3 (0.13s)
        --- PASS: TestConformance/HTTPRouteListenerHostnameMatching/7_request_to_'no.matching.host/'_should_receive_one_of_[] (0.14s)
        --- PASS: TestConformance/HTTPRouteListenerHostnameMatching/3_request_to_'boo.bar.com/'_should_go_to_infra-backend-v3 (0.15s)
        --- PASS: TestConformance/HTTPRouteListenerHostnameMatching/5_request_to_'multiple.prefixes.foo.com/'_should_go_to_infra-backend-v3 (0.15s)
        --- PASS: TestConformance/HTTPRouteListenerHostnameMatching/1_request_to_'foo.bar.com/'_should_go_to_infra-backend-v2 (0.15s)
        --- PASS: TestConformance/HTTPRouteListenerHostnameMatching/0_request_to_'bar.com/'_should_go_to_infra-backend-v1 (0.15s)
        --- PASS: TestConformance/HTTPRouteListenerHostnameMatching/4_request_to_'multiple.prefixes.bar.com/'_should_go_to_infra-backend-v3 (0.15s)
```

### DIFF
--- a/pkg/gateway/model/model_build_listener_test.go
+++ b/pkg/gateway/model/model_build_listener_test.go
@@ -431,7 +431,7 @@ func Test_buildListenerTags(t *testing.T) {
 			},
 			externalManagedTags: []string{"ExternalTag", "ManagedByTeam"},
 			expectedTags:        nil,
-			expectedErr:         errors.New("external managed tag key ExternalTag cannot be specified"),
+			expectedErr:         errors.New("external managed tag key"),
 		},
 	}
 
@@ -448,6 +448,7 @@ func Test_buildListenerTags(t *testing.T) {
 			if tt.expectedErr != nil {
 				assert.Error(t, err)
 				assert.Contains(t, err.Error(), tt.expectedErr.Error())
+				assert.True(t, strings.Contains(err.Error(), "ExternalTag") || strings.Contains(err.Error(), "ManagedByTeam"))
 				assert.Nil(t, got)
 			} else {
 				assert.NoError(t, err)
--- a/pkg/gateway/routeutils/descriptor.go
+++ b/pkg/gateway/routeutils/descriptor.go
@@ -28,10 +28,9 @@ type routeMetadataDescriptor interface {
 	// When a route attaches to multiple listeners on the same port, hostnames are aggregated.
 	// When a route attaches to listeners on different ports, each port has its own hostname list.
 	GetCompatibleHostnamesByPort() map[int32][]gwv1.Hostname
-	// SetCompatibleHostnamesByPort sets the compatible hostnames for each listener port.
-	// This is called after route attachment validation to store the port-specific hostnames
-	// that should be used for routing rules and precedence sorting.
-	SetCompatibleHostnamesByPort(map[int32][]gwv1.Hostname)
+	// setCompatibleHostnamesByPort is a package-private method to set compatible hostnames.
+	// This is called by the loader after route attachment validation.
+	setCompatibleHostnamesByPort(map[int32][]gwv1.Hostname)
 }
 
 type routeLoadError struct {
--- a/pkg/gateway/routeutils/http.go
+++ b/pkg/gateway/routeutils/http.go
@@ -141,7 +141,7 @@ func (httpRoute *httpRouteDescription) GetCompatibleHostnamesByPort() map[int32]
 	return httpRoute.compatibleHostnamesByPort
 }
 
-func (httpRoute *httpRouteDescription) SetCompatibleHostnamesByPort(hostnamesByPort map[int32][]gwv1.Hostname) {
+func (httpRoute *httpRouteDescription) setCompatibleHostnamesByPort(hostnamesByPort map[int32][]gwv1.Hostname) {
 	httpRoute.compatibleHostnamesByPort = hostnamesByPort
 }
 
--- a/pkg/gateway/routeutils/grpc.go
+++ b/pkg/gateway/routeutils/grpc.go
@@ -148,7 +148,7 @@ func (grpcRoute *grpcRouteDescription) GetCompatibleHostnamesByPort() map[int32]
 	return grpcRoute.compatibleHostnamesByPort
 }
 
-func (grpcRoute *grpcRouteDescription) SetCompatibleHostnamesByPort(hostnamesByPort map[int32][]gwv1.Hostname) {
+func (grpcRoute *grpcRouteDescription) setCompatibleHostnamesByPort(hostnamesByPort map[int32][]gwv1.Hostname) {
 	grpcRoute.compatibleHostnamesByPort = hostnamesByPort
 }
 
--- a/pkg/gateway/routeutils/loader.go
+++ b/pkg/gateway/routeutils/loader.go
@@ -197,7 +197,7 @@ func (l *loaderImpl) loadChildResources(ctx context.Context, preloadedRoutes map
 			}
 		}
 		if len(hostnamesByPort) > 0 {
-			route.SetCompatibleHostnamesByPort(hostnamesByPort)
+			route.setCompatibleHostnamesByPort(hostnamesByPort)
 		}
 	}
 
--- a/pkg/gateway/routeutils/loader_test.go
+++ b/pkg/gateway/routeutils/loader_test.go
@@ -41,7 +41,7 @@ func (m *mockRoute) GetCompatibleHostnamesByPort() map[int32][]gwv1.Hostname {
 	return m.CompatibleHostnamesByPort
 }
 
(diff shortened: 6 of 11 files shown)


pkg/gateway/routeutils/utils.go
```diff
@@ -122,20 +123,29 @@ func IsHostNameInValidFormat(hostName string) (bool, error) {
 // isHostnameCompatible checks if given two hostnames are compatible with each other
 // this function is used to check if listener hostname and Route hostname match
 func isHostnameCompatible(hostnameOne, hostnameTwo string) bool {
+	_, isCompatible := getCompatibleHostname(hostnameOne, hostnameTwo)
+	return isCompatible
+}
+
+// getCompatibleHostname returns the more specific hostname if two hostnames are compatible
+// Returns the compatible hostname and true if compatible, empty string and false otherwise
+func getCompatibleHostname(hostnameOne, hostnameTwo string) (string, bool) {
```
shraddhabang, line 138: Can you write unit tests for this?

shuqz: yea i forgot it somehow, will add it

pkg/gateway/routeutils/descriptor.go
```diff
@@ -21,6 +21,8 @@ type routeMetadataDescriptor interface {
 	GetRouteListenerRuleConfigRefs() []gwv1.LocalObjectReference
 	GetRouteGeneration() int64
 	GetRouteCreateTimestamp() time.Time
+	GetCompatibleHostnames() []gwv1.Hostname
+	SetCompatibleHostnames([]gwv1.Hostname)
```
zac-nixon: It would be great to add a comment saying what a "compatible hostname" is.

pkg/gateway/routeutils/http.go
```diff
@@ -51,9 +51,10 @@ func (t *convertedHTTPRouteRule) GetListenerRuleConfig() *elbv2gw.ListenerRuleCo
 /* Route Description */
 
 type httpRouteDescription struct {
-	route           *gwv1.HTTPRoute
-	rules           []RouteRule
-	ruleAccumulator attachedRuleAccumulator[gwv1.HTTPRouteRule]
+	route               *gwv1.HTTPRoute
+	rules               []RouteRule
+	ruleAccumulator     attachedRuleAccumulator[gwv1.HTTPRouteRule]
+	compatibleHostnames []gwv1.Hostname
```
zac-nixon: How does this work for a route that attaches to two different listeners in a Gateway?

pkg/gateway/routeutils/listener_attachment_helper.go
```diff
@@ -171,8 +171,17 @@ func (attachmentHelper *listenerAttachmentHelperImpl) kindCheck(listener gwv1.Li
 }
 
 func (attachmentHelper *listenerAttachmentHelperImpl) hostnameCheck(listener gwv1.Listener, route preLoadRouteDescriptor) (bool, error) {
-	// A route can attach to listener if it does not have hostname or listener does not have hostname
-	if listener.Hostname == nil || len(route.GetHostnames()) == 0 {
+	// If route has no hostnames but listener does, use listener hostname
+	if len(route.GetHostnames()) == 0 {
```
zac-nixon: Don't we need a nil check still?

pkg/gateway/routeutils/descriptor.go
zac-nixon: From gateway documentation, 
```
Hostnames defines a set of SNI names that should match against the
SNI attribute of TLS ClientHello message in TLS handshake. This matches
the RFC 1123 definition of a hostname with 2 notable exceptions:
1. IPs are not allowed in SNI names per RFC 6066.
2. A hostname may be prefixed with a wildcard label (*.). The wildcard
label must appear by itself as the first label.
If a hostname is specified by both the Listener and TLSRoute, there
must be at least one intersecting hostname for the TLSRoute to be
attached to the Listener. For example:
A Listener with test.example.com as the hostname matches TLSRoutes
that have either not specified any hostnames, or have specified at
least one of test.example.com or *.example.com.
A Listener with *.example.com as the hostname matches TLSRoutes
that have either not specified any hostnames or have specified at least
one hostname that matches the Listener hostname. For example,
test.example.com and *.example.com would both match. On the other
hand, example.com and test.example.net would not match.
If both the Listener and TLSRoute have specified hostnames, any
TLSRoute hostnames that do not match the Listener hostname MUST be
ignored. For example, if a Listener specified *.example.com, and the
TLSRoute specified test.example.com and test.example.net,
test.example.net must not be considered for a match.
If both the Listener and TLSRoute have specified hostnames, and none
match with the criteria above, then the TLSRoute is not accepted. The
implementation must raise an 'Accepted' Condition with a status of
False in the corresponding RouteParentStatus.
```

pkg/gateway/model/model_build_listener.go
```diff
@@ -509,8 +509,17 @@ func mapGatewayListenerConfigsByPort(gw *gwv1.Gateway, routes map[int32][]routeu
 
 		if listenerRoutes != nil {
 			for _, route := range listenerRoutes {
-				for _, routeHostname := range route.GetHostnames() {
-					gwListenerConfigs[port].hostnames.Insert(string(routeHostname))
+				// Use compatible hostnames (intersection) instead of raw route hostnames
+				compatibleHostnamesByPort := route.GetCompatibleHostnamesByPort()[port]
```
zac-nixon, line 513: I don't think we need the `SetCompatible...` function anymore if the calculation is done within the loader, or atleast it doesn't need to be exposed `setCompatible..`

zac-nixon (approved):

pkg/gateway/routeutils/descriptor.go
shuqz: synced offline

pkg/gateway/model/model_build_listener.go
shuqz: i think i prefer to go with keep setter and make it private. if i remove setXxxx, and with the fact that Getxxx is got be called in multiple places, i need to add CompatibleHostnamesByPort field to LoaderResult and pass it as param when needed. felt either method works

pkg/gateway/model/model_build_listener_test.go
```diff
@@ -448,6 +448,7 @@ func Test_buildListenerTags(t *testing.T) {
 			if tt.expectedErr != nil {
 				assert.Error(t, err)
 				assert.Contains(t, err.Error(), tt.expectedErr.Error())
+				assert.True(t, strings.Contains(err.Error(), "ExternalTag") || strings.Contains(err.Error(), "ManagedByTeam"))
```
shuqz, line 451: this test has been flaky 
```
--- FAIL: Test_buildListenerTags (0.00s)
--- FAIL: Test_buildListenerTags/external_managed_tags_specified_by_user_should_cause_error (0.00s)
model_build_listener_test.go:450:
Error Trace: /Users/shuqz/go/src/github.com/sigs.k8s.io/shuqz-aws-load-balancer-controller/pkg/gateway/model/model_build_listener_test.go:450
Error: "external managed tag key ManagedByTeam cannot be specified" does not contain "external managed tag key ExternalTag cannot be specified"
Test: Test_buildListenerTags/external_managed_tags_specified_by_user_should_cause_error
FAIL
```

PR #4418, opened 2025-10-28
Support URLRewrite in HTTPRoutes

zac-nixon:
### Description

Supports path and host header rewrites using URLRewrite found in HTTPRouteRule. https://gateway-api.sigs.k8s.io/guides/http-redirect-rewrite/#rewrites

I put lots of descriptions in the code. The TL;DR is that we can fully support both path and header based rewrites using the newly launched ALB URLRewrite feature. One piece that still needs figuring out is that ALB has more powerful rewrite capabilities than listed here. Originally, I was going to use the ListenerRuleConfiguration object to specify it. However, to keep this change manageable, I have only implemented the Gateway API spec.

### Checklist
- [x] Manually tested

### DIFF
--- a/test/e2e/gateway/alb_ip_target_test.go
+++ b/test/e2e/gateway/alb_ip_target_test.go
@@ -4,9 +4,6 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"strings"
-	"time"
-
 	awssdk "github.com/aws/aws-sdk-go-v2/aws"
 	elbv2types "github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2/types"
 	"github.com/gavv/httpexpect/v2"
@@ -24,6 +21,8 @@ import (
 	"sigs.k8s.io/aws-load-balancer-controller/test/framework/utils"
 	"sigs.k8s.io/aws-load-balancer-controller/test/framework/verifier"
 	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+	"strings"
+	"time"
 )
 
 var _ = Describe("test k8s alb gateway using ip targets reconciled by the aws load balancer controller", func() {
@@ -195,7 +194,6 @@ var _ = Describe("test k8s alb gateway using ip targets reconciled by the aws lo
 			})
 		})
 	})
-
 	Context("with ALB ip target configuration with HTTPRoute specified matches", func() {
 		BeforeEach(func() {})
 		It("should provision internet-facing load balancer resources", func() {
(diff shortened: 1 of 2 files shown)


pkg/gateway/routeutils/route_rule_transform.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+const (
+	replaceWholeHostHeaderRegex           = ".*"
+	replaceWholePathMinusQueryParamsRegex = "^([^?]*)"
+)
+
+func BuildRoutingRuleTransforms(gwRoute RouteDescriptor, gwRule RulePrecedence) []elbv2model.Transform {
+	switch gwRoute.GetRouteKind() {
+	case HTTPRouteKind:
```
shraddhabang, line 17: I am supposing we will add Transforms for GRPCRoute through LrCfg right?

pkg/gateway/routeutils/route_rule_transform.go
```diff
@@ -0,0 +1,119 @@
...
+			},
+		},
+	}
+}
+
+func generatePrefixReplacementRegex(httpMatch *gwv1.HTTPRouteMatch, replacement string) (string, string) {
+	match := *httpMatch.Path.Value
+
+	/*
+		If we're being asked to replace a prefix with "", we still need to keep one '/' to form a valid path.
```
shraddhabang, line 92: Great job on explaining this in details. 
We should also add good documentation with these samples in our docs for this feature once we implement it with ELB gaps.

pkg/gateway/routeutils/route_validator_test.go
```diff
@@ -0,0 +1,260 @@
...
+							{
+								Path: &gwv1.HTTPPathMatch{
+									Type:  &exact,
+									Value: awssdk.String("/foo"),
+								},
+							},
+						},
+						Filters: []gwv1.HTTPRouteFilter{
+							{
+								URLRewrite: &gwv1.HTTPURLRewriteFilter{
```
shraddhabang: I would have preferred if the spec added this validations instead of us doing this for our implementation. Any reason that spec is not doing this on their side?

shraddhabang:
Can you also add E2E tests for this?

pkg/gateway/routeutils/route_rule_transform.go
zac-nixon: It doesn't make sense to support path rewrites with GRPC. But, yes we could support header re-writes for GRPC traffic using the lrc.

pkg/gateway/routeutils/route_rule_transform.go
```diff
@@ -0,0 +1,119 @@
...
+	replaceWholeHostHeaderRegex           = ".*"
+	replaceWholePathMinusQueryParamsRegex = "^([^?]*)"
+)
+
+func BuildRoutingRuleTransforms(gwRoute RouteDescriptor, gwRule RulePrecedence) []elbv2model.Transform {
+	switch gwRoute.GetRouteKind() {
+	case HTTPRouteKind:
+		return buildHTTPRuleTransforms(gwRule.CommonRulePrecedence.Rule.GetRawRouteRule().(*gwv1.HTTPRouteRule), gwRule.HTTPMatch)
+	default:
+		return []elbv2model.Transform{}
```
shuqz, line 20: for any other cases, do we want to add a v(1) logging or warning saying only httpRoute is supported for transform?

pkg/gateway/routeutils/route_rule_transform.go
```diff
@@ -0,0 +1,119 @@
...
+					transforms = append(transforms, generateHostHeaderRewriteTransform(*rf.URLRewrite.Hostname))
+				}
+			}
+		}
+	}
+
+	return transforms
+}
+
+func generateHostHeaderRewriteTransform(hostname gwv1.PreciseHostname) elbv2model.Transform {
```
shuqz, line 44: maybe rename it to generateHttpHostHeaderRewriteTransform, same for generateURLRewritePathTransform

pkg/gateway/routeutils/route_rule_transform.go
```diff
@@ -0,0 +1,119 @@
...
+					Regex:   replacementRegex,
+					Replace: replacement,
+				},
+			},
+		},
+	}
+}
+
+func generatePrefixReplacementRegex(httpMatch *gwv1.HTTPRouteMatch, replacement string) (string, string) {
+	match := *httpMatch.Path.Value
```
shuqz, line 89: i did not see where we are checking nil for httpMatch

pkg/gateway/routeutils/route_rule_transform.go
```diff
@@ -0,0 +1,119 @@
...
+		A couple examples:
+
+		Without the capture group, e.g. (^%s)
+		input path = '/foo/', prefixRegex = '(^/foo)', replacement value = '/cat/' results in '/cat//'
+
+		To extend the example, now consider using having the capture group and always adding that to the result.
+		input path = '/foo/', prefixRegex = '(^/foo(/)?)', replacement value = '/cat/$2' results in (again) '/cat//'
+
+		Without the capture group, we would have one '/' too few.
+		input path = '/foo/bar', prefixRegex = '(^/foo(/)?)', replacement value = '/cat$2' results in '/catbar'
```
shuqz, line 109: thanks for all explanation. can you verify if `*httpMatch.Path.Value`(match) can be something like `/foo/` or replacement can be with Trailing Slash? those cases will lead to double slash right?

pkg/gateway/routeutils/loader.go
```diff
@@ -108,16 +110,20 @@ func (l *loaderImpl) LoadRoutesForGateway(ctx context.Context, gw gwv1.Gateway,
...
 		}
 	}
 
-	// validate listeners configuration and get listener status
+	// 2. Validate listeners configuration and get listener status
 	listenerValidationResults := ValidateListeners(gw, controllerName, ctx, l.k8sClient)
 
-	// 2. Remove routes that aren't granted attachment by the listener.
+	// 3. Validate route configuration, filter out bad configurations.
+	validatedRoutes, validationErrors := l.routeValidator.filterToValidRoutes(gw, foo)
```
shuqz: can you attach a screenshot of this status update in route to show a failed case with correct failed reason?

pkg/gateway/routeutils/route_rule_transform.go
zac-nixon: I don't think so. This is already documented within the gateway api

zac-nixon:
E2E test result -
```
Ran 2 of 18 Specs in 915.365 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 16 Skipped
PASS
```

pkg/gateway/routeutils/route_rule_transform.go
zac-nixon: any explanation here? :)

pkg/gateway/routeutils/route_rule_transform.go
zac-nixon: it's done in the routeValidator

pkg/gateway/routeutils/route_rule_transform.go
zac-nixon: Yup it can, I actually have an explicit test case for this, name = `prefix path rewrite with explicit '/' on suffix`. To answer your question, no we won't add a double slash, because we check if the replacement ends with a `/`, in that case we don't append the optional capture group.

pkg/gateway/routeutils/loader.go
zac-nixon: Turns out route validator is not needed - 
```
# httproutes.gateway.networking.k8s.io "http-app-1" was not valid:
# * spec.rules[0]: Invalid value: "object": When using URLRewrite filter with path.replacePrefixMatch, exactly one PathPrefix match must be specified
```

pkg/gateway/routeutils/route_validator_test.go
zac-nixon: Turns out route validator is not needed - 
```
# httproutes.gateway.networking.k8s.io "http-app-1" was not valid:
# * spec.rules[0]: Invalid value: "object": When using URLRewrite filter with path.replacePrefixMatch, exactly one PathPrefix match must be specified
```

test/e2e/gateway/alb_ip_target_test.go
```diff
@@ -666,7 +664,139 @@ var _ = Describe("test k8s alb gateway using ip targets reconciled by the aws lo
...
+										},
+									},
+								},
+							},
+						},
+						Priority: 1,
+					},
+				})
+				Expect(err).NotTo(HaveOccurred())
+			})
```
shraddhabang, line 797: Can we verify the transform are applied in the response by sending the http request?

zac-nixon: We don't do this for other listener rules, I don't see a reason to do it here.

pkg/gateway/routeutils/route_rule_transform.go
shuqz: because it feels like this function is only for HTTPRoute, but it is a nit change

PR #4432, opened 2025-11-05
Correct enum for healthCheckProtocol field

dex4er:
### Description

The enumeration doesn't match validation.

Currently `https` is silently replaced with `HTTP` internally then it is not possible to set the health check for non-HTTP protocol.

### Checklist
- [ ] Manually tested


zac-nixon:
/ok-to-test

oops :)

PR #4434, opened 2025-11-06
[feat aga] Add AGA accelerator deployer

shraddhabang:
### Description

This PR introduces support for AGA deployer allowing Kubernetes users to manage Global Accelerator resources through Kubernetes custom resources. I have divided this PR into smaller commits to make it easier to review. 

Here is the brief description on each of them

1. Setup AGA SDK Client (a39933f0)
Adds AWS Global Accelerator SDK integration with the controller. Implements client interfaces, mock implementations, and cloud provider extensions to interact with the Global Accelerator service. Updates dependencies to include required AWS SDK packages for Global Accelerator.

2. Add AGA CRD Status Updates Utils (c34f2eb0)
Implements status updating utilities for Global Accelerator CRDs. Provides mechanisms to reflect the state of AWS Global Accelerator resources back to Kubernetes custom resources, enabling users to track resource status, lifecycle events, and error conditions directly in Kubernetes.

3. Add AGA Tags Reconciler (9484b803)
Implements tagging functionality for Global Accelerator resources to maintain proper resource ownership and tracking. Provides interfaces and implementations for applying, updating, and reconciling tags on AWS Global Accelerator resources with comprehensive test coverage. Please note I have moved tracking tags to deployer instead of builders. Also I have added one more region specific tags so that multiple controller in same account in different region do not override each other

4. Add AGA Controller Config Flags (96951375)
I have disabled AGAController for now until its ready for GA. We will make this enabled by default when we are ready to release this for prod. We have also added exponential back-off duration flags for AGA controller which customer can configure. I wanna also know if we should make requeue time configurable as well. 

5. Add AGA Deployer (47a7becc)
Implements the core deployment logic for Global Accelerator resources. This includes:
    - Accelerator Manager for coordinating resource lifecycle operations
    - Synthesizer for transforming Kubernetes model objects into AWS resources
    - Stack deployer for managing resource dependencies and deployment ordering
    - Error handling and lifecycle management for Global Accelerator resources
    
6. update controller-gen version(8dfe5223d)
Updating the controller-gen version to use v0.19.0
To fix the failing CRD verifications in CI/CD

TODO: I will add e2e tests once we build all the resources separately as this was already getting very big PR. 
TODO: I have not added IAM policies. I wanna add when we are done with the upcoming release. 

### Checklist
- [x] Manually tested


pkg/aga/model_build_accelerator.go
```diff
@@ -39,6 +40,7 @@ var _ acceleratorBuilder = &defaultAcceleratorBuilder{}
 type defaultAcceleratorBuilder struct {
 	trackingProvider tracking.Provider
 	clusterName      string
+	clusterRegion    string
```
shraddhabang, line 43: There is a possibility that the same account may host the clusters in different regions and each may have controller running in that cluster. In that situations only cluster name would not suffice to identify the resources as they can have cluster with same name in different regions. So we will need the region as well to uniquely identify the resource managed by controller and avoid potential overrides.

pkg/aga/model_build_accelerator.go
```diff
@@ -126,14 +129,5 @@ func (b *defaultAcceleratorBuilder) buildAcceleratorTags(_ context.Context, stac
 		return nil, err
 	}
 
-	// Add tracking tags (includes cluster tag and stack tag)
```
shraddhabang, line 129: We are adding these tracking tags in deployer phase now. So dont need these here.

pkg/deploy/aga/tagging_manager.go
```diff
@@ -0,0 +1,236 @@
...
+	defer m.resourceTagsCacheMutex.Unlock()
+
+	// Check if the ARN is in cache
+	if rawTagsCacheItem, exists := m.resourceTagsCache.Get(arn); exists {
+		tagsCacheItem := rawTagsCacheItem.(map[string]string)
+		return tagsCacheItem, nil
+	}
+
+	// ARN not in cache, need to fetch from RGT API
+	tags, err := m.describeResourceTagsFromRGT(ctx, arn)
```
shraddhabang, line 160: For AGAController, we will need RGT APIs to be enabled by default so using these here as first check and fallback to AGA API in case of failures.

pkg/status/aga/status_updater.go
```diff
@@ -0,0 +1,307 @@
...
+	// Reason constants
+	ReasonAcceleratorReady        = "AcceleratorReady"
+	ReasonAcceleratorProvisioning = "AcceleratorProvisioning"
+	ReasonAcceleratorDisabling    = "AcceleratorDisabling"
+	ReasonAcceleratorDeleting     = "AcceleratorDeleting"
+
+	// Status constants
+	StatusDeployed   = "DEPLOYED"
+	StatusInProgress = "IN_PROGRESS"
+	StatusDeleting   = "Deleting"
```
wweiwei-li: for consistentcy, should this be StatusDeleting   = "DELETING"

pkg/status/aga/status_updater.go
```diff
@@ -0,0 +1,307 @@
...
+	var needPatch bool
+
+	// Update observed generation
+	if ga.Status.ObservedGeneration == nil || *ga.Status.ObservedGeneration != ga.Generation {
+		ga.Status.ObservedGeneration = &ga.Generation
+		needPatch = true
+	}
+
+	// Set status to "Deleting" to indicate it's in the process of being deleted
+	if ga.Status.Status == nil || *ga.Status.Status != StatusDeleting {
```
wweiwei-li, line 229: is the intent here : if status isn't already 'Deleting', then set it to 'Deleting ?   

API seems like does not have 'Delete' status.  Should the CRD match API ?
```
Status -> (string)

Describes the deployment status of the accelerator.

Possible values:

DEPLOYED
IN_PROGRESS
```

pkg/status/aga/status_updater.go
```diff
@@ -0,0 +1,307 @@
...
+	ReasonAcceleratorDeleting     = "AcceleratorDeleting"
+
+	// Status constants
+	StatusDeployed   = "DEPLOYED"
+	StatusInProgress = "IN_PROGRESS"
+	StatusDeleting   = "Deleting"
+)
+
+// StatusUpdater handles GlobalAccelerator resource status updates
+type StatusUpdater interface {
```
wweiwei-li, line 34: I like we have different folder for status updates

pkg/status/aga/status_updater.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+
+	// Update DNS name
+	if ga.Status.DNSName == nil || *ga.Status.DNSName != accelerator.Status.DNSName {
+		ga.Status.DNSName = &accelerator.Status.DNSName
+		needPatch = true
+	}
+
+	// Update dual stack DNS name if available
+	if accelerator.Status.DualStackDNSName != "" {
```
wweiwei-li, line 97: In a lot of places.  We are just setting status.  For example, set DualStackDNSName here.  Do we also need to clear DualStackDNSName in some case ? 

I checked how ingress update status.  They are replace entire arrays of status, which automatically clear .

pkg/deploy/aga/tagging_manager.go
```diff
@@ -0,0 +1,236 @@
+package aga
```
wweiwei-li, line 1: Just out of curiosity, would it be possible to have deploy/tagging so code can be shared with all resources? If it's a lot of changes, ignore me. I did see each resource has its own tagging_manager.

pkg/aga/model_build_accelerator.go
wweiwei-li: I found an interesting thing.  AGA can has same name with different ARN.   I start wondering what the use case is here ? @emdelaney. is this because they want to use same name for different use cases ? 

In that case, I am thinking instead of adding region to name.  could we add a cluster region tag so we only find resource in the cluster regions

```
aws globalaccelerator list-accelerators --region us-west-2
{
    "Accelerators": [
        {
            "AcceleratorArn": "arn:aws:globalaccelerator::875258973867:accelerator/ffdd5de1-55e4-4701-a667-caf0db9f1621",
            "Name": "testAga",
            "IpAddressType": "IPV4",
            "Enabled": true,
            "IpSets": [
                {
                    "IpFamily": "IPv4",
                    "IpAddresses": [
                        "52.223.1.197",
                        "166.117.219.76"
                    ],
                    "IpAddressFamily": "IPv4"
                }
            ],
            "DnsName": "a5f6fd3469a8e384f.awsglobalaccelerator.com",
            "Status": "DEPLOYED",
            "CreatedTime": 1762556585.0,
            "LastModifiedTime": 1762556586.0
        },
        {
            "AcceleratorArn": "arn:aws:globalaccelerator::875258973867:accelerator/35365048-8014-45f4-80db-3e13b2311bbd",
            "Name": "testAga",
            "IpAddressType": "IPV4",
            "Enabled": true,
            "IpSets": [
                {
                    "IpFamily": "IPv4",
                    "IpAddresses": [
                        "52.223.11.213",
                        "166.117.185.124"
                    ],
                    "IpAddressFamily": "IPv4"
                }
            ],
            "DnsName": "a5cc273802627ab3a.awsglobalaccelerator.com",
            "Status": "DEPLOYED",
            "CreatedTime": 1762554553.0,
            "LastModifiedTime": 1762554553.0
        }
    ]
}
```

controllers/aga/globalaccelerator_controller.go
```diff
@@ -212,25 +232,92 @@ func (r *globalAcceleratorReconciler) reconcileGlobalAcceleratorResources(ctx co
...
+	if err := acceleratorManager.Delete(ctx, acceleratorWithTags); err != nil {
+		// Check if it's an AcceleratorNotDisabledError
+		var notDisabledErr *agadeploy.AcceleratorNotDisabledError
+		if errors.As(err, &notDisabledErr) {
+			// Update status to indicate we're waiting for the accelerator to be disabled
+			if updateErr := r.statusUpdater.UpdateStatusDeletion(ctx, ga); updateErr != nil {
+				r.logger.Error(updateErr, "Failed to update status during accelerator deletion")
+			}
+			// Requeue after 30 seconds to check again
+			return ctrlerrors.NewRequeueNeededAfter("Waiting for accelerator to be disabled", 30*time.Second)
```
emdelaney: Is there a pattern for exponential backoff available? Some kind of ramp up would be good if we have a pattern for it already because most updates are fully applied within a minute or two, but sometimes changes get delayed for longer. Some kind of exponential or planned backoff would provide a better (faster) experience when changes are flowing at the usual rate.

controllers/aga/globalaccelerator_controller.go
```diff
@@ -212,25 +232,92 @@ func (r *globalAcceleratorReconciler) reconcileGlobalAcceleratorResources(ctx co
...
+	if acceleratorARN == "" {
+		r.logger.Info("Empty accelerator ARN in status, nothing to clean up", "globalAccelerator", k8s.NamespacedName(ga))
+		return nil
+	}
+
+	// 2. Delete the accelerator using accelerator delete manager
+	acceleratorManager := r.stackDeployer.GetAcceleratorManager()
+	r.logger.Info("Deleting accelerator", "acceleratorARN", acceleratorARN, "globalAccelerator", k8s.NamespacedName(ga))
+
+	// Create accelerator with ARN to delete
```
emdelaney: (nit) A different phrasing than "create accelerator" might be good here, since we aren't creating an accelerator in the AWS-resource sense.

pkg/aga/model_build_accelerator.go
emdelaney: AGA's name field is similar to EC2's instance name: a convenient, human-readable name for the accelerator. The Name of an accelerator can be updated at any time: https://docs.aws.amazon.com/global-accelerator/latest/api/API_UpdateAccelerator.html#globalaccelerator-UpdateAccelerator-request-Name

I would not rely on the name for identification of a specific resource in the controller, if we can avoid it; the ARN is the only immutable + unique identifier for an accelerator. The tag-based approach sounds great for tracking which accelerators were created by a given cluster in a given region. 

Having said all that, in the Global Accelerator console, the Name is the only customer-configured field that is visible in the list of all accelerators worldwide. So being able to go from the Name to the specific cluster that generated it would be useful to humans who are trying to find the accelerator for a given cluster in the console. If we need the region to do that, it's probably worth spending 15 characters (out 64 available) on it.

pkg/deploy/aga/accelerator_manager.go
```diff
@@ -0,0 +1,271 @@
...
+	return m.buildAcceleratorStatus(updatedAccelerator), nil
+}
+
+func (m *defaultAcceleratorManager) Delete(ctx context.Context, sdkAccelerator AcceleratorWithTags) error {
+	acceleratorARN := *sdkAccelerator.Accelerator.AcceleratorArn
+	m.logger.Info("Deleting accelerator", "acceleratorARN", acceleratorARN)
+
+	// Step 1: Try to disable the accelerator first if it's enabled
+	m.logger.Info("Disabling accelerator before deletion", "acceleratorARN", acceleratorARN)
+	isAlreadyDeleted, err := m.disableAccelerator(ctx, acceleratorARN)
```
emdelaney: Can we avoid a second describe call in disableAccelerator if we pass in the sdkAccelerator here? (or is it stale and in need of refreshing?)

pkg/deploy/aga/tagging_manager.go
emdelaney: Only accelerators can be tagged.

pkg/status/aga/status_updater.go
emdelaney: AGA doesn't have a service-side construct of deleting - once it's deleted, it's gone.

shraddhabang: This is used for displaying on CRD while its being deleted since it may take some time for accelerator to be disabled. So till that time, the CRD resource will be in deleting status.

pkg/status/aga/status_updater.go
shraddhabang: I specifically want to show deleting status on our resource instead of In_progess because its much more clear.  Deletion through the controller is two step process(Disable and Delete which can take some time) which we handle for them. And disabling can take upto 5 minutes. So they need to differentiate between create/updates/delete calls.

pkg/deploy/aga/tagging_manager.go
shraddhabang: Yes thats why we only need one tagging_manager. Thanks Matt.

pkg/status/aga/status_updater.go
shraddhabang: Good catch. Yes we need to clear this.

pkg/aga/model_build_accelerator.go
shraddhabang: @wweiwei-li I am not relying on names at all for identifying the accelerator. That is done purely through ARN. 
Matt has explained it well. There are two reasons I added this for.
- In `buildAcceleratorName()`, the `clusterRegion` is included in the SHA256 hash that generates unique names. This helps with human identification in the console listing as the accelerators from different regions with same name would be uniquely named by controllers. 

- Adding a cluster region tag for tracking and filtering for customers if the hash is not enough. They can always group the accelerators through tags to identify their cluster managed resources.

controllers/aga/globalaccelerator_controller.go
shraddhabang: Yes. This is good catch. This is leftover from my testing. I have fixed this. 
We do have a default exponential backoff strategy in controller runtime. I have implemented it [here](https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4434/files#diff-bd38cfbeee6cccf2996a0644452455cf377231ed9f9f897aac40b472137cc478R349). The base-delay I have set for now it 5 ms and default max back-off is 16ms (This is configurable through controller flag). We use these values for our other load balancer controllers too. So if the customer wants faster feedback, they can always lower their value for the max-back off. What do you think about this? Is this too aggressive? Do we wanna increase base delay?

pkg/status/aga/status_updater.go
shraddhabang: Updated.

controllers/aga/globalaccelerator_controller.go
shraddhabang: Done.

pkg/deploy/aga/accelerator_manager.go
shraddhabang: You are right. We can avoid the call when we have the enabled status returned already from the describe calls when we land here from PostSynthesis. But when we land here from clean up, we dont have the current status of the accelerator. we only know the ARN from status. So we need to make sure that it is disabled before we can attempt the deletions. I have modified the logic to reflect this.

controllers/aga/globalaccelerator_controller.go
```diff
@@ -212,25 +232,92 @@ func (r *globalAcceleratorReconciler) reconcileGlobalAcceleratorResources(ctx co
...
+	r.logger.Info("Cleaning up GlobalAccelerator resources", "globalAccelerator", k8s.NamespacedName(ga))
+
+	// TODO we will handle cleaning up dependent resources when we implement those
+	// 1. Find the accelerator ARN from the CRD status
+	if ga.Status.AcceleratorARN == nil {
+		r.logger.Info("No accelerator ARN found in status, nothing to clean up", "globalAccelerator", k8s.NamespacedName(ga))
+		return nil
+	}
+
+	acceleratorARN := *ga.Status.AcceleratorARN
```
zac-nixon, line 285: I think we need a fallback mechanism. Suppose that we provision an accelerator but are unable to persist the ARN. If before the next reconcile run happens and the customer deletes the resource from the cluster, we will orphan the accelerator because of this line.

docs/deploy/configurations.md
```diff
@@ -106,6 +106,8 @@ Currently, you can set only 1 namespace to watch in this flag. See [this Kuberne
...
 | targetgroupbinding-max-concurrent-reconciles                                    | int                       | 3                                          | Maximum number of concurrently running reconcile loops for targetGroupBinding                                                                                                 |
 | targetgroupbinding-max-exponential-backoff-delay                                | duration              | 16m40s                                     | Maximum duration of exponential backoff for targetGroupBinding reconcile failures                                                                                             |
+| globalaccelerator-max-concurrent-reconciles                                     | int                       | 3                                          | Maximum number of concurrently running reconcile loops for GlobalAccelerator objects                                                                                          |
```
zac-nixon: 3 is probably overkill here, maybe default 1?

pkg/aws/provider/default_aws_clients_provider.go
```diff
@@ -76,15 +79,23 @@ func NewDefaultAWSClientsProvider(cfg aws.Config, endpointsResolver *endpoints.R
 		}
 	})
 
+	globalAcceleratorClient := globalaccelerator.NewFromConfig(cfg, func(o *globalaccelerator.Options) {
+		o.Region = "us-west-2" // Global Accelerator is a global service that requires us-west-2
```
zac-nixon, line 83: If this is the case, we shouldn't be spinning up the controller in other partitions like us-gov and cn.

controllers/aga/globalaccelerator_controller.go
emdelaney: Let's increase the base delay to 5 seconds for AGA (it takes more than milliseconds for the change to be applied worldwide).

pkg/deploy/aga/tagging_manager.go
```diff
@@ -0,0 +1,236 @@
...
+	// Check if the resource was found
+	for _, resource := range resources {
+		resourceArn := awssdk.ToString(resource.ResourceARN)
+		if resourceArn == arn {
+			return services.ParseRGTTags(resource.Tags), nil
+		}
+	}
+
+	// If resource not found in RGT, fall back to direct API
+	m.logger.V(1).Info("Resource not found in RGT, falling back to direct API", "arn", arn)
```
zac-nixon: This kinda defeats the purpose of using RGT if we can't use it as a valid source of truth.

shraddhabang: I was on fence for this too. I wanna keep RGT as source of truth only. But was worried if there are cases where it may fail. If you also feel the same, I can remove this.

pkg/aws/provider/default_aws_clients_provider.go
shraddhabang: Yeah that makes sense.

controllers/aga/globalaccelerator_controller.go
shraddhabang: I tried implementing this after your suggestion but realized that since we will do look up based on tags fro finding the accelerator, we may end up getting multiple resources if the customer configures these resource tags on any other accelerator manually (very rare but they could) and we might delete the unwanted resource due to this tags error and disrupt the traffic. Since this is edge case, I don't think its worth to delete the orphaned resource instead of cleaning up a resource which might be serving traffic. What do you think?

pkg/aws/provider/default_aws_clients_provider.go
shraddhabang: Updated

docs/deploy/configurations.md
shraddhabang: Done

PR #4436, opened 2025-11-10
[feat aga] Add AGA listener support without auto-discovery

shraddhabang:
### Description

This PR introduces the AGA listener lifecycle management in AGA controller. It has two commits as follows. 

## 1.  [[feat aga] Add AGA listener builder without auto-discovery](https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4436/commits/b2d62b341ee2e9640ed5267ff28acd2afb354985)
This commit introduces the model builder changes for AGA Listener Resources & Port Range Validation

- __Model Builder for Listener Resources__: Updated the listener model builder to directly use values specified in the CRD spec. We plan to set these values based on endpoints with auto-discovery if not specified. We will implement that later. This is simple implementation which directly builds object from CRD specs. 

- __Port Range Validation Webhook__: Implemented validation logic that prevents overlapping port ranges between listeners with the same protocol if specified in CRD. We will extend the same validation during auto-discovery. 

  - Groups port ranges by protocol (TCP/UDP)
  - Checks for overlaps within each protocol group
  - Returns validation errors when overlapping port ranges are detected

## 2. [[feat aga] Add AGA listener deployer with clean up](https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4436/commits/b8e33364fabc4e93c4670b52199a08c89d92cd7a)

This PR implements the complete AGA Listener deployer, which reconciles listener resources defined in Kubernetes with the AGA service with Intelligent Matching Algorithm. 

## Key Features

- __Two-phase Listener Matching Algorithm__: Implements a sophisticated matching algorithm that:

  1. First attempts to find exact matches based on protocol and port ranges
  2. Then falls back to similarity-based matching for remaining unmatched listeners
  3. Uses a scoring system that considers protocol, port overlap, and client affinity
  4. Minimizes resource churn by reusing existing listeners where possible

- __Optimized Operation Order__: The deployer performs operations in an order that minimizes service disruption:

  1. Delete only conflicting listeners that block updates
  2. Update existing matched listeners
  3. Delete non-conflicting unmatched listeners
  4. Create new listeners as needed

- __Comprehensive Test Suite__: Includes extensive test cases that verify:

  - Port conflict detection in various scenarios
  - Key generation for consistent matching
  - Similarity score calculation for partial matches
  - The complete matching workflow with both exact and similarity matches

## Cleanup Logic

The current implementation includes a simple cleanup mechanism that safely removes listeners while cleaning up the accelerators. A more sophisticated cleanup strategy that considers endpoint groups and other dependent resources will be implemented separately once all Global Accelerator resource types are supported. Future PRs will enhance the cleanup logic with more sophisticated resource dependency tracking once all GA resource types are supported.

### Checklist
- [x] Manually tested


pkg/aga/model_build_listener.go
```diff
@@ -0,0 +1,128 @@
...
+		ClientAffinity: clientAffinity,
+	}, nil
+}
+
+// buildListenerProtocol determines the protocol for the listener
+func buildListenerProtocol(_ context.Context, listener agaapi.GlobalAcceleratorListener) (agamodel.Protocol, error) {
+	if listener.Protocol == nil {
+		// TODO: Auto-discovery feature - Auto-determine protocol from endpoints if nil
+		// For now, default to TCP
+		return agamodel.ProtocolTCP, nil
```
emdelaney: What do you think about returning an error if the protocol isn't specified (until we have auto-discovery)?

pkg/aga/model_build_listener.go
```diff
@@ -0,0 +1,128 @@
...
+		return "", errors.Errorf("unsupported protocol: %s", *listener.Protocol)
+	}
+}
+
+// buildListenerPortRanges determines the port ranges for the listener
+func buildListenerPortRanges(_ context.Context, listener agaapi.GlobalAcceleratorListener) ([]agamodel.PortRange, error) {
+	if listener.PortRanges == nil {
+		// TODO: Auto-discovery feature - Auto-determine port ranges from endpoints if nil
+		// For now, default to port 80
+		return []agamodel.PortRange{{
```
emdelaney: Same here, I'd expect an error if the port isn't provided and we can't auto-discover it.

pkg/deploy/aga/listener_synthesizer.go
```diff
@@ -0,0 +1,532 @@
...
+	listenerManager ListenerManager
+	logger          logr.Logger
+	stack           core.Stack
+}
+
+func (s *listenerSynthesizer) Synthesize(ctx context.Context) error {
+	var resListeners []*agamodel.Listener
+	s.stack.ListResources(&resListeners)
+
+	// If no listeners are defined in the model, there's nothing to do
```
emdelaney: If there are listeners on the existing accelerator but not on the model, does that delete need to be handled here? Or does it happen somewhere else?

pkg/deploy/aga/listener_synthesizer.go
```diff
@@ -0,0 +1,532 @@
...
+		if !listenerIsConflicting[arn] {
+			nonConflictingListeners = append(nonConflictingListeners, sdkListener)
+		}
+	}
+
+	// STEP 2: Execute operations in correct order
+
+	// First, delete ONLY conflicting listeners (those that would block updates)
+	// TODO: When we implement endpoint groups, for a more comprehensive solution, we might also want to add the ability to
+	// migrate endpoint groups from these conflicting listeners to non-conflicting ones as much as possible.
```
emdelaney, line 154: AGA Endpoint Groups are a sub-resource of a specific listener, so they can't be moved to another listener unfortunately.

pkg/deploy/aga/listener_synthesizer.go
```diff
@@ -0,0 +1,532 @@
...
+	exactMatches, remainingResListeners, remainingSDKListeners := s.findExactMatches(resListeners, sdkListeners)
+
+	// For remaining listeners, use similarity-based matching
+	similarityMatches, unmatchedResListeners, unmatchedSDKListeners := s.findSimilarityMatches(
+		remainingResListeners, remainingSDKListeners)
+
+	// Combine exact and similarity matches
+	matchedPairs := append(exactMatches, similarityMatches...)
+
+	s.logger.V(1).Info("Matched listeners",
```
emdelaney, line 274: It might be useful to log the key+arn for the exact and similarity matches here? That way, if we ever do a match that the user finds surprising, they could understand why a given listener was updated or deleted.

pkg/aga/model_build_listener.go
shraddhabang: It will be done before GA hence I left it at default for now.

pkg/deploy/aga/listener_synthesizer.go
shraddhabang: By migration I meant, adding those endpoints first to relevant existing  listeners if applicable before deleting those. Because these deletions may lead to temp loss of endpoints from accelerator.

pkg/deploy/aga/listener_synthesizer.go
shraddhabang: Good catch. This needs to be handled.

pkg/deploy/aga/listener_synthesizer.go
shraddhabang: Good idea. let me add that.

apis/aga/v1beta1/globalaccelerator_types.go
```diff
@@ -48,6 +48,7 @@ const (
 )
 
 // PortRange defines the port range for Global Accelerator listeners.
+// +kubebuilder:validation:XValidation:rule="self.fromPort <= self.toPort",message="FromPort must be less than or equal to ToPort"
```
zac-nixon, line 51: How about making `ToPort`optional, that way users can just specify one value rather than two when they just want to expose a single port?

pkg/aga/model_build_listener.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	return portRanges, nil
+}
+
+// buildListenerClientAffinity determines the client affinity for the listener
+func buildListenerClientAffinity(_ context.Context, listener agaapi.GlobalAcceleratorListener) agamodel.ClientAffinity {
+	switch listener.ClientAffinity {
+	case agaapi.ClientAffinitySourceIP:
+		return agamodel.ClientAffinitySourceIP
+	case agaapi.ClientAffinityNone:
```
zac-nixon: nit: you can drop the `case agaapi.ClientAffinityNone:`

pkg/deploy/aga/accelerator_manager.go
```diff
@@ -162,7 +164,29 @@ func (m *defaultAcceleratorManager) Delete(ctx context.Context, sdkAccelerator A
...
+	if err != nil {
+		var apiErr *agatypes.AcceleratorNotFoundException
+		if errors.As(err, &apiErr) {
+			m.logger.Info("Accelerator not found, assuming already deleted", "acceleratorARN", acceleratorARN)
+			return nil
+		}
+		return fmt.Errorf("failed to list listeners for accelerator: %w", err)
+	}
+
+	for _, listener := range listeners {
```
zac-nixon, line 180: Is this necessary? I just tried out deleting an accelerator in the console and I didn't have to delete the listeners.

pkg/deploy/aga/listener_manager.go
```diff
@@ -0,0 +1,247 @@
...
+	// before deleting the listener (when those features are implemented)
+
+	m.logger.Info("Deleting listener", "listenerARN", listenerARN)
+
+	// First check if the listener exists to avoid errors on already deleted resources
+	descInput := &globalaccelerator.DescribeListenerInput{
+		ListenerArn: aws.String(listenerARN),
+	}
+
+	_, err := m.gaService.DescribeListenerWithContext(ctx, descInput)
```
zac-nixon: This seems expensive to do, we 2x the call volume. Could you just delete the listener always and then handle the error if it occurs?

zac-nixon: It seems you do handle the error below anyways. IMO, this check is not needed.

pkg/deploy/aga/listener_manager.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	m.logger.Info("Successfully deleted listener", "listenerARN", listenerARN)
+	return nil
+}
+
+// isSDKListenerSettingsDrifted checks if the listener configuration has drifted from the desired state
+func (m *defaultListenerManager) isSDKListenerSettingsDrifted(resListener *agamodel.Listener, sdkListener *ListenerResource) bool {
+	// Check if protocol differs
+	if string(resListener.Spec.Protocol) != string(sdkListener.Listener.Protocol) {
```
zac-nixon, line 193: Is there any implicit defaults that we need to worry about here? E.g. if the protocol is missing then it's assumed to be TCP. If that's the case, we will always say that the listener setting has drifted. We run into this problem with quite a few ELB features.

pkg/deploy/aga/listener_manager.go
```diff
@@ -0,0 +1,247 @@
...
+	return false
+}
+
+// arePortRangesEqual compares port ranges from the resource model and SDK
+func (m *defaultListenerManager) arePortRangesEqual(modelPortRanges []agamodel.PortRange, sdkPortRanges []agatypes.PortRange) bool {
+	if len(modelPortRanges) != len(sdkPortRanges) {
+		return false
+	}
+
+	// Since port ranges are unordered, we need to compare them as sets
```
zac-nixon, line 216: nit: We already import the set package, you can use it here.

pkg/deploy/aga/listener_synthesizer.go
```diff
@@ -0,0 +1,555 @@
...
+// listenerSynthesizer is responsible for synthesize Listener resources for a stack.
+type listenerSynthesizer struct {
+	gaClient        services.GlobalAccelerator
+	listenerManager ListenerManager
+	logger          logr.Logger
+	stack           core.Stack
+}
+
+func (s *listenerSynthesizer) Synthesize(ctx context.Context) error {
+	var resListeners []*agamodel.Listener
```
zac-nixon: this should go after the accelerator checks, IMO.

pkg/deploy/aga/accelerator_manager.go
shraddhabang: The console handles the deletion of dependent resources hence you did not notice it. We also need to delete all the dependent resources before we can actually delete the accelerator. 
Ref: https://docs.aws.amazon.com/global-accelerator/latest/api/API_DeleteAccelerator.html#:~:text=Delete%20an%20accelerator.%20Before%20you%20can%20delete%20an%20accelerator%2C%20you%20must%20disable%20it%20and%20remove%20all%20dependent%20resources%20(listeners%20and%20endpoint%20groups).%20To%20disable%20the%20accelerator%2C%20update%20the%20accelerator%20to%20set%20Enabled%20to%20false.

pkg/deploy/aga/listener_synthesizer.go
```diff
@@ -0,0 +1,555 @@
...
+	}
+
+	// Match resource listeners with existing SDK listeners
+	// - matchedResAndSDKListeners: pairs of resource and SDK listeners that will be updated
+	// - unmatchedResListeners: resource listeners that don't match any SDK listeners and will be created
+	// - unmatchedSDKListeners: SDK listeners that don't match any resource listeners and will be deleted
+	matchedResAndSDKListeners, unmatchedResListeners, unmatchedSDKListeners := s.matchResAndSDKListeners(resListeners, sdkListeners)
+
+	// Improved operation order to minimize traffic disruption:
+	// 1. Delete only conflicting listeners (that would block updates)
```
zac-nixon, line 99: Why can't you use update listener for conflicting listeners?
https://docs.aws.amazon.com/cli/latest/reference/globalaccelerator/update-listener.html

pkg/deploy/aga/listener_synthesizer.go
```diff
@@ -0,0 +1,555 @@
...
+			s.logger.Info("Accelerator not found in AWS, skipping listener listing",
+				"acceleratorARN", accARN)
+			return nil, nil
+		}
+		return nil, errors.Wrapf(err, "failed to list listeners for accelerator %s", accARN)
+	}
+
+	var listeners []*ListenerResource
+	for _, listener := range sdkListeners {
+		// Clone the listener as the range variable address is reused
```
zac-nixon: This seemed suspicious to me, this is what Q told me:
```
In Go 1.22+, this works correctly because loop variables are now per-iteration scoped. For earlier Go versions, your current approach with 
explicit cloning is the right solution.

If you need to support Go < 1.22, keep your current code. If you're on Go 1.22+, you can use the simplified version above.
```

Are you intending for this to be a shallow or deep copy? The current code is only shallow.

pkg/deploy/aga/listener_synthesizer.go
```diff
@@ -0,0 +1,555 @@
...
+
+// listenerPairScore holds a potential match with its similarity score
+type listenerPairScore struct {
+	resListener *agamodel.Listener
+	sdkListener *ListenerResource
+	score       int
+}
+
+// findSimilarityMatches matches remaining listeners based on similarity score
+func (s *listenerSynthesizer) findSimilarityMatches(resListeners []*agamodel.Listener, sdkListeners []*ListenerResource) (
```
zac-nixon, line 377: interesting approach, this would be useful for updating ELB listener rules as well. Is this similarity score based off of a well-known algorithm?

pkg/deploy/aga/listener_synthesizer.go
shraddhabang: The port range overlap is not allowed between listeners with same protocol. So the update calls will fail if we have listeners which have conflicting ranges which are not needed anymore. So to avoid that, we are finding the conflicting listeners which are intended to be deleted before update calls.

pkg/deploy/aga/utils.go
```diff
@@ -0,0 +1,117 @@
...
+// ResPortRangesToSet adds all ports from resource model PortRange slices to the provided portSet map
+func ResPortRangesToSet(portRanges []agamodel.PortRange, portSet map[int32]bool) {
+	for _, pr := range portRanges {
+		PortRangesToSet(pr.FromPort, pr.ToPort, portSet)
+	}
+}
+
+// GetAWSInt32Value safely gets the value from an AWS SDK Int32 pointer
+// Returns the value if pointer is not nil, or defaultValue otherwise
+func GetAWSInt32Value(ptr *int32, defaultValue int32) int32 {
```
zac-nixon: nit: this seem like valuable general utilities that could go among the shared utilities.

pkg/deploy/aga/accelerator_manager.go
zac-nixon: good to know. thanks for the doc link.

pkg/deploy/aga/listener_manager.go
emdelaney: Client Affinity defaults to NONE: https://docs.aws.amazon.com/global-accelerator/latest/api/API_CreateListener.html

The ports and protocols must be specified.

shraddhabang: Thanks. I was just confirming this. We will always set Protocols and Ports from our model build. And Client Affinity is None by default in our case too. So we wont see the false drift problems.

pkg/deploy/aga/listener_synthesizer.go
shraddhabang: The idea came from hamming distance algo and Jaccard Index which are standard similarity algorithms. I had this  algorithm idea when I was working with rules management for ELB. But it seemed complex for it since our rules are much more complex. I was worried about performance at that time. So never actually implemented it.
Since the AGA listeners were much more easier structures, it was easier to implement the algo on them and I did not want to repeat our rules problem again. But I think we can definitely give it a try now for our rules management as well.

apis/aga/v1beta1/globalaccelerator_types.go
shraddhabang: I thought about this. But I want to keep this structure similar to match the accelerator APIs. Consistency in API design makes it easier for users familiar with AGA to work with controller. IMO,  making `ToPort` optional creates ambiguity about whether it means a single port or an open-ended range. I don't intuitively feel very convinced that it means single port. I understand the desire to simplify single-port configuration, but I believe explicit is better than implicit here.

pkg/deploy/aga/listener_synthesizer.go
shraddhabang: We only intend to use shallow copy since we wont modify sdkListeners. These are only used for comparison. Good catch. Since we are using Go 1.24, we can use simplified version. Updated

pkg/deploy/aga/utils.go
shraddhabang: These are leftovers and not used anywhere in the code. So removed those.

pkg/deploy/aga/listener_synthesizer.go
zac-nixon: Thanks for the thorough explanation. I find it helpful to detail the algorithm in the code itself. That way developers don't need to come in find these PR comments when working on this piece of code.

shraddhabang: Sure good call. I have added that.

zac-nixon (approved):

PR #4446, opened 2025-11-13
[feat gw-api]update reference grant check and route status

shuqz:
### Description
- Modified route status behavior
   - Before: when resolvedRef is false, we mark accepted as false too. Now (aligns with conformance test requirement): only resolvedRef reason can impact resolvedRef status. same for Accepted. 
   - this change got us passed following test cases: 
   - HTTPRouteInvalidBackendRefUnknownKind
   - HTTPRouteInvalidCrossNamespaceBackendRef
   - HTTPRouteInvalidNonExistentBackendRef
   - HTTPRoutePartiallyInvalidViaInvalidReferenceGrant
- Modified reference grant check logic
   - we are returning 503 instead 500 when no backend service, conformance test is expecting 500
   - we did not do From.Group and To.Group check, added it
   - this change got us passed following test cases: 
   - HTTPRouteInvalidReferenceGrant
   - HTTPRouteReferenceGrant
- Modified e2e test
   - based on above logic modification, changed corresponding e2e test
   - explicitly set targetType as instance in alb/nlb instance test cases. we were not explicitly set instance type and reply on using default value. but when i was doing test, since i have `--default-target-type=ip` in deployment, it is overriding the default value. i want to eliminating similar issue to e2e tests by explicitly setting it to instance (just like we set it to ip in ip tests)
   - minor: added tls section in listeners, we do not process it but it needs to be provided to pass validation and continue test (found it during e2e testing)

### Checklist
- [x] Manually tested
```
ginkgo -v -r test/e2e/gateway --  -ginkgo.focus="with ALB ip target configuration with basic HTTPRoute"
Ran 1 of 28 Specs in 552.775 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 27 Skipped
PASS

ginkgo -v -r test/e2e/gateway --  -ginkgo.focus="with ALB instance target configuration with basic HTTPRoute"
Ran 1 of 28 Specs in 554.888 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 27 Skipped
PASS

ginkgo -v -r test/e2e/gateway --  -ginkgo.focus="with NLB ip target configuration, using readiness gates false"
Ran 1 of 28 Specs in 963.706 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 27 Skipped
PASS
```

### DIFF
--- a/docs/guide/gateway/gateway.md
+++ b/docs/guide/gateway/gateway.md
@@ -127,7 +127,7 @@ spec:
 
 When `my-http-service` or the configured service port can't be found,
 the target group will not be materialized on any ALBs that the route attaches to.
-An [503 Fixed Response](https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_FixedResponseActionConfig.html)
+An [500 Fixed Response](https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_FixedResponseActionConfig.html)
 will be added to any Listener Rules that would have referenced the invalid backend.
 
 ## Specify out-of-band Target Groups


controllers/gateway/route_reconciler.go
```diff
@@ -226,69 +226,48 @@ func (d *routeReconcilerImpl) resolveRefGateway(parentRef gwv1.ParentReference,
...
-				LastTransitionTime: timeNow,
-				ObservedGeneration: route.GetGeneration(),
-			},
-		}
-		return
+		conditions = append(conditions, metav1.Condition{
+			Type:               string(gwv1.RouteConditionAccepted),
+			Status:             metav1.ConditionTrue,
+			Reason:             string(gwv1.RouteReasonAccepted),
+			Message:            "",
```
shraddhabang, line 264: No message?

pkg/gateway/model/model_build_listener.go
```diff
@@ -381,16 +381,16 @@ func buildL7ListenerDefaultActions() []elbv2model.Action {
 	return []elbv2model.Action{action404}
 }
 
-// returns 503 when no backends are configured
+// returns 500 when no backends are configured
```
shraddhabang, line 384: Need to update our docs as well for this. 
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/guide/gateway/gateway.md#:~:text=When%20my%2Dhttp%2Dservice%20or%20the%20configured%20service%20port%20can%27t%20be%20found%2C%20the%20target%20group%20will%20not%20be%20materialized%20on%20any%20ALBs%20that%20the%20route%20attaches%20to.%20An%20503%20Fixed%20Response%20will%20be%20added%20to%20any%20Listener%20Rules%20that%20would%20have%20referenced%20the%20invalid%20backend.

controllers/gateway/route_reconciler.go
shuqz: i felt the status and reason combined is pretty intuitive (accepted=true and resolvedref=true). adding a message is basically just repeating the same thing

zac-nixon: agreed with Shelley, seems intuitive you don't need additional info for an accepted status.

pkg/gateway/model/model_build_listener.go
shuqz: updated

PR #4450, opened 2025-11-14
Support BYOIP

wweiwei-li:
### Description

Support BYOIP (Bring Your Own IP) for Global Accelerator

This change enables users to specify custom IP addresses from their own IP address ranges when creating Global Accelerators, instead of relying solely on AWS-provided IP addresses.
 
Note:  Update BYOIP is not supported 

### Checklist
- [x ] Added tests that cover your change (if possible)
- [x ] Manually tested

### DIFF
--- a/pkg/deploy/aga/accelerator_manager.go
+++ b/pkg/deploy/aga/accelerator_manager.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
+
 	"github.com/aws/aws-sdk-go-v2/aws"
 	awssdk "github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/globalaccelerator"
@@ -60,10 +62,10 @@ func (m *defaultAcceleratorManager) buildSDKCreateAcceleratorInput(_ context.Con
 		IdempotencyToken: aws.String(idempotencyToken),
 	}
 
-	//TODO: BYOIP feature
-	//if len(resAccelerator.Spec.IpAddresses) > 0 {
-	//	createInput.IpAddresses = resAccelerator.Spec.IpAddresses
-	//}
+	// BYOIP feature: Set IP addresses if provided
+	if len(resAccelerator.Spec.IpAddresses) > 0 {
+		createInput.IpAddresses = resAccelerator.Spec.IpAddresses
+	}
 
 	// Add tags
 	tags := m.trackingProvider.ResourceTags(resAccelerator.Stack(), resAccelerator, resAccelerator.Spec.Tags)
@@ -103,11 +105,8 @@ func (m *defaultAcceleratorManager) buildSDKUpdateAcceleratorInput(ctx context.C
 		IpAddressType:  agatypes.IpAddressType(resAccelerator.Spec.IPAddressType),
 		Enabled:        resAccelerator.Spec.Enabled,
 	}
+	// BYOIP is only supported during accelerator creation, not updates
 
-	//TODO: BYOIP feature
-	//if len(resAccelerator.Spec.IpAddresses) > 0 {
-	//	updateInput.IpAddresses = resAccelerator.Spec.IpAddresses
-	//}
 	return updateInput
 }
 
@@ -272,8 +271,41 @@ func (m *defaultAcceleratorManager) isSDKAcceleratorSettingsDrifted(resAccelerat
 		return true
 	}
 
-	//TODO : BYOIP feature
+	// Check if user attempts to change IP addresses (BYOIP only supported at creation)
+	if len(resAccelerator.Spec.IpAddresses) > 0 && !m.areIPAddressesEqual(resAccelerator.Spec.IpAddresses, sdkAccelerator.Accelerator.IpSets) {
+		m.logger.Info("IP addresses cannot be updated after accelerator creation, ignoring IP address changes")
+	}
+
 	return false
+}
+
+func (m *defaultAcceleratorManager) areIPAddressesEqual(desiredIPs []string, actualIPSets []agatypes.IpSet) bool {
+
+	// IPv6 BYOIP is not supported at this time
+	return m.areIPv4AddressesEqual(desiredIPs, actualIPSets)
+}
+
+// areIPv4AddressesEqual compares desired IPv4 addresses with actual IP sets from AWS
+func (m *defaultAcceleratorManager) areIPv4AddressesEqual(desiredIPs []string, actualIPSets []agatypes.IpSet) bool {
+	actualIPv4s := extractIPv4Addresses(actualIPSets)
+	if len(desiredIPs) != len(actualIPv4s) {
+		return false
+	}
+
+	slices.Sort(desiredIPs)
+	slices.Sort(actualIPv4s)
+	return slices.Equal(desiredIPs, actualIPv4s)
+}
+
+// extractIPv4Addresses extracts IPv4 addresses from IPSets
+func extractIPv4Addresses(ipSets []agatypes.IpSet) []string {
+	ips := make([]string, 0)
+	for _, ipSet := range ipSets {
+		if ipSet.IpAddressFamily == "IPv4" {
+			ips = append(ips, ipSet.IpAddresses...)
+		}
+	}
+	return ips
 }
 
 func (m *defaultAcceleratorManager) getIdempotencyToken(resAccelerator *agamodel.Accelerator) string {
--- a/pkg/deploy/aga/accelerator_manager_test.go
+++ b/pkg/deploy/aga/accelerator_manager_test.go
@@ -3,11 +3,13 @@ package aga
 import (
 	"context"
 	"errors"
-	"k8s.io/apimachinery/pkg/types"
 	"testing"
+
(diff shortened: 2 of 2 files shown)


pkg/deploy/aga/accelerator_manager_test.go
```diff
@@ -721,3 +723,86 @@ func Test_defaultAcceleratorManager_disableAccelerator(t *testing.T) {
...
+			actualIPSets:   makeIPSets([]string{"169.254.8.13", "99.82.158.217"}),
+			expectedResult: false,
+			description:    "Desired BYOIP not in current",
+		},
+		{
+			name:           "Changing from one BYOIP to another - drift detected",
+			desiredIPs:     []string{"169.254.8.16"},
+			actualIPSets:   makeIPSets([]string{"169.254.9.13", "99.82.158.217"}),
+			expectedResult: false,
+			description:    "Different BYOIP, count matches but IP doesn't",
```
shraddhabang, line 980: NIT: Update the description or test case. Count does not match here.

pkg/deploy/aga/accelerator_synthesizer.go
```diff
@@ -183,7 +198,48 @@ func (s *acceleratorSynthesizer) isAcceleratorNotFound(err error) bool {
 
 // isSDKAcceleratorRequiresReplacement checks whether a sdk Accelerator requires replacement to fulfill an Accelerator resource.
 func isSDKAcceleratorRequiresReplacement(sdkAccelerator AcceleratorWithTags, resAccelerator *agamodel.Accelerator) bool {
```
shraddhabang, line 185: Can you add unit tests for `isSDKAcceleratorRequiresReplacement` so that we can cover different test cases ?
Also can we comment the behavior here on when we consider the replacement or add a task for us to document for it for later?

pkg/deploy/aga/accelerator_synthesizer.go
```diff
@@ -183,7 +198,48 @@ func (s *acceleratorSynthesizer) isAcceleratorNotFound(err error) bool {
...
-	// The accelerator will only need replacement in BYOIP scenarios. I will implement this later as a separate PR
-	// TODO : BYOIP feature
+	if len(resAccelerator.Spec.IpAddresses) == 0 {
```
shraddhabang: What happens if we had applied BYOIP initially and now we removed it? Do we need a replacement here since we are going from BYOIP to non-BYOIP?

pkg/deploy/aga/accelerator_synthesizer.go
```diff
@@ -183,7 +198,48 @@ func (s *acceleratorSynthesizer) isAcceleratorNotFound(err error) bool {
...
+		if ipSet.IpAddressFamily == "IPv4" {
+			ips = append(ips, ipSet.IpAddresses...)
+		}
+	}
+	return ips
+}
+
+// HasSharedIPv4CIDR checks if current and desired IPs share any IPv4 CIDR block.
+func HasSharedIPv4CIDR(currentIPSets []agatypes.IpSet, desiredIPs []string) bool {
+	if len(desiredIPs) == 0 {
```
shraddhabang: This is not possible since we have already compared length of `desiredIPs` aka `resAccelerator.Spec.IpAddresses` to zero and return false before we call this function right?

pkg/deploy/aga/accelerator_synthesizer_test.go
```diff
@@ -648,3 +648,106 @@ func Test_acceleratorSynthesizer_handleUpdateAccelerator(t *testing.T) {
...
+func Test_HasSharedIPv4CIDR(t *testing.T) {
+	tests := []struct {
+		name           string
+		currentIPSets  []agatypes.IpSet
+		desiredIPs     []string
+		expectedResult bool
+		description    string
+	}{
+		{
+			name:           "No desired IPs - has shared CIDR",
```
shraddhabang: This test case wont be required if we remove zero len check for desired IPs.

pkg/deploy/aga/accelerator_manager.go
```diff
@@ -58,10 +60,10 @@ func (m *defaultAcceleratorManager) buildSDKCreateAcceleratorInput(_ context.Con
 		IdempotencyToken: aws.String(idempotencyToken),
 	}
 
-	//TODO: BYOIP feature
-	//if len(resAccelerator.Spec.IpAddresses) > 0 {
-	//	createInput.IpAddresses = resAccelerator.Spec.IpAddresses
-	//}
+	// BYOIP feature: Set IP addresses if provided
+	if len(resAccelerator.Spec.IpAddresses) > 0 {
```
shuqz, line 66: do we need to do nil check for resAccelerator.Spec and resAccelerator.Spec.IpAddresses

pkg/deploy/aga/accelerator_synthesizer.go
```diff
@@ -72,8 +74,21 @@ func (s *acceleratorSynthesizer) Synthesize(ctx context.Context) error {
 
 	// Accelerator exists, determine if it needs replacement or update
 	if isSDKAcceleratorRequiresReplacement(sdkAccelerator, resAccelerator) {
+		currentIPv4s := extractIPv4Addresses(sdkAccelerator.Accelerator.IpSets)
+		desiredBYOIPs := resAccelerator.Spec.IpAddresses
+		if len(desiredBYOIPs) == 1 {
+			s.logger.Info("Accelerator requires replacement: BYOIP CIDR change",
```
shuqz: do we want to keep it as v(1) logging

pkg/deploy/aga/accelerator_synthesizer.go
shuqz: +1

pkg/deploy/aga/accelerator_manager_test.go
wweiwei-li: Yes, Let me add one more ip to desiredIPs

pkg/deploy/aga/accelerator_synthesizer.go
wweiwei-li: Yes, Adding it

pkg/deploy/aga/accelerator_synthesizer.go
wweiwei-li: Yes, since we have checked it at  isSDKAcceleratorRequiresReplacement level.  We don't need it

pkg/deploy/aga/accelerator_synthesizer_test.go
wweiwei-li: Yes, removing it

pkg/deploy/aga/accelerator_manager.go
wweiwei-li: We don't need to do nil check for resAccelerator.Spec and resAccelerator.Spec.IpAddresses.  For resAccelerator.Spec, it is a struct. so It can't be nil.  No nil check needed.  For resAccelerator.Spec.IpAddresses, it is a slice,  the len(nil) will return 0 in go.

pkg/deploy/aga/accelerator_synthesizer.go
wweiwei-li: I am thinking we should keep it as it is since accelerator replacement is a significant operation that we want it to be accessible by default.  Let me know what do you think

pkg/deploy/aga/accelerator_synthesizer.go
wweiwei-li: For BYOIP to non-BYOIP transitions, updating is not supported. BYOIP to non-BYOIP will always need replacement. However, if we made it return true, then every non-BYOIP to non-BYOIP transition would cause replacement unless we have a mechanism like below to check if current IPs include BYOIP (we discussed this internally we try not adding this). 
```
if len(resAccelerator.Spec.IpAddresses) == 0 && hasBYOIP(sdkAccelerator) {
return true
}
```

so I was thinking for this case we need to document that users need to remove the accelerator first and recreate a new one. This is my intent. Let me try it out to check if it is easy and put together internal documentation for the behavior for discussion. Also I wonder how likely they want to from For BYOIP to non-BYOIP transitions

shraddhabang: Agreed. Lets take it offline.

pkg/deploy/aga/accelerator_synthesizer.go
shraddhabang: Yes I think we need this log. This is important one.

pkg/deploy/aga/accelerator_manager.go
```diff
@@ -103,11 +105,8 @@ func (m *defaultAcceleratorManager) buildSDKUpdateAcceleratorInput(ctx context.C
 		IpAddressType:  agatypes.IpAddressType(resAccelerator.Spec.IPAddressType),
 		Enabled:        resAccelerator.Spec.Enabled,
 	}
+	// BYOIP is only supported during accelerator creation, not updates
```
shraddhabang, line 108: Do we want to put a verbose warning log here for this if we find any IP addresses during updates? Just so that it's clear to customers.

shraddhabang: Ok I see if you have put it in drift check. Nvm

PR #4451, opened 2025-11-14
mention pod identity

zac-nixon:
We support Pod Identity, but don't mention it in our docs.


docs/deploy/installation.md
```diff
@@ -121,7 +121,13 @@ Example condition for cluster name resource tag:
     --approve
     ```
 
-### Option B: Attach IAM policies to nodes
+### Option B: Recommended, Pod Identity
```
shuqz, line 124: i see option A is also recommended, so both recommended?

PR #4458, opened 2025-11-18
[feat aga] Implement AGA endpoint resource references loading and monitoring

shraddhabang:
This PR implements two key components for the AWS Global Accelerator controller:

1. **Endpoint Loaders**: Dynamically loads and resolves endpoints from Kubernetes resources
2. **Resource Monitoring**: Watches referenced resources for changes to trigger reconciliation

## Commit 1: [[feat aga] Implement endpoint loader with DNS resolution](https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/b5097b5cb94aa4e1a4a120b349f06e66114fb61f)
This commit implements the endpoint loading system for the AGA controller. It provides:

- Dynamic loading of endpoints from Kubernetes resources (Services, Ingresses, Gateways)
- DNS resolution to AWS load balancer ARNs
- Efficient LRU caching with TTL-based invalidation
- Detailed error reporting and status tracking
- Comprehensive unit tests

The endpoint loader enables GlobalAccelerator resources to reference Kubernetes objects and automatically resolve them to the appropriate AWS resources.

## Commit 2: [[feat aga] Implement resource monitoring for referenced resources](https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/54f87b61ba129fe878fa5fe2553cac2641db6f71)
This commit implements the resource monitoring system for the AGA controller. It provides:

- Dynamic watching of referenced Kubernetes resources
- Event handling to trigger reconciliation when resources change
- Only watches resources that are actively referenced
- Reference tracking between resources and GlobalAccelerators
- Efficient watch management with cleanup for unreferenced resources
- Unit tests for all monitoring components

This monitoring system ensures that when a referenced resource changes (e.g., a Service gets a new load balancer), the GlobalAccelerator is automatically reconciled to use the updated endpoint.

## Note for temporary limitations for cross namespace reference

We want to allow references to Kubernetes resources (Services, Ingresses, Gateways) that exist in different namespaces from the GlobalAccelerator CR itself. This enables more flexible architectural patterns but requires careful security considerations. We will implement this later as we will need to come up with a proper cross-namespace reference system keeping security concerns in mind. For now in the current implementation cross-namespace references are detected but only result in a warning - this means:
   - The references won't work (the endpoint is marked as warning)
   - The GlobalAccelerator CR won't be rejected
   

### Checklist
- [x] Manually tested

### DIFF
--- a/controllers/aga/eventhandlers/resource_events.go
+++ b/controllers/aga/eventhandlers/resource_events.go
@@ -1,0 +1,107 @@
+package eventhandlers
+
+import (
+	"context"
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	networking "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/aga"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+// NewEnqueueRequestsForResourceEvent creates a new handler for generic resource events
+func NewEnqueueRequestsForResourceEvent(
+	resourceType aga.ResourceType,
+	referenceTracker *aga.ReferenceTracker,
+	logger logr.Logger,
+) handler.EventHandler {
+	return &enqueueRequestsForResourceEvent{
+		resourceType:     resourceType,
+		referenceTracker: referenceTracker,
+		logger:           logger,
+	}
+}
+
+// enqueueRequestsForResourceEvent handles resource events and enqueues reconcile requests for GlobalAccelerators
+// that reference the resource
+type enqueueRequestsForResourceEvent struct {
+	resourceType     aga.ResourceType
+	referenceTracker *aga.ReferenceTracker
+	logger           logr.Logger
+}
+
+// The following methods implement handler.TypedEventHandler interface
+
+// Create handles Create events with the typed API
+func (h *enqueueRequestsForResourceEvent) Create(ctx context.Context, evt event.TypedCreateEvent[client.Object], queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	h.handleResource(ctx, evt.Object, "created", queue)
+}
+
+// Update handles Update events with the typed API
+func (h *enqueueRequestsForResourceEvent) Update(ctx context.Context, evt event.TypedUpdateEvent[client.Object], queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	h.handleResource(ctx, evt.ObjectNew, "updated", queue)
+}
+
+// Delete handles Delete events with the typed API
+func (h *enqueueRequestsForResourceEvent) Delete(ctx context.Context, evt event.TypedDeleteEvent[client.Object], queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	h.handleResource(ctx, evt.Object, "deleted", queue)
+}
+
+// Generic handles Generic events with the typed API
+func (h *enqueueRequestsForResourceEvent) Generic(ctx context.Context, evt event.TypedGenericEvent[client.Object], queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	h.handleResource(ctx, evt.Object, "generic event", queue)
+}
+
+// handleTypedResource handles resource events for the typed interface
+func (h *enqueueRequestsForResourceEvent) handleResource(_ context.Context, obj interface{}, eventType string, queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	var namespace, name string
+
+	// Extract namespace and name based on the object type
+	switch res := obj.(type) {
+	case *corev1.Service:
+		namespace = res.Namespace
+		name = res.Name
+	case *networking.Ingress:
+		namespace = res.Namespace
+		name = res.Name
+	case *gwv1.Gateway:
+		namespace = res.Namespace
+		name = res.Name
+	default:
+		h.logger.Error(nil, "Unknown resource type", "type", h.resourceType)
+		return
+	}
+
+	resourceKey := aga.ResourceKey{
+		Type: h.resourceType,
+		Name: types.NamespacedName{
+			Namespace: namespace,
+			Name:      name,
+		},
+	}
+
+	// If this resource is not referenced by any GA, no need to queue reconciles
+	if !h.referenceTracker.IsResourceReferenced(resourceKey) {
+		return
+	}
+
+	// Get all GAs that reference this resource
+	gaRefs := h.referenceTracker.GetGAsForResource(resourceKey)
+
+	// Queue reconcile for affected GAs
+	for _, gaRef := range gaRefs {
+		h.logger.V(1).Info("Enqueueing GA for reconcile due to resource event",
+			"resourceType", h.resourceType,
+			"resourceName", resourceKey.Name,
+			"eventType", eventType,
+			"ga", gaRef)
+
+		queue.Add(reconcile.Request{NamespacedName: gaRef})
+	}
+}
--- a/controllers/aga/globalaccelerator_controller.go
+++ b/controllers/aga/globalaccelerator_controller.go
@@ -35,9 +35,12 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	agaapi "sigs.k8s.io/aws-load-balancer-controller/apis/aga/v1beta1"
+	"sigs.k8s.io/aws-load-balancer-controller/controllers/aga/eventhandlers"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/aga"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/config"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/deploy"
@@ -50,6 +53,7 @@ import (
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/model/core"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/runtime"
 	agastatus "sigs.k8s.io/aws-load-balancer-controller/pkg/status/aga"
+	gwclientset "sigs.k8s.io/gateway-api/pkg/client/clientset/versioned"
 )
 
 const (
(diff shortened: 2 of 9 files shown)


pkg/aga/dns_resolver.go
```diff
@@ -0,0 +1,95 @@
...
+	"sync"
+	"time"
+
+	elbv2sdk "github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2"
+	"github.com/hashicorp/golang-lru"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/aws/services"
+)
+
+// DNSResolver resolves load balancer DNS names to ARNs
+type DNSResolver struct {
```
emdelaney: Can we call this DNSToArnResolver or DNSToLoadBalancerResolver? "DNS Resolver" makes me think we're mapping a DNS name to IP addresses, which is not what we're doing here.

pkg/aga/resource_watcher.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Stop stops the reflector
+func (w *ResourceWatcher) Stop() {
+	close(w.stopCh)
+}
+
+// AddConsumer adds a consumer (GlobalAccelerator) to the watcher
+func (w *ResourceWatcher) AddConsumer(consumerID string) {
+	w.consumers.Insert(consumerID)
```
wweiwei-li, line 84: I am thinking we should add mutex to protect consumers from concurrent access

pkg/aga/dns_resolver.go
shraddhabang: Sure. Makes sense.

pkg/aga/resource_watcher.go
shraddhabang: `ResourceWatcher` is already being used in a thread-safe manner from `endpoint_resources_manager.go` (which uses its own mutex). This change will simply make ResourceWatcher more robust on its own. I can do that.

controllers/aga/eventhandlers/resource_events.go
```diff
@@ -0,0 +1,111 @@
...
+	case *corev1.Service:
+		namespace = res.Namespace
+		name = res.Name
+	case *networking.Ingress:
+		namespace = res.Namespace
+		name = res.Name
+	case *gwv1.Gateway:
+		namespace = res.Namespace
+		name = res.Name
+	case *unstructured.Unstructured:
```
zac-nixon: What is this? why would you use it?

controllers/aga/globalaccelerator_controller.go
```diff
@@ -335,21 +406,91 @@ func (r *globalAcceleratorReconciler) SetupWithManager(ctx context.Context, mgr
...
+	// Add watches using the channel sources with event handlers
+	if err := c.Watch(source.Channel(r.serviceEventChan, serviceHandler)); err != nil {
+		return err
+	}
+
+	if err := c.Watch(source.Channel(r.ingressEventChan, ingressHandler)); err != nil {
+		return err
+	}
+
+	if err := c.Watch(source.Channel(r.gatewayEventChan, gatewayHandler)); err != nil {
```
zac-nixon: This will break if the user has decided to not add the Gateway CRDs to their cluster. I would suggest this:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/controllers/ingress/group_controller.go#L339

pkg/aga/dns_to_load_balancer_resolver.go
```diff
@@ -0,0 +1,95 @@
...
+type cacheEntry struct {
+	arn      string
+	expireAt time.Time
+}
+
+// NewDNSToLoadBalancerResolver creates a new DNSToLoadBalancerResolver
+func NewDNSToLoadBalancerResolver(elbv2Client services.ELBV2) (*DNSToLoadBalancerResolver, error) {
+	// AWS Global Accelerator has a quota of 420 endpoints per AWS account (can be increased)
+	// Using 420 provides headroom while efficiently caching DNS-to-ARN resolutions
+	cache, err := lru.New(420)
```
zac-nixon: nit: magic number to constant

pkg/aga/dns_to_load_balancer_resolver.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &DNSToLoadBalancerResolver{
+		elbv2Client: elbv2Client,
+		cache:       cache,
+		ttl:         5 * time.Minute, // Default TTL of 5 minutes
```
zac-nixon, line 44: IMO, this is too short. DNS names / ARNs do not change that frequently. What do you think about increasing the cache time?

pkg/aga/dns_to_load_balancer_resolver_test.go
```diff
@@ -0,0 +1,273 @@
...
+	tests := []struct {
+		name        string
+		fields      fields
+		dnsName     string
+		wantARN     string
+		wantErr     bool
+		setupFields func(fields fields)
+	}{
+		{
+			name: "successfully resolves DNS to ARN",
```
zac-nixon, line 37: nit: Could you add a test case for an NLB ARN / DNS name?

pkg/aga/endpoint_loader.go
```diff
@@ -0,0 +1,432 @@
...
+		namespace = *endpoint.Namespace
+	}
+
+	// Set up the default result with basic information
+	name := ""
+	if endpoint.Name != nil {
+		name = *endpoint.Name
+	}
+
+	weight := int32(128) // Default weight
```
zac-nixon: nit: magic number to const

pkg/aga/endpoint_loader.go
```diff
@@ -0,0 +1,432 @@
...
+}
+
+// extractServiceDNS extracts DNS from Services
+func extractServiceDNS(obj client.Object) (string, error) {
+	svc, ok := obj.(*corev1.Service)
+	if !ok {
+		return "", fmt.Errorf("object is not a Service")
+	}
+
+	if svc.Spec.Type != corev1.ServiceTypeLoadBalancer {
```
zac-nixon, line 238: I believe NodePort types are supported too.
https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/service/nlb/

pkg/aga/endpoint_loader.go
```diff
@@ -0,0 +1,432 @@
...
+		return "", fmt.Errorf("object is not an Ingress")
+	}
+
+	if len(ing.Status.LoadBalancer.Ingress) == 0 {
+		return "", fmt.Errorf("ingress %v does not have a LoadBalancer", k8s.NamespacedName(ing))
+	}
+
+	for _, ingress := range ing.Status.LoadBalancer.Ingress {
+		if ingress.Hostname != "" {
+			return ingress.Hostname, nil
```
zac-nixon: This chooses a random Hostname between ALB / NLB if using Frontend NLBs. Weiwei created functionality to determine if the hostname is NLB / ALB.

pkg/aga/endpoint_loader.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	logger.V(1).Info("Endpoint loading statistics",
+		"total", len(endpoints),
+		"loaded", loaded,
+		"warnings", warning,
+		"fatal", fatal)
+
+	// Log individual endpoints
+	for i, endpoint := range endpoints {
```
zac-nixon, line 429: nit: this will potentially construct a lot of strings. i would be mindful of leaving this in here.

pkg/aga/endpoint_resources_manager.go
```diff
@@ -0,0 +1,229 @@
...
+			continue
+		}
+
+		switch endpoint.Type {
+		case agaapi.GlobalAcceleratorEndpointTypeService:
+			ref := ktypes.NamespacedName{Namespace: endpoint.Namespace, Name: endpoint.Name}
+			serviceRefs.Insert(ref.String())
+
+			// Start watching this service if not already watched
+			if _, exists := m.serviceWatches[ref]; !exists {
```
zac-nixon, line 103: IMO, this could be error prone. I really like abstracting away these kind of critical data structures such that no one can accidentally modify the internal state without being forced to acquire the mutex first. If this a big refactor, no worries just add a TODO.

pkg/aga/reference_tracker.go
```diff
@@ -0,0 +1,138 @@
...
+
+		currentResources.Insert(resourceKey)
+
+		// Update resource -> GA mapping
+		if _, exists := t.resourceMap[resourceKey]; !exists {
+			t.resourceMap[resourceKey] = sets.NewString()
+		}
+		t.resourceMap[resourceKey].Insert(gaKey.String())
+
+		t.logger.V(1).Info("Resource referenced by GA",
```
zac-nixon, line 59: same comment about a potential lot of strings of constructed

controllers/aga/eventhandlers/resource_events.go
shraddhabang: This was a leftover before I set the Gateway Client. Good catch I will remove this. We should not be handling any other objects than the ones we support. We will never get here to be honest since we have CEL, but there is no need for this.

pkg/aga/dns_to_load_balancer_resolver.go
shraddhabang: I kept it small because I wanted to remove the older dns from the cache and keep the cache upto date. Having large TTL will keep the older dns and invalid (Deleted) ones for longer time. If we encounter any older dns because the our k8s were not updated for some reason, we will resolve that to invalid arn. Hence I kept it for 5 minutes so that we get faster failure feedback.  What do you think?

pkg/shared_utils/aga_utils.go
```diff
@@ -0,0 +1,36 @@
...
+	// First check if AGA controller is enabled via feature gate
+	if !featureGates.Enabled(config.AGAController) {
+		return false
+	}
+
+	// Global Accelerator is only available in standard AWS partition
+	// Not available in specialized AWS partitions
+	regionLower := strings.ToLower(region)
+
+	// Check for non-standard AWS partitions where Global Accelerator is not available
```
zac-nixon: it's only available in `aws` partition right?

shraddhabang: Yes.

pkg/aga/endpoint_loader.go
shraddhabang: Good catch. Updating.

pkg/aga/endpoint_loader.go
shraddhabang: Yes. I am aware of that. But I think this is crucial to log this info for troubleshooting since the endpoints work dynamically. And we need to know which endpoints we filtered out so that we can debug this better. We are not expecting many endpoints per accelerator so I think verbose level 1 logs are okay for this.

pkg/aga/reference_tracker.go
shraddhabang: Same reasoning as above.

pkg/aga/endpoint_resources_manager.go
shraddhabang: I understand the concern about direct map access potentially leading to race conditions if not properly synchronized. The current implementation does correctly lock the mutex before accessing the map. I've added a TODO comment for now in the `defaultEndpointResourcesManager` definition to document the need to refactor this pattern in future since this is time-sensitive PR.

controllers/aga/globalaccelerator_controller.go
shraddhabang: Good catch. I will add a check for gateway CRD availability.

PR #4463, opened 2025-11-23
[feat aga] Implement endpoint group management with port override conflict resolution

shraddhabang:
This PR introduces comprehensive endpoint group management capabilities for AGA in the AWS Load Balancer Controller. The implementation includes endpoint group reconciliation logic, conflict detection and resolution, and full lifecycle management (creation, updates, and deletion) within the AGA controller.

This PR is divided in multiple commits for easier review. (The earlier PR is not yet merged so please ignore the duplicate commits from the older PR only focus on the following commits for this PR)

#### [[feat aga] Implement model builder for endpoint groups](https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/9f463bb3fd6b8629cff870055ac97711c9c74db3) 

- Added the model builder for endpoint groups to build model object for endpoint group management
- This also includes the  validations to detect failures early. 

#### [[feat aga] few bugfixes and code refactoring](https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/5e57e3f86669a108b18a7ea1eaaf2b8fb22dc727) 
- This includes the bug fix for getting the region info from cloud service instead if controller flag. The user may not always set controller flag for regions. So to handle that, we are resolving region of the cluster from cloud svc which use EC2 metadata in absence of configured aws region
- Refactored some code to include some detailed logging for model build and deploy failures. 
- Found the infinite reconcile loop happening during model deploy failed error as we were setting the message of the status condition for accelerator for each failure which was causing the status to update after each reconcile . Added a fixed message to avoid that. 

#### [[feat aga] Implement endpoint group deployer with port override conf](https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/3f693791a6305799dd68ebbb890b7b3b1c491ad0)

This commit includes two parts. 
##### Endpoint Group Management

- Added region-based reconciliation to map endpoint groups with their corresponding listeners
- Implemented three-way reconciliation (update, create, delete) for endpoint groups in `synthesizeEndpointGroupsOnListener`
- Enhanced resource matching logic using region as the unique key within listener context
- Added robust error handling and detailed logging throughout the reconciliation process

##### Port Override Conflict Resolution

- Implemented `detectConflictsWithSDKEndpointGroups` to identify region-specific port conflicts

- Created two-phase conflict resolution algorithm in `ProcessEndpointGroupPortOverrides`:

  1. Analysis phase: Collects and evaluates all endpoint groups and potential conflicts
  2. Execution phase: Updates endpoint groups to remove conflicting port overrides

- Added validation rules ensuring endpoint ports don't overlap with listener port ranges

- Ensured listener ports in port overrides remain within listener port range specifications

### Checklist
- [x] Manually tested


controllers/aga/globalaccelerator_controller.go
```diff
@@ -224,6 +269,29 @@ func (r *globalAcceleratorReconciler) buildModel(ctx context.Context, ga *agaapi
 
 func (r *globalAcceleratorReconciler) reconcileGlobalAcceleratorResources(ctx context.Context, ga *agaapi.GlobalAccelerator) error {
 	r.logger.Info("Reconciling GlobalAccelerator resources", "globalAccelerator", k8s.NamespacedName(ga))
+
+	// Get all endpoints from GA
+	endpoints := aga.GetAllEndpointsFromGA(ga)
```
emdelaney, line 274: I'd consider calling this GetDesiredEndpointsFromGA; so that it's obvious these are coming from the spec and not from the current state of the accelerator.

pkg/aga/reference_tracker.go
```diff
@@ -0,0 +1,138 @@
...
+func NewReferenceTracker(logger logr.Logger) *ReferenceTracker {
+	return &ReferenceTracker{
+		resourceMap: make(map[ResourceKey]sets.String),
+		gaRefMap:    make(map[types.NamespacedName]sets.Set[ResourceKey]),
+		logger:      logger,
+	}
+}
+
+// UpdateReferencesForGA updates the tracking information for a GlobalAccelerator
+func (t *ReferenceTracker) UpdateReferencesForGA(ga *agaapi.GlobalAccelerator, endpoints []EndpointReference) {
```
emdelaney, line 38: Assuming `endpoints` is existing / current / actual Endpoints, let's specify that here.

pkg/deploy/aga/listener_synthesizer.go
```diff
@@ -97,55 +98,15 @@ func (s *listenerSynthesizer) synthesizeListenersOnAccelerator(ctx context.Conte
...
 	// 1. Delete only conflicting listeners (that would block updates)
-	// 2. Update matched listeners
-	// 3. Delete unneeded (non-conflicting) listeners
-	// 4. Create new listeners
+	// 2. Process all port overrides which may block listener updates
+	//   - Remove endpoint port overrides that overlap with any new listener port ranges
+	//   - Remove listener port overrides from existing listener which is outside desired listener port ranges
+	// 3. Update matched listeners
+	// 4. Delete unneeded (non-conflicting) listeners
+	// 5. Create new listeners
```
emdelaney, line 106: Do we need a step 6, add the port overrides that used to overlap, but don't anymore? (or is that handled by the endpoint group synthesizer later?)

shraddhabang: Yeah that will be handled by endpoint group synth as its part of endpoint groups. We will always achieve what is desired state for epgs in its synthesizer. So any port overrides that are removed (but desired) from listeners synth will end up in epgs later. This will be temp loss to make listener updates to go through.

pkg/aga/reference_tracker.go
shraddhabang: Makes sens. These are desired ones coming from spec. I will rename these.

controllers/aga/globalaccelerator_controller.go
shraddhabang: Yes makes sense

PR #4471, opened 2025-11-26
[feat aga] Implement endpoint management for endpoint groups in accelerator

shraddhabang:
This PR implements optimized endpoint management for AWS Global Accelerator resources with intelligent handling of endpoint states and API interactions.

#### Commit 1: [[feat aga] Implement endpoints builder filtering for partially loaded…](https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/28ea3e0b41be7651ae9c5acd9e8f56a6d25f52ac)
- Implemented intelligent filtering of endpoints in warning state when building the desired endpoint configuration
- Added clear separation between warning and fully loaded endpoints to maintain clean model building
- Ensured endpoints in warning state are tracked but excluded from configuration to prevent deployment failures
- Enhanced logging to provide visibility into endpoint state transitions

This change ensures that warning state endpoints don't block the successful deployment of healthy endpoints. While we want to eventually deploy these endpoints when they become healthy, we don't want them to prevent the deployment of already-healthy endpoints.

#### Commit 2: [[feat aga] Implement endpoints deployer with requeue logic for partia…](https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/328d317b20414efd49691ba7aa9c3ef8374ede0e)

- Implemented comprehensive endpoint drift detection:

  - Identifies endpoints that need to be added, updated, or removed
  - Detects property changes (weight, clientIPPreservation) requiring updates

- Optimized API usage pattern:

  - Uses efficient add-endpoints/remove-endpoints APIs for endpoint presence changes
  - Uses update-endpoint-group API for property updates (since add-endpoints doesn't support updates)

- Added intelligent error handling with flip-flop pattern:

  - Handles LimitExceededException by removing existing endpoints before adding new ones

- Implemented warning state requeuing:

  - Sets requeue intervals for resources with warning endpoints
  - Logs detailed information about warning endpoints for easier troubleshooting

- Added comprehensive unit tests for all new functionality

### Checklist
- [x] Manually tested

### DIFF
--- a/controllers/aga/globalaccelerator_controller.go
+++ b/controllers/aga/globalaccelerator_controller.go
@@ -64,9 +64,15 @@ const (
 	agaResourcesGroupVersion = "aga.k8s.aws/v1beta1"
 	globalAcceleratorKind    = "GlobalAccelerator"
 
-	// Requeue constants for provisioning state monitoring
-	requeueMessage          = "Monitoring provisioning state"
-	statusUpdateRequeueTime = 1 * time.Minute
+	// Requeue constants for state monitoring
+	// requeueReasonAcceleratorInProgress indicates that the reconciliation is being requeued because
+	// the Global Accelerator is still in progress state
+	requeueReasonAcceleratorInProgress = "Waiting for Global Accelerator %s  with status 'IN_PROGRESS' to complete"
+
+	// requeueReasonEndpointsInWarningState indicates that the reconciliation is being requeued because
+	// there are endpoints in warning state that need to be periodically rechecked
+	requeueReasonEndpointsInWarningState = "Retrying endpoints for Global Accelerator %s  which did load successfully - will check availability again soon"
+	statusUpdateRequeueTime              = 1 * time.Minute
 
 	// Metric stage constants
 	MetricStageFetchGlobalAccelerator     = "fetch_globalAccelerator"
@@ -251,8 +257,8 @@ func (r *globalAcceleratorReconciler) cleanupGlobalAccelerator(ctx context.Conte
 	return nil
 }
 
-func (r *globalAcceleratorReconciler) buildModel(ctx context.Context, ga *agaapi.GlobalAccelerator) (core.Stack, *agamodel.Accelerator, error) {
-	stack, accelerator, err := r.modelBuilder.Build(ctx, ga)
+func (r *globalAcceleratorReconciler) buildModel(ctx context.Context, ga *agaapi.GlobalAccelerator, loadedEndpoints []*aga.LoadedEndpoint) (core.Stack, *agamodel.Accelerator, error) {
+	stack, accelerator, err := r.modelBuilder.Build(ctx, ga, loadedEndpoints)
 	if err != nil {
 		r.eventRecorder.Event(ga, corev1.EventTypeWarning, k8s.GlobalAcceleratorEventReasonFailedBuildModel, fmt.Sprintf("Failed build model due to %v", err))
 		return nil, nil, err
@@ -279,7 +285,7 @@ func (r *globalAcceleratorReconciler) reconcileGlobalAcceleratorResources(ctx co
 	r.endpointResourcesManager.MonitorEndpointResources(ga, endpoints)
 
 	// Validate and load endpoint status using the endpoint loader
-	_, fatalErrors := r.endpointLoader.LoadEndpoints(ctx, ga, endpoints)
+	loadedEndpoints, fatalErrors := r.endpointLoader.LoadEndpoints(ctx, ga, endpoints)
 	if len(fatalErrors) > 0 {
 		err := fmt.Errorf("failed to load endpoints: %v", fatalErrors[0])
 		r.eventRecorder.Event(ga, corev1.EventTypeWarning, k8s.GlobalAcceleratorEventReasonFailedEndpointLoad, fmt.Sprintf("Failed to reconcile due to %v", err))
@@ -295,7 +301,7 @@ func (r *globalAcceleratorReconciler) reconcileGlobalAcceleratorResources(ctx co
 	var accelerator *agamodel.Accelerator
 	var err error
 	buildModelFn := func() {
-		stack, accelerator, err = r.buildModel(ctx, ga)
+		stack, accelerator, err = r.buildModel(ctx, ga, loadedEndpoints)
 	}
 	r.metricsCollector.ObserveControllerReconcileLatency(controllerName, MetricStageBuildModel, buildModelFn)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


controllers/aga/globalaccelerator_controller.go
```diff
@@ -64,9 +64,15 @@ const (
...
-	statusUpdateRequeueTime = 1 * time.Minute
+	// Requeue constants for state monitoring
+	// requeueReasonAcceleratorInProgress indicates that the reconciliation is being requeued because
+	// the Global Accelerator is still in progress state
+	requeueReasonAcceleratorInProgress = "Waiting for Global Accelerator %s  with status 'IN_PROGRESS' to complete"
+
+	// requeueReasonEndpointsInWarningState indicates that the reconciliation is being requeued because
+	// there are endpoints in warning state that need to be periodically rechecked
+	requeueReasonEndpointsInWarningState = "Retrying endpoints for Global Accelerator %s  which did load successfully - will check availability again soon"
+	statusUpdateRequeueTime              = 1 * time.Minute
```
wweiwei-li, line 75: Q:  Just thinking, should we have different `statusUpdateRequeueTime`  for `requeueReasonAcceleratorInProgress` and `requeueReasonEndpointsInWarningState`.  would Endpoint issue (service is not ready, DNS is not ready)  take more longer to resolve ?

pkg/aga/model_build_endpoint_group.go
```diff
@@ -248,3 +337,23 @@ func (b *defaultEndpointGroupBuilder) validatePortOverrides(listener *agamodel.L
 
 	return nil
 }
+
+// buildEndpointConfiguration creates an EndpointConfiguration from a GlobalAcceleratorEndpoint
+// This helper function consolidates the repeated code for creating endpoint configurations
+func buildEndpointConfigurationFromEndpoint(endpoint *agaapi.GlobalAcceleratorEndpoint) agamodel.EndpointConfiguration {
```
wweiwei-li: `buildEndpointConfigurationFromEndpoint` is not used .

pkg/deploy/aga/endpoint_group_manager.go
```diff
@@ -238,3 +270,299 @@ func (m *defaultEndpointGroupManager) arePortOverridesEqual(modelPortOverrides [
...
+			}
+			// If no endpoints to remove but hit limit, just return the original error
+			return fmt.Errorf("failed to add endpoints due to limit and no endpoints available to remove for endpoint group %s: %w", endpointGroupARN, err)
+		} else if err != nil {
+			// For any other error, return it directly
+			return fmt.Errorf("failed to add endpoints for endpoint group %s: %w", endpointGroupARN, err)
+		}
+	}
+
+	// Now remove endpoints that are no longer needed
```
wweiwei-li, line 460: I like the way you how you handle add/removing endpoints.  I feel we could make it even better to avoid zero downtime. Maybe even in the Flip-Flop approach.  We can consider mot removing all endpoints at once.  something like batched remove and add.    But I am also thinking if only few customer will hit into this case, we can use what you have now. We don't want to over engineer it.

shraddhabang: Yeah thats exactly the reason I kept it simple for now. We can always improve this later. It will be very difficult to find out which to remove first to avoid downtime entirely. So it will be very difficult to achieve zero downtime anyways.

controllers/aga/globalaccelerator_controller.go
shraddhabang: I did not want to give many configurations in first iteration hence kept the re-queue time to be uniform for now. I think this will be occurring for very edge cases since we don't usually do provisions that often.

pkg/aga/model_build_endpoint_group.go
shraddhabang: Good catch removing.

pkg/aga/model_build_endpoint_group_test.go
```diff
@@ -2,15 +2,337 @@ package aga
...
+			want:        "Service/test-namespace/",
+		},
+		{
+			name: "endpoint with both nil namespace and name",
+			endpoint: agaapi.GlobalAcceleratorEndpoint{
+				Type:      agaapi.GlobalAcceleratorEndpointTypeService,
+				Namespace: nil,
+				Name:      nil,
+			},
+			gaNamespace: "default",
```
zac-nixon, line 87: can you test w/ a non-default namespace?

zac-nixon: also, what happens if the namespace is omitted (implicit default)?

pkg/aga/model_build_endpoint_group.go
```diff
@@ -146,14 +159,90 @@ func (b *defaultEndpointGroupBuilder) buildEndpointGroupSpec(ctx context.Context
...
+// generateEndpointKey creates a consistent string key for endpoint lookup
+func generateEndpointKey(ep agaapi.GlobalAcceleratorEndpoint, gaNamespace string) string {
+	namespace := gaNamespace
+	if ep.Namespace != nil {
+		namespace = awssdk.ToString(ep.Namespace)
+	}
+	name := awssdk.ToString(ep.Name)
+
+	if ep.Type == agaapi.GlobalAcceleratorEndpointTypeEndpointID {
+		return fmt.Sprintf("%s/%s", ep.Type, awssdk.ToString(ep.EndpointID))
```
zac-nixon, line 186: do you need a region in here? I am not sure about AGA's id generation guarantees across region.

pkg/aga/model_build_endpoint_group_test.go
shraddhabang: We will always get a namespace from the AGA CRD even if its omitted which defaults to `default`. It works in that case as well. I have tested it. Is that what you mean?

pkg/aga/model_build_endpoint_group.go
shraddhabang: The endpoint ID here will be ARN of the LB right. So it should have region info in it which will be unique identifier. So I dont think we need region here again for making it unique.

controllers/aga/globalaccelerator_controller.go
```diff
@@ -326,14 +332,37 @@ func (r *globalAcceleratorReconciler) reconcileGlobalAcceleratorResources(ctx co
...
-	// Update GlobalAccelerator status after successful deployment
+	// Check if any endpoints have warning status and collect them
+	hasWarningEndpoints := false
+	for _, ep := range loadedEndpoints {
+		if ep.Status == aga.EndpointStatusWarning {
+			hasWarningEndpoints = true
+		}
+	}
+
+	// Update GlobalAccelerator status after successful deployment, including warning endpoints
```
emdelaney, line 343: So it's a success even if some endpoints still have a warning status?

pkg/aga/model_build_endpoint_group.go
emdelaney: EC2 instance endpoints support the `i-12345...` format in the endpoint ID, so if we ever support those, we'd need a way to guarantee uniqueness across regions. (such as by using the ARN of the instance, or by reading the region off of the endpoint group and using it in the key.)

pkg/aga/model_build_endpoint_group.go
```diff
@@ -146,14 +159,90 @@ func (b *defaultEndpointGroupBuilder) buildEndpointGroupSpec(ctx context.Context
...
+			if loadedEndpoint.Status == EndpointStatusLoaded {
+				// Create a base configuration with the loaded endpoint's ARN
+				endpointConfig := agamodel.EndpointConfiguration{
+					EndpointID: loadedEndpoint.ARN,
+				}
+				endpointConfig.Weight = awssdk.Int32(loadedEndpoint.Weight)
+				endpointConfig.ClientIPPreservationEnabled = ep.ClientIPPreservationEnabled
+				endpointConfigurations = append(endpointConfigurations, endpointConfig)
+			} else {
+				// Log warning for endpoints which are not loaded successfully during loading and has Warning status
```
emdelaney, line 228: Do users expect this to succeed with only some of the requested endpoints (if some cannot be loaded)?

pkg/deploy/aga/endpoint_group_manager.go
emdelaney: It's worth considering whether we can add-before-remove in at least some cases to prevent downtime during full replacements. We can discuss more offline.

shraddhabang: Got it. Sure.

pkg/aga/model_build_endpoint_group.go
shraddhabang: Yeah that makes sense. We will cross that bridge when we do the support for it.

controllers/aga/globalaccelerator_controller.go
shraddhabang: Yes. I made this design choice intentional. The one with warning status are usually the endpoints which are non-existent or does not have valid DNS yet (May be in progress of provisioning, may be in error state), both these factors are outside of our global accelerator concern. We re-queue reconcile for these type  of endpoints in the hope that user will fix the error on these so that we can register them with our accelerator.  
I don't think we should block traffic on other good endpoints because of one bad endpoint. So I divided them in three categories. Fatal errors will be blocked immediately (If k8s API are failing somehow which means its a cluster issue), warning ones will be skipped but logged for troubleshooting and loaded endpoints will go through successful reconciles.

pkg/aga/model_build_endpoint_group.go
shraddhabang: Ye refer to my earlier comment for this design decision.

pkg/aga/model_build_endpoint_group_test.go
zac-nixon: I meant there should be unit tests that use other namespaces rather than just default.

PR #4476, opened 2025-11-29
[feat aga] Implement auto-discovery feature for supported endpoints

shraddhabang:
This PR adds an auto-discovery feature for AWS Global Accelerator that automatically detects and configures protocol and port ranges based on the underlying endpoint's configuration. This "one-click accelerator" approach simplifies configuration for users by eliminating the need to manually specify these settings.

[[feat aga] Implement autp-discovery feature for supported endpoints](https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/a087ab647b77f30b70d4d1be0a9191c26b7bd626)

The auto-discovery feature detects protocol and port ranges from the referenced endpoint, enabling users to create accelerators with minimal configuration. The feature:

1. Automatically detects TCP and UDP protocols from endpoint listeners

2. Discovers port ranges based on the endpoint type:

   - Services: Uses Service port definitions and protocols
   - Ingresses: Uses ALB listener ports (with support for ALB annotations)
   - Gateways: Uses Gateway API listener configuration
   - EndpointID: Resolves load balancer ARN and queries AWS for listener information

3. Intelligently consolidates ports into ranges for efficient AWS Global Accelerator configuration

4. Works exclusively with simple accelerator configurations (single listener with a single endpoint group containing a single endpoint)

## Usage

Users can now create a Global Accelerator with minimal configuration:

```yaml
kind: GlobalAccelerator
metadata:
  name: autodiscovery-accelerator
  namespace: my-namespace
spec:
  name: "autodiscovery-test-accelerator"
  listeners:
    - # Protocol and portRanges will be auto-discovered
      endpointGroups:
        - endpoints:
            - type: Ingress  # Can also be Service, Gateway, or EndpointID
              name: web-ingress
              namespace: my-namespace
              weight: 200
```

This feature is currently limited to configurations which has exactly one endpoint referenced. 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/aga/endpoint_discovery.go
+++ b/pkg/aga/endpoint_discovery.go
@@ -2,19 +2,18 @@ package aga
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	awssdk "github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2"
 	elbv2types "github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2/types"
 	"github.com/go-logr/logr"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/aws/services"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/shared_utils"
 
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 
 	agaapi "sigs.k8s.io/aws-load-balancer-controller/apis/aga/v1beta1"
-	elbv2api "sigs.k8s.io/aws-load-balancer-controller/apis/elbv2/v1beta1"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/annotations"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/k8s"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -51,7 +50,7 @@ func NewEndpointDiscovery(client client.Client, logger logr.Logger, elbv2Client 
 // 1. Identify the endpoint type (Service, Ingress, Gateway, or LoadBalancer via EndpointID)
 // 2. Extract protocol and port information from the stored K8s resource or AWS API
 // 3. For Service endpoints, handle both TCP and UDP protocols based on the Service definition
-// 4. For Ingress endpoints, parse annotations to get port configurations
+// 4. For Ingress endpoints, extract ports from the load balancer status
 // 5. For Gateway endpoints, map Gateway protocols to GlobalAccelerator protocols
 // 6. For LoadBalancer (EndpointID) endpoints, query AWS API to get listener information
 func (d *EndpointDiscovery) FetchProtocolPortInfo(ctx context.Context, endpoint *LoadedEndpoint) ([]ProtocolPortInfo, error) {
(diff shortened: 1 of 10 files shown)


pkg/aga/endpoint_discovery.go
```diff
@@ -0,0 +1,410 @@
...
+	// Check if there's a certificate (from annotations or IngressClassParams)
+	hasCert, err := d.ingressHasCertificate(ctx, ing)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check for ingress certificates: %w", err)
+	}
+
+	// Check for listen-ports annotation
+	var tcpPorts []int32
+	rawListenPorts := ""
+	if d.annotationParser.ParseStringAnnotation(annotations.IngressSuffixListenPorts, &rawListenPorts, ing.Annotations) {
```
zac-nixon: If you have the ARN already, I wonder if it's more durable (and less error prone) to just query the ELB API to get the listener ports. I feel like implementing this would actually remove a lot of the code here.

pkg/aga/endpoint_discovery.go
```diff
@@ -0,0 +1,410 @@
...
+	if err := d.client.Get(ctx, client.ObjectKey{Name: ingressClass.Spec.Parameters.Name}, ingressClassParams); err != nil {
+		return false, fmt.Errorf("failed to get IngressClassParams %s: %w", ingressClass.Spec.Parameters.Name, err)
+	}
+
+	// Check if certificate ARNs are defined
+	return len(ingressClassParams.Spec.CertificateArn) > 0, nil
+}
+
+// getProtocolPortFromELBListener get the protocol and port info from ELB listener
+func (d *EndpointDiscovery) getProtocolPortFromELBListener(ctx context.Context, lbARN string) ([]ProtocolPortInfo, error) {
```
zac-nixon, line 235: Do we need any sort of caching here? This might pose some throttling issues if we have to run this describe on every loop iteration.

pkg/aga/endpoint_discovery.go
```diff
@@ -0,0 +1,410 @@
...
+			// Any other protocols are not supported by Global Accelerator
+			return nil, fmt.Errorf("listener protocol %s is not supported by Global Accelerator for load balancer %s",
+				listenerProtocol, lbARN)
+		}
+	}
+
+	// Create protocol port info objects
+	var protocolPortsInfo []ProtocolPortInfo
+	if len(tcpPorts) > 0 {
+		// When test expectations are checking for multiple protocol groups, split each port into its own group
```
zac-nixon: Can you elaborate more on this comment? We should be doing the most performant implementation.

pkg/aga/endpoint_discovery.go
```diff
@@ -0,0 +1,410 @@
...
+
+	// Return TCP protocol with discovered ports
+	return []ProtocolPortInfo{
+		{Protocol: agaapi.GlobalAcceleratorProtocolTCP, Ports: tcpPorts},
+	}, nil
+}
+
+// ingressHasCertificate checks if an Ingress has certificates defined either in
+// annotations or IngressClassParams
+func (d *EndpointDiscovery) ingressHasCertificate(ctx context.Context, ing *networkingv1.Ingress) (bool, error) {
```
wweiwei-li: Other than annotations or IngressClassParams,  should we also consider `ing.Spec.TLS` as an indictor for their intent to have secure listener 443

pkg/aga/endpoint_discovery.go
```diff
@@ -0,0 +1,410 @@
...
+	}
+
+	return protocolPortsInfo, nil
+}
+
+// fetchLoadBalancerProtocolPortInfo extracts protocol and port information from a LoadBalancer ARN
+// This uses the AWS API to retrieve ELBv2 listener information
+func (d *EndpointDiscovery) fetchLoadBalancerProtocolPortInfo(ctx context.Context, endpoint *LoadedEndpoint) ([]ProtocolPortInfo, error) {
+	lbARN := endpoint.ARN
+	if lbARN == "" {
```
wweiwei-li: nit: duplicate check since we already checked it by caller

pkg/aga/endpoint_discovery.go
```diff
@@ -0,0 +1,410 @@
...
+	for port := range udpPortsMap {
+		udpPorts = append(udpPorts, port)
+	}
+
+	// Group by protocol for a more efficient implementation
+	// This matches what would be done in production, where we want to minimize the number of listeners
+	var protocolPortsInfo []ProtocolPortInfo
+
+	// Create a single protocol group for each protocol type
+	if len(tcpPorts) > 0 {
```
wweiwei-li: nit: we have used this protocol grouping logic three times.  Maybe we could use a helper.

pkg/aga/model_build_listener.go
```diff
@@ -111,6 +137,153 @@ func buildListenerPortRanges(_ context.Context, listener agaapi.GlobalAccelerato
...
+	if len(protocolPortsInfo) == 0 {
+		err := errors.New("no protocol or port information found for auto-discovery")
+		b.logger.Error(err, "unable to auto-discover listener configuration",
+			"endpoint", loadedEndpoint.Name,
+			"accelerator", k8s.NamespacedName(ga))
+		return nil, err
+	}
+
+	// Create new listeners for each protocol detected
+	var listenersToProcess []agaapi.GlobalAcceleratorListener
```
wweiwei-li: nit:  how about use pre allocated slice for `listenersToProcess` since we know the length, it is same as len(protocolsToCreate),

pkg/aga/model_build_listener.go
```diff
@@ -111,6 +137,153 @@ func buildListenerPortRanges(_ context.Context, listener agaapi.GlobalAccelerato
...
+		}
+	} else if len(protocolPortsInfo) == 1 {
+		// Single protocol - create one listener
+		protocolsToCreate = []agaapi.GlobalAcceleratorProtocol{protocolPortsInfo[0].Protocol}
+	}
+
+	// Create listeners for each protocol
+	for _, protocol := range protocolsToCreate {
+		// Get matching ports for this protocol
+		var matchingPorts []int32
```
wweiwei-li, line 195: would there be a case that no matching port for the protocol ?  Should we skip listener creation for it here

pkg/aga/model_build_listener_test.go
```diff
@@ -96,6 +109,180 @@ func TestDefaultListenerBuilder_Build(t *testing.T) {
...
+				return mockClient, mockElbv2Client
+			},
+			wantListeners:   1,
+			wantProtocol:    agamodel.ProtocolTCP,
+			wantPortCount:   2,
+			wantErr:         false,
+			isAutoDiscovery: true,
+		},
+		{
+			name: "with auto-discovery for Service endpoint with mixed TCP/UDP protocols",
```
wweiwei-li, line 256: Nit:  for mixed case, can we add verification to make sure both protocols exist

pkg/aga/utils.go
```diff
@@ -45,3 +47,96 @@ func IsAGAControllerEnabled(featureGates config.FeatureGates, region string) boo
...
+}
+
+// CanApplyAutoDiscoveryForGA checks if auto-discovery can be applied for the GlobalAccelerator
+// Auto-discovery is only applicable if:
+// 1. There's exactly one listener
+// 2. The listener has exactly one endpoint group
+// 3. The endpoint group has exactly one endpoint
+// 4. The protocol or port ranges are not specified (needing discovery)
+// 5. The loaded endpoint is usable (successful loading with valid ARN)
+func CanApplyAutoDiscoveryForGA(ga *agaapi.GlobalAccelerator, loadedEndpoints []*LoadedEndpoint) bool {
```
zac-nixon: I don't think this needs to be exported.

pkg/aga/utils.go
```diff
@@ -45,3 +47,96 @@ func IsAGAControllerEnabled(featureGates config.FeatureGates, region string) boo
 
 	return true
 }
+
+// consolidatePortRanges combines consecutive ports into ranges
+func ConsolidatePortRanges(ports []int32) []agamodel.PortRange {
```
zac-nixon: Same, doesn't need to be exported.

pkg/aga/endpoint_discovery.go
shraddhabang: Thats a valid point and I did think about it and let me explain why Matt and I agreed on this route. 
Having an valid ARN does not necessarily guarantee us the desired state of expected resources. We might end up in reconcile errors on our Ingress and may not have the desired state on our load balancers. As user is referring an ingress, we should assume they always want us to use the desired state on the resources itself for their configuration in accelerator. Hence I am relying on the desired state of listener ports from annotations rather than the actual state through ELB API.

pkg/aga/endpoint_discovery.go
shraddhabang: I chose not to do caching here because I want to always get the actual ports. We don't have event driven system like our other k8s resources for ARN endpoints. So the reconciles are the only true way to get the latest info from these resources so I wanted to have fresh data on these as much as possible. What do you think?

pkg/aga/model_build_listener_test.go
shraddhabang: We verify number of expected listeners for mixed cases which is 2. What other verification you mean?

pkg/aga/endpoint_discovery.go
shraddhabang: No unless we specify the listen port annotations specifically to include HTTPS, the `ing.Spec.TLS` has no affect on setting secure listeners.

pkg/aga/endpoint_discovery.go
shraddhabang: This is a leftover comment. Removed it. I was building this in different way earlier and changed the approach later.

pkg/aga/model_build_listener.go
shraddhabang: We will have matching port for protocol if we fetch it from endpoint.  There might be case where the listener has configured a protocol but we dont get matching port from our endpoint for that protocol, If there are no matching port for a protocol, Our listener that we create to process will have empty port ranges which will be caught later during building listener spec. So here we only wanna build the listener from the info we have. But the actual error will be caught eventually during the building spec like we do for non-autodiscovery endpoints.

pkg/aga/endpoint_discovery.go
zac-nixon: I would really prefer to not re-implement this logic.

1/ It's potentially error prone, as you have to write the logic twice.
2/ Any updates to our listener creation logic must be reflected here.
3/ In terms of separation of concerns, the AGA controller shouldn't have this intricate Ingress logic.

To address your point, since your working on the desired state of the Ingress rather than the actual state you might misconfigure the AGA target, suppose that the ALB has a desired port of 443 but it's not materialized. You now configure the AGA with port 443 and start black holing traffic.

shraddhabang: Discussed this offline. Implemented the newer approach suggested by Zac for extracting the ports from Ingress and Service status instead of duplicating the parsing logic.

Issue #4477, opened 2025-11-29
Private IPv4 targets are not registered when cross-zone load balancing is enabled


JakubSzuber:
**Bug Description**

I have 2 peered VPCs. The first one is for general-purpose, where among others, ALB nodes reside, and the second one is for IPs of EKS nodes and pods only.

I use `ip-mode` for my target group, so the traffic from ALB is routed directly to IPv4 addresses of pods.

Everything was working fine until I added a new value to `alb.ingress.kubernetes.io/target-group-attributes` - `load_balancing.cross_zone.enabled=false`. After disabling the cross-zone load balancing none of the targets were registered. In the logs of the LB controller deployment, I can see an example error message:
```json
{
   "level":"error",
   "ts":"2025-11-29T22:12:06Z",
   "msg":"Reconciler error",
   "controller":"targetGroupBinding",
   "controllerGroup":"elbv2.k8s.aws",
   "controllerKind":"TargetGroupBinding",
   "TargetGroupBinding":{
      "name":"k8s-gobacken-gobacken-fc702f12bc",
      "namespace":"go-backend"
   },
   "namespace":"go-backend",
   "name":"k8s-gobacken-gobacken-fc702f12bc",
   "reconcileID":"99181ed3-1c2a-4a0d-a534-98c9ce3af0d5",
   "error":"operation error Elastic Load Balancing v2: RegisterTargets, https response error StatusCode: 400, RequestID: dee398b3-b0b0-419c-a209-240689556b50, api error ValidationError: When cross-zone load balancing is disabled on this target group, you must specify an Availability Zone for IP target '172.16.0.108' since it is outside of the VPC"
}
```

The relevant part:
```
ValidationError: When cross-zone load balancing is disabled on this target group, you must specify an Availability Zone for IP target '172.16.0.108' since it is outside of the VPC
```

This is partially expected because the AWS API requires explicitly specifying the AZ when we try to register a private IPv4 from a peered VPC as a target. However, the LB controller should automatically discover AZ of the underlying EC2 instance and provide this data when registering a new target, but it seems that it doesn't do that. I also didn't find any information about this problem of the LB controller anywhere.

When I manually go into the AWS console and register a target (private IPv4 of pod) with specifying a proper AZ, then the target is successfully registered and everything works perfectly fine. Also, when I go back to my previous setup where cross-zone is disabled, then everything works fine too.

[...]

PR #4496, opened 2025-12-11
handle cross zone disabled for alb

shuqz:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4477

### Description
For ALB, if cross_zone=disabled and AZ is either not specified or all, we get error from api 
```
ValidationError: When cross-zone load balancing is disabled on this target group, you must specify an Availability Zone for IP target '172.16.0.108' since it is outside of the VPC
```
it only succeed when az is specified. this does not apply to NLB
Fix implemented: 
- introduced a cache: needsPodAZCache tracks TGBs that require pod availability zones during target registration.
- catch error from registerTarget, if it is validationError, try register with pod AZ, if still fails, fail with error

### Checklist
- [x] Manually tested
created peered VPC, trying to register pod from VPC 1 to target group in VPC2. Before fix, no target is registered. After fix, target is registered. tested with yaml (`load_balancing.cross_zone.enabled=false`)
```
---
apiVersion: v1
kind: Namespace
metadata:
  name: game-2048
---
apiVersion: apps/v1
kind: Deployment
metadata:
  namespace: game-2048
  name: deployment-2048
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: app-2048
  replicas: 5
  template:
    metadata:
      labels:
        app.kubernetes.io/name: app-2048
    spec:
      containers:
        - image: public.ecr.aws/l6m2t8p7/docker-2048:latest
          imagePullPolicy: Always
          name: app-2048
          ports:
            - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  namespace: game-2048
  name: service-2048
spec:
  ports:
    - port: 80
      targetPort: 80
      protocol: TCP
  type: NodePort
  selector:
    app.kubernetes.io/name: app-2048
---
kind: Ingress
metadata:
  namespace: game-2048
  name: ingress-2048
  annotations:
    alb.ingress.kubernetes.io/target-group-attributes: load_balancing.cross_zone.enabled=false
spec:
  ingressClassName: alb
  rules:
    - http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: service-2048
                port:
                  number: 80
---
kind: TargetGroupBinding
metadata:
  name: peered-tgb
  namespace: game-2048
spec:
  serviceRef:
    name: service-2048
    port: 80
  targetGroupARN: arn:aws:elasticloadbalancing:us-west-2::targetgroup/test-tg/
  vpcID: vpc- # The peered VPC ID (different from cluster VPC)
```
Screenshot 2025-12-12 at 5 41 23 PM
Screenshot 2025-12-12 at 5 41 53 PM

### DIFF
--- a/pkg/targetgroupbinding/resource_manager.go
+++ b/pkg/targetgroupbinding/resource_manager.go
@@ -641,7 +641,11 @@ func (m *defaultResourceManager) prepareRegistrationCall(ctx context.Context, en
 		}
 		if doAzOverride(podIP) {
 			if usePodAZ && !usingCrossAccount {
-				if az := m.getPodAvailabilityZone(ctx, endpoint.Pod); az != nil {
+				az, err := m.getPodAvailabilityZone(ctx, endpoint.Pod)
+				if err != nil {
+					return sdkTargets, err
+				}
+				if az != nil {
 					target.AvailabilityZone = az
 				} else {
 					m.logger.Info("Failed to get pod AZ, falling back to 'all'", "pod", endpoint.Pod.Key)
@@ -911,32 +915,32 @@ func isAZValidationError(err error) bool {
 	return false
 }
 
-func (m *defaultResourceManager) getPodAvailabilityZone(ctx context.Context, pod k8s.PodInfo) *string {
+func (m *defaultResourceManager) getPodAvailabilityZone(ctx context.Context, pod k8s.PodInfo) (*string, error) {
 	if pod.NodeName == "" {
-		return nil
+		return nil, errors.Errorf("pod %s has no node assigned", pod.Key)
 	}
 
 	m.nodeAZCacheMutex.RLock()
 	if cachedAZ, ok := m.nodeAZCache.Get(pod.NodeName); ok {
 		m.nodeAZCacheMutex.RUnlock()
 		az := cachedAZ.(string)
-		return &az
+		return &az, nil
 	}
 	m.nodeAZCacheMutex.RUnlock()
 
 	node := &corev1.Node{}
 	if err := m.k8sClient.Get(ctx, types.NamespacedName{Name: pod.NodeName}, node); err != nil {
-		return nil
+		return nil, err
 	}
 
 	az, ok := node.Labels[corev1.LabelTopologyZone]
 	if !ok {
-		return nil
+		return nil, nil
 	}
 
 	m.nodeAZCacheMutex.Lock()
 	m.nodeAZCache.Set(pod.NodeName, az, m.nodeAZCacheTTL)
 	m.nodeAZCacheMutex.Unlock()
 
-	return &az
+	return &az, nil
 }
--- a/pkg/targetgroupbinding/resource_manager_test.go
+++ b/pkg/targetgroupbinding/resource_manager_test.go
@@ -773,14 +773,6 @@ func Test_defaultResourceManager_getPodAvailabilityZone(t *testing.T) {
 			nodeLabels: map[string]string{},
 			wantAZ:     nil,
 		},
-		{
-			name: "pod without node assigned",
-			pod: k8s.PodInfo{
-				Key:      types.NamespacedName{Namespace: "default", Name: "my-pod"},
-				NodeName: "",
-			},
-			wantAZ: nil,
-		},
 	}
 
 	for _, tt := range tests {
@@ -808,7 +800,8 @@ func Test_defaultResourceManager_getPodAvailabilityZone(t *testing.T) {
 			}
 
 			ctx := context.Background()
-			got := m.getPodAvailabilityZone(ctx, tt.pod)
+			got, err := m.getPodAvailabilityZone(ctx, tt.pod)
+			assert.NoError(t, err)
 
 			if tt.wantAZ == nil {
 				assert.Nil(t, got)
@@ -850,7 +843,8 @@ func Test_defaultResourceManager_getPodAvailabilityZone_Cache(t *testing.T) {
 	ctx := context.Background()
 
 	// First call - should fetch from K8s API
-	az1 := m.getPodAvailabilityZone(ctx, pod)
+	az1, err := m.getPodAvailabilityZone(ctx, pod)
+	assert.NoError(t, err)
 	assert.NotNil(t, az1)
 	assert.Equal(t, "us-east-1a", *az1)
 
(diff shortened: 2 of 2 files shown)


pkg/targetgroupbinding/resource_manager.go
```diff
@@ -844,3 +864,57 @@ func (m *defaultResourceManager) getMaxNewTargets(newTargetCount int, currentTar
...
+func (m *defaultResourceManager) getServiceAnnotations(ctx context.Context, tgb *elbv2api.TargetGroupBinding) (map[string]string, error) {
+	svcKey := buildServiceReferenceKey(tgb, tgb.Spec.ServiceRef)
+	svc := &corev1.Service{}
+	if err := m.k8sClient.Get(ctx, svcKey, svc); err != nil {
+		return nil, err
+	}
+	return svc.Annotations, nil
+}
+
+func (m *defaultResourceManager) isALBIngress(svcAnnotations map[string]string) bool {
```
zac-nixon: This might not always work. Customers aren't required to add ingress annotations to their Services.

It's more reliable to use the TG protocol: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/apis/elbv2/v1beta1/targetgroupbinding_types.go#L143-L146

pkg/targetgroupbinding/resource_manager.go
```diff
@@ -844,3 +864,57 @@ func (m *defaultResourceManager) getMaxNewTargets(newTargetCount int, currentTar
...
+	for key := range svcAnnotations {
+		if strings.HasPrefix(key, annotations.AnnotationPrefixIngress) {
+			return true
+		}
+	}
+	return false
+}
+
+func (m *defaultResourceManager) isCrossZoneDisabled(svcAnnotations map[string]string, isALB bool) bool {
```
zac-nixon: For Ingress objects, this could go on the Ingress or Service, meaning that this could be incorrect

pkg/targetgroupbinding/resource_manager.go
```diff
@@ -844,3 +864,57 @@ func (m *defaultResourceManager) getMaxNewTargets(newTargetCount int, currentTar
...
+	if err := m.k8sClient.Get(ctx, types.NamespacedName{Name: pod.NodeName}, node); err != nil {
+		return "", err
+	}
+
+	az, ok := node.Labels[corev1.LabelTopologyZone]
+	// fallback: try legacy/deprecated label failure-domain.beta.kubernetes.io/zone
+	if !ok {
+		az, ok = node.Labels["failure-domain.beta.kubernetes.io/zone"]
+	}
+	if !ok {
```
zac-nixon, line 937: This will block reconciles on non-EKS clusters. I suspect that's not what we want.

pkg/targetgroupbinding/resource_manager.go
```diff
@@ -844,3 +864,57 @@ func (m *defaultResourceManager) getMaxNewTargets(newTargetCount int, currentTar
...
+		}
+	} else {
+		if attrs, ok := svcAnnotations[annotations.SvcBetaAnnotationPrefix+"/"+annotations.SvcLBSuffixLoadBalancerAttributes]; ok {
+			return strings.Contains(attrs, crossZoneDisabled)
+		}
+	}
+	return false
+}
+
+func (m *defaultResourceManager) getPodAvailabilityZone(ctx context.Context, pod k8s.PodInfo) (string, error) {
```
zac-nixon: What do you think about using a node-az cache? That way we don't have to recalculate the node AZ each reconcile loop iteration.

pkg/targetgroupbinding/resource_manager.go
zac-nixon: I realized that this only works for the Ingress API. We will want a solution that works ALB Gateways too.

pkg/targetgroupbinding/resource_manager.go
zac-nixon: See my other comment, this will never work for ALB Gateways.

pkg/targetgroupbinding/resource_manager.go
shuqz: i thought about TG protocol too, but did not realize they are not required to add annotation. 
yea i forgot gateway api integrations, i will include it

pkg/targetgroupbinding/resource_manager.go
shuqz: i can remove this fallback part

pkg/targetgroupbinding/resource_manager.go
zac-nixon: If the protocol isn't added by user, then we infer and inject it.
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/webhooks/elbv2/targetgroupbinding_mutator.go#L82-L84

pkg/targetgroupbinding/resource_manager.go
shuqz: the other more striaghtforward way is to use describe api get cross-zone value.

pkg/targetgroupbinding/resource_manager.go
```diff
@@ -844,3 +890,31 @@ func (m *defaultResourceManager) getMaxNewTargets(newTargetCount int, currentTar
 
 	return newTargetCount
 }
+
+func isAZValidationError(err error) bool {
+	var apiErr smithy.APIError
+	if errors.As(err, &apiErr) {
+		isMatch := apiErr.ErrorCode() == "ValidationError" &&
+			strings.Contains(apiErr.ErrorMessage(), "you must specify an Availability Zone")
```
shuqz, line 912: i am not sure if they will change error message but checking ValidationError alone might not enough

pkg/annotations/constants.go
```diff
@@ -81,6 +81,7 @@ const (
 
 	// NLB annotation suffixes
 	// prefixes service.beta.kubernetes.io, service.kubernetes.io
+	SvcBetaAnnotationPrefix                              = "service.beta.kubernetes.io"
```
zac-nixon: nit: no longer used.

pkg/targetgroupbinding/resource_manager.go
zac-nixon: I don't see the node cache here, it might be useful to reduce duplicate calculations.

pkg/targetgroupbinding/resource_manager.go
zac-nixon: This still blocks on non-eks clusters

zac-nixon (approved):
/lgtm
/approved

PR #4502, opened 2025-12-15
Add check for duplicate endpoints

wweiwei-li:
### Description

1. Added checkForDuplicateEndpoints() function to detect duplicate endpoints
- Added getEndpointKey() helper to generate unique keys for endpoints (type/namespace/name)
- Validates on both Create and Update operations
- Handles different endpoint types (Service, Ingress, Gateway, EndpointID)

2. Added 4 test cases covering:
- Valid: unique endpoints
- Invalid: duplicate endpoints (same name/namespace)
- Valid: same name in different namespaces
- Valid: same name with different endpoint types

3. Added GlobalAccelerator webhook to the helm template (it was in config but missed in helm) ! It's conditionally rendered when the feature gate is enabled

4. E2E Test Improvements
- Fixed GatewayClass creation to handle AlreadyExists errors - This is for preparing in parallel test 
- Added proper namespace cleanup in ingress tests
- Fixed listener verification to be order-independent - This fix sometimes multiple listener tests failed due to API return order
- Updated skip messages

5. Logging Improvements
- Changed "no drift" log messages from Info to V(1) (debug level) in accelerator/listener/endpoint managers

6. Minor Fixes
- Changed kubectl column name from "NAME" to "ACCELERATOR-NAME" for clarity. This avoids the confusion of having two "NAME" columns. 

### Checklist
- [x] Manually tested

### DIFF
--- a/test/e2e/gateway/common_resource_stack.go
+++ b/test/e2e/gateway/common_resource_stack.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	elbv2gw "sigs.k8s.io/aws-load-balancer-controller/apis/gateway/v1beta1"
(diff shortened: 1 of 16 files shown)


test/e2e/gateway/common_resource_stack.go
```diff
@@ -208,7 +209,14 @@ func deleteReferenceGrants(ctx context.Context, f *framework.Framework, refGrant
 
 func createGatewayClass(ctx context.Context, f *framework.Framework, gwc *gwv1.GatewayClass) error {
```
zac-nixon, line 210: Curious what clean up looks like here. All GW tests will clean up the GWC, which could potentially interfere with the parallel execution.

zac-nixon (approved):

test/e2e/gateway/common_resource_stack.go
wweiwei-li: It happened when I was trying to test `ginkgo -timeout 3h -v -p -r test/e2e` locally. With `-p`, it actually parallelizes individual test within each suite. In this case, I noticed my aga gateway endpoints test failed at creating gateway class due to "already exists" error.

Issue #3915, opened 2024-10-28
EIP allocations and subnet ID order is not respected


dgadodia:
**Describe the bug**
We are using `service.beta.kubernetes.io/aws-load-balancer-eip-allocations` and `service.beta.kubernetes.io/aws-load-balancer-subnets` annotations to create a mapping between EIP and subnets.
As per [docs](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/annotations/) , EIP allocations has this side note `Length/order must match subnets`

However this order is not respected because we end up sorting the list of subnets.
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/990ee2e98224e96ef8ec443e027dcb6ada3bb4f8/pkg/networking/subnet_resolver.go#L461

As a result, we end up mapping EIP to subnets which we don't intend (want) to

**Steps to reproduce**

Deployed k8s manifest example
```
apiVersion: v1
kind: Service
metadata:
  annotations:
    kubernetes.io/ingress.class: alb
    service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags: service=myapp-gateway,vertical=platform,cost_center=infrastructure
    service.beta.kubernetes.io/aws-load-balancer-alpn-policy: HTTP2Preferred
    service.beta.kubernetes.io/aws-load-balancer-attributes: deletion_protection.enabled=true,access_logs.s3.enabled=true,access_logs.s3.bucket=myapp-logs-internal-us-west-2,access_logs.s3.prefix=myapp-gateway-access-logs
    service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: 'false'
    service.beta.kubernetes.io/aws-load-balancer-eip-allocations: eip-w,eip-x,eip-y,eip-z
    service.beta.kubernetes.io/aws-load-balancer-ip-address-type: dualstack
    service.beta.kubernetes.io/aws-load-balancer-name: myapp-gateway
    service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
    service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: '*'
    service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
    service.beta.kubernetes.io/aws-load-balancer-ssl-cert: arn:aws:acm:us-west-2:1234567890:certificate/some-random-uuid
[...]

PR #4504, opened 2025-12-15
Fix: preserve requested order for subnets when using aws-load-balancer-subnets annotation

nelsen129:
### Issue

#3915 

### Description

#3925 previously attempted to fix the issue of the subnet ID and EIP allocation order not matching. It did this by removing sorting from the output of AWS `DescribeSubnets`. However, since `DescribeSubnets` doesn't guarantee an order, this change made the order go from being deterministically wrong to being completely non-deterministic.

Now, this is using a map and re-generating the slice at the end of the `listSubnetsByNameOrIDs` to ensure that the output list is in the same order as the input.

### Checklist
- [X] Manually tested

### DIFF
--- a/pkg/networking/subnet_resolver.go
+++ b/pkg/networking/subnet_resolver.go
@@ -240,6 +240,7 @@ func (r *defaultSubnetsResolver) ResolveViaNameOrIDSlice(ctx context.Context, su
 }
 
 // listSubnetsByNameOrIDs lists subnets within vpc matching given ID or name.
+// The returned subnets will be in the same order as the input subnetNameOrIDs slice.
 func (r *defaultSubnetsResolver) listSubnetsByNameOrIDs(ctx context.Context, subnetNameOrIDs []string) ([]ec2types.Subnet, error) {
 	var subnetIDs []string
 	var subnetNames []string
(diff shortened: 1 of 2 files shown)


nelsen129:
@shuqz @zac-nixon can I get a review on this? Thanks!

zac-nixon:
I don't quite understand what this is fixing. Do you have an example / reproduction?

nelsen129:
We use pre-created EIP allocations and subnets. At creation time, we internally "assign" the EIP allocations to subnets (e.g., `eip-aaa` should map to `subnet-aaa` and `eip-bbb` should map to `subnet-bbb`). We expect that using the `service.beta.kubernetes.io/aws-load-balancer-eip-allocations` and `service.beta.kubernetes.io/aws-load-balancer-subnets` annotations, the order should be preserved between two lists.

However, the current behavior is that there is a `DescribeSubnets` API call, and the returning list is the correct order. But `DescribeSubnets` does not guarantee an order:

```
> aws ec2 describe-subnets --query "Subnets[*].SubnetId" --output text --subnet-ids "subnet-0b92XXXXXXXXXXXX" "subnet-0bb8XXXXXXXXXXXX"
subnet-0bb8XXXXXXXXXXXX        subnet-0b92XXXXXXXXXXXX
```

So as a result, the order is functionally random, and we'd have to add extra logic later to figure out which EIPs match to which subnets

Note that in the v2.2 docs, they stated that [the order is preserved](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/annotations/). This was removed in v2.3. If this was removed because of an intentional decision to not preserve the order, this is a breaking change for us.

Previously, the order was wrong, but deterministic. https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3925 removed a sort that did put it in the wrong order, but at least we could predict that behavior.

This PR changes the order to be consistent with how it's defined in the annotation.

wweiwei-li (approved):

zac-nixon (approved):
/lgtm
/approved

Thanks for fixing this. Sorry about the delayed review.

Issue #4517, opened 2025-12-22
Helm Chart 1.17.0 includes duplicated CRD globalaccelerators, causing kustomize render to fail


motoki317:
**Bug Description**
`aws-load-balancer-controller` helm chart 1.17.0 includes a duplicated CRD definition of `globalaccelerators.aga.k8s.aws`.

**Steps to Reproduce**
1. Create `kustomization.yaml`.
```yaml
helmCharts:
  - repo: https://aws.github.io/eks-charts
    name: aws-load-balancer-controller
    version: "1.17.0"
    includeCRDs: true
    namespace: aws-load-balancer-controller
    releaseName: aws-load-balancer-controller
    valuesFile: values.yaml
```

2. Create `values.yaml`.
```yaml
clusterName: "test"
```

3. Run `kustomize build --enable-helm`.

**Expected Behavior**
Helm is rendered correctly.

**Actual Behavior**
Kustomize fails with the following error:
```
Error: could not parse rnode slice into resource map: may not add resource with an already registered id: CustomResourceDefinition.v1.apiextensions.k8s.io/globalaccelerators.aga.k8s.aws.[noNs]
```

(Note that `helm template` still works fine; helm doesn't seem to check for duplicated resources. I'm not sure what would happen with `helm install` because I'm not using helm commands directly - I'm simply rendering helm charts from kustomization.yaml and applying the templated manifests via ArgoCD.)

**Regression**
Kustomize rendering was working fine in helm chart 1.16.0.
The bug was probably introduced in this commit: https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/54f05c862a28c03a7e6d0fd669bf893d7231f1cb

**Current Workarounds**
Use helm chart 1.16.0.

**Environment**
- `kustomize version`: v5.8.0
- `helm version`: version.BuildInfo{Version:"v3.19.1", GitCommit:"v3.19.1", GitTreeState:"", GoVersion:"go1.25.4"}

**Possible Solution (Optional)**
The chart has duplicated definitions of the globalaccelerators CRD at both:
- https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/c35898ad3f4bdc83564592e1ccba3bea4f41a2d9/helm/aws-load-balancer-controller/crds/crds.yaml
[...]

PR #4518, opened 2025-12-22
fix: duplicated CRD in helm kustomization

wweiwei-li:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4517

### Description
Issue: The globalaccelerators.aga.k8s.aws CRD is duplicated in the Helm chart

Included in crds.yaml
Also listed separately in aga-crds.yaml
Both referenced in crds/kustomization.yaml
Impact:
✅ Direct Helm users: No impact - helm install and helm upgrade work normally.
❌ Kustomize/ArgoCD users with includeCRDs: true : Deployment fails with duplicate resource error.

This fix is to remove `aga-crds.yaml` from kustomization.yaml
-->

### Checklist
- [ ] Manually tested


Makefile
```diff
@@ -83,7 +81,6 @@ crds: manifests
 	$(KUSTOMIZE) build config/crd/aga > config/crd/aga/aga-crds.yaml
 	echo '---' > config/crd/aga/aga-crds.yaml
 	$(KUSTOMIZE) build config/crd/aga >> config/crd/aga/aga-crds.yaml
-	$(COPY_AGA_CRDS_TO_HELM)
```
shuqz, line 86: we have 
```
	$(KUSTOMIZE) build config/crd/gateway > config/crd/gateway/gateway-crds.yaml
	echo '---' > config/crd/gateway/gateway-crds.yaml
	$(KUSTOMIZE) build config/crd/gateway >> config/crd/gateway/gateway-crds.yaml
$(COPY_GATEWAY_CRDS_TO_HELM)
``` 
for gateway api because we want to separate out gateway api crds into its own directory. now since want to consider aga at same level as bases, i am wondering should we move aga crd definitions to /bases instead of /aga

wweiwei-li: Yeah, good question. I was debating.  but I am thinking just keep it as it is, so it is in separate/aga for organization, but included by default in  https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/config/crd/kustomization.yaml (like bases)
```
resources:
  - bases/elbv2.k8s.aws_targetgroupbindings.yaml
  - bases/elbv2.k8s.aws_ingressclassparams.yaml
  - bases/elbv2.k8s.aws_albtargetcontrolconfigs.yaml
  - aga/aga-crds.yaml
 ```

PR #4524, opened 2025-12-26
fix missing status for stuck TargetGroupBinding deletion

elizhabiboullah:
### Description

This PR improves the observability of `TargetGroupBinding` resources during deletion. Previously, if the controller failed to clean up AWS resources (e.g., due to missing permissions or network issues), the kubectl delete command would return successfully, but the resource would remain stuck in a "Deleting" state (blocked by finalizer) with no visible feedback in the Status.

### Checklist
- [ ] Manually tested

### DIFF
--- a/config/crd/bases/elbv2.k8s.aws_targetgroupbindings.yaml
+++ b/config/crd/bases/elbv2.k8s.aws_targetgroupbindings.yaml
@@ -190,50 +190,6 @@ spec:
           status:
             description: TargetGroupBindingStatus defines the observed state of TargetGroupBinding
             properties:
-              conditions:
-                description: Conditions describe the current conditions of the TargetGroupBinding.
-                items:
-                  description: Condition contains details for one aspect of the current state of this API Resource.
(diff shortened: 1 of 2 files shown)


zac-nixon:
Is using the events a suitable replace for using conditions?

elizhabiboullah:
@zac-nixon Using Events alone for the blocked state isn't great for observability when you're running at scale. Events are transient (TTL-bound) and not designed for representing the persistent state of a resource. If a TargetGroupBinding gets stuck during deletion for a while, the original failure event might get garbage collected before anyone sees it, leaving you with a resource that's just stuck there with no visible explanation.

If we expose this as a Condition instead, the error state is visible as long as the problem exists. This lets external automation detect stuck resources using kubectl wait or JSONPath, rather than having to parse through event logs.

zac-nixon:
We will continually retry the deletion, so the cause should be found within the events. Nonetheless, adding a status is good change IMO.

PR #4531, opened 2026-01-06
[gw api] Refactor listener loading

zac-nixon:
### Description

In attempting to implement the [ListenerSet](https://gateway-api.sigs.k8s.io/geps/gep-1713/) construct I ran into a lot of issues. Our Gateway API implementation makes a lot of assumptions that ties the Gateway and Listeners. With ListenerSet, we need to be flexible and remove the assumption that the listener lives on the Gateway. This PR is step 1 of that change.

Summary of changes:

- [REFACTOR] Refactor Listener Status logic, the current logic was shared between the controller and route utils package. I separated it such that status generation is done via route utils and status setting is done in the controller package. The goal is to have the ListenerValidation object give all the info needed to generate a valid Listener status.
- [ADDITION] Added a listener discovery component, right now this just parses the gateway listeners, in an upcoming iteration this will handle listener set discovery.
- [REFACTOR] Refactored the listener logic to utilize discovered listeners in the controllers and routeutils package. In an upcoming iteration, I will do the same logic with the model builder.
- [REFACTOR] Added the Kind to parent ref when we generate the de-dupe map for setting the Route Status. This is needed because ListenerSet is a potential parent ref kind.
- [BUG FIX] There was a bug when calculating the # of routes attached to each listener. The bug was that the final result put to a listener was the number of routes attached to the listener port, this was incorrect. We need to be reporting the number of routes that attach to a individual listener.

### Checklist
- [x] Manually tested


pkg/gateway/routeutils/listener_discovery.go
```diff
@@ -0,0 +1,35 @@
...
+
+// DiscoveredListener represents a listener with its index for efficient access
+type DiscoveredListener struct {
+	Listener   gwv1.Listener
+	Parent     interface{}
+	ParentType ListenerParentType
+}
+
+// discoverListeners discovers available listeners for this Gateway.
+func discoverListeners(gw gwv1.Gateway) []DiscoveredListener {
```
shuqz, line 22: i am assuming we will include discover listeners from listener sets later?

pkg/gateway/routeutils/listener_utils.go
```diff
@@ -1,20 +1,19 @@
...
-	"context"
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 	gateway_constants "sigs.k8s.io/aws-load-balancer-controller/pkg/gateway/constants"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
 type ListenerValidationResult struct {
```
shuqz, line 11: if this result is gonna be shared by gateway and listenerSet? if yes, maybe do separate reasons in case of status is different for gateway and listener? or it is ok to modify it in followup PRs when we implement listenerSet

controllers/gateway/listener_status_utils_test.go
```diff
@@ -59,40 +38,47 @@ func Test_buildListenerStatus(t *testing.T) {
 				},
 			},
 			attachedRoutesMap: map[gwv1.SectionName]int32{"listener1": 1},
```
shuqz, line 40: thanks for fixing the count issue in attached map! can you add some unit test around it actually? i do not see we have thorough testing cases regarding attachedRoutesMap. and also want to double check, our e2e test is still passing with this new change right? (it is a foundational change so i want to be careful)

pkg/gateway/routeutils/listener_discovery.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+type ListenerParentType string
+
+const (
+	ListenerParentTypeGateway     = "Gateway"
+	ListenerParentTypeListenerSet = "ListenerSet"
```
shuqz, line 11: want to confirm, so we have one LB for one gateway (1:1), and having new listenerSet does not introduce more LB, it will share the same LB as gateway. and if gateway does not exist, listenerSet is not valid right?

shraddhabang:
Looks good to me.

Issue #4541, opened 2026-01-12
Certificate outage due to bug in rotation logic


zac-nixon:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4359#issuecomment-3737995686

```
This change introduces a critical bug. The `aws-load-balancer-tls` `Secret`, which was created using the now-removed self-signed issuer, still exists, but the Certificate that owned and renewed it no longer does.

As a result, the certificate stored in the secret eventually expires, causing TLS negotiation failures in the webhook. This, in turn, prevents the AWS Load Balancer Controller from updating target group bindings and ultimately leads to a complete outage.

In my opinion, this should have been treated as a breaking change.

We also need to apply more critical thinking before blindly accepting changes generated by Claude or other coding assistants. These tools often lack the full context or a deep understanding of the underlying system, which can lead to subtle but severe failures like this.
```

frittentheke:
FWIW, to me it does not make much sense to keep the `keepTLSSecret` logic and conditionals at all.
Re-issuing certs with the new hierarchy is a one time thing and should, as @portswigger-tim stated, just be a documented breaking change for the next release. Let's keep this already complex matter simple and clean in the templates / code.

zac-nixon:
I need to take some time to really understand what the original issue was, what this change was attempting to do, and then how to fix it. In the meantime, I am going to revert the commit that we know is bad.

portswigger-tim:
The "fix" for me was to delete the orphaned TLS secrets and re-deploy the controller to put the new CA chain in place. I'm still not 100% sure _why_ the `cert-manager` didn't update the secret.

A symptom I saw after moving to the newer chart was that every time a `helm diff` ran, it would want to create the `Certificate` object - but applying the change did not create the `Certificate` and there was no error returned. This happened across all clusters with AWS LB Controller installed. It could, of course, be a bug in `cert-manager` and how it processes changes to `Certificate` resources.

frittentheke:
@portswigger-tim also see my comment https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4542#issuecomment-3743113632 to the revert @zac-nixon did.

I simply set `keepTLSSecret=false` to use the new hierachy.
Cert-Manager will then simply recognize that the issuer of the `Certificate` pointing to the secret has changed ...

[...]

PR #4542, opened 2026-01-12
Revert "fix: Race condition in webhook certificate renewal with cert-…

zac-nixon:
There is a bug with this logic: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4541


frittentheke:
@zac-nixon @shraddhabang  ... a revert of a breaking change seems not the most sensible approach to fix this.
What about people that already adopted the new CA -> serving cert structure (new installs without existing secret or `keepTLSSecret=false`).

Just removing the conditional and documenting the breaking change in a new release seemed much cleaner TBH.

zac-nixon:
@frittentheke I've reverted the code so that I may fully understand it before attempting to fix it. I prefer to not have buggy code checked in :). Will have a PR out sometime this week with a fix.

Issue #4543, opened 2026-01-13
Helm installation fails with schema validation error: unknown model reference "k8s.io~1apimachinery~1pkg~1apis~1meta~1v1.ObjectMeta"


oliverpark999:
**Bug Description**
Helm installation of aws-load-balancer-controller fails with schema validation error. The error occurs when attempting to install the controller using the standard helm command, resulting in "unknown model in reference: k8s.io~1apimachinery~1pkg~1apis~1meta~1v1.ObjectMeta" error.

**Steps to Reproduce**
- Step-by-step guide to reproduce the bug:
  1. Add EKS helm repository: `helm repo add eks https://aws.github.io/eks-charts`
  2. Update helm repo: `helm repo update`
  3. Attempt to install aws-load-balancer-controller with the following command:
```bash
  helm install aws-load-balancer-controller eks/aws-load-balancer-controller \
    --set clusterName=foobar-eks \
    -n kube-system \
    --set serviceAccount.create=false \
    --set serviceAccount.name=aws-load-balancer-controller
```

- Manifests applied while reproducing the issue: N/A (using Helm chart)
- Controller logs/error messages while reproducing the issue:
```
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: error validating "": error validating data: SchemaError(gitlab.com/aws/eks-extension-metrics-apiserver/pkg/apis/metrics/v1.ETCD.metadata): unknown model in reference: "k8s.io~1apimachinery~1pkg~1apis~1meta~1v1.ObjectMeta"
```

**Expected Behavior**
The Helm chart should install successfully without schema validation errors, as it did in previous installations.

**Actual Behavior**
The Helm installation fails with a schema validation error referencing `k8s.io/apimachinery/pkg/apis/meta/v1.ObjectMeta`. The error appears to be related to CRD validation in the `gitlab.com/aws/eks-extension-metrics-apiserver` component.

- Impact: Controller installation fails completely, preventing new installations
- Frequency: Always (100% reproduction rate)

**Regression**
Was the functionality working correctly in a previous version ? **Yes**

The same helm install command was working successfully until approximately 24 hours ago. No changes were made to the cluster configuration or installation command.

**Current Workarounds**
Using `--disable-openapi-validation` flag bypasses the validation error:
```bash
helm install aws-load-balancer-controller eks/aws-load-balancer-controller \
  --set clusterName=foobar-eks \
  -n kube-system \
[...]

PR #4544, opened 2026-01-14
Fix AZ mismatch in refreshUnhealthyTargets causing orphaned targets

MinhNguyen-at:
### Issue

When refreshing unhealthy targets, `refreshUnhealthyTargets` queries AWS using the cached `AvailabilityZone`. If the cached AZ differs from what AWS has recorded, the target isn't found and is incorrectly treated as "not registered." This can cause orphaned targets in the ELB.

### Description

**Reproduction**
1. Pod registers with `AvailabilityZone: nil` (common for IP targets within VPC)
2. AWS auto-assigns an AZ (e.g., `us-east-1d`) to the target
3. Cache stores target with `AvailabilityZone: nil`
4. Pod becomes unhealthy or is deleted
5. `refreshUnhealthyTargets` queries AWS with `AvailabilityZone: nil`
6. AWS doesn't find the target (AZ mismatch)
7. Target filtered out by `IsNotRegistered()` check
8. Controller never deregisters the target → orphan in ELB

The controller has mitigations for AZ handling (`getPodAvailabilityZone`, `needsPodAZCache`, node topology lookups), but they don't cover all cases:

1. **Pod deletion race**: If the pod is deleted before its AZ is cached or between cache expiry and refresh, the lookup fails
2. **Node label missing**: `getPodAvailabilityZone` relies on `topology.kubernetes.io/zone` label - if missing, AZ defaults to nil
3. **Cache expiry**: After TTL, all targets refresh. If pod is gone, can't look up its AZ anymore

**Proposed Solution**

Strip AvailabilityZone when querying AWS in refreshUnhealthyTargets. Querying without AvailabilityZone finds the target regardless of what AZ AWS has recorded, avoiding cache mismatch issues.

### Checklist
- [x] Manually tested

### BONUS POINTS checklist
- [x] Backfilled missing tests for code in same general area
- [ ] Refactored something and made the world a better place

### DIFF
--- a/pkg/targetgroupbinding/targets_manager.go
+++ b/pkg/targetgroupbinding/targets_manager.go
@@ -86,7 +86,7 @@ func (m *cachedTargetsManager) RegisterTargets(ctx context.Context, tgb *elbv2ap
 	for _, targetsChunk := range targetsChunks {
 		req := &elbv2sdk.RegisterTargetsInput{
 			TargetGroupArn: aws.String(tgARN),
-			Targets:        pointerizeTargetDescriptions(targetsChunk),
+			Targets:        cloneTargetDescriptionSlice(targetsChunk),
 		}
 		m.logger.Info("registering targets",
 			"arn", tgARN,
@@ -115,7 +115,7 @@ func (m *cachedTargetsManager) DeregisterTargets(ctx context.Context, tgb *elbv2
 	for _, targetsChunk := range targetsChunks {
 		req := &elbv2sdk.DeregisterTargetsInput{
 			TargetGroupArn: aws.String(tgARN),
-			Targets:        pointerizeTargetDescriptions(targetsChunk),
+			Targets:        cloneTargetDescriptionSlice(targetsChunk),
 		}
 		m.logger.Info("deRegistering targets",
 			"arn", tgARN,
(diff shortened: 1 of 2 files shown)


zac-nixon:
/ok-to-test

Quite the interesting issue. I was able to reproduce it too -> https://gist.github.com/zac-nixon/5473d2277ee81afdc07270d7e0996364

Just out of curiosity, how often do you change the zone of an ip address?

zac-nixon:
IMO, after some testing, it looks like there's not really a good reason to ever pass in the AZ when calling DescribeTargetHealth.

What do you think about updating this function to not pass in the AZ for targets given to the DTH call? https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/ecc69caead57523dcc2c0801ae5abe6962d0baae/pkg/targetgroupbinding/targets_manager.go#L210-L223

MinhNguyen-at:
> IMO, after some testing, it looks like there's not really a good reason to ever pass in the AZ when calling DescribeTargetHealth.

I agree, I was just more confident in the refreshUnhealthyTargets case as a narrower scope.

> What do you think about updating this function to not pass in the AZ for targets given to the DTH call?

I'll update the PR, sounds better.

MinhNguyen-at:
Also renamed pointerizeTargetDescriptions which is no logner accurate with SDK v2

MinhNguyen-at:
> Just out of curiosity, how often do you change the zone of an ip address?

I think almost never in an AWS VPC only setup. We hit the nil case more often.

MinhNguyen-at:
/retest-required

test failure seems unrelated
```
  unable to build kubernetes objects from release manifest:
  error validating data: SchemaError(gitlab.com/aws/eks-extension-metrics-apiserver/pkg/apis/metrics/v1.ETCD.metadata):
  unknown model in reference: "k8s.io~1apimachinery~1pkg~1apis~1meta~1v1.ObjectMeta"
```

zac-nixon:
/retest

related issue: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4543

zac-nixon:
Thank you!

MinhNguyen-at:
Appreciate the quick review

PR #4547, opened 2026-01-16
[feat aga] Add cross-namespace reference support for AGA

shraddhabang:
This PR adds support for cross-namespace references to the AWS Global Accelerator controller, allowing a GlobalAccelerator resource to reference endpoints (Ingress, Service, Gateway) located in different namespaces using ReferenceGrant resources from the Gateway API.

## Changes

- Implemented cross-namespace validation for GlobalAccelerator endpoints
- Added support for resolving references to resources in different namespaces
- Updated endpoint loader to respect ReferenceGrant permissions
- Added ReferenceGrant event handlers to trigger reconciliation when grants are created/updated/deleted
- Added e2e tests for cross-namespace references with various endpoint types

## How It Works

The controller now checks for the existence of ReferenceGrant resources that explicitly allow cross-namespace references. When a GlobalAccelerator attempts to reference an endpoint in another namespace:

1. If no ReferenceGrant exists, the reference is rejected with a warning
2. When a valid ReferenceGrant is created, the controller reconciles the GlobalAccelerator and establishes the endpoint connection

## Testing

- Added comprehensive e2e tests for:

  - Ingress endpoints in different namespaces
  - Service endpoints in different namespaces
  - Gateway API endpoints in different namespaces
  - Auto-discovery with cross-namespace references
  - Mixed endpoint types across multiple namespaces

### Checklist
- [x] Manually tested


controllers/aga/eventhandlers/reference_grant_events.go
```diff
@@ -0,0 +1,177 @@
...
+		return
+	}
+
+	totalMatched := 0
+
+	// Process each impacted GlobalAccelerator namespace
+	for _, from := range impactedFroms {
+
+		var gaList agaapi.GlobalAcceleratorList
+		if err := h.k8sClient.List(ctx, &gaList, &client.ListOptions{Namespace: string(from.Namespace)}); err != nil {
```
zac-nixon, line 103: Is there any caching you can do here to prevent calling List() on the same namespace multiple times? I think the client should handle this case, but adding an explicit cache might make things cleaner here.

docs/guide/globalaccelerator/aga-controller.md
```diff
@@ -177,7 +177,60 @@ endpointGroups:
...
+  - endpointGroups:
+    - endpoints:
+      - namespace: web-ns  # Target namespace
+        name: web-service  # Service in target namespace
+        type: Service
+```
+
+!!!note "To use cross-namespace references"
+
+    1. The Gateway API CRDs must be installed in your cluster (specifically the ReferenceGrant CRD)
```
zac-nixon: You can be explicit here and say that Reference grant is installed via "standard crds" https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/gateway/gateway/#prerequisites

pkg/aga/cross_namespace_validator.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+	}
+
+	// No matching grant found
+	return fmt.Errorf("cross-namespace reference not allowed for reference %s/%s - no matching reference grant found",
+		toNamespace, toName)
+}
+
+// grantAllowsReference checks if a specific ReferenceGrant allows the reference
+func (v *ReferenceGrantValidator) grantAllowsReference(grant gwv1beta1.ReferenceGrant,
```
zac-nixon: Can you refactor this reference grant logic and the gateway reference grant logic into a shared resource?

zac-nixon: If you choose to not refactor this, I think it should have private visibility.

pkg/aga/endpoint_loader.go
```diff
@@ -187,13 +193,41 @@ func (l *endpointLoaderImpl) loadResourceWithDNS(
...
+			group = shared_constants.GatewayAPIResourcesGroup
+			kind = shared_constants.GatewayApiKind
+		default:
+			return NewFatalError(UnsupportedResourceTypeMsg,
+				fmt.Errorf("unsupported resource type: %s", resourceType),
+				result.EndpointRef, parentNamespace)
+		}
+
+		// Validate cross-namespace reference using ReferenceGrant
+		if err := l.crossNamespaceValidator.ValidateCrossNamespaceReference(ctx,
```
zac-nixon: I suppose you want a more graceful failure mode here when the user is attempting to use cross namespace referencing but they have forgotten to install the proper CRD.

test/e2e/globalaccelerator/gateway_endpoint_test.go
```diff
@@ -217,4 +221,352 @@ var _ = Describe("GlobalAccelerator with Gateway endpoint", func() {
...
+			gatewayNamespace string
+			gwStack          *gateway.NLBTestStack
+			gatewayName      string
+			acceleratorName  string
+			gaName           string
+			refGrantName     string
+		)
+
+		BeforeEach(func() {
+			if tf.Options.IPFamily == framework.IPv6 {
```
zac-nixon, line 410: why is ipv6 not supported?

test/e2e/globalaccelerator/ingress_endpoint_test.go
```diff
@@ -51,6 +56,27 @@ var _ = Describe("GlobalAccelerator with Ingress endpoint", func() {
 		ingStack = ingress.NewResourceStack([]*appsv1.Deployment{deployment}, []*corev1.Service{svc}, []*networkingv1.Ingress{ing})
 		err = ingStack.Deploy(ctx, tf)
 		Expect(err).NotTo(HaveOccurred())
+
+		currentTestName := CurrentSpecReport().LeafNodeText
+		if !strings.Contains(currentTestName, "cross-namespace") {
```
zac-nixon, line 61: nit: magic string

controllers/aga/eventhandlers/reference_grant_events.go
shraddhabang: Ohh are you concerned they might have multiple same namespace references in single grant or you thinking across grants?

test/e2e/globalaccelerator/gateway_endpoint_test.go
shraddhabang: The dualstack NLBs with Client IP preservations enabled are not supported hence we have disabled those in our tests.

controllers/aga/eventhandlers/reference_grant_events.go
zac-nixon: I took a look at the Reference Grant structure, and the comment I made doesn't make sense (sorry). I see the existing reference grant logic also does this same listing.

pkg/aga/endpoint_loader.go
shraddhabang: Yes added.

pkg/aga/cross_namespace_validator.go
shraddhabang: Sure refactored it.

pkg/aga/endpoint_loader.go
```diff
@@ -187,13 +191,45 @@ func (l *endpointLoaderImpl) loadResourceWithDNS(
...
+		allowed, err := shared_utils.ValidateCrossNamespaceReference(ctx, l.k8sClient,
+			parentNamespace, shared_constants.GlobalAcceleratorResourcesGroup, shared_constants.GlobalAcceleratorKind,
+			group, kind, result.Namespace, result.Name)
+		if err != nil {
+			return NewFatalError(APIServerErrorMsg,
+				err, result.EndpointRef, parentNamespace)
+		}
+		if !allowed {
+			// Return a warning error if reference is not allowed
+			return NewWarningError(CrossNamespaceReferenceMsg, err, result.EndpointRef, parentNamespace)
```
zac-nixon, line 224: ReferenceGrants should not leak information, like the resource exists but you don't have access to it. Better to return not found. [Reference](https://gateway-api.sigs.k8s.io/api-types/referencegrant/#implementation-guidelines)

shraddhabang: We only say cross-namespace reference not allowed.  We dont say whether it exists or reference not found. We do the same for gateway as well right. 
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/gateway/routeutils/backend_gateway.go#L120

PR #4553, opened 2026-01-26
feat: Add support for preserving listeners with specific tags during reconciliation

paulzhang97:
### Description

This change introduces an optional PREVENT_DELETION_TAG_NAME environment variable that allows operators to mark certain listeners on managed load balancers as externally managed, preventing the controller from deleting them during reconciliation.

Use case: This is useful when multiple systems need to manage listeners on the same NLB, such as when setting up AWS PrivateLink endpoints that require additional listeners beyond what the controller creates for Kubernetes Services.

When `PREVENT_DELETION_TAG_NAME` is set, the controller will skip deletion of any unmatched listeners that have the specified tag key, logging the skip action for visibility.

When `PREVENT_DELETION_TAG_NAME` is not set, behavior remains unchanged - all unmatched listeners are deleted as before.

### Checklist
- [✅] Manually tested
- [✅] Made sure the title of the PR is a good description that can go into the release notes

### DIFF
--- a/pkg/deploy/elbv2/listener_synthesizer.go
+++ b/pkg/deploy/elbv2/listener_synthesizer.go
@@ -2,6 +2,8 @@ package elbv2
 
 import (
 	"context"
+	"os"
+
 	awssdk "github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -71,8 +73,17 @@ func (s *listenerSynthesizer) synthesizeListenersOnLB(ctx context.Context, lbARN
 	if err != nil {
 		return err
 	}
+	preventDeletionTagName := os.Getenv("PREVENT_DELETION_TAG_NAME")
 	matchedResAndSDKLSs, unmatchedResLSs, unmatchedSDKLSs := matchResAndSDKListeners(resLSs, sdkLSs)
 	for _, sdkLS := range unmatchedSDKLSs {
+		if preventDeletionTagName != "" {
+			if _, found := sdkLS.Tags[preventDeletionTagName]; found {
+				if sdkLS.Listener.ListenerArn != nil {
+					s.logger.Info("skip deleting listener", "arn", *sdkLS.Listener.ListenerArn)
+				}
+				continue
+			}
+		}
 		if err := s.lsManager.Delete(ctx, sdkLS); err != nil {
 			return err
 		}


shraddhabang:
Can you please elaborate your use-case? Why does private-link require extra listener controller managed NLB? Can you point me to docs if there are any?

paulzhang97:
> Can you please elaborate your use-case? Why does private-link require extra listener controller managed NLB? Can you point me to docs if there are any?

In our use case, AWS Private Link ([ref1](https://docs.aws.amazon.com/vpc/latest/privatelink/concepts.html), [ref2](https://docs.aws.amazon.com/vpc/latest/privatelink/create-endpoint-service.html)), we create listeners for Private Link endpoint service on the same NLB managed by AWS LB controller for cost savings. The listeners are created outside of AWS LB controller, using Terraform.  The PREVENT_DELETION_TAG_NAME feature allows marking externally-managed listeners with a tag, instructing the controller to preserve them during reconciliation while still managing its own listeners.

shraddhabang:
> > Can you please elaborate your use-case? Why does private-link require extra listener controller managed NLB? Can you point me to docs if there are any?
> 
> In our use case, AWS Private Link ([ref1](https://docs.aws.amazon.com/vpc/latest/privatelink/concepts.html), [ref2](https://docs.aws.amazon.com/vpc/latest/privatelink/create-endpoint-service.html)), we create listeners for Private Link endpoint service on the same NLB managed by AWS LB controller for cost savings. The listeners are created outside of AWS LB controller, using Terraform. The PREVENT_DELETION_TAG_NAME feature allows marking externally-managed listeners with a tag, instructing the controller to preserve them during reconciliation while still managing its own listeners.

I totally see the cost-saving logic in sharing the NLB! but I’m curious—are we managing that listener in Terraform because the standard Service API limits us to listeners from a single Service. If thats the case, have  you considered moving toward the[ Gateway API](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v3.0/guide/gateway/l4gateway/). It would allow the AWS LB Controller to manage that shared NLB and all its listeners (across different services) in one place saving us from running into split-brain scenario.

paulzhang97:
> I totally see the cost-saving logic in sharing the NLB! but I’m curious—are we managing that listener in Terraform because the standard Service API limits us to listeners from a single Service. If thats the case, have you considered moving toward the[ Gateway API](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v3.0/guide/gateway/l4gateway/). It would allow the AWS LB Controller to manage that shared NLB and all its listeners (across different services) in one place saving us from running into split-brain scenario.

It does not seem Gateway API solve our use case, as it can't create VPC Endpoint Services for Private Link. Also not sure whether it can use an existing NLB. We deploy and mange all Private Link resources in one place with Terraform.  But I will take a look further.

shraddhabang:
> > I totally see the cost-saving logic in sharing the NLB! but I’m curious—are we managing that listener in Terraform because the standard Service API limits us to listeners from a single Service. If thats the case, have you considered moving toward the[ Gateway API](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v3.0/guide/gateway/l4gateway/). It would allow the AWS LB Controller to manage that shared NLB and all its listeners (across different services) in one place saving us from running into split-brain scenario.
> 
> It does not seem Gateway API solve our use case, as it can't create VPC Endpoint Services for Private Link. Also not sure whether it can use an existing NLB. We deploy and mange all Private Link resources in one place with Terraform. But I will take a look further.

Yeah we don't support using existing NLB with Gateway API. You will require to do a migration for that. But you could definitely manage multiple listeners. You could manage your private link service outside but atleast do listener management on NLB through controller using GW API.

zac-nixon:
Have you seen this? https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/use_cases/self_managed_lb/

It inverts control, so the NLB is managed by terraform, but your targets will still be managed by Kubernetes.

paulzhang97:
> Have you seen this? https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/use_cases/self_managed_lb/
> 
> It inverts control, so the NLB is managed by terraform, but your targets will still be managed by Kubernetes.

Thanks for the link. If we go this route, we’ll need to manage the load balancers directly, and LBC would only handle the `TargetGroupBindings` in our case. While this would work, it means we wouldn’t be able to take full advantage of LBC and would require more effort on our side to manage the load balancers.

zac-nixon:
>Thanks for the link. If we go this route, we’ll need to manage the load balancers directly, and LBC would only handle the TargetGroupBindings in our case. While this would work, it means we wouldn’t be able to take full advantage of LBC and would require more effort on our side to manage the load balancers.

Hmm, I may be missing something. The PR adds the ability to ignore listeners on an NLB, so you are managing these listeners elsewhere. Is your use-case that you have one NLB that is maintained by multiple sources (e.g. Terraform, Kubernetes?)

paulzhang97:
> Hmm, I may be missing something. The PR adds the ability to ignore listeners on an NLB, so you are managing these listeners elsewhere. Is your use-case that you have one NLB that is maintained by multiple sources (e.g. Terraform, Kubernetes?)

We create NLB through annotating k8s service with `service.beta.kubernetes.io/aws-load-balancer-` annotations. Then we use the same NLB for Private Link. Our Terraform code takes the NLB to create VPC endpoint service, listeners, TargetGroupBindings.

paulzhang97:
@shraddhabang @zac-nixon Any further comments?

paulzhang97:
> @shraddhabang @zac-nixon Any further comments?

Any other thought that can address our needs?

shraddhabang:
@paulzhang97 We need some time to discuss this internally as this is anti-pattern to have same resource managed by two different platforms. May be we can sync up to discuss more on your use-case so that we can give you a good solution for this.

paulzhang97:
Close this PR and look forward to [Private Link Support.](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1859)

Issue #4494, opened 2025-12-11
[Gateway API] Support syncing Kubernetes secrets into ACM


starlightromero:
**Describe the feature you are requesting**

Currently, the [documentation](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/gateway/gateway/#certificate-discovery-for-secure-listeners) states:
> The caveat is that configuration of TLS certificates can not be done via the certificateRefs field of a Gateway Listener, as the controller only supports certificate references via an ARN. In the future, we may support syncing Kubernetes secrets into ACM.

This issue is to request and track the support of syncing Kubernetes secrets into ACM.

**Motivation**

When using cert-manager to generate AWS PCA certificates, a Kubernetes secret of type kubernetes.io/tls is produced. Currently, the secrets has be be copied to a new secret of type opaque (a limitation of the ACK ACM Controller; [related issue](https://github.com/aws-controllers-k8s/community/issues/2721#issuecomment-3642660105)). Then imported into ACM, potentially utilizing ACK ACM's certificate import. 

This workaround is very cumbersome and can be completely negated if the AWS Load Balancer Controller supported syncing Kubernetes secrets into ACM.

**Describe the proposed solution you'd like**

AWS ALB Controller is able to import Kubernetes secrets into ACM

**Describe alternatives you've considered**

n/a

**Contribution Intention (Optional)**

-[ ] Yes, I am willing to contribute a PR to implement this feature
-[x] No, I cannot work on a PR at this time

iAnomaly:
I would love this feature as well but until then we are using this to sync cert-manager TLS secrets to AWS ACM: https://github.com/robertlestak/cert-manager-sync?tab=readme-ov-file#aws-acm

starlightromero:
> I would love this feature as well but until then we are using this to sync cert-manager TLS secrets to AWS ACM: https://github.com/robertlestak/cert-manager-sync?tab=readme-ov-file#aws-acm

Yeah my current workaround is a Kyverno Policy that generates a new opaque secret from the TLS secret. Then using ACK ACM to import the certificate from the opaque secret

zac-nixon:
We're working with the AWS security team on getting a solution approved for this project. Stay tuned 🚀

grafanalf:
[...]

PR #4554, opened 2026-01-27
feat(acm): add ACM certificate management feature

### LINKED ISSUE #2509: Automatically provision ACM certificates and attach to ALB based on spec

Right now if you want a dynamic environment based on configured Ingresses, but also includes TLS protection (for example development or demo environments provisioned on-demand) it seems the only option is to use a combination of the Nginx Ingress Controller, cert-manager (using LetsEncrypt) and external-dns. This works but is clunky and it would be nicer to use the AWS Load Balancer Controller, ALBs and ACM. It seems that this controller is most of the way there, but just needs the ability to provision a cert in ACM based on the configured host names - attaching it to the ALB is already possible.

When an Ingress object is created, the controller inspects the configured hostnames and requests a certificate be provisioned in ACM. The certificate is then attached to the configured load balancer.

As mentioned, the only workable alternative is using another Ingress controller altogether. In our case Ingresses are created by some other tooling that doesn't have the capability of creating ACM certificates and providing the ARN to this controller. This functionality is already present when using the combination of nginx-ingress/cert-manager/external-dns so there is clearly a use case that has been solved by other projects.

the-technat:
### Issue

Closes #2509 

### Description

This PR adds support for automatically provisioning ACM certificates based on ingress annotations.

An initial design idea is described in the issue: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2509#issuecomment-3737553294.

How it works: see [the prepared user-docs](https://github.com/the-technat/aws-load-balancer-controller/blob/issue/2509/docs/guide/ingress/acm_certificates.md) for a description of the feature.

I focused on ingress objects in this feature to keep the change "small". But I think most of the Synthesizer / Manager could be reused for implementing the same thing for Gateway API. Maybe as an alternative approach for #4494 (skipping cert-manager and directly manage the cert).

#### Internals
Some internals not mentioned in the user-facing docs:
- due to a certain delay between certificate requesting and issuance, the controller has a (currently non-user-configurable) time to wait for the certificate to be come issued. This was implemented similar to the wait behavior of the ELBv2 service
  - if the timeout hits before the certificate got issued, the reconciliation will fail with an error and retry according to the controller's standard retry mechanism
  - a subsequent reconciliation attempt will discover the already requested certificate and wait another round for it to become issued, unless the already requested certificate has exceeded a certain age (currently 5 minutes), in this case the certificate is recreated
- certificate state tracking is based on the Tagging Manager interface used for other resources as well, thus orphaned certificates are cleaned up as part of the PostSynthesize phase. Due to time delay between switching certificates for a listener and the listener actually releasing the previously used certificate the deletion of an orphaned certificate is in any case retried a couple of times if it's still in use.
- certificate validation (only required for Amazon Issued certificates) is only support using DNS Method and Route53. The validation records are created automatically and pruned when a certificate is deleted. Due to the possibility of multiple certificates using the same CNAME record, we ignore "not found" and "other value" errors, as another certificate not managed by the controller having the same domains / CA will use the exact same CNAME record for validation, just with another value
- ingress objects that are modified and whoose hosts set change, will trigger a new certificate to be requested with more Subject Alternative Names

#### Tests

I added unit tests where possible and feasible.

All cases have been manually tested multiple times in a real environment using EKS, PCA, ACM and an intermediate build of the AWS Load Balancer Controller

#### API rate limiting

To visualize the impact of this feature on API requests I tried collecting the number of API requests that occur **per** reconciliation attempt of **one** ingress object.

I took the rates from here: [API rate quotas](https://docs.aws.amazon.com/acm/latest/userguide/acm-limits.html#api-rate-limits) and identified the following ones as used by my feature: `RequestCertificate`, `ListTagsForCertificate`, `ListCertificates`, `DeleteCertificate` and `DescribeCertificate`. All of those operations have a rate-limit of either 5 or 10 queries per second.

Here's how often they are called:
  - `ListTagsForCertificate` & `ListCertificates`: one request every minute to rebuild the in-memory cache in the Certificate Discovery part & ACM Tagging Manager
  - `RequestCertificate`: at most once per reconciliation. If an existing matching certificate is found we'll use this and rather wait for it's issuance than request another certificate.
  - `DescribeCertificate`: is used at 3 different places:
     - DNS validation records: after requesting a certificate we wait for it to provide DNS records values. This uses a retry-mechanism with a request every 5s up to a `30s` timeout -> at most 6 requested with a `5s` delay in-between
    - certificate waiter: waits for the certificate to be issued: has timeout of `5m` and a minDelay between requests of `60s` increasing exponentially up a maxDelay of `120s`
    - Deleting Certificates: to obtain the validation records values we need to clean up, one request per certificate deletion attempt is made too
  - `DeleteCertificate`: has a retry-mechanism that tries to delete the certificate if one request fails with a retry wait interval of `5s` and a total timeout of `30s` -> per certificate there's at most 6 requests with a `5s` delay in-between
  
In addition to the ACM API rate quotas the ones for Route53 might also be relevant.

These rates are taken from here: [Route53 API request limits](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html#limits-api-requests).

Here's how often they are called:
  - `ListHostedZones`: one request every 5 minutes to rebuild the in-memory cache in the Route53 service
  - `ChangeResourceRecordSets`: one request per certificate issue request and one request per certificate deletion attempt

### Checklist
- [x] Manually tested


zac-nixon:
Hello, I appreciate that you took the time to look into this issue. In the current state of the PR, it is going to be very hard to review this properly. I would recommend when doing these large features you break up the feature into logical chunks and then commit it that way. It makes reviewing the code much easier.

the-technat:
Hi @zac-nixon, yeah I already assumed that it might be a bit too big to review. Did I understood you correct that you suggest splitting it into multiple commits but still in the same PR?

Or shall I split it into several PRs?

the-technat:
@zac-nixon I split the PR into multiple commits. Hope this helps with reviewing!

I'll probably add some more commits since there are some unit tests still failing due to the change to using `core.ResourceToken` for certificate references, and I might also add some missing unit tests, but the feature itself should be ready.

docs/guide/ingress/certificate_management.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Disable TLS certificates
+
+If you choose to opt-out of this feature by removing the annotation from an ingress object, the controller will fallback to the [certificate discovery mechanism](cert_discovery.md) or if configured, explicitly certificate ARNs.
+
+This reconciliation might fail if there are no certificates present to autodiscover. The certificate discovery mechanism is configured to ignore certificates created by the controller. Thus create a matching certificate manually for the controller to reconcile the ingress object back to using a non-managed certificate. 
+
+### Certificate Validation
+
+Amazon Issued certificates are currently validated using DNS Method and Route53 records. E-Mail validation is not supported due to significant higher delays between requesting a certificate and it's issuance. When using a PCA, certificates don't have to be validated.
```
shraddhabang: Certificate validation can take up to 30 minutes (not 5 minutes as currently implemented)
The synchronous waiting blocks the entire ingress reconciliation, preventing updates to other listeners or ingress parameters. Can we do this asynchronously?

docs/deploy/configurations.md
```diff
@@ -77,6 +77,8 @@ The --cluster-name flag is mandatory and the value must match the name of the ku
...
 | cluster-name                                                                    | string                          |                                            | Kubernetes cluster name                                                                                                                                                       |
 | default-ssl-policy                                                              | string                          | ELBSecurityPolicy-2016-08                  | Default SSL Policy that will be applied to all Ingresses or Services that do not have the SSL Policy annotation                                                               |
+| enable-acm-certificates                                                              | boolean                          | false                  | Whether to enable the [ACM Certificates feature](../guide/ingress/acm_certificates.md)|
```
shraddhabang: Can you make these a feature gate flags instead of controller flags as they are easier to manage?

shraddhabang:
1. Can you also add what e2e tests you performed?
2. I found a bug while testing that the LBC keeps on requesting new certs for ingress if the validation exceeds. It does not clean up the previous certs. Can you please fix that?

shraddhabang:
Can you add configurable e2e tests which we can run conditionally? These does not need to be enabled by default.

pkg/deploy/acm/certificate_synthesizer.go
```diff
@@ -0,0 +1,240 @@
...
+				unmatchedSDKCerts = append(unmatchedSDKCerts, sdkCert)
+				unmatchedResCerts = append(unmatchedResCerts, resCert)
+				continue
+			}
+			matchedCerts = append(matchedCerts, resAndSDKCertificatePair{
+				resCert: resCert,
+				sdkCert: sdkCert,
+			})
+			foundMatch = true
+			if !foundMatch {
```
shraddhabang: This will never be executed.

the-technat:
Thanks for the feedback @shraddhabang! Will add an E2E test.

What exactly do you mean by exceeded validation? The reissueWaitTime mechanism that I added that should recreate the cert if it's not issued within 5 minutes?

shraddhabang:
When its recreating a new cert, it should be deleting the older one which failed validation. My certs are not getting cleaned up which are failing validations.

the-technat:
@shraddhabang the bug should be fixed with the latest commit. I'm now deleting certificates whose reissueWaitTime has expired before requesting a new one, instead of adding them to the PostSynthesize queue, given that the PostSynthesize stage will only be reached when a certificate was successfully issued, thus never cleaning up previous attempts. This is safe to do since certificates not yet issued can't be in use yet. I also added another test case to the Synthesizer test to cover this.

Regarding E2E test scenarios, here's what I've tested:
- Happy path creating an Amazon issued certificate for an ingress who has no certificates before
- Happy path but for an ingress that had previously auto-discovered certificates attached
- Adding the `acm-pca-arn` annotation to an ingress that already managed it's certificate, resulting in a switch from an amazon_issued to a private certificate
- removing the `create-acm-cert` annotation, causing the controller to fall-back to certificate discovery
- adding both the `create-acm-cert` annotation as well as the `certificate-arn` annotation pinning to an existing certificate -> the `certificate-arn` takes precedence and no certificate is created
- updating an ingress object by adding another ingress host rule causing the controller to request a new certificate with more SANs
- adding the `default-pca-arn` flag on the controller and issuing certificates from this PCA. Tried overriding this using `acm-pca-arn` annotation and value of empty string, but that doesn't work. Overriding with another PCA ARN should work though

Regarding your requested E2E test I might need some guidance. Looking at the E2E testing directory I guess I'm simply going to add another test file with some test cases for my specific case right?

Shall I use amazon issued certificates (require a domain in route53 for validation) or private ones (require a PCA to be preprovisioned) in this E2E test?

shraddhabang:
Thank you for the detailed info.

> 
> Regarding your requested E2E test I might need some guidance. Looking at the E2E testing directory I guess I'm simply going to add another test file with some test cases for my specific case right?
> 
> Shall I use amazon issued certificates (require a domain in route53 for validation) or private ones (require a PCA to be preprovisioned) in this E2E test?

Yes please add a new test file covering e2e tests for this feature. You can provide options to provide the domain or PCA to set while running the tests. If none are provided, then we can skip the tests.

shraddhabang (changes requested):

docs/deploy/configurations.md
the-technat: I moved it to the feature gates section. The `--default-pca-arn` is still a flag on the ingress config. Shall I move this somewhere else too?

docs/guide/ingress/certificate_management.md
the-technat: I've added a note to the docs mentioning these 30 minutes and raised the timeout in the controller to 30 minutes. Regarding asynchronously waiting for certificates, I might need some more guidance.

As far as I understand the controller-runtime there's already an option to reconcile multiple ingress objects concurrently thats disabled by default. Reading your comment carefully I see you didn't exactly meant that but concurrency inside the reconciliation of a single ingress object, specifically inside the listeners part. To avoid one reconciler being reconciled within seconds and one blocking reconciliation I guess the only real option would to be to put reconciliation of the listeners in background goroutines and return, which would still be problematic since PostSynthesize can't continue cleaning up orphaned resources that might still be in use. So we'd have to put that into background as well and somehow find these processes again if another reconciliation attempt is made in the mean time to avoid spawning too many processes.

Maybe I made it more complex than it is, but right now I don't see how this cold work🤷.

the-technat:
@shraddhabang I added an E2E test that verifies a certificate is created either using amazon issued or private certificates. It can be used by running the script with env vars:

```console
ENABLE_CERT_MGMT_TESTS=true ROUTE53_VALIDATION_DOMAIN=albc.technat.dev  ./scripts/ci_e2e_test.sh 

## or 

ENABLE_CERT_MGMT_TESTS=true PCA_ARN=arn:aws:... ./scripts/ci_e2e_test.sh 
```

The needed IAM policy is then deployed, the flag on the controller is enabled and the respective test is not skipped.

I've skipped complex reconciliation cases as that's already covered by the `TestSynthesize` unit test. Let me know if we need more test cases in the E2E test as well.

shraddhabang:
@the-technat Thanks for adding the tests. I will take a look at this early next week.

the-technat:
@shraddhabang did you already had a chance to review the PR?

shraddhabang:
> @shraddhabang did you already had a chance to review the PR?

Hey I am still working through it. Thanks for checking in.

shraddhabang:
Hey I found one more bug in cert reuse. I create a cert for an ingress with a hostname. It succeeds, but now if I create an ingress with same hostname, the cert discovery finds the same cert LBC created earlier and attaches it. If i delete the original ingress, it will try to clean up the cert as well. Although the deletion fails at this time. But the deletion fails, we get this error. "Failed deploy model due to waited too long for the ALB to release the certificate: timed out waiting for the condition". I think we should fix the cert-discovery logic to not allow re-using the same cert for another ingress. The cert must be tied to single ingress to avoid such cases.

the-technat:
> Hey I found one more bug in cert reuse. I create a cert for an ingress with a hostname. It succeeds, but now if I create an ingress with same hostname, the cert discovery finds the same cert LBC created earlier and attaches it. If i delete the original ingress, it will try to clean up the cert as well. Although the deletion fails at this time. But the deletion fails, we get this error. "Failed deploy model due to waited too long for the ALB to release the certificate: timed out waiting for the condition". I think we should fix the cert-discovery logic to not allow re-using the same cert for another ingress. The cert must be tied to single ingress to avoid such cases.

Good catch! That should be avoided. I already added some extensions to the cert discovery to exclude certs that are owned by the controller but that only takes into account certs that the controller created for this specific ingress resource, not other ingress resources. Will extend this accordingly.

the-technat:
@shraddhabang  I adjusted the existing cert discovery filter slightly to only check the tag keys but not their values. This effectively filters out any certificate from cert discovery that has stack tags on them, no matter what stack / cluster it is.

shraddhabang:
@the-technat I appreciate your patience for this and the time you spent on this. Please note that since this is a big feature we need to follow the detailed security review for this one internally. This may take some time to merge this completely till then. Just want to update you. I will continue to post the updates on this based on that and our own testing.

pkg/deploy/acm/certificate_manager.go
```diff
@@ -0,0 +1,288 @@
...
+		req.SubjectAlternativeNames = certModel.Spec.SubjectAlternativeNames
+	}
+	if certModel.Spec.KeyAlgorithm != "" {
+		req.KeyAlgorithm = certModel.Spec.KeyAlgorithm
+	}
+
+	c.logger.Info("requesting certificate", "resourceID", certModel.ID())
+	resp, err := c.acmClient.RequestCertificateWithContext(ctx, req)
+	if err != nil {
+		return nil, err
```
shraddhabang, line 185: When the number of hostnames exceeds the ACM SAN quota, the error surfaced to the user is the raw SDK error which is noisy and hard to parse:
```
operation error ACM: RequestCertificate, exceeded maximum number of attempts, 10, 
https response error StatusCode: 400, RequestID: 248c12ab-..., 
LimitExceededException: Certificate has too many domains.
```
Two asks:

- catch `LimitExceededException` after the `RequestCertificateWithContext` call and wrap it with a cleaner message: something like this
```
resp, err := c.acmClient.RequestCertificateWithContext(ctx, req)
if err != nil {
    var limitErr *acmtypes.LimitExceededException
    if errors.As(err, &limitErr) {
        return nil, fmt.Errorf("ACM certificate request failed: too many Subject Alternative Names (SANs): %w", err)
    }
    return nil, err
}
```

- Add a note in the user-facing docs under a "Limitations" or "Known Limits" section mentioning the SAN quota:
```
The number of hostnames (Subject Alternative Names) you can specify in spec.tls[].hosts on a single ingress is limited by your AWS account's ACM SAN quota. The default limit is 10 SANs per certificate. If you need more, request a quota increase via the AWS Service Quotas console
```

pkg/deploy/acm/certificate_manager.go
```diff
@@ -0,0 +1,288 @@
...
+										Value: opts.ResourceRecord.Value,
+									},
+								},
+							},
+						},
+					},
+				},
+			}
+			_, err = c.route53Client.ChangeRecordsWithContext(ctx, input)
+			if err != nil && !strings.Contains(err.Error(), "exist") { // ignore other cert having the same validation records
```
shraddhabang: Lets add better logging here. 
```
if err != nil && strings.Contains(err.Error(), "exist") {
    c.logger.Info("validation record already exists, skipping creation",
        "certificateARN", resp.CertificateArn,
        "record", opts.ResourceRecord.Name)
} else if err != nil {
    return nil, err
}
```

docs/deploy/configurations.md
```diff
@@ -96,7 +97,7 @@ The --cluster-name flag is mandatory and the value must match the name of the ku
...
 | ingress-max-concurrent-reconciles                                               | int                             | 3                                          | Maximum number of concurrently running reconcile loops for ingress                                                                                                            |
-| kubeconfig                                                                      | string                          | in-cluster config                          | Path to the kubeconfig file containing authorization and API server information                                                                                               |
+| kubecble-aconfig                                                                      | string                          | in-cluster config                          | Path to the kubeconfig file containing authorization and API server information                                                                                               |
```
shraddhabang: Typo: `kubecble-aconfig`

docs/deploy/configurations.md
```diff
@@ -194,7 +195,9 @@ There are a set of key=value pairs that describe AWS load balancer controller fe
 | GlobalAcceleratorController           | string                          | false        | Enable the Global Accelerator controller for managing AWS Global Accelerator resources through Kubernetes CRDs                                                                                                                                                    |
 | EnhancedDefaultBehavior              | string                          | false        | Enable this feature to allow the controller to remove Provisioned Capacity or mTLS settings by removing the corresponding annotation.                                                                                                                             |
 | EnableDefaultTagsLowPriority          | string                          | false        | If enabled, tags supplied via `--default-tags` will be overridden by tags specified in other manners, like via annotations.                            |
+<<<<<<< HEAD
```
shraddhabang: Fix this please

shraddhabang:
@the-technat Good news. We are getting ready to merge your PR. Can you address the latest comments and please resolve the merge conflicts?

pkg/deploy/acm/certificate_manager.go
the-technat: Good point. I added the error catch & limitations section to the docs of my feature.

pkg/deploy/acm/certificate_manager.go
the-technat: Should be added

the-technat:
> @the-technat Good news. We are getting ready to merge your PR. Can you address the latest comments and please resolve the merge conflicts?

🎉 looking forward to using the feature soon! I've rebased and addressed your comments.

docs/guide/ingress/certificate_management 2.md
```diff
@@ -0,0 +1,52 @@
+# ACM Certificate Management
```
shraddhabang: You have duplicated files. Can you fix this?

the-technat: Should be fixed, sorry that was a leftover from some rebase that went wrong.

the-technat:
@shraddhabang had to rebase once more, but as far as I can see now it looks good.

shraddhabang:
Hey @the-technat , Did you run E2E test suite? I found few problems while running it myself. These are test bugs only. But can you check once?
1. `ExpectOneLBProvisionedForIngress` timeout too short

The cert management tests call `ExpectOneLBProvisionedForIngress` which uses IngressReconcileTimeout (1 minute). For cert management ingresses, the ALB can't be created until the certificate is issued, which takes 1-2 minutes for DNS validation. The 1-minute timeout expires before the ALB is ready.

Fix: Use CertReconcileTimeout (30 min) for the LB provisioning check in cert management tests.

2. ExpectCertToBeInUse has no retry

`ExpectCertToBeInUse` calls `DescribeCertificate` once and checks `InUseBy`. I think ACM's InUseBy field has eventual consistency — it may not reflect a newly attached ALB immediately. The check fails intermittently.

Fix: Wrap in Eventually with PollTimeoutMedium (5 min) and PollIntervalMedium (10s).

3. FindCertificateByHostnames matches wrong cert

FindCertificateByHostnames in `certificate.go`
 searches by hostname only. If the AWS account has an old cert with the same hostname (e.g., from a previous manual test), it returns that cert instead of the one the controller just created. This causes ExpectCertToBeInUse to fail because the old cert isn't attached to the new ALB.

Fix: Add a stackTag parameter and filter by `ingress.k8s.aws/stack` tag to match the specific cert created for this ingress

the-technat:
@shraddhabang I did run the E2E tests a couple of times and they eventually passed, but I  was also in a hurry to finish this feature so it's very likely that they aren't good enough yet. I also used a fresh account for testing, which is probably not a very realistic scenario.

I added your suggestions and I'm currently rerunning the tests to see if everything still works. Maybe you could try running them again as well?

For the FindCertificatesByHostname helper I copied some code from `tagging_manager.go` to avoid having a dependency on it while testing, but that creates redundant code. Let me know if you think that's the bigger issue and I'm happy to rewrite it to use the Tagging Manager's code for finding and filtering the certs...

zac-nixon (approved):
/lgtm
/approved

Thank you for hard work here! It's a huge change.

shraddhabang:
/approved
Thanks a lot @the-technat. This is awesome work.

the-technat:
Thanks for the patient and throughout review! Can't wait to tell my team about this new feature. Also feel free to tag me on bug reports that relate to this feature, I'm happy to help with maintaining the feature too.

Issue #4557, opened 2026-01-29
When a LoadBalancerConfiguration is attached to a GatewayClass, the controller pods will panic.


leynebe:
Sorry if this is a duplicate, I couldn't immediately find a similar issue.

**Bug Description**
When a LoadBalancerConfiguration is attached to a GatewayClass, the controller pods will panic.

**Steps to Reproduce**

- Step-by-step guide to reproduce the bug:
  * use kubernetes aws eks 1.34.2 and kubectl 1.34.2
     ```
        Client Version: v1.34.2
        Kustomize Version: v5.7.1
        Server Version: v1.34.2-eks-b3126f4
     ```
  * deploy **experimental** gateway-api 1.3.0 (https://github.com/kubernetes-sigs/gateway-api/config/crd/experimental?timeout=120&ref=v1.3.0)
  * deploy aws lbc v3.0.0
  * deploy custom manifests below
- Manifests applied while reproducing the issue:
```
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: example
spec:
  controllerName: gateway.k8s.aws/nlb
  parametersRef:
    group: gateway.k8s.aws
    kind: LoadBalancerConfiguration
    name: example-global
---
apiVersion: gateway.k8s.aws/v1beta1
kind: LoadBalancerConfiguration
metadata:
  name: example-global
spec:
  ipAddressType: ipv4
```
- Controller logs/error messages while reproducing the issue:
```
...
{"level":"error","ts":"2026-01-29T09:52:01Z","msg":"Reconciler error","controller":"aws-lbc-gateway-class-controller","object":{"name":"example"},"namespace":"","name":"example","reconcileID":"32030b9d-442c-46f7-8196-776f6e7b1a0c","error":"Namespace must be specified in ParametersRef"}
[...]

PR #4562, opened 2026-01-31
Fix NPE on invalid parameters ref, add more debugging details to accepted status

zac-nixon:
### Issue

[](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4557)

### Description

When using LoadBalancerConfiguration on GatewayClass, it's possible to crash the controller by forgetting to specify a namespace on the loadbalancerconfiguration. This fixes all occurrences of NPEs on forgotten namespace, it also correctly sets the accepted status to false for the gateway class.

I also found that we are setting the accepted status = true once and never changing it for gateways. I changed the logic to make the accepted status be true IIF we can successfully deploy the current gateway config. I wanted to add more debug logs to the Gateway status too.

This code was generated by Claude, but I was able to perform E2E tests and clean it up a bit so it's not all AI slop :)

### Checklist
- [x] Manually tested

### DIFF
--- a/controllers/gateway/utils.go
+++ b/controllers/gateway/utils.go
@@ -109,19 +109,19 @@ func prepareGatewayConditionUpdate(gw *gwv1.Gateway, targetConditionType string,
 		}
 	}
 
-	// 32768 is the max message limit
 	truncatedMessage := truncateMessage(message)
 
 	if indxToUpdate != -1 {
-		if derivedCondition.Status != newStatus || derivedCondition.Message != truncatedMessage || derivedCondition.Reason != reason || derivedCondition.ObservedGeneration != gw.Generation {
-			gw.Status.Conditions[indxToUpdate].LastTransitionTime = metav1.NewTime(time.Now())
-			gw.Status.Conditions[indxToUpdate].ObservedGeneration = gw.Generation
-			gw.Status.Conditions[indxToUpdate].Status = newStatus
-			gw.Status.Conditions[indxToUpdate].Message = truncatedMessage
-			gw.Status.Conditions[indxToUpdate].Reason = reason
-			return true
+		if derivedCondition.Status == newStatus && derivedCondition.Message == truncatedMessage && derivedCondition.Reason == reason && derivedCondition.ObservedGeneration == gw.Generation {
+			return false
 		}
-		return false
+
+		gw.Status.Conditions[indxToUpdate].LastTransitionTime = metav1.NewTime(time.Now())
+		gw.Status.Conditions[indxToUpdate].ObservedGeneration = gw.Generation
+		gw.Status.Conditions[indxToUpdate].Status = newStatus
+		gw.Status.Conditions[indxToUpdate].Message = truncatedMessage
+		gw.Status.Conditions[indxToUpdate].Reason = reason
+		return true
 	}
 
 	// Condition doesn't exist, create it
--- a/controllers/gateway/gateway_controller.go
+++ b/controllers/gateway/gateway_controller.go
@@ -259,6 +259,7 @@ func (r *gatewayReconciler) reconcileHelper(ctx context.Context, req reconcile.R
 	stack, lb, newAddOnConfig, backendSGRequired, secrets, err := r.buildModel(ctx, gw, mergedLbConfig, allRoutes, currentAddOns, isDeleting)
 
 	if err != nil {
+		r.handleReconcileError(ctx, gw, err)
 		return err
 	}
 	r.logger.V(1).Info("Got this addon config", "current", currentAddOns, "new addon", newAddOnConfig)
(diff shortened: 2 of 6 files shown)


controllers/gateway/utils.go
```diff
@@ -25,6 +26,19 @@ const (
 
 	// The max message that can be stored in a condition
 	maxMessageLength = 32700
+
+	// messageSimilarityThreshold is the minimum similarity ratio for two messages
+	// to be considered equivalent. This allows for dynamic parts like request IDs,
+	// timestamps, or other identifiers to change without triggering status updates.
+	// Set to 70% to be more relaxed - prefer skipping status updates over risking
+	// an infinite reconcile loop from constantly changing dynamic content.
```
shraddhabang: We will be missing the request ids here which might help them to troubleshoot. What do you think? I this request ids will be helpful.

controllers/gateway/utils.go
```diff
@@ -25,6 +26,19 @@ const (
...
+	messageSimilarityThreshold = 0.70
+
+	// messageOnlyUpdateInterval is the minimum time that must pass before updating
+	// a condition when only the message has changed (status, reason, and generation
+	// are the same). This prevents rapid status updates from dynamic content like
+	// request IDs while still eventually reflecting the latest error message.
+	messageOnlyUpdateInterval = 1 * time.Minute
```
shraddhabang: Is this too high? Will we miss the status updates if the messages are not similar within this period?

controllers/gateway/utils.go
zac-nixon: I don't follow the comment. We are still putting the request ID into the condition. This check is to prevent a infinite reconcile loop. Basically, everytime we update the Gateway (even if's a condition) we trigger another reconcile run. If we always update the condition with a new request ID, then we will run the loop way too often.

In my testing, this threshold is good enough to catch just the request ID changing. For error messages that truly just change the request ID the likeness is usually 90-95%.

controllers/gateway/utils.go
zac-nixon: Maybe this should be an OR? 

```
			// Update only if transition time is old enough AND messages are dissimilar
			if timeSinceLastTransition >= messageOnlyUpdateInterval || !messagesSimilar {
				gw.Status.Conditions[indxToUpdate].LastTransitionTime = metav1.NewTime(time.Now())
				gw.Status.Conditions[indxToUpdate].Message = truncatedMessage
				return true
			}
```

What do you think?

controllers/gateway/utils.go
shraddhabang: I meant wouldn't the customer want latest request ids to give to service team for the debugging? If we dont update the request Ids, their will be mismatch between what customer is seeing on the status and what the actual failed request is. I do agree that changing the status just because the request IDs has changed will cause frequent unnecessary updates. If we want to get around that, should we just remove these dynamic parts from the message all together and only put the error message. That way they will know what the actual error message is and then can dive into logs for finding the request Ids instead of taking an older one. This will remove the frequent updates caused because of the dynamic parts.

controllers/gateway/utils.go
shraddhabang: Yeah this has to be OR. but I was thinking if we only put message in the condition then we may not need to check the intervals for rapid updates.

controllers/gateway/utils.go
zac-nixon: I don't think there's a deterministic way to remove the dynamic parts of a message, hence why I used the edit distance as the proxy of what has change. I like putting the request ID in the condition, that way someone can easily give us a request id rather than digging through the logs which has been a pain point.

controllers/gateway/utils.go
zac-nixon: Actually, I was thinking on this more. I think we want to keep this as an AND. Suppose that we hit a new error message (messagesSimilar = false) BUT the transition time is 30 seconds, so we don't write a new condition.

The reconcile loop will retry later anyways (thats how we do all reconcile errors) and this time we will correct the condition message. The delays new error messages by ~1 minute, but also prevents run away reconcile loops.

controllers/gateway/utils.go
zac-nixon: Edit distance also handles other dynamic fields, e.g. if an error message adds a timestamp into the message.

shraddhabang: I understand your intent behind putting request IDs in the conditions directly. But I think the current logic creates the opposite problem: stale request IDs that create debugging confusion. For example this case `message only change with old transition time but similar message - should NOT update`, the AND logic written below will essentially never update the message with the latest request IDs even if we pass the transition time. We will be always looking at the older requests(the first time they hit this error) rather than the latest one and then will have to request the logs anyway eventually to dive deeper to find the latest request ID.

controllers/gateway/utils.go
shraddhabang: If `messagesSimilar = false`, that means we have genuinely new, different information. Why should we even delay communicating that to the user by one minute? The whole point of the similarity check is to distinguish between "same problem, different request ID" (ignore/delay updates) vs. "different problem" (report). 

Can you please reconsider using OR logic here which will solve our problem of rapid status updates.

- __If messages are dissimilar:__ Update immediately (catches new error types)
- __If messages are similar but transition time has passed:__ Update for periodic freshness.

Issue #4566, opened 2026-02-03
Single Ingress with 46 paths only produces 7 ListenerRules in built model


neilkuan:
# Single Ingress with 46 paths only produces 7 ListenerRules in built model

## Description

When a single Ingress has **one host**, **one** `spec.rules` entry, and **46** entries in `spec.rules[0].http.paths`, the controller's built model only contains **7** `AWS::ElasticLoadBalancingV2::ListenerRule` resources (`443:1` through `443:7`). The remaining 39 paths do not get their own listener rules.

Is this expected? If so, is there a way to get one ALB rule per path, or an option to disable path merging?

## Environment

- AWS Load Balancer Controller version: v2.16.0
- Kubernetes version: 1.34
- Ingress: single Ingress, `ingressClassName: alb`
- Host: `a.example.com`
- Path count in `spec.rules[0].http.paths`: **46**
- Path types: mix of `Exact`, `Prefix`, and `ImplementationSpecific` (regex)

## Expected behavior
- deployed yaml
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: my-app-new-sidebar
  namespace: be-app
  labels:
    helm.sh/chart: be-app-6.1.0-rc.1
    app: my-app
    service: be-app
    app.kubernetes.io/managed-by: Helm
  annotations:
    alb.ingress.kubernetes.io/actions.default-action: |
      {"Type":"fixed-response","FixedResponseConfig":{"ContentType":"text/plain","StatusCode":"403","MessageBody":"403 permission denied"}}
    alb.ingress.kubernetes.io/actions.first-order-rules: |
      {"Type":"fixed-response","FixedResponseConfig":{"ContentType":"text/plain","StatusCode":"404","MessageBody":"404"}}
    alb.ingress.kubernetes.io/actions.doeing-302: |
      {"Type": "redirect", "RedirectConfig": { "host": "b.example.com", "path": "/#{path}", "query": "#{query}", "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_302"}}
    alb.ingress.kubernetes.io/actions.doeing-default-pin-code-302: |
      {"Type": "redirect", "RedirectConfig": { "host": "b.example.com", "path": "/w/pin_code", "query": "#{query}", "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_302"}}
[...]

PR #4569, opened 2026-02-05
Fix rule optimizer incorrectly removing regex-based listener rules

shraddhabang:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4566

### Description

ALB rule optimizer removes redundant rules to save space. However, it was only checking literal path patterns and completely ignored regex patterns. When customers used regex paths, the optimizer saw empty path lists and incorrectly assumed all rules were the same, removing most of them. This PR fixes this issue. 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ingress/rule_optimizer.go
+++ b/pkg/ingress/rule_optimizer.go
@@ -98,6 +98,10 @@ func isInfiniteRedirectRule(port int32, protocol elbv2model.Protocol, rule Rule)
 		case condition.Field == elbv2model.RuleConditionFieldHostHeader && condition.HostHeaderConfig != nil:
 			ruleHosts.Insert(condition.HostHeaderConfig.Values...)
 		case condition.Field == elbv2model.RuleConditionFieldPathPattern && condition.PathPatternConfig != nil:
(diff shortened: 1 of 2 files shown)


pkg/ingress/rule_optimizer.go
```diff
@@ -146,7 +146,11 @@ func isSupersetConditions(lhsConditions []elbv2model.RuleCondition, rhsCondition
 		case condition.Field == elbv2model.RuleConditionFieldHostHeader && condition.HostHeaderConfig != nil:
```
zac-nixon, line 150: Do you need to add the regex values [here](https://github.com/shraddhabang/aws-load-balancer-controller/blob/4ec10863b98aa32272a3d97705a52cc9f3faa977/pkg/ingress/rule_optimizer.go#L101)?

shraddhabang: Yes we need to skip this check for regex paths. Good catch.

wweiwei-li (approved):

PR #4588, opened 2026-02-14
Fix subnet resolution when multiple subnets share the same Name tag

bjhaid:
### Description
The order preservation fix in commit defc1c17 inadvertently broke support for having multiple subnets with the same Name tag across different availability zones. This is actually a valid AWS configuration that we rely on in our setup.

The bug happened because the code used a simple map to store subnets by their Name tag. When multiple subnets shared the same name, only the last one would be kept in the map. Then when the user specified that name multiple times (expecting to get different subnets in different AZs), they'd get the same subnet repeated, which would fail validation with "multiple subnets in same Availability Zone".

This fix changes the storage from a simple map to a multi-map (map to slice), so all subnets with the same name can be tracked. When the same name appears multiple times in the user's input, different subnets from that list are handed out in order. This way both use cases work:
- Users can have multiple subnets with the same Name tag (now fixed)
- Subnet order is preserved as requested (still works)

The AWS query now deduplicates names before sending them to AWS, which avoids redundant lookups while keeping behavior deterministic. The validation was updated to check that at least one subnet exists for each unique name, rather than doing a strict count match that doesn't make sense when names can be duplicated.

This change adds test coverage for the duplicate name scenario and verifies all existing tests still pass.

Fixes: defc1c17 ("Fix: preserve requested order for subnets when using aws-load-balancer-subnets annotation")

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/networking/subnet_resolver.go
+++ b/pkg/networking/subnet_resolver.go
@@ -253,7 +253,7 @@ func (r *defaultSubnetsResolver) listSubnetsByNameOrIDs(ctx context.Context, sub
 	}
 
 	subnetByID := make(map[string]ec2types.Subnet)
-	subnetByName := make(map[string]ec2types.Subnet)
+	subnetsByName := make(map[string][]ec2types.Subnet)
 
 	if len(subnetIDs) > 0 {
 		subnets, err := r.listSubnetsByIDs(ctx, subnetIDs)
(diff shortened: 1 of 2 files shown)


bjhaid:
@shraddhabang what do we need to do to merge this PR?

shuqz (approved):

PR #4602, opened 2026-02-25
add auto detection of gateway api resources

zac-nixon:
### Description

Auto-detect Gateway API resources, in order to auto enable Gateway API controllers.

{"Gateway", "GatewayClass", "HTTPRoute", "GRPCRoute"} => Enable ALB Gateway Controller
{"TCPRoute", "UDPRoute", "TLSRoute"} => Enable NLB Gateway Controller

This also changes the default values to be true, if the resources aren't detected we disable the feature flag, thereby not interacting with the gateway api resources. Further, it still allows installations to disable the feature flag by setting it explicitly false.

### Checklist
- [x] Manually tested


main.go
```diff
@@ -175,6 +177,10 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Gateway API CRD auto-detection: check which CRDs are installed and disable
+	// feature flags for missing CRDs before any controller setup reads them.
```
shuqz, line 181: i have a question, so if user explicitly set `--feature-gates=NLBGatewayAPI=true` but CRDs does not exist for NLB, do we force it to be disabled?

pkg/gateway/crddetect/gateway_feature_flags.go
```diff
@@ -0,0 +1,57 @@
...
+		logger.Info("Disabling ALBGatewayAPI: missing standard Gateway API CRDs",
+			"groupVersion", standardResult.GroupVersion,
+			"missing", standardResult.MissingKinds)
+		featureGates.Disable(config.ALBGatewayAPI)
+	} else {
+		logger.Info("All standard Gateway API CRDs detected, ALBGatewayAPI remains enabled",
+			"groupVersion", standardResult.GroupVersion)
+	}
+
+	if !standardResult.AllPresent || !experimentalResult.AllPresent {
```
shuqz, line 47: i thought standardResult includes HTTPRoute and GRPCRoutes, why should NLB be disabled if it misses any of these

pkg/gateway/crddetect/gateway_feature_flags.go
```diff
@@ -0,0 +1,57 @@
...
+	standardResult := k8s.DetectCRDs(client, GatewayV1GroupVersion, StandardCRDKinds)
+	experimentalResult := k8s.DetectCRDs(client, GatewayV1Alpha2GroupVersion, ExperimentalCRDKinds)
+
+	ApplyGatewayFeatureFlags(standardResult, experimentalResult, featureGates, logger)
+}
+
+// ApplyGatewayFeatureFlags applies the Gateway CRD detection results to the
+// feature gates. Extracted for testability — accepts pre-computed results.
+func ApplyGatewayFeatureFlags(standardResult, experimentalResult k8s.CRDGroupResult, featureGates config.FeatureGates, logger logr.Logger) {
+	if !standardResult.AllPresent {
```
shuqz, line 37: i am a bit confused on why we use AllPresent here. in current logic, even if i have gw, gwclass, httpRoute, it is still disabled since i am missing grpcroute. why?

main.go
zac-nixon: Yes. This is because it will cause the controller to crash.

pkg/gateway/crddetect/gateway_feature_flags.go
zac-nixon: standardResult includes gateway and gatewayclass, which is required for NLB.

pkg/gateway/crddetect/gateway_feature_flags.go
zac-nixon: If you have a corrupted install (missing GRPCRoute) then the controller will crash.

pkg/gateway/crddetect/gateway_feature_flags.go
shuqz: yes gateway and gwclass is required for NLB. but not HTTPRoute and GRPCRoute. i see standardResult includes HTTPRoute and GRPCRoute and you are using `!standardResult.AllPresent`.
though i can understand if they have GW & GWClass installed, it is likely they are HTTPRoute GRPCRoute installed too since they are installed together as "standard crd"

pkg/gateway/crddetect/gateway_feature_flags.go
shuqz: i see

shuqz:
i was originally thinking it auto discovers based on actual custom resource instances exist. but now i see it is based on whether the CRD is installed. 
/lgtm

Issue #4025, opened 2025-01-21
Fairer allocation of targets to target groups when quota exceeded


cunningr:
**Describe the bug**

Note: This may not always be considered a bug but there may be an improvement in behaviour for this abnormal condition.

The condition is hit the AWS quota limit for the number of targets.  As we slowly scale up a cluster that is the target for an NLB with multiple ports, each time a node is added it will be added (+1) to each of the target groups/ports.  As we are targeting the Instance, there is a target in each group for every node.  If we now exceed the quota for targets, so long as we have added the nodes one-by-one we would have roughly an even number of targets in each target group and 'things would look ok'.

Now if we have exceed the quota x4 (easily done when you have a target group multiplier) and we then start to replace nodes (ASG image replacement for example), the aws-loadbalancer-controller will fill up the first target group first, then the second, then the third, and so on until the quota is exceeded.  If the quota is exceeded when filling up the second target group for example, the third and the fourth will be starved of targets and no traffic will be served for those ports.

For example, a cluster with 250 nodes and an NLB with 4 ports will end up with 250 targets in TG1, 250 in TG2, 0 in TG3 and 0 TG4 if the quota is 500 targets.

A better/fairer way would be to add a new target across all target groups before adding the next target.  This would result in a fairer allocation of the available target quota and prevent potential outages.  The above example would then look like:

125 in TG1, 125 in TG2, 125 in TG3 and 125 TG4.

**Steps to reproduce**

Deploy an NLB with multiple ports/target groups, for example 5.  Scale the cluster nodes up when targeting Instances from the LoadBalancer then replace each node in the ASG.  The LB controller will allocate to the first target-groups first until the quota is exceeded and then starve the rest.

**Expected outcome**

Targets should be added evenly across target-groups so that even when the quota is exceeded there are an equal number of nodes in each target group.

**Environment**

* AWS Load Balancer controller version: 1.6.0 
* Kubernetes version: 1.24
* Using EKS (yes/no), if so version? yes 1.24

**Additional Context**:

yykkibbb:
Hi @cunningr,

I'd like to work on this issue. The problem you've described makes sense - when targets are registered sequentially per target group rather than distributed evenly, a quota limit can cause some target groups to be completely starved.

My initial approach would be:
1. Identify where targets are registered to target groups in the codebase
2. Modify the registration logic to use a round-robin or interleaved approach across all target groups
[...]

PR #4604, opened 2026-02-25
Add RegisterTargetsInterleaved for fair target allocation across target groups

yykkibbb:
## Summary

This PR adds a new `RegisterTargetsInterleaved` method to `TargetsManager` that registers targets to multiple target groups in an interleaved manner, ensuring fair distribution when AWS quota limits are reached.

**Problem (Issue #4025):**
When replacing nodes in a cluster with an NLB serving multiple ports, the controller registers all targets to TG1 first, then TG2, etc. If the quota is exceeded midway, later target groups receive zero targets, causing service outages for those ports.

Example with 250 nodes, 4 ports, quota=500:
- **Current behavior:** TG1=250, TG2=250, TG3=0, TG4=0 (ports 3,4 have no traffic)
- **With interleaved:** TG1=125, TG2=125, TG3=125, TG4=125 (all ports work)

## Changes

- Added `TargetGroupTargets` struct to represent a TGB and its targets
- Added `RegisterTargetsInterleaved` method that:
  - Registers targets in chunks across all target groups before moving to the next chunk
  - Continues with other target groups even if one fails (fair distribution)
  - Falls back to standard registration for single target group
- Added comprehensive tests including:
  - 4-port NLB scenario (exact issue scenario)
  - Partial failure handling
  - Different target counts per TG

## Note

This PR adds the foundational API. The method is not yet called from the reconciler. Integration with `resource_manager.go` to actually use this method for multi-TGB scenarios would be the next step. I'd appreciate feedback on the preferred approach for that integration.

## Test plan

- [x] Unit tests for `RegisterTargetsInterleaved` pass
- [x] All existing tests pass

Related to: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4025

### DIFF
--- a/pkg/targetgroupbinding/targets_manager.go
+++ b/pkg/targetgroupbinding/targets_manager.go
@@ -22,10 +22,22 @@ const (
 	defaultNodeAZCacheTTL             = 60 * time.Minute
 )
 
+// TargetGroupTargets represents a target group binding and its targets to register.
+type TargetGroupTargets struct {
+	TGB     *elbv2api.TargetGroupBinding
+	Targets []elbv2types.TargetDescription
+}
+
 // TargetsManager is an abstraction around ELBV2's targets API.
 type TargetsManager interface {
 	// Register Targets into TargetGroup.
 	RegisterTargets(ctx context.Context, tgb *elbv2api.TargetGroupBinding, targets []elbv2types.TargetDescription) error
+
+	// RegisterTargetsInterleaved registers targets to multiple target groups in an interleaved manner.
+	// This ensures fair distribution of targets across target groups when quota limits are reached.
+	// Instead of registering all targets to TG1, then TG2, etc., it registers one target to each TG
+	// before moving to the next target, resulting in even distribution.
+	RegisterTargetsInterleaved(ctx context.Context, tgbTargets []TargetGroupTargets) error
 
 	// Deregister Targets from TargetGroup.
 	DeregisterTargets(ctx context.Context, tgb *elbv2api.TargetGroupBinding, targets []elbv2types.TargetDescription) error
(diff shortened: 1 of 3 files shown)


zac-nixon:
Thanks for opening this. The algorithm makes sense to me. I think there is some more discussion needed here, if we want to accept this.

1/ TargetGroupBindings only know of their TargetGroup.
2/ How do I handle addition of another target group? Do I un-register some targets? That is not clear to me.

yykkibbb:
Hi @zac-nixon, thank you for taking the time to review and for the insightful questions. They helped me reconsider  some aspects of the design that I hadn't fully thought through.

After going through your feedback and revisiting the codebase (especially the existing maxTargetsPerTargetGroup  mechanism in resource_manager.go and the IndexKeyServiceRefName index), I was thinking about a different approach that might fit the architecture better: sibling-aware self-limiting within the existing single-TGB reconciler.

The rough idea would be:
  - During reconcileWithInstanceTargetType, count sibling TGBs for the same service using the existing IndexKeyServiceRefName index
  - Compute a fair-share limit per TG based on a configured total quota and the sibling count
  - Each TGB independently limits its own registration — no cross-TGB coordination needed

This would build on the existing maxTargetsPerTargetGroup mechanism rather than introducing a new registration path, and it stays within the current architecture where each TGB reconciles independently.

For the "new TG added" scenario: when numSiblings increases, each TGB's fair share would automatically decrease on the next reconcile cycle. Existing targets wouldn't be un-registered, but new registrations would respect the reduced limit, and natural node turnover (e.g., ASG replacements) would gradually ebalance the distribution over time.

Of course, this is just an initial thought — I'd love to hear if this direction makes sense or if there are concerns I'm still  missing. Happy to revise the PR or take a completely different approach if needed.

Thanks again for the guidance!

PR #4606, opened 2026-02-26
feat: Support for deploying LB to multiple regions using `region` property on `LoadBalancerConfiguration`

aleksanderaleksic:
### Description

**Motivation:** Support deploying edge ALBs in AWS regions closer to end users, enabling TLS termination at the edge to reduce latency. For example, a cluster running in `us-east-1` can now deploy an ALB in `ap-northeast-1` via the Gateway API, terminating TLS in Tokyo rather than routing encrypted traffic back to the US.

#### What this PR does

Adds multi-region support to the AWS Load Balancer Controller's Gateway API implementation. A new optional `region` field on `LoadBalancerConfiguration` allows users to specify an AWS region for the ALB/NLB, and the controller handles all cross-region resource management automatically.

**Key changes:**

1. **`LoadBalancerConfiguration` API** (`apis/gateway/v1beta1/loadbalancerconfig_types.go`)
   - Added `Region` and `VpcID` fields to `LoadBalancerConfigurationSpec`, allowing users to target a specific AWS region and VPC for the load balancer.

2. **`ReconcileContext` / Cloud Provider** (`pkg/gateway/cloud_provider.go`)
   - New `ReconcileContext` encapsulates region-specific AWS clients (ELBV2, EC2, ACM) and resolvers (subnets, VPC info, security groups, cert discovery, tagging manager, TG ARN mapper).
   - `ReconcileContext` is always passed to `Build()` — for the default region it wraps the existing clients/resolvers; for non-default regions all fields are region-scoped.
   - `IsCrossRegion() bool` on the context lets downstream code branch on cross-region without comparing VPC IDs or regions directly.
   - `GetReconcileContext` auto-discovers the VPC in the target region (via `vpcId`, `vpcSelector`, or first subnet) and caches the result.

3. **Gateway Controller** (`controllers/gateway/gateway_controller.go`)
   - Always obtains a `ReconcileContext` from `CloudProvider` and passes it through to the model builder and stack deployer.
   - Uses `reconcileContext.IsCrossRegion()` to select the region-scoped stack deployer and backend SG provider.

4. **Model Builder** (`pkg/gateway/model/base_model_builder.go`)
   - `Build()` takes a required `ReconcileContextInterface` parameter (no variadic/optional). All effective clients and resolvers come from the RC unconditionally.
   - `ReconcileContextInterface` exposes `IsCrossRegion()` — used to disable backend SG allocation and clear the backend SG token on `TargetGroupBinding` resources for cross-region gateways (cross-region SG references are not supported by AWS).

5. **Region-aware AWS clients** (`pkg/aws/region.go`)
   - `NewCloudForRegion`, `NewEC2ClientForRegion`, `NewELBV2ForRegion` — factory functions for creating region-scoped AWS service clients without EC2 metadata.
   - `regionStubCloud` — minimal `Cloud` implementation for ELBV2-only operations (webhooks, ARN validation).

6. **TGB Webhooks** (`webhooks/elbv2/targetgroup_helper.go`, `targetgroupbinding_mutator.go`, `targetgroupbinding_validator.go`)
   - `ELBV2ClientProvider` type and `resolveELBV2ForTGB` helper extract the region from a target group ARN and return the correct regional ELBV2 client.
   - Mutator and validator use the regional client for `DescribeTargetGroups` calls, preventing `ValidationError` when validating cross-region TG ARNs.

7. **TGB Reconciler** (`pkg/targetgroupbinding/targets_manager.go`, `resource_manager.go`)
   - `cachedTargetsManager.resolveELBV2()` dynamically selects the correct ELBV2 client based on target group ARN region for `RegisterTargets`, `DeregisterTargets`, and `DescribeTargetHealth`.
   - `generateOverrideAzFn` now handles cross-region VPCs (not just cross-account) — when the TGB's VPC differs from the controller's VPC and `DescribeVpcs` returns `InvalidVpcID.NotFound`, target AZs are overridden to "all".

8. **Security Group Synthesizer** (`pkg/deploy/ec2/security_group_synthesizer.go`)
   - Handles cross-region SG cleanup gracefully.

9. **Documentation** (`docs/guide/gateway/loadbalancerconfig.md`, `docs/guide/gateway/spec.md`)
   - Documents the new `region` and `vpcID` fields with usage examples.

10. **Tests**
    - `webhooks/elbv2/targetgroup_helper_test.go` — `regionFromTGARN` and `resolveELBV2ForTGB` (18 cases)
    - `pkg/targetgroupbinding/targets_manager_resolve_test.go` — `cachedTargetsManager.resolveELBV2` (7 cases)
    - `pkg/aws/region_test.go` — `regionStubCloud` interface compliance and getters (12 cases)
    - `pkg/gateway/cloud_provider_test.go` — `ReconcileContext` getters, `IsCrossRegion()`, `GetReconcileContext` default path, `resolveVPCForRegion`, `resolveVPCFromSelector`, `resolveVPCFromFirstSubnet` (21 cases)
    - Updated `pkg/targetgroupbinding/resource_manager_test.go` — cross-region VPC override AZ test case
    - Updated `pkg/gateway/model/model_build_target_group_binding_network_test.go` — nil backend SG token (cross-region) case

#### Example usage

```yaml
metadata:
  name: edge-tokyo
spec:
  region: ap-northeast-1
  scheme: internet-facing
  # vpcID is auto-discovered from the target region
```

### Checklist
- [x] Manually tested

### BONUS POINTS checklist
- [x] Backfilled missing tests for code in same general area
- [x] Refactored something and made the world a better place

### DIFF
--- a/apis/gateway/v1beta1/loadbalancerconfig_types.go
+++ b/apis/gateway/v1beta1/loadbalancerconfig_types.go
@@ -198,6 +198,20 @@ type LoadBalancerConfigurationSpec struct {
 	// +optional
 	MergingMode *LoadBalancerConfigMergeMode `json:"mergingMode,omitempty"`
 
+	// region is the AWS region where the load balancer will be deployed. When unset, the controller's default region is used.
+	// When set to a different region, vpcId, vpcSelector, or loadBalancerSubnets with identifiers must be set so the VPC can be resolved.
+	// +optional
+	Region *string `json:"region,omitempty"`
+
+	// vpcId is the VPC ID in the target region. Used when region is set (and especially when it differs from the controller default).
+	// +optional
+	VpcID *string `json:"vpcId,omitempty"`
+
+	// vpcSelector selects the VPC in the target region by tags. Each key is a tag name; the value list is the allowed tag values.
+	// A VPC matches if it has each tag key with one of the corresponding values. Exactly one VPC must match in the target region.
+	// +optional
+	VpcSelector *map[string][]string `json:"vpcSelector,omitempty"`
+
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=32
 	// loadBalancerName defines the name of the LB to provision. If unspecified, it will be automatically generated.
--- a/apis/gateway/v1beta1/zz_generated.deepcopy.go
+++ b/apis/gateway/v1beta1/zz_generated.deepcopy.go
@@ -655,6 +655,36 @@ func (in *LoadBalancerConfigurationSpec) DeepCopyInto(out *LoadBalancerConfigura
 		*out = new(LoadBalancerConfigMergeMode)
 		**out = **in
 	}
+	if in.Region != nil {
+		in, out := &in.Region, &out.Region
+		*out = new(string)
+		**out = **in
+	}
+	if in.VpcID != nil {
+		in, out := &in.VpcID, &out.VpcID
+		*out = new(string)
+		**out = **in
+	}
+	if in.VpcSelector != nil {
+		in, out := &in.VpcSelector, &out.VpcSelector
+		*out = new(map[string][]string)
+		if **in != nil {
+			in, out := *in, *out
+			*out = make(map[string][]string, len(*in))
+			for key, val := range *in {
+				var outVal []string
+				if val == nil {
+					(*out)[key] = nil
+				} else {
+					inVal := (*in)[key]
+					in, out := &inVal, &outVal
+					*out = make([]string, len(*in))
+					copy(*out, *in)
+				}
+				(*out)[key] = outVal
+			}
+		}
+	}
 	if in.LoadBalancerName != nil {
 		in, out := &in.LoadBalancerName, &out.LoadBalancerName
 		*out = new(string)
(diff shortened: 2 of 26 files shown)


zac-nixon:
Hi. I wish you would have reached out to discuss this, prior to writing all this code :(. We are unlikely to accept this contribution as AWS offers CDN capabilities already (like CloudFront). What is your use-case?

aleksanderaleksic:
Hi @zac-nixon. 
Sorry for creating this PR without much context. 

To be honest I did not realize that we could use Cloudfront to get similar results so thanks for enlightening me about that.

We have our workloads based in `us-east-1` and we have some customers in ASIA and want to make sure the TLS handshake does not have to go to `us-east-1` every time, so we want to terminate TLS in `ap-northeast-1`. 
Cloudfront would solve that for us but we also have some requirements around Anycast Static IPs, and while Cloudfront supports that we dont want to pay $3k per month for that feature. So instead we have a global accelerator for this.

Also we dont want to deploy a EKS cluster in ASIA as of now, so we accept the latency between the edge ALBs and the instances in `us-east-1`. In fact our benchmarks shows that we have cut the overall request latency by 50% and that the TLS handshake is now 68x faster than before, so for us this is a huge win.

I hope this gives you a better understanding of why we want to de able to manage ALBs from a different region.

Given that the aws-load-balancer-controller does not support multiple deployments in a single cluster and its bound to a single region, we where not able to achieve the same results as with the suggested approach in any other way.

Issue #4589, opened 2026-02-17
[Feature] Want to set default targetgroupconfiguration for gateway


regis-underdog:
**Describe the feature you are requesting**
We would like to specify a _default_ `targetGroupConfiguration` that is applied globally to every target group that is served by a `gateway`. (Or perhaps a `gatewayClass` if you prefer). In our use case, for example, we have many services. Each service is in a different namespace and requires different `targetGroupConfiguration`s per service and per namespace. We would like to administer a single default in the admin namespace that applies to all target groups joining a gateway. Any local overrides per service would take precedence. For, example, we have copied an example from [the docs](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/gateway/customization/#customizing-services-target-groups-using-targetgroupconfiguration-crd) but wish to make it more generic so that any service with an `httpRoute` or `grpcRoute` would not require a new `targetGroupConfiguration` object.

**Motivation**
We wish to accomplish several objectives:
1. Remove the need to deploy `targetGroupConfiguration` per service and per namespace
2. Set sane, common defaults that we would like to manage for all target groups on this `gateway` (or `gatewayClass`, your choice)
3. Manage settings in a single location and let consumers override their own settings as needed
4. Allow existing helm charts and third party `httpRoute`/`grpcRoute` entities to function without forcing them to apply a custom resource they are not aware of

**Describe the proposed solution you'd like**
```yaml
apiVersion: gateway.k8s.aws/v1beta1
kind: TargetGroupConfiguration
metadata:
  name: alb-gateway-internal
  namespace: alb-gateway-system
spec:
  targetReference:
    group: gateway.networking.k8s.io
    kind: Gateway
    name: alb-gateway-internal
  defaultConfiguration:
    healthCheckConfig:
      healthCheckInterval: 5
      healthyThresholdCount: 2
    targetGroupAttributes:
      - key: deregistration_delay.timeout_seconds
        value: "31"
    targetType: ip
```
Notice the `targetReference` is the `gateway` itself to apply a default configuration for any new `targetGroups` that want to join.

If you look above, we have ideally set the `targetReference` to point toward the gateway itself. Alternatively, you could set this on the `loadbalancerCongifuration`, or on the `gateway` object, or the `gatewayClass` itself.

**Describe alternatives you've considered**
The existing use case is cumbersome and requires explicit controls for every service and namespace.

**Contribution Intention (Optional)**
[...]

PR #4609, opened 2026-02-27
[feat gw-api]gateway level targetgroupconfiguration

shuqz:
### Issue
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4589

### Description
Added support for gateway level tgc definition. Priority order: 
- Service-level TGC exists → use its defaultConfiguration + its routeConfigurations (gateway TGC ignored entirely)
- No service-level TGC, gateway-level TGC exists → use gateway TGC's defaultConfiguration + its routeConfigurations
- Neither exists → controller defaults

Changes summary
- add webhook to verify only one gateway level TGC allowed in same namespace, error will occur when apply 
```
Error from server (Forbidden): error when creating "gateway-level-tgc.yaml": admission webhook "vtargetgroupconfiguration.gateway.k8s.aws" denied the request: TargetGroupConfiguration referencing Gateway "my-gateway" already exists in namespace "gw-tgc-test" (gateway-defaults), only one is allowed per Gateway
```
- add `setGatewayDefaultTGConfig` so it will only need to be loaded once per reconcile. i thought about passes gatewayNamespacedName down the chain, and serviceLoader does a k8s List call every time it needs the fallback. With N service backends and none having their own TGC, that's N identical List calls per reconcile. so ended up not choosing this method 
- modify documentation 
- added test case to e2e test

### Checklist
```
ginkgo -v -r test/e2e/gateway -- --kubeconfig=$KUBECONFIG --cluster-name=awslbc-xx --aws-region=us-west-2 --aws-vpc-id=vpc-xxx --enable-gateway-tests -ginkgo.focus="Gateway-level default TGC"

Ran 8 of 49 Specs in 2272.136 seconds
SUCCESS! -- 8 Passed | 0 Failed | 0 Pending | 41 Skipped
PASS

Ginkgo ran 1 suite in 38m25.478929708s
Test Suite Passed
```
- [x] Manually tested
```
##############################################
# 1. Namespace
##############################################
apiVersion: v1
kind: Namespace
metadata:
  name: gw-tgc-test
---
##############################################
# 2. GatewayClass (ALB)
##############################################
kind: GatewayClass
metadata:
  name: gw-tgc-test-class
spec:
---
##############################################
# 3. Gateway
##############################################
kind: Gateway
metadata:
  name: my-gateway
  namespace: gw-tgc-test
spec:
  gatewayClassName: gw-tgc-test-class
  listeners:
    - name: http
      port: 80
      protocol: HTTP
---
##############################################
# 4. LoadBalancerConfiguration
##############################################
metadata:
  name: my-gateway
  namespace: gw-tgc-test
spec:
  scheme: internet-facing
---
##############################################
# 5. Gateway-level TargetGroupConfiguration
#    (targets the Gateway, not a Service)
##############################################
metadata:
  name: gateway-defaults
  namespace: gw-tgc-test
spec:
  targetReference:
    group: gateway.networking.k8s.io
    kind: Gateway
    name: my-gateway
  defaultConfiguration:
    targetType: ip
    healthCheckConfig:
      healthCheckProtocol: HTTP
      healthCheckPath: /gw-healthcheck
      healthCheckInterval: 15
      healthyThresholdCount: 2
      unhealthyThresholdCount: 3
---
metadata:
  name: gateway-defaults-should-fail
  namespace: gw-tgc-test
spec:
  targetReference:
    group: gateway.networking.k8s.io
    kind: Gateway
    name: my-gateway
  defaultConfiguration:
    targetType: ip
    healthCheckConfig:
      healthCheckProtocol: HTTP
      healthCheckPath: /gw-healthcheck
      healthCheckInterval: 10
      healthyThresholdCount: 10
      unhealthyThresholdCount: 10
---
##############################################
# 6. Deployment — svc1 (no service-level TGC,
#    should inherit from gateway TGC)
##############################################
apiVersion: apps/v1
kind: Deployment
metadata:
  name: svc1
  namespace: gw-tgc-test
spec:
  replicas: 2
  selector:
    matchLabels:
      app: svc1
  template:
    metadata:
      labels:
        app: svc1
    spec:
      containers:
        - name: app
          image: public.ecr.aws/nginx/nginx:latest
          ports:
            - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: svc1
  namespace: gw-tgc-test
spec:
  selector:
    app: svc1
  ports:
    - port: 80
      targetPort: 80
---
##############################################
# 7. Deployment — svc2 (has its own service-level
#    TGC that overrides the gateway TGC)
##############################################
apiVersion: apps/v1
kind: Deployment
metadata:
  name: svc2
  namespace: gw-tgc-test
spec:
  replicas: 2
  selector:
    matchLabels:
      app: svc2
  template:
    metadata:
      labels:
        app: svc2
    spec:
      containers:
        - name: app
          image: public.ecr.aws/nginx/nginx:latest
          ports:
            - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: svc2
  namespace: gw-tgc-test
spec:
  selector:
    app: svc2
  ports:
    - port: 80
      targetPort: 80
---
##############################################
# 8. Service-level TGC for svc2 only
#    (overrides gateway TGC for this service)
##############################################
metadata:
  name: svc2-tgc
  namespace: gw-tgc-test
spec:
  targetReference:
    kind: Service
    name: svc2
  defaultConfiguration:
    targetType: ip
    healthCheckConfig:
      healthCheckProtocol: HTTP
      healthCheckPath: /svc-healthcheck
      healthCheckInterval: 10
      healthyThresholdCount: 3
      unhealthyThresholdCount: 2
---
##############################################
# 9. HTTPRoute — routes traffic to both services
##############################################
kind: HTTPRoute
metadata:
  name: test-route
  namespace: gw-tgc-test
spec:
  parentRefs:
    - name: my-gateway
      sectionName: http
  rules:
    - backendRefs:
        - name: svc1
          port: 80
        - name: svc2
          port: 80

```

### DIFF
--- a/docs/guide/gateway/targetgroupconfig.md
+++ b/docs/guide/gateway/targetgroupconfig.md
@@ -18,10 +18,55 @@ spec:
 Defines the Kubernetes object to attach the Target Group settings to.
 
 - **group**: The group of the referent. For example, "gateway.networking.k8s.io". When unspecified or empty string, core API group is inferred.
-- **kind**: The Kubernetes resource kind of the referent. For example "Service". Defaults to "Service" when not specified.
+- **kind**: The Kubernetes resource kind of the referent. For example "Service". Defaults to "Service" when not specified. Supported values: `Service`, `Gateway`.
 - **name**: The name of the referent.
 
 **Default** No default, required field
+
+## Gateway-Level Default TargetGroupConfiguration
+
+You can create a TargetGroupConfiguration that references a Gateway instead of a Service. This acts as a default configuration for all Service backends attached to that Gateway via HTTPRoute, GRPCRoute, TCPRoute, UDPRoute, or TLSRoute.
+
+This is useful when you want to set common target group defaults (such as `targetType: ip`, health check settings, or deregistration delay) for all services on a Gateway without requiring each service to have its own TargetGroupConfiguration.
+
+### Example
+
+```yaml
+apiVersion: gateway.k8s.aws/v1beta1
+kind: TargetGroupConfiguration
+metadata:
+  name: gateway-defaults
+  namespace: gateway-system   # must be in the same namespace as the Gateway
+spec:
+  targetReference:
+    group: gateway.networking.k8s.io
+    kind: Gateway
+    name: my-alb-gateway
+  defaultConfiguration:
+    targetType: ip
+    healthCheckConfig:
+      healthCheckInterval: 5
+      healthyThresholdCount: 2
+    targetGroupAttributes:
+      - key: deregistration_delay.timeout_seconds
+        value: "31"
+```
+
+### Resolution Priority
+
+When the controller builds a target group for a Service backend, it resolves configuration in this order:
+
+1. **Service-level TargetGroupConfiguration** — A TGC in the Service's namespace with `targetReference.kind: Service` (or unset) and `targetReference.name` matching the Service name. If found, this is used exclusively.
+2. **Gateway-level TargetGroupConfiguration** — A TGC in the Gateway's namespace with `targetReference.kind: Gateway` and `targetReference.name` matching the Gateway name. Used as a fallback when no Service-level TGC exists.
+3. **Controller defaults** — Hardcoded defaults (e.g., `targetType: instance`) and the `--default-target-type` controller flag.
+
+There is no cross-TGC field-level merging. If a Service has its own TGC, the Gateway-level TGC is not consulted at all.
+
+### Constraints
+
+- Only one Gateway-scoped TargetGroupConfiguration is allowed per Gateway in the same namespace.
+- The `routeConfigurations` field works the same way as for Service-level TGCs — you can use it to provide route-specific overrides within the Gateway-level TGC.
+
 
 ### DefaultConfiguration
 
--- a/pkg/gateway/routeutils/backend_test.go
+++ b/pkg/gateway/routeutils/backend_test.go
@@ -374,7 +374,7 @@ func TestCommonBackendLoader_Service(t *testing.T) {
 				assert.NoError(t, err, fmt.Sprintf("%+v", g))
 			}
 
-			result, warningErr, fatalErr := commonBackendLoader(context.Background(), k8sClient, tc.backendRef, tc.routeIdentifier, kind)
+			result, warningErr, fatalErr := commonBackendLoader(context.Background(), k8sClient, tc.backendRef, tc.routeIdentifier, kind, (*elbv2gw.TargetGroupConfiguration)(nil))
 
 			if tc.expectWarning {
 				assert.Error(t, warningErr)
@@ -441,7 +441,7 @@ func TestCommonBackendLoader_TargetGroupName(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			k8sClient := testutils.GenerateTestClient()
 
-			result, warningErr, fatalErr := commonBackendLoader(context.Background(), k8sClient, tc.backendRef, tc.routeIdentifier, HTTPRouteKind)
+			result, warningErr, fatalErr := commonBackendLoader(context.Background(), k8sClient, tc.backendRef, tc.routeIdentifier, HTTPRouteKind, (*elbv2gw.TargetGroupConfiguration)(nil))
 
 			if tc.expectWarning {
 				assert.Error(t, warningErr)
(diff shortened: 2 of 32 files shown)


docs/guide/gateway/targetgroupconfig.md
```diff
@@ -18,11 +18,56 @@ spec:
...
+
+### Resolution Priority
+
+When the controller builds a target group for a Service backend, it resolves configuration in this order:
+
+1. **Service-level TargetGroupConfiguration** — A TGC in the Service's namespace with `targetReference.kind: Service` (or unset) and `targetReference.name` matching the Service name. If found, this is used exclusively.
+2. **Gateway-level TargetGroupConfiguration** — A TGC in the Gateway's namespace with `targetReference.kind: Gateway` and `targetReference.name` matching the Gateway name. Used as a fallback when no Service-level TGC exists.
+3. **Controller defaults** — Hardcoded defaults (e.g., `targetType: instance`) and the `--default-target-type` controller flag.
+
+There is no cross-TGC field-level merging. If a Service has its own TGC, the Gateway-level TGC is not consulted at all.
```
shraddhabang, line 63: Why are we not supporting the merge? What if I want different granular customization (for example health check) for few of of my service but I want same target type for each service?

docs/guide/gateway/targetgroupconfig.md
```diff
@@ -18,11 +18,56 @@ spec:
 Defines the Kubernetes object to attach the Target Group settings to.
 
 - **group**: The group of the referent. For example, "gateway.networking.k8s.io". When unspecified or empty string, core API group is inferred.
-- **kind**: The Kubernetes resource kind of the referent. For example "Service". Defaults to "Service" when not specified.
+- **kind**: The Kubernetes resource kind of the referent. For example "Service". Defaults to "Service" when not specified. Supported values: `Service`, `Gateway`.
```
shraddhabang, line 21: Lets also update the How it works section. 
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/guide/gateway/customization.md
We will also need to update the diagram showing CRD mappings to gateway resources.

webhooks/elbv2/targetgroupconfig_validator.go
```diff
@@ -0,0 +1,90 @@
...
+	return v.checkGatewayTGCUniqueness(ctx, tgc)
+}
+
+func (v *targetGroupConfigurationValidator) ValidateDelete(_ context.Context, _ runtime.Object) error {
+	return nil
+}
+
+// checkGatewayTGCUniqueness ensures at most one TargetGroupConfiguration with
+// targetReference.kind=Gateway exists per Gateway in the same namespace.
+func (v *targetGroupConfigurationValidator) checkGatewayTGCUniqueness(ctx context.Context, tgc *elbv2gw.TargetGroupConfiguration) error {
```
shraddhabang, line 59: Why are we handling this through the webhook? Can we handle this in `LookUpTargetGroupConfiguration` instead to avoid re-listing. That way we can do the same uniqueness for ServiceKind as well?

docs/guide/gateway/targetgroupconfig.md
shuqz: yea you are right, i was thinking they still have the ability to define routeConfiguration in this gateway level tgc. so it would be something like 
```
spec:
  targetReference:
    group: gateway.networking.k8s.io
    kind: Gateway
    name: my-alb-gateway
  defaultConfiguration:
    targetType: ip
    healthCheckConfig:
      healthCheckPath: /healthz
      healthCheckInterval: 15
      healthyThresholdCount: 2
    targetGroupAttributes:
        value: "30"
  routeConfigurations:
    # api-service gets a custom health check path, everything else from defaultConfiguration
    - routeIdentifier:
        kind: HTTPRoute
        namespace: app-ns
        name: api-route
      targetGroupProps:
        healthCheckConfig:
          healthCheckPath: /api/health
    # admin-service gets a custom path + longer interval
    - routeIdentifier:
        kind: HTTPRoute
        namespace: app-ns
        name: admin-route
      targetGroupProps:
        healthCheckConfig:
          healthCheckPath: /admin/health
          healthCheckInterval: 30
    # web-service has no entry here — it just gets defaultConfiguration as-is
```
but now i think it has two downsides
1. if it is scenario that two services within same route needs to have difference config, then they need to specify a service-level tgc, and that needs to contain everything (defined in gateway-level tgc)
2. it is likely each service team wants to have a different health check config, and it does not make sense if they all need to modify this gateway level tgc

i will update behavior

webhooks/elbv2/targetgroupconfig_validator.go
shuqz: we can, both places works to catch this uniqueness. it is more about different user experience, putting it in webhook rejects faster during apply file time and they see the explicit message. for LookUpTargetGroupConfiguration, they need to check in logs

docs/guide/gateway/targetgroupconfig.md
zac-nixon: What if we added a field to the LB Config that accepted this value? That way we don't need to worry about a webhook to validate TGC configuration?

zac-nixon: The reason why we added the reference field to TGC is because Service is an API we have no influence over. As we control the LB config CRD, we have the flexibility to make this experience better for users.

pkg/gateway/routeutils/backend_test.go
```diff
@@ -374,7 +374,7 @@ func TestCommonBackendLoader_Service(t *testing.T) {
 				assert.NoError(t, err, fmt.Sprintf("%+v", g))
 			}
 
-			result, warningErr, fatalErr := commonBackendLoader(context.Background(), k8sClient, tc.backendRef, tc.routeIdentifier, kind)
```
zac-nixon, line 377: Im surprised that you have to cast nil?

pkg/gateway/routeutils/grpc.go
```diff
@@ -55,6 +55,11 @@ type grpcRouteDescription struct {
 	rules                     []RouteRule
 	ruleAccumulator           attachedRuleAccumulator[gwv1.GRPCRouteRule]
 	compatibleHostnamesByPort map[int32][]gwv1.Hostname
+	gatewayDefaultTGConfig    *elbv2gw.TargetGroupConfiguration
+}
+
+func (grpcRoute *grpcRouteDescription) setGatewayDefaultTGConfig(cfg *elbv2gw.TargetGroupConfiguration) {
```
zac-nixon, line 61: If this belongs to a Gateway, why are you setting this per route?

pkg/gateway/routeutils/backend_test.go
shuqz: nil is good, i will update it. missed it then i have tests auto generated.

shuqz:
This is old PR, not needed anymore. New one is merged: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4611

PR #4611, opened 2026-03-03
[feat gw-api]add gateway and gatewayclass level tgc support

shuqz:
### Issue
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4589

### Description
Added support for gateway level and gateway class level tgc definition by modifying LoadBalancerConfiguration
- for details of expected behavior, check updates on documentations pages
- One thing to call out is TargetReference is now optional. (it was required before, so it is backward compatible)

### Checklist
- [x] Manually tested and added E2E tests

**example log error when trying to use a tgc with targetReference as default tgc**

```
"error":"failed to resolve default TargetGroupConfiguration \"my-gateway-defaults\" referenced by Gateway LoadBalancerConfiguration \"my-lb-config\" in namespace \"echoserver\": TargetGroupConfiguration \"my-gateway-defaults\" in namespace \"echoserver\" has targetReference set and cannot be used as a defaultTargetGroupConfiguration — a default TGC must not have targetReference"} 
```

**Example test yaml**
```
# GatewayClass-Level Default Target Group Configuration
#
# Chain: GatewayClass -> LBC (via parametersRef) -> default TGC (via defaultTargetGroupConfiguration)
#
# mergingMode: prefer-gateway means Gateway-level LBC wins over GatewayClass-level LBC.
#
# Expected final configuration:
#
# team-a/my-service (no Gateway-level LBC, no service TGC — inherits everything from GatewayClass):
#   targetType: ip
#   healthCheckPath: /healthz
#   healthCheckInterval: 30
#   healthCheckProtocol: HTTP
#   targetGroupAttributes: [deregistration_delay.timeout_seconds=60]
#
# team-b/my-service (Gateway-level LBC overrides, no service TGC):
#   targetType: ip                                    (from team-b gateway default TGC — not set, so falls back to GatewayClass TGC)
#   healthCheckPath: /team-b/health                   (from team-b gateway default TGC)
#   healthCheckInterval: <AWS default>                (not set in team-b TGC, GatewayClass TGC NOT merged because healthCheckConfig is replaced wholesale)
#   healthCheckProtocol: <AWS default>                (same reason)
#   targetGroupAttributes: [deregistration_delay.timeout_seconds=60]  (inherited from GatewayClass TGC — team-b TGC didn't set any)

---
# 1. Admin's default TGC in kube-system
metadata:
  name: org-tg-defaults
  namespace: kube-system
spec:
  defaultConfiguration:
    targetType: ip
    healthCheckConfig:
      healthCheckPath: /healthz
      healthCheckInterval: 30
      healthCheckProtocol: HTTP
    targetGroupAttributes:
        value: "60"

---
# 2. GatewayClass LBC references the default TGC
metadata:
  name: org-lb-config
  namespace: kube-system
spec:
  mergingMode: prefer-gateway
  scheme: internal
  defaultTargetGroupConfiguration:
    name: org-tg-defaults

---
# 3. GatewayClass points to the LBC
kind: GatewayClass
metadata:
  name: gw-tgc-test-class
spec:
  parametersRef:
    group: gateway.k8s.aws
    name: org-lb-config
    namespace: kube-system

---
# ===== Team A: inherits everything from GatewayClass =====

apiVersion: v1
kind: Namespace
metadata:
  name: team-a

---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-service
  namespace: team-a
spec:
  replicas: 2
  selector:
    matchLabels:
      app: my-service
  template:
    metadata:
      labels:
        app: my-service
    spec:
      containers:
        - name: app
          image: public.ecr.aws/eks-distro/kubernetes/pause:3.7
          ports:
            - containerPort: 80

---
apiVersion: v1
kind: Service
metadata:
  name: my-service
  namespace: team-a
spec:
  selector:
    app: my-service
  ports:
    - port: 80
      targetPort: 80

---
kind: Gateway
metadata:
  name: team-a-gateway
  namespace: team-a
spec:
  gatewayClassName: gw-tgc-test-class
  listeners:
    - name: http
      port: 80
      protocol: HTTP

---
kind: HTTPRoute
metadata:
  name: team-a-route
  namespace: team-a
spec:
  parentRefs:
    - name: team-a-gateway
  rules:
    - backendRefs:
        - name: my-service
          port: 80

---
# ===== Team B: overrides at Gateway level =====

apiVersion: v1
kind: Namespace
metadata:
  name: team-c
---
apiVersion: v1
kind: Namespace
metadata:
  name: team-b

---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-service
  namespace: team-b
spec:
  replicas: 2
  selector:
    matchLabels:
      app: my-service
  template:
    metadata:
      labels:
        app: my-service
    spec:
      containers:
        - name: app
          image: public.ecr.aws/eks-distro/kubernetes/pause:3.7
          ports:
            - containerPort: 80

---
apiVersion: v1
kind: Service
metadata:
  name: my-service
  namespace: team-b
spec:
  selector:
    app: my-service
  ports:
    - port: 80
      targetPort: 80

---
# Team B's default TGC — overrides health check path
metadata:
  name: team-b-tg-defaults
  namespace: team-c
spec:
  defaultConfiguration:
    healthCheckConfig:
      healthCheckPath: /team-b/health

---
# Team B's LBC references their own default TGC
metadata:
  name: team-b-lb-config
  namespace: team-b
spec:
  scheme: internet-facing
  defaultTargetGroupConfiguration:
    name: team-b-tg-defaults

---
kind: Gateway
metadata:
  name: team-b-gateway
  namespace: team-b
spec:
  gatewayClassName: gw-tgc-test-class
  infrastructure:
    parametersRef:
      group: gateway.k8s.aws
      name: team-b-lb-config
  listeners:
    - name: http
      port: 80
      protocol: HTTP

---
kind: HTTPRoute
metadata:
  name: team-b-route
  namespace: team-b
spec:
  parentRefs:
    - name: team-b-gateway
  rules:
    - backendRefs:
        - name: my-service
          port: 80

```


controllers/gateway/eventhandlers/target_group_configuration_events.go
```diff
@@ -65,6 +66,9 @@ func (h *enqueueRequestsForTargetGroupConfigurationEvent) Generic(ctx context.Co
 }
 
 func (h *enqueueRequestsForTargetGroupConfigurationEvent) enqueueImpactedObject(ctx context.Context, tgconfig *elbv2gw.TargetGroupConfiguration, queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	if tgconfig.Spec.TargetReference == nil {
```
zac-nixon, line 74: It's not enough to return here -- you're dropping potential updates.

You'll need to check what (if any) LoadBalanacerConfigurations are using this TGC and enqueue an update. This is similar to how we handle routes.
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/controllers/gateway/eventhandlers/service_events.go#L89

controllers/gateway/config_resolver.go
```diff
@@ -66,29 +73,106 @@ func (resolver *gatewayConfigResolverImpl) getLoadBalancerConfigForGateway(ctx c
...
+		mergedLBConfig = resolver.configMergeFn(*gatewayClassLBConfig, *gatewayLBConfig)
+	}
+
+	return mergedLBConfig, resolvedDefaultTGC, nil
+}
+
+// resolveAndMergeDefaultTGCs resolves the default TGC from both the GatewayClass LBC and Gateway LBC,
+// then merges their props based on mergingMode. For the Gateway LBC, the TGC must be in the same namespace
+// as the Gateway; otherwise it is treated as nil.
+func (resolver *gatewayConfigResolverImpl) resolveAndMergeDefaultTGCs(ctx context.Context, k8sClient client.Client, gwClassLBC *elbv2gw.LoadBalancerConfiguration, gwLBC *elbv2gw.LoadBalancerConfiguration) *elbv2gw.TargetGroupConfiguration {
```
zac-nixon: We have established the pattern of doing this merge logic in pkg/gateway, infact there is already a TGC merger. Can you please refactor the code so that the logic lives there? Ideally the delniation is

pkg/gateway -> gateway api implementation stuff
controller/gateway -> handle k8s api interactions and updates.

docs/guide/gateway/customization.md
```diff
@@ -82,6 +82,25 @@ The `TargetGroupConfiguration` CRD enables granular customization of the AWS Tar
 
 For a comprehensive overview of configurable parameters, please refer to the  [TargetGroupConfiguration CRD documentation](./targetgroupconfig.md).
 
+`TargetGroupConfiguration` can be applied at three levels, listed from broadest to most specific:
+
+1. **GatewayClass-level defaults** — Org-wide defaults applied to all Gateways in a GatewayClass.
```
zac-nixon: Do we use org-wide elsewhere? Org-wide is pretty specific, I'd prefer to use broader language maybe
"Defaults that apply to all Gateways that belong to a GatewayClass"

docs/guide/gateway/loadbalancerconfig.md
```diff
@@ -769,4 +769,29 @@ See applicable [considerations](https://docs.aws.amazon.com/elasticloadbalancing
...
+kind: LoadBalancerConfiguration
+metadata:
+  name: example-config
+  namespace: echoserver
+spec:
+  defaultTargetGroupConfiguration:
+    name: my-default-tgc
+```
+
+References a TargetGroupConfiguration by name in the same namespace as this LoadBalancerConfiguration. The referenced TGC provides default target group properties for all Service backends attached to the Gateway. Service-level TGCs override these defaults on a per-field basis.
```
zac-nixon: This can _technically_ be a Gateway too (NLB -> ALB). Same comment for "Service-level TGCs"

docs/guide/gateway/loadbalancerconfig.md
```diff
@@ -769,4 +769,29 @@ See applicable [considerations](https://docs.aws.amazon.com/elasticloadbalancing
...
+
+The referenced TGC should omit `targetReference` — it serves as a standalone set of defaults only used when an LBC points to it.
```
zac-nixon: I suppose there is nothing stopping them from having a TGC in-use by an LBC and referencing a service?

docs/guide/gateway/loadbalancerconfig.md
```diff
@@ -769,4 +769,29 @@ See applicable [considerations](https://docs.aws.amazon.com/elasticloadbalancing
...
+
+When both a GatewayClass LBC and a Gateway LBC define a `defaultTargetGroupConfiguration`, the controller resolves both TGCs and merges their `defaultConfiguration` props field-by-field. The `mergingMode` on the GatewayClass LBC controls which one wins on overlapping fields — non-overlapping fields from both are always preserved. See [TargetGroupConfiguration — GatewayClass + Gateway Default TGC Merging](targetgroupconfig.md#gatewayclass--gateway-default-tgc-merging) for details.
+
+> **Important:** The referenced TGC must exist in the same namespace as the LoadBalancerConfiguration. Cross-namespace references are not supported — the `defaultTargetGroupConfiguration` field only accepts a `name`, not a `namespace`. If the referenced TGC is not found in the LBC's namespace, the controller logs an error and proceeds without defaults for that level.
```
shraddhabang: Why are we failing silently here? 
It will mess up the existing traffic if they accidentally delete the TGC referenced by LBC. I think we should fail here and not allow proceeding until they fix the error by fixing their TGC reference.

controllers/gateway/eventhandlers/target_group_configuration_events.go
```diff
@@ -65,6 +66,9 @@ func (h *enqueueRequestsForTargetGroupConfigurationEvent) Generic(ctx context.Co
 }
 
 func (h *enqueueRequestsForTargetGroupConfigurationEvent) enqueueImpactedObject(ctx context.Context, tgconfig *elbv2gw.TargetGroupConfiguration, queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
```
shraddhabang, line 73: We should also be enqueuing the impacted LBCs referencing this TGC now to propagate the changes to gateway level tgc

controllers/gateway/targetgroup_configuration_controller.go
```diff
@@ -111,6 +112,12 @@ func (r *targetgroupConfigurationReconciler) handleDelete(tgConf *elbv2gw.Target
 		}
 	}
 
+	if tgConf.Spec.TargetReference == nil {
+		// TGC without targetReference is a default TGC (used via LBC reference).
+		// It doesn't reference any service, so always eligible for finalizer removal.
+		return r.finalizerManager.RemoveFinalizers(context.Background(), tgConf, shared_constants.TargetGroupConfigurationFinalizer)
```
shraddhabang, line 126: Do we need to check if its being referenced by LBC before we remove the finalizers?

controllers/gateway/eventhandlers/target_group_configuration_events.go
shuqz: good catch, i missed it

controllers/gateway/targetgroup_configuration_controller.go
shuqz: yes i will add it

docs/guide/gateway/loadbalancerconfig.md
shuqz: right, it is valid. i will update wording here

docs/guide/gateway/loadbalancerconfig.md
shuqz: yea i thought about this too. but in the end, i think the default TGC is optional not required. and backends can have their own TGCs. A GatewayClass LBC with a default TGC may serve many gateways across namespaces. Failing hard would block all of them over a missing optional config. and plus, the finalizer should prevents accidental deletion if TGC is referred by lbc. that's why i ended up with soft-fail with a log. and it is not a one way door decision, we can adjust it later.

controllers/gateway/eventhandlers/target_group_configuration_events.go
```diff
@@ -98,7 +106,73 @@ func (h *enqueueRequestsForTargetGroupConfigurationEvent) enqueueImpactedObject(
...
+	gwControllerSet := sets.New(h.gwController)
+
+	for i := range lbConfigList.Items {
+		lbConfig := &lbConfigList.Items[i]
+		if lbConfig.Spec.DefaultTargetGroupConfiguration == nil || lbConfig.Spec.DefaultTargetGroupConfiguration.Name != tgconfig.Name {
+			continue
+		}
 
+		// Check Gateways that directly reference this LBC
+		gateways, err := gatewayutils.GetImpactedGatewaysFromLbConfig(ctx, h.k8sClient, lbConfig, h.gwController)
```
shraddhabang: Instead of enqueuing gateways and gwclasses from here, lets just enqueue LBcfgs that way lbcfg evet handlers will enqueue the respective impacted resources. We dont need to duplicate this code.

docs/guide/gateway/loadbalancerconfig.md
shraddhabang: The finalizers will help with the accidental deletions but this will not catch the cases where the default TGC is not even created or present but referencing other backend(customer error). The customers will be expecting the TGC cfg on the backends which does not have any tgc applied. But those will still end up on the controller defaults which can lead to confusions. If they are referencing something from LBC we need to make sure that its is valid. What do you think? We wont be able to enforce this later because we will be breaking customer who dont have valid TGC in their LBCfg. They will be required to either remove the reference or create the defaultTGC.

controllers/gateway/eventhandlers/target_group_configuration_events.go
shuqz: that's my first think and i tried this approach too. lbConfigEventChan only checks impacted gateway, it does not check gatewayclass. so i would need to wiring 2 new channels (lbConfigEventChan + gwClassEventChan) through the gateway controller setup, creates new handler instances in both ALB and NLB setup functions. the whole implementation is actually more complex than current one.

zac-nixon: You could probably could refactor the existing code so that this event handler:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/controllers/gateway/eventhandlers/gatewayclass/load_balancer_configuration_events.go

Is shared among Gateway / GatewayClass.

controllers/gateway/eventhandlers/target_group_configuration_events.go
```diff
@@ -98,7 +106,73 @@ func (h *enqueueRequestsForTargetGroupConfigurationEvent) enqueueImpactedObject(
...
+		for _, gw := range gateways {
+			h.logger.V(1).Info("enqueue gateway for default targetgroupconfiguration event",
+				"targetgroupconfiguration", k8s.NamespacedName(tgconfig),
+				"loadbalancerconfiguration", k8s.NamespacedName(lbConfig),
+				"gateway", k8s.NamespacedName(gw))
+			queue.Add(reconcile.Request{NamespacedName: k8s.NamespacedName(gw)})
+		}
+
+		// Check GatewayClasses that reference this LBC, then find their Gateways
+		gwClasses, err := gatewayutils.GetImpactedGatewayClassesFromLbConfig(ctx, h.k8sClient, lbConfig, gwControllerSet)
```
zac-nixon: You could potentially double enqueue a gateway that has the tgc and the a gateway that has a gwc that references the same tgc (weird case)

controllers/gateway/targetgroup_configuration_controller.go
```diff
@@ -111,6 +112,19 @@ func (r *targetgroupConfigurationReconciler) handleDelete(tgConf *elbv2gw.Target
...
 
+	if tgConf.Spec.TargetReference == nil {
+		// TGC without targetReference is a default TGC (used via LBC reference).
+		// Check if any LoadBalancerConfiguration still references this TGC before allowing deletion.
+		inUse, err := r.isDefaultTGCInUse(context.Background(), tgConf)
+		if err != nil {
+			return err
+		}
+		if inUse {
+			return fmt.Errorf("default targetgroup configuration [%+v] is still in use", k8s.NamespacedName(tgConf))
```
zac-nixon: Would be helpful to tell users what is using this tgc.

docs/guide/gateway/customization.md
```diff
@@ -82,6 +82,25 @@ The `TargetGroupConfiguration` CRD enables granular customization of the AWS Tar
 
 For a comprehensive overview of configurable parameters, please refer to the  [TargetGroupConfiguration CRD documentation](./targetgroupconfig.md).
 
+`TargetGroupConfiguration` can be applied at three levels, listed from broadest to most specific:
+
+1. **GatewayClass-level defaults** — Defaults that apply to all Gateways that belong to a GatewayClass.
+2. **Gateway-level defaults** — Per-Gateway defaults applied to all Service backends behind that Gateway.
```
zac-nixon, line 88: Will we support per-route configuration still?

docs/guide/gateway/targetgroupconfig.md
```diff
@@ -18,10 +18,151 @@ spec:
...
+  targetReference:
+    name: service-a
+  defaultConfiguration:
+    healthCheckConfig:
+      healthCheckPath: /service-a/health
+```
+
+Effective configuration for `service-a`:
+- `targetType: ip` — inherited from Gateway default TGC
+- `healthCheckConfig: { healthCheckPath: /service-a/health }` — from Service TGC (replaces entire health check block)
```
zac-nixon, line 127: This is a bit confusing IMO, if an org is dedicating that everyone use a specific health check config, it doesn't make sense to allow to an individual team to override it?

zac-nixon: These examples are a very nice addition tho!

pkg/gateway/routeutils/http.go
```diff
@@ -61,7 +62,7 @@ func (httpRoute *httpRouteDescription) GetAttachedRules() []RouteRule {
 	return httpRoute.rules
 }
 
-func (httpRoute *httpRouteDescription) loadAttachedRules(ctx context.Context, k8sClient client.Client) (RouteDescriptor, []routeLoadError) {
+func (httpRoute *httpRouteDescription) loadAttachedRules(ctx context.Context, k8sClient client.Client, gatewayDefaultTGConfig *elbv2gw.TargetGroupConfiguration) (RouteDescriptor, []routeLoadError) {
```
zac-nixon, line 65: nit: instead of changing the constructor for all of these routes, you could probably do something like this in the route loader
```
generatedRoute, loadAttachedRulesErrors := preloadedRoute.loadAttachedRules(ctx, l.k8sClient)
			if len(loadAttachedRulesErrors) > 0 { ... }
generatedRoute.setDefaultConfiguration(gatewayDefaultTGConfig)
```

controllers/gateway/eventhandlers/target_group_configuration_events.go
shuqz: i think it is handled 
> TypedRateLimitingInterface[reconcile.Request] in [controller-runtime](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile) ensures that duplicate reconcile.Request objects (namespace/name) are efficiently handled within the workqueue, typically by ensuring that a unique resource is added to the queue only once, even if multiple events occur. 

but i am gonna refactor this part of logic anyway based on previous comment

docs/guide/gateway/customization.md
shuqz: good callout, i forgot to hanle this one when we shift to new approach

docs/guide/gateway/targetgroupconfig.md
shuqz: emmm this default tgc is more like a fallback default, not enforced value mechanism, let me rewording a bit

docs/guide/gateway/loadbalancerconfig.md
shuqz: ok, it is better to go from strict to loose than the other way around

pkg/gateway/routeutils/http.go
shuqz: we need defaultTGC during loadAttachedRules, not after tho.

controllers/gateway/config_resolver.go
```diff
@@ -14,50 +18,60 @@ import (
...
+				return elbv2gw.LoadBalancerConfiguration{}, nil, errors.Errorf("failed to add finalizers on load balancer configuration %s", k8s.NamespacedName(gatewayClassLBConfig))
 			}
 		}
 		storedVersion := getStoredProcessedConfig(gwClass)
-		if storedVersion == nil || *storedVersion != gatewayClassLBConfig.ResourceVersion {
+		latestVersion := gatewayClassLBConfig.ResourceVersion
+		if gatewayClassLBConfig.Spec.DefaultTargetGroupConfiguration != nil {
+			tgc, err := lookUpDefaultTGCByName(ctx, k8sClient, gatewayClassLBConfig.Spec.DefaultTargetGroupConfiguration.Name, gatewayClassLBConfig.Namespace)
+			if err == nil && tgc != nil {
+				latestVersion = latestVersion + "-" + tgc.ResourceVersion
```
zac-nixon: Can you refactor this to be a shared helper? We want to make sure ID generation like this is consistent, especially because of line 69.

shuqz: done

controllers/gateway/targetgroup_configuration_controller.go
```diff
@@ -111,6 +113,19 @@ func (r *targetgroupConfigurationReconciler) handleDelete(tgConf *elbv2gw.Target
 		}
 	}
 
+	if tgConf.Spec.TargetReference == nil {
```
zac-nixon, line 116: I think there's a bug here. If (for some reason)
`tgConf.Spec.TargetReference != nil` (e.g it's in use by a service) we'll remove the finalizer even if it's in-use by an lb config.

shuqz: synced offline. instead of fixing above lines, we are enforcing default tgc with targetReference not allowed.

PR #4613, opened 2026-03-04
Upgrade to v1.5 of Gateway API

zac-nixon:
### Description

Upgrade to use v1.5 of Gateway API. Main change is that TLSRoute moves alpha2 to v1. Need to work through the guidance for customers still, as the current Gateway CRDs remove TLSRoute entirely from alpha2 (effectively breaking users)

Also, threw in a small refactor to support listenersets (next pr, it will be ~largeish)

### Checklist
- [ ] Manually tested

### DIFF
--- a/docs/guide/gateway/gateway.md
+++ b/docs/guide/gateway/gateway.md
@@ -11,16 +11,25 @@ The LBC Gateway API implementation supports the following Gateway API routes:
 * L4 (NLBGatewayAPI): UDPRoute, TCPRoute, TLSRoute >=v2.13.3
 * L7 (ALBGatewayAPI): HTTPRoute, GRPCRoute >= 2.14.0
 
-The LBC is built for Gateway API version v1.3.0.
+The LBC is built for Gateway API version v1.5.0.
 
 ## Prerequisites
 * LBC >= v2.13.0
 * For `ip` target type:
     * Pods have native AWS VPC networking configured. For more information, see the [Amazon VPC CNI plugin](https://github.com/aws/amazon-vpc-cni-k8s#readme) documentation.
 * Installation of Gateway API CRDs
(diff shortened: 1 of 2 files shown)


docs/guide/gateway/gateway.md
```diff
@@ -11,17 +11,26 @@ The LBC Gateway API implementation supports the following Gateway API routes:
...
 * Installation of LBC Gateway API specific CRDs: `kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/refs/heads/main/config/crd/gateway/gateway-crds.yaml`
 
+
+## Upgrading Gateway API CRDs
+
+Most releases of the Gateway API are backwards compatible, meaning you can apply new CRDs without updating the Load Balancer Controller.
+There a few backwards incompatible release.
+
+- Going from a version less than v1.5
+  - v1.5 brings TLSRoute out of alpha and into the v1 API version. Once this happens, old versions of the controller will not work with TLSRoutes.
```
wweiwei-li, line 32: > v1.5 brings TLSRoute out of alpha and into the v1 API version. Once this happens, old versions of the controller will not work with TLSRoutes.

Is the statement true?  The older version of the controller will not work with v1 TLSRoutes, but they'll still work with v1alpha2 TLSRoutes , right ?.  We would better say users must ensure their TLSRoute match the API version their controller expects.

zac-nixon: Unfortunately the statement is true. The Gateway API community decide to set storage = false, serve = false for TLSRoutes on alpha2, meaning that the API won't return them. I am thinking for our v3.2.0 release, we will update the CRDs to set storage = true, to allow seamless transition for users.

shuqz:
outside scope of this PR, if we are updating gateway api version, we need to run conformance test again

PR #4615, opened 2026-03-06
fix route status reconciler for gw v1.5 upgrade

### LINKED ISSUE #4600: [Gateway API] Error in logs with two parentRefs within one httproute
**Bug Description**
When HTTPRoute has a config with two parentRefs (for example, http and https), there are errors in LBC logs
```
 {"level":"error","ts":"2026-02-24T18:59:32Z","logger":"routeReconciler","msg":"Failed to update route status","error":"HTTPRoute.gateway.networking.k8s.io \"vm-cluster-a-victoria-metrics-cluster-vminsert\" is invalid: [status.parents[1].conditions: Required value, <nil>: Invalid value: \"null\": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]"}
```

**Steps to Reproduce**
Create HTTPRoute with spec like:
```
spec:
  hostnames:
  - prod-vminsert-vmcluster-a.dummy.domain
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: cluster-internal
    namespace: kube-system
    sectionName: https
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: cluster-internal
    namespace: kube-system
    sectionName: http
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: vm-cluster-a-victoria-metrics-cluster-vminsert
      port: 8480
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /insert
```

**Expected Behavior**
No error log records

**Actual Behavior**
Error in logs. While HTTPRoute works

**Regression**
Not sure. We are on v3.0.0.

**Current Workarounds**
Not sure if this can be counted as a workaround, but removing one parentRef fixes the issue.

**Environment**
- AWS Load Balancer controller version: v3.0.0
- Kubernetes version: 1.34
- Using EKS (yes/no), if so version?: yes, 1.34
- Using Service or Ingress: HTTPRoute
- AWS region: us-east-1
- How was the aws-load-balancer-controller installed: with a helm chart deployed by ArogCD 
```
  Containers:
   aws-load-balancer-controller:
    Image:       eks/aws-load-balancer-controller:v3.0.0
    Ports:       9443/TCP, 8080/TCP
    Host Ports:  0/TCP, 0/TCP
    Args:
      --cluster-name=general
      --ingress-class=alb
      --enable-shield=false
      --enable-waf=false
      --enable-wafv2=true
      --enable-backend-security-group=false
      --feature-gates=ALBGatewayAPI=true
```

zac-nixon:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4600

### Description

A couple fixes:
1/ Add a in 'fast unit test' mode which is much faster than standard 'make test.
2/ Fix route reconciler to not operate on route parents outside the scope of the reconcile cycle. This fixes #4600.
3/ Refactor Feature Flags to allow the controller to know if a feature flag was set or it has been defaulted.
4/ Refactor CRD detection logic to use one API call rather than two.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/config/feature_gates.go
+++ b/pkg/config/feature_gates.go
@@ -11,7 +11,9 @@ import (
 type Feature string
 
 type FeatureStatus struct {
-	Enabled     bool
+	// Enabled - Is this feature enabled?
+	Enabled bool
+	// IsDefaulted - Did the user specify this flag, or have we fallen back to the default feature state?
 	IsDefaulted bool
 }
 
@@ -107,11 +109,11 @@ func (f *defaultFeatureGates) Enabled(feature Feature) bool {
 }
 
 func (f *defaultFeatureGates) Enable(feature Feature) {
-	f.featureState[feature] = generateFeatureStatus(true, false)
+	f.featureState[feature] = generateSetFeatureStatus(true)
 }
 
 func (f *defaultFeatureGates) Disable(feature Feature) {
-	f.featureState[feature] = generateFeatureStatus(false, false)
+	f.featureState[feature] = generateSetFeatureStatus(false)
 }
 
 func (f *defaultFeatureGates) String() string {
@@ -153,7 +155,7 @@ func (f *defaultFeatureGates) Set(value string) error {
 		if !ok {
 			return fmt.Errorf("unknown feature: %v", k)
 		}
-		f.featureState[Feature(k)] = generateFeatureStatus(v, false)
+		f.featureState[Feature(k)] = generateSetFeatureStatus(v)
 	}
 	return nil
 }
@@ -162,16 +164,16 @@ func (f *defaultFeatureGates) Type() string {
 	return "mapStringBool"
 }
 
-func generateFeatureStatus(enabled, isDefault bool) FeatureStatus {
-	return FeatureStatus{
-		Enabled:     enabled,
-		IsDefaulted: isDefault,
-	}
-}
-
 func generateDefaultFeatureStatus(enabled bool) FeatureStatus {
 	return FeatureStatus{
 		Enabled:     enabled,
 		IsDefaulted: true,
+	}
+}
+
+func generateSetFeatureStatus(enabled bool) FeatureStatus {
+	return FeatureStatus{
+		Enabled:     enabled,
+		IsDefaulted: false,
 	}
 }
--- a/pkg/gateway/crddetect/gateway_feature_flags.go
+++ b/pkg/gateway/crddetect/gateway_feature_flags.go
@@ -30,7 +30,7 @@ func ApplyGatewayCRDDetection(client k8s.DiscoveryClient, featureGates config.Fe
 		return nil
 	}
 
-	availableResources, err := k8s.DetectCRDs(client, sets.New(GatewayV1Alpha2GroupVersion, GatewayV1GroupVersion, "foo"))
+	availableResources, err := k8s.DetectCRDs(client, sets.New(GatewayV1Alpha2GroupVersion, GatewayV1GroupVersion))
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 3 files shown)


pkg/gateway/crddetect/gateway_feature_flags.go
```diff
@@ -14,44 +15,74 @@ const (
...
 // disables the corresponding feature flags when required CRDs are missing.
 // It is called from main() after the k8s client is ready and before any
 // controller reads the feature flags.
-func ApplyGatewayCRDDetection(client k8s.DiscoveryClient, featureGates config.FeatureGates, logger logr.Logger) {
-	standardResult := k8s.DetectCRDs(client, GatewayV1GroupVersion, StandardCRDKinds)
-	experimentalResult := k8s.DetectCRDs(client, GatewayV1Alpha2GroupVersion, ExperimentalCRDKinds)
+func ApplyGatewayCRDDetection(client k8s.DiscoveryClient, featureGates config.FeatureGates, logger logr.Logger) error {
 
-	ApplyGatewayFeatureFlags(standardResult, experimentalResult, featureGates, logger)
+	if !featureGates.GetFeatureStatus(config.ALBGatewayAPI).IsDefaulted && !featureGates.GetFeatureStatus(config.NLBGatewayAPI).IsDefaulted {
```
wweiwei-li, line 28: Q: if a user sets both alb gateway and nlb gateway true,  this line will be true, will do nothing.  What if user set them true but they don't have gateway CRD installed ?

pkg/gateway/crddetect/gateway_feature_flags.go
```diff
@@ -14,44 +15,74 @@ const (
...
+		// User set this flags directly, do nothing.
+		return nil
+	}
+
+	availableResources, err := k8s.DetectCRDs(client, sets.New(GatewayV1Alpha2GroupVersion, GatewayV1GroupVersion, "foo"))
```
wweiwei-li: foo, left over ?

pkg/config/feature_gates.go
```diff
@@ -10,6 +10,11 @@ import (
 
 type Feature string
 
+type FeatureStatus struct {
+	Enabled     bool
+	IsDefaulted bool
```
wweiwei-li, line 17: can you add a comment regarding what does IsDefaulted mean, means untouch, or it means currently equal to controller default value

pkg/gateway/crddetect/gateway_feature_flags.go
zac-nixon: It will crash the controller. I think it is the preferred behavior because the alternative is that you think the Gateway API is enabled when it's really not.

This also helps with installation scenarios where the user had the order wrong. (start controller, then install gateway api crds) as the controller will crash and then pick up that the crds are now installed.

pkg/gateway/crddetect/gateway_feature_flags.go
zac-nixon: indeed, sorry. I was testing the logic out.

pkg/config/feature_gates.go
zac-nixon: Sure thing.

pkg/gateway/crddetect/gateway_feature_flags.go
```diff
@@ -14,44 +15,74 @@ const (
...
+			if !availableKinds.Has(kind) {
+				missing = append(missing, kind)
+			}
+		}
+	}
+
+	return missing
+}
+
+func generateCombinedRequest(productSpecific map[string][]string, req map[string]sets.Set[string]) {
```
shuqz: where is this function used?

zac-nixon: Left over from old impl. will remove.

PR #4617, opened 2026-03-09
Add flag supporting configurable targetgroupbinding requeue duration

cezarsa:
### Description

The targetgroupbinding controller currently has a hardcoded requeue interval of 15s. That requeue interval mostly triggers when using pod readiness gates and waiting for targets in the target group to become healthy.

In some situations, a 15s interval can be too long, especially when rolling update services with a large number of pods and low maxSurge/maxUnavailable values. In these situations, 15s for each pod can add many minutes to the rollout process.

This PR add a new flag to allow experimenting with lower intervals and potentially reducing the time needed for the pod to become fully ready.

### Checklist
- [ ] Manually tested


pkg/targetgroupbinding/resource_manager.go
```diff
@@ -31,10 +31,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const (
-	defaultRequeueDuration = 15 * time.Second
-	invalidVPCTTL          = 60 * time.Minute
```
cezarsa, line 36: This wasn't referenced anywhere so I just deleted it.

zac-nixon:
/ok-to-test

Thanks for opening this. Are you aware you may encounter throttles from AWS APIs with this change?

cezarsa:
> Are you aware you may encounter throttles from AWS APIs with this change?

Yeah, from what I gathered, we're going to see more requests to the `DescribeTargetHealth` API. But it has a pretty generous bucket refill rate of 10 req/s, so my guess is we can likely reduce the interval a bit before we see too many errors.

In any case, we're going to monitor the controller logs for rate-limit errors if we try to decrease from the default.

shuqz (approved):

PR #4620, opened 2026-03-12
add listenerset, fix bug with route reconciler, add proper unit tests

zac-nixon:
### Description

Adds rbac to the lbc role to allow for listener set events.
Fixes an infinite reconcile issue where the route reconciler accidentally removes route statuses.
Adds proper unit tests 

### Checklist
- [ ] Manually tested

### DIFF
--- a/controllers/gateway/route_reconciler.go
+++ b/controllers/gateway/route_reconciler.go
@@ -107,7 +107,6 @@ func (d *routeReconcilerImpl) handleRouteStatusUpdate(routeData routeutils.Route
 
 	// compare it with original status, patch if different
 	if !d.isRouteStatusIdentical(routeOld, route) {
-		d.logger.Info("Patching route", "route", route)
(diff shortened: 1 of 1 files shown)


controllers/gateway/route_reconciler.go
```diff
@@ -322,6 +324,8 @@ func (d *routeReconcilerImpl) isRouteStatusIdentical(routeOld client.Object, rou
 		newStatusMap[key] = status
 	}
 
+	d.logger.Info("Old map, new map", "old", oldStatusMap, "new", newStatusMap)
```
wweiwei-li: Thanks for the fix.  LGTM.  Just wanted to check if you are adding new two` d.logger.Info` level log on purpose.  Looks to me,  it could be put into debug level

zac-nixon: Thanks for catching this, good call. I'll just remove the logging because it may be expensive.

Issue #4341, opened 2025-09-17
Deletion of ingresses and ingress classes do not get eventual consistency if the ingress class is deleted first


james-alford-ln:
**Bug Description**

Deletion of ingresses and ingress classes do not get eventual consistency if the ingress class is deleted first. 

The ingress is orphaned and cannot be cleaned up until the line `ingressClassName: test-ingress-class` is removed from the ingress manifest.

This differs from behaviour using nginx ingress where the ingresses can be removed post ingress class deletion without issue

**Steps to Reproduce**

deploy.yaml
```
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
  namespace: default
  labels:
    app.kubernetes.io/name: test-pod
spec:
  containers:
  - name: nginx
    image: nginx:stable
    ports:
      - containerPort: 80
        name: http-web-svc

---
apiVersion: v1
kind: Service
metadata:
  name: test-service
  namespace: default
spec:
  selector:
    app.kubernetes.io/name: test-pod
  ports:
  - name: http-web-svc
    protocol: TCP
    port: 80
    targetPort: http-web-svc

---
apiVersion: elbv2.k8s.aws/v1beta1
kind: IngressClassParams
metadata:
  name: test-ingress-class-params
spec:
  certificateArn: []
  group:
    name: test-ingress-class-params
  inboundCIDRs:
    - 10.0.0.0/8
  ipAddressType: ipv4
  loadBalancerAttributes: []
  scheme: internal
  subnets:
    ids:
      - subnet-1
      - subnet-2
      - subnet-3
  targetType: ip

---
apiVersion: networking.k8s.io/v1
kind: IngressClass
metadata:
  name: test-ingress-class
  namespace: default
spec:
  controller: ingress.k8s.aws/alb
  parameters:
    apiGroup: elbv2.k8s.aws
    kind: IngressClassParams
    name: test-ingress-class-params
    scope: Cluster

---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test-ingress
  namespace: default
  labels:
    app.kubernetes.io/name: ingress
  annotations:
    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS":443}]'
    alb.ingress.kubernetes.io/ssl-redirect: '443'
spec:
  ingressClassName: test-ingress-class
  rules:
    - host: subdomain.domain.com
      http:
        paths:
          - path: /
[...]

PR #4624, opened 2026-03-15
Fix ingress stuck in deletion when IngressClass is removed first

aydosman:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4341

### Description

When an IngressClass is deleted before the Ingresses that reference it,
the reconciler correctly marks those Ingresses as inactive members and
attempts to remove their group finalizers via a PATCH/UPDATE. That update
triggers the ValidateUpdate admission webhook, which tries to look up the
now-missing IngressClass and denies the request. The finalizer is never
removed, leaving the Ingress permanently stuck with a DeletionTimestamp
and no way to clean it up short of manually editing the spec.

The fix is to skip all ValidateUpdate checks when the Ingress already has
a DeletionTimestamp set. An ingress in that state has no active invariants
to enforce — the only meaningful operation remaining is cleanup. This is
consistent with ValidateDelete, which already returns nil unconditionally.

Adds a test covering both the fixed path (deleting ingress with missing
IngressClass is allowed) and the unchanged path (non-deleting ingress with
missing IngressClass is still denied).

### Checklist
- [x] Manually tested

### DIFF
--- a/webhooks/networking/ingress_validator.go
+++ b/webhooks/networking/ingress_validator.go
@@ -83,6 +83,10 @@ func (v *ingressValidator) ValidateCreate(ctx context.Context, obj runtime.Objec
 
 func (v *ingressValidator) ValidateUpdate(ctx context.Context, obj runtime.Object, oldObj runtime.Object) error {
 	ing := obj.(*networking.Ingress)
+	// skip validation for ingresses being deleted to allow finalizer removal when IngressClass is already gone.
+	if !ing.DeletionTimestamp.IsZero() {
+		return nil
+	}
 	oldIng := oldObj.(*networking.Ingress)
 	if skip, err := v.checkIngressClass(ctx, ing); skip || err != nil {
 		v.metricsCollector.ObserveWebhookValidationError(apiPathValidateNetworkingIngress, "checkIngressClass")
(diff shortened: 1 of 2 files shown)


aydosman:
Hi @zac-nixon @shraddhabang — this is a fix for #4341. The issue was that `ValidateUpdate` blocks finalizer removal when the `IngressClass` has already been deleted, leaving the `Ingress` permanently stuck. The fix skips webhook validation when `DeletionTimestamp` is set, consistent with how `ValidateDelete` already behaves. Unit tests included. Would appreciate an `/ok-to-test` when you get a chance.

webhooks/networking/ingress_validator.go
```diff
@@ -83,6 +83,10 @@ func (v *ingressValidator) ValidateCreate(ctx context.Context, obj runtime.Objec
 
 func (v *ingressValidator) ValidateUpdate(ctx context.Context, obj runtime.Object, oldObj runtime.Object) error {
 	ing := obj.(*networking.Ingress)
+	// skip validation for ingresses being deleted to allow finalizer removal when IngressClass is already gone.
```
zac-nixon, line 86: You're not really checking if the IngressClass is gone here. If the intended logic is to allow Ingress deletion to proceed when IngressClass is already gone, then adding a check for a `NotFoundError` is probably better.
https://github.com/aydosman/aws-load-balancer-controller/blob/5d4c3e1288ccfa840027d12c028ca3651c18de9d/webhooks/networking/ingress_validator.go#L90-L94

aydosman: @zac-nixon Thank you; I was going for the simplest solution. I will look at this at some point this week.

PR #4625, opened 2026-03-16
[Gateway API] ListenerSet Loader

zac-nixon:
### Description

Implements a ListenerSet loader which utilizes the standard Gateway API attachment rules to load ListenerSets which attach to a Gateway. I took the opportunity to refactor the attachment logic into a separate utility, as we had it implemented twice already, and a third time seemed overkill. While doing the refactor, I found a small bug which incorrectly used the Gateway namespace instead of the route namespace while performing the first part part of the route <-> gateway handshake.

The ListenerSet attachment rules are documented here: https://gateway-api.sigs.k8s.io/geps/gep-1713/. The next step is implement route loading based off of what listener sets are loaded.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/gateway/routeutils/listener_set.go
+++ b/pkg/gateway/routeutils/listener_set.go
@@ -1,1 +1,119 @@
 package routeutils
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+type handshakeState string
+
+const (
+	// acceptedHandshakeState - both resource and gateway agree to attachement
+	acceptedHandshakeState handshakeState = "accepted"
+	// gatewayRejectedHandshakeState - the gateway configuration rejects this configuration
+	gatewayRejectedHandshakeState handshakeState = "rejected"
+	// irrelevantResourceHandshakeState - the resource has no relation to the gateway
+	irrelevantResourceHandshakeState handshakeState = "irrelevant"
+)
+
+type listenerSetLoader interface {
+	retrieveListenersFromListenerSets(ctx context.Context, gateway gwv1.Gateway) ([]gwv1.Listener, error)
+}
+
+type listenerSetLoaderImpl struct {
+	k8sClient         client.Client
+	namespaceSelector namespaceSelector
+	logger            logr.Logger
+}
+
+func newListenerSetLoader(k8sClient client.Client, logger logr.Logger) listenerSetLoader {
+	return &listenerSetLoaderImpl{
+		k8sClient:         k8sClient,
+		namespaceSelector: newNamespaceSelector(k8sClient),
+		logger:            logger.WithName("listener-set-loader"),
+	}
+}
+
+func (l *listenerSetLoaderImpl) retrieveListenersFromListenerSets(ctx context.Context, gateway gwv1.Gateway) ([]gwv1.Listener, error) {
+	return []gwv1.Listener{}, nil
+
+}
+
+func (l *listenerSetLoaderImpl) getListenerSets(ctx context.Context, gateway gwv1.Gateway) ([]gwv1.Listener, error) {
+	listenerSets := &gwv1.ListenerSetList{}
+	err := l.k8sClient.List(ctx, listenerSets)
+	if err != nil {
+		return nil, err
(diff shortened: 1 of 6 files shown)


pkg/gateway/routeutils/listener_set.go
```diff
@@ -0,0 +1,119 @@
...
+func newListenerSetLoader(k8sClient client.Client, logger logr.Logger) listenerSetLoader {
+	return &listenerSetLoaderImpl{
+		k8sClient:         k8sClient,
+		namespaceSelector: newNamespaceSelector(k8sClient),
+		logger:            logger.WithName("listener-set-loader"),
+	}
+}
+
+func (l *listenerSetLoaderImpl) retrieveListenersFromListenerSets(ctx context.Context, gateway gwv1.Gateway) ([]gwv1.Listener, error) {
+	return []gwv1.Listener{}, nil
```
shraddhabang, line 42: Why empty?
We should be calling `getListenerSets` here right or expose that instead of this.

pkg/gateway/routeutils/listener_set.go
```diff
@@ -0,0 +1,119 @@
...
+	listenerSets := &gwv1.ListenerSetList{}
+	err := l.k8sClient.List(ctx, listenerSets)
+	if err != nil {
+		return nil, err
+	}
+
+	var result []gwv1.Listener
+	for _, item := range listenerSets.Items {
+		handshake, err := l.listenerSetGatewayHandshake(ctx, item, gateway)
+		if err != nil {
```
shraddhabang, line 56: Error due to one bad listener_set may block the good ones. Should we instead skip this and continue for the rest?

pkg/gateway/routeutils/listener_set.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (l *listenerSetLoaderImpl) retrieveListenersFromListenerSets(ctx context.Context, gateway gwv1.Gateway) ([]gwv1.Listener, error) {
+	return []gwv1.Listener{}, nil
+
+}
+
+func (l *listenerSetLoaderImpl) getListenerSets(ctx context.Context, gateway gwv1.Gateway) ([]gwv1.Listener, error) {
+	listenerSets := &gwv1.ListenerSetList{}
+	err := l.k8sClient.List(ctx, listenerSets)
```
shraddhabang, line 48: Can we scope this list call to allowedListeners namespaces instead of fetching the clusterwide? We do the check later during handshake logic anyway. So we could return early if the gateway does not allow any listener sets.

pkg/gateway/routeutils/route_attachment_helper.go
```diff
@@ -27,23 +27,8 @@ func newRouteAttachmentHelper(logger logr.Logger) routeAttachmentHelper {
...
-		if parentRef.Kind != nil && *parentRef.Kind != "Gateway" {
-			continue
-		}
-
-		var namespaceToCompare string
-
-		if parentRef.Namespace != nil {
-			namespaceToCompare = string(*parentRef.Namespace)
-		} else {
-			namespaceToCompare = gw.Namespace
```
shraddhabang, line 41: Good catch!

pkg/gateway/routeutils/listener_set.go
wweiwei-li: Non-blocking: we could add if gateway doesn't allow any ListenerSets, skip getListenerSets entirely.

shuqz: is retrieveListenersFromListenerSets more of a placeholder here? and details are gonna be filled in follow-up PR? same thing with `getListenerSets` cuz i did not see how it handling duplicate listeners and conflict ones, so i felt those 2 functions may be completed later

zac-nixon: Yeah exactly @shuqz. This is just a place holder.

pkg/gateway/routeutils/listener_set.go
zac-nixon: The error is from Kubernetes API errors, not misconfiguration of the listener set.

pkg/gateway/routeutils/listener_set.go
zac-nixon: We need to do a cluster-wide list because we need to be able to update the status of listener sets that are trying to attach to the Gateway, but the attachment policy doesn't allow it.

shraddhabang: I see. Makes sense.

PR #4630, opened 2026-03-19
[feat i2g]implement annotation translate

shuqz:
### Description
- it is a large file but test files takes a large portion of it because i have unit test file for almost all each new files. 
- this PR implememnts translation of annotations, pkg/ingress2gateway/translate/annotation_coverage_test.go file summarized which annotations are handled by this PR. anything related to ListenerRuleConfiguration is not handled, along with some other features. 
- please focus on changes under directory `ingress2gateway/translate`
- handles priority chain (ingressparam > service > ingress)
- LoadBalancerConfiguration and TargetGroupConfiguration are only generated when there are actual annotations to map — empty configs are not emitted
- Migration tracking tag (gateway.k8s.aws/migrated-from) is added to LB and TG configs only when they have real content
- defaultAction has a unique behavior, see documentation for detail 
- added a comment handle in reader, make it resilient to comments

Some safety test checks: 
- TestAllIngressAnnotationsCovered — verifies all 66 annotation suffixes from constants.go are categorized as implemented, planned, or not applicable. Fails if a new annotation is added without being categorized.
- TestAllIngressClassParamsFieldsHandled — uses reflection to verify every field in IngressClassParamsSpec is handled. Fails if a new field is added without being mapped.

### Checklist
- [x] Manually tested

input ingress manifest
```
# This example demonstrates the annotation priority chain:
# IngressClassParams > Service annotations > Ingress annotations
#
# - Ingress says scheme=internal, but IngressClassParams overrides to internet-facing
# - Ingress says target-type=instance, but Service overrides to ip
# - healthcheck-path comes from Ingress (no override)
---
apiVersion: v1
kind: Namespace
metadata:
  name: demo
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: echoserver
  namespace: demo
spec:
  replicas: 2
  selector:
    matchLabels:
      app: echoserver
  template:
    metadata:
      labels:
        app: echoserver
    spec:
      containers:
        - name: echoserver
          image: public.ecr.aws/docker/library/httpd:2.4
          ports:
            - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: echoserver
  namespace: demo
  annotations:
    # Service overrides Ingress: target-type ip wins over instance
spec:
  type: NodePort
  selector:
    app: echoserver
  ports:
    - port: 80
      targetPort: 80
      protocol: TCP
---
kind: IngressClass
metadata:
  name: alb
spec:
  controller: ingress.k8s.aws/alb
  parameters:
    apiGroup: elbv2.k8s.aws
    kind: IngressClassParams
    name: alb-params
---
kind: IngressClassParams
metadata:
  name: alb-params
spec:
  # IngressClassParams overrides Ingress: internet-facing wins over internal
  scheme: internet-facing
  tags:
    - key: ManagedBy
      value: platform-team
---
kind: Ingress
metadata:
  name: echo
  namespace: demo
  annotations:
    # This will be overridden by IngressClassParams → internet-facing
    # This will be overridden by Service annotation → ip
    alb.ingress.kubernetes.io/target-type: instance
    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80},{"HTTP": 81}]'
    alb.ingress.kubernetes.io/healthcheck-path: /migrate
    alb.ingress.kubernetes.io/healthcheck-interval-seconds: "30"
    alb.ingress.kubernetes.io/load-balancer-attributes: idle_timeout.timeout_seconds=120
    alb.ingress.kubernetes.io/target-group-attributes: deregistration_delay.timeout_seconds=30
    alb.ingress.kubernetes.io/tags: App=echo
spec:
  ingressClassName: alb
  # defaultBackend: catch-all for requests not matching any rule
  defaultBackend:
    service:
      name: echoserver
      port:
        number: 80
  rules:
    - host: echo.example.com
      http:
        paths:
          - path: /migrate
            pathType: Prefix
            backend:
              service:
                name: echoserver
                port:
                  number: 80
          - path: /health
            pathType: Exact
            backend:
              service:
                name: echoserver
                port:
                  number: 80
```

generated gateway manifest 
```
kind: GatewayClass
metadata:
  name: aws-alb
spec:
---
metadata:
  name: echo-lb-confi-ad8c9c7aee
  namespace: demo
spec:
  loadBalancerAttributes:
  - key: idle_timeout.timeout_seconds
    value: "120"
  scheme: internet-facing
  tags:
    App: echo
    ManagedBy: platform-team
    gateway.k8s.aws/migrated-from: ingress/demo/echo
---
kind: Gateway
metadata:
  name: echo-gateway-de5d75dc9e
  namespace: demo
spec:
  gatewayClassName: aws-alb
  infrastructure:
    parametersRef:
      group: gateway.k8s.aws
      name: echo-lb-confi-ad8c9c7aee
  listeners:
  - name: http-80
    port: 80
    protocol: HTTP
  - name: http-81
    port: 81
    protocol: HTTP
---
kind: HTTPRoute
metadata:
  name: echo-route-793e7248e5
  namespace: demo
spec:
  hostnames:
  - echo.example.com
  parentRefs:
  - name: echo-gateway-de5d75dc9e
    sectionName: http-80
  - name: echo-gateway-de5d75dc9e
    sectionName: http-81
  rules:
  - backendRefs:
    - name: echoserver
      port: 80
    matches:
    - path:
        type: PathPrefix
        value: /migrate
  - backendRefs:
    - name: echoserver
      port: 80
    matches:
    - path:
        type: Exact
        value: /health
---
kind: HTTPRoute
metadata:
  name: echo-default-9aeb32758f
  namespace: demo
spec:
  parentRefs:
  - name: echo-gateway-de5d75dc9e
    sectionName: http-80
  - name: echo-gateway-de5d75dc9e
    sectionName: http-81
  rules:
  - backendRefs:
    - name: echoserver
      port: 80
---
metadata:
  name: echoserv-tg-confi-05e819a62c
  namespace: demo
spec:
  defaultConfiguration:
    healthCheckConfig:
      healthCheckInterval: 30
      healthCheckPath: /migrate
    tags:
      App: echo
      gateway.k8s.aws/migrated-from: ingress/demo/echo
    targetGroupAttributes:
      value: "30"
    targetType: ip
  targetReference:
    name: echoserver
```

### DIFF
--- a/cmd/lbc-migrate/main.go
+++ b/cmd/lbc-migrate/main.go
@@ -120,8 +120,7 @@ func runMigrate(ctx context.Context, opts *ingress2gateway.MigrateOptions) error
 		return resources, nil
 	}
 
-	translateFunc := translate.Translate
 	writeFunc := writer.Write
 
-	return ingress2gateway.Migrate(ctx, *opts, readFunc, translateFunc, writeFunc)
+	return ingress2gateway.Migrate(ctx, *opts, readFunc, translate.Translate, writeFunc)
 }
--- a/docs/guide/gateway/migrate_from_ingress.md
+++ b/docs/guide/gateway/migrate_from_ingress.md
@@ -6,6 +6,10 @@
 
 `lbc-migrate` is a CLI tool that helps migrate AWS Load Balancer Controller (LBC) Ingress resources to Gateway API equivalents. It reads Ingress, Service, IngressClass, and IngressClassParams resources from YAML/JSON files or a live Kubernetes cluster, translates annotations to Gateway API CRD fields, and writes the output manifests.
 
+## Prerequisites
+
+Before applying the generated manifests, ensure the Gateway API CRDs are installed in your cluster at a compatible version. See the [Gateway API installation guide](https://gateway-api.sigs.k8s.io/guides/getting-started/#installing-gateway-api) and the [AWS LBC Gateway API documentation](gateway.md) for supported versions.
+
 ## Installation
 
 Build from source (requires Go):
@@ -95,18 +99,8 @@ Gateway API has no `defaultBackend` equivalent ([upstream docs](https://gateway-
     - **Deployment** — unchanged, continues running your application pods.
     - **Service** — unchanged, the new HTTPRoute `backendRefs` reference it by name.
 
-    After migration, apply the generated Gateway API manifest alongside your existing workload:
-    ```bash
-    # Your existing workload (no changes)
-    kubectl apply -f deployment.yaml
-    kubectl apply -f service.yaml
+    Once equivalent gateway manifest is generated from ingress manifests, apply the generated Gateway API manifest alongside your existing workload.
 
-    # NEW: replaces your old ingress.yaml
-    kubectl apply -f gateway-resources.yaml
-
-    # Remove the old Ingress (after verifying the Gateway ALB works)
-    kubectl delete -f ingress.yaml
-    ```
 
 ### Annotation Priority Chain
 
--- a/pkg/ingress2gateway/reader/file_reader.go
+++ b/pkg/ingress2gateway/reader/file_reader.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	goyaml "gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
 	networking "k8s.io/api/networking/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -89,8 +90,12 @@ func decodeResources(data []byte, resources *ingress2gateway.InputResources) err
 			continue
 		}
 
-		// Skip documents that are only comments (no actual YAML content)
-		if isCommentOnly(doc) {
+		// Skip documents with no meaningful content (comments-only, empty maps, etc.)
+		empty, err := isEmptyDocument(doc)
+		if err != nil {
+			return fmt.Errorf("error parsing document: %w", err)
+		}
+		if empty {
 			continue
 		}
 
@@ -133,16 +138,13 @@ func decodeResources(data []byte, resources *ingress2gateway.InputResources) err
 	return nil
 }
 
-// isCommentOnly returns true if the YAML document contains only comments and whitespace.
-func isCommentOnly(doc []byte) bool {
-	for _, line := range bytes.Split(doc, []byte("\n")) {
-		trimmed := bytes.TrimSpace(line)
-		if len(trimmed) == 0 {
-			continue
-		}
-		if trimmed[0] != '#' {
-			return false
-		}
+// isEmptyDocument returns true if the YAML document has no meaningful content
+// (e.g., only comments, whitespace, or an empty mapping).
+// JSON files don't have comments, so this check is only relevant for YAML.
+func isEmptyDocument(doc []byte) (bool, error) {
+	var m map[string]interface{}
+	if err := goyaml.Unmarshal(doc, &m); err != nil {
+		return false, err
 	}
-	return true
+	return len(m) == 0, nil
 }
--- a/pkg/ingress2gateway/translate/annotation_coverage_test.go
+++ b/pkg/ingress2gateway/translate/annotation_coverage_test.go
@@ -7,6 +7,20 @@ import (
 	annotations "sigs.k8s.io/aws-load-balancer-controller/pkg/annotations"
 )
 
+// Resource type categories for annotation mapping.
+const (
+	LoadBalancerConfig = "LoadBalancerConfiguration"
+	ListenerConfig     = "ListenerConfiguration"
+	TargetGroupConfig  = "TargetGroupConfiguration"
+	HealthCheck        = "HealthCheckConfig"
+	HTTPRouteConfig    = "HTTPRoute"
+	IngressGrouping    = "IngressGrouping"
+	Routing            = "Routing"
+	Authentication     = "Authentication"
+	FrontendNLB        = "FrontendNLB"
+	NotApplicable      = "NotApplicable"
+)
+
 // TestAllIngressAnnotationsCovered ensures every Ingress annotation suffix defined in
 // pkg/annotations/constants.go is accounted for in the migration tool.
 //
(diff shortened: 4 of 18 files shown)


cmd/lbc-migrate/main.go
```diff
@@ -119,7 +120,8 @@ func runMigrate(ctx context.Context, opts *ingress2gateway.MigrateOptions) error
 		return resources, nil
 	}
 
+	translateFunc := translate.Translate
```
zac-nixon: nit (your ide might also say this) you can just pass `translate.Translate` into `ingress2gateway.Migrate` rather than creating a new variable.

docs/guide/gateway/migrate_from_ingress.md
```diff
@@ -70,16 +69,55 @@ You can combine `-f` and `--input-dir`, but `--from-cluster` cannot be used with
...
+    - **Namespace** — already exists, no changes needed.
+    - **Deployment** — unchanged, continues running your application pods.
+    - **Service** — unchanged, the new HTTPRoute `backendRefs` reference it by name.
+
+    After migration, apply the generated Gateway API manifest alongside your existing workload:
+    ```bash
+    # Your existing workload (no changes)
+    kubectl apply -f deployment.yaml
+    kubectl apply -f service.yaml
+
```
zac-nixon: If no changes, why are you suggesting to apply the files? :)

docs/guide/gateway/migrate_from_ingress.md
```diff
@@ -70,16 +69,55 @@ You can combine `-f` and `--input-dir`, but `--from-cluster` cannot be used with
...
+    # NEW: replaces your old ingress.yaml
+    kubectl apply -f gateway-resources.yaml
+
+    # Remove the old Ingress (after verifying the Gateway ALB works)
+    kubectl delete -f ingress.yaml
+    ```
```
zac-nixon: Let's NOT suggest destructive actions.

docs/guide/gateway/migrate_from_ingress.md
```diff
@@ -70,16 +69,55 @@ You can combine `-f` and `--input-dir`, but `--from-cluster` cannot be used with
...
+
+### Annotation Priority Chain
+
+LBC resolves annotations with a priority chain (highest wins):
+
+1. **IngressClassParams** (cluster-level policy)
+2. **Service annotations** (per-backend override)
+3. **Ingress annotations** (per-Ingress config)
+4. **LBC controller defaults**
```
zac-nixon, line 112: LBC controller is redundant (Load Balancer Controller Controller)

docs/guide/gateway/migrate_from_ingress.md
```diff
@@ -1,11 +1,10 @@
 # Migrate from Ingress
```
zac-nixon, line 1: Might be good to mention that they are responsible for applying the various CRDs from the correct Gateway API version.

pkg/ingress2gateway/reader/file_reader.go
```diff
@@ -89,6 +89,11 @@ func decodeResources(data []byte, resources *ingress2gateway.InputResources) err
 			continue
 		}
 
+		// Skip documents that are only comments (no actual YAML content)
```
zac-nixon: Interesting -- Does the yaml decoder handle this case?

I would suggest implementing the check this way - https://go.dev/play/p/OyfnZvpHka1

zac-nixon: We don't want to support json objects too?

pkg/ingress2gateway/translate/annotation_coverage_test.go
```diff
@@ -0,0 +1,134 @@
...
+// TestAllIngressAnnotationsCovered ensures every Ingress annotation suffix defined in
+// pkg/annotations/constants.go is accounted for in the migration tool.
+//
+// WHY THIS TEST EXISTS:
+// If a new annotation is added to pkg/annotations/constants.go but not categorized here,
+// this test fails — forcing the developer to decide how to handle it in the migration tool.
+func TestAllIngressAnnotationsCovered(t *testing.T) {
+
+	// Implemented: annotations handled in the translate package.
+	implemented := []string{
```
zac-nixon: I would suggest using a map here, instead of relying on comments, e.g.

```
implemented := map[RESOURCE_TYPE][]string
implemented[LB_RESOURCE] = [annotations.IngressSuffixScheme, annotations.IngressSuffixLoadBalancerName, ...]
implemented[TG_RESOURCE] = [annotations.IngressSuffixTargetType, ...]
```

pkg/ingress2gateway/translate/httproute.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+		for _, path := range rule.HTTP.Paths {
+			routeRule := gwv1.HTTPRouteRule{}
+
+			// Build match
+			match := gwv1.HTTPRouteMatch{}
+			if path.Path != "" {
+				pathType := toGatewayPathType(path.PathType, useRegex)
+				pathValue := path.Path
+				// When using regex path match with ImplementationSpecific, the Ingress controller
```
zac-nixon, line 70: nice catch!

pkg/ingress2gateway/translate/httproute.go
```diff
@@ -0,0 +1,183 @@
...
+					pathValue = pathValue[1:]
+				}
+				match.Path = &gwv1.HTTPPathMatch{
+					Type:  &pathType,
+					Value: &pathValue,
+				}
+			}
+			routeRule.Matches = []gwv1.HTTPRouteMatch{match}
+
+			// Build backendRef
```
zac-nixon, line 83: You need to consider the 'use-annotation' service here, https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/ingress/annotations/#actions

It will add a lot of complexity that you shouldn't tackle now, but a TODO should suffice.

pkg/ingress2gateway/translate/httproute.go
```diff
@@ -0,0 +1,183 @@
...
+		return gwv1.PathMatchExact
+	case networking.PathTypePrefix:
+		return gwv1.PathMatchPathPrefix
+	case networking.PathTypeImplementationSpecific:
+		if useRegex {
+			return gwv1.PathMatchRegularExpression
+		}
+		return gwv1.PathMatchPathPrefix
+	default:
+		return gwv1.PathMatchPathPrefix
```
zac-nixon: IMO, defaulting is not good here. It may to lead to un-intended translations.

pkg/ingress2gateway/translate/httproute.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+}
+
+func deduplicateHostnames(hostnames []gwv1.Hostname) []gwv1.Hostname {
+	seen := make(map[string]struct{})
```
zac-nixon: nit: use the `set` object here.

pkg/ingress2gateway/translate/icp_overrides.go
```diff
@@ -0,0 +1,165 @@
...
+	if icp.Spec.IPAddressType != nil {
+		ipType := gatewayv1beta1.LoadBalancerIpAddressType(*icp.Spec.IPAddressType)
+		spec.IpAddressType = &ipType
+	}
+
+	if icp.Spec.LoadBalancerName != "" {
+		spec.LoadBalancerName = &icp.Spec.LoadBalancerName
+	}
+
+	if icp.Spec.SSLPolicy != "" {
```
zac-nixon, line 39: Is it safe to add an SSL policy to a non-secure listener? I assume we just ignore it, but it's kinda weird.

pkg/ingress2gateway/translate/icp_overrides.go
```diff
@@ -0,0 +1,165 @@
...
+						})
+					}
+				}
+			}
+		}
+		spec.ListenerConfigurations = &lcs
+	}
+}
+
+// applyIngressClassParamsToTGProps applies IngressClassParams overrides directly to TargetGroupProps.
```
zac-nixon, line 159: How does this work for multiple TargetGroups?

pkg/ingress2gateway/translate/lb_config.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	var meaningful []gatewayv1beta1.ListenerConfiguration
+	for _, lp := range listenPorts {
+		protocolPort := gatewayv1beta1.ProtocolPort(fmt.Sprintf("%s:%d", lp.Protocol, lp.Port))
+		lc := gatewayv1beta1.ListenerConfiguration{
+			ProtocolPort: protocolPort,
+		}
+
+		isSecure := lp.Protocol == "HTTPS" || lp.Protocol == "TLS"
```
zac-nixon: Why TLS?

pkg/ingress2gateway/translate/tg_config.go
```diff
@@ -0,0 +1,159 @@
...
+		props.EnableMultiCluster = v
+	}
+
+	if tags := getStringMap(annos, annotations.IngressSuffixTags); len(tags) > 0 {
+		props.Tags = &tags
+	}
+
+	// target-control-port.${serviceName}.${servicePort}
+	if serviceName != "" && servicePort > 0 {
+		tcpSuffix := fmt.Sprintf("%s.%s.%d", annotations.IngressSuffixTargetControlPort, serviceName, servicePort)
```
zac-nixon: tcpSuffix is a little weird (tcp is normally the protocol :))

pkg/ingress2gateway/translate/translate.go
```diff
@@ -0,0 +1,226 @@
...
+		gw := buildGateway(gatewayName, namespace, lbConfig, listenPorts)
+		out.Gateways = append(out.Gateways, gw)
+
+		// Build HTTPRoute(s) — may produce a separate route for defaultBackend
+		routes := buildHTTPRoutes(ing, namespace, gatewayName, listenPorts)
+		out.HTTPRoutes = append(out.HTTPRoutes, routes...)
+
+		// Step 7: Build TargetGroupConfigurations for each unique service
+		for _, svcRef := range getServiceRefs(ing) {
+			tgcKey := fmt.Sprintf("%s/%s", namespace, svcRef.name)
```
zac-nixon: I think the key needs to include service port too? The route references the service by service name / namespace / port.

pkg/ingress2gateway/translate/translate.go
```diff
@@ -0,0 +1,226 @@
...
+
+// mergeTGAnnotations merges service-level annotations for TG-specific fields.
+// Service annotations take priority over Ingress annotations for per-backend fields.
+func mergeTGAnnotations(ingressAnnotations map[string]string, servicesByKey map[string]corev1.Service, namespace, svcName string) map[string]string {
+	merged := make(map[string]string)
+	for k, v := range ingressAnnotations {
+		merged[k] = v
+	}
+
+	svcKey := fmt.Sprintf("%s/%s", namespace, svcName)
```
zac-nixon, line 142: same comment about using service port, although I think we don't need it here please double check.

pkg/ingress2gateway/translate/translate.go
```diff
@@ -0,0 +1,226 @@
...
+				merged[key] = v
+			}
+		}
+	}
+
+	return merged
+}
+
+// serviceRef holds a service name and port referenced by an Ingress backend.
+type serviceRef struct {
```
zac-nixon, line 153: I think we need the namespace too, it's just the namespace of the ingress file.

pkg/ingress2gateway/utils/helpers.go
```diff
@@ -0,0 +1,70 @@
...
+	"strings"
+)
+
+// JoinStrings joins a string slice with commas.
+func JoinStrings(ss []string) string {
+	return strings.Join(ss, ",")
+}
+
+// GetListenerName generates a valid Gateway API SectionName string from protocol and port.
+// The result is lowercase to satisfy the SectionName regex: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
```
zac-nixon, line 16: Nit: GetSectionName, you use two different nomenclatures (Listener and Section name)

pkg/ingress2gateway/utils/helpers.go
```diff
@@ -0,0 +1,70 @@
...
+	return fmt.Sprintf("%s-%d", strings.ToLower(protocol), port)
+}
+
+// resourceName builds a deterministic, collision-resistant resource name.
+// Pattern: {name}-{suffix}-{hash10}
+// The hash is SHA-256 of namespace/name/suffix, truncated to 10 hex chars.
+// This mirrors the controller's naming pattern (e.g., "k8s-demo-echogate-ddb87892de").
+// Name is truncated to 8 chars and suffix to 8 chars; trailing hyphens from
+// truncation are trimmed to produce valid K8s names.
+func resourceName(namespace, name, suffix string) string {
```
zac-nixon, line 27: It would be cool to somehow include the original resource name in here. For example, if I have an ingress `foo`, my gateway could be named `foo` as well.

pkg/ingress2gateway/reader/file_reader.go
shuqz: i thought we cannot add comments in json so did not handle it

docs/guide/gateway/migrate_from_ingress.md
shuqz: yea let me be more thoughtful on this part of documentation, i will modify whole part

pkg/ingress2gateway/translate/annotation_coverage_test.go
shuqz: lol i struggled on this, in the end i felt like i only want to care if the annotation has been handled. the comment part is more for future developers convenience and understanding. but i can update it. 
FYI: i am planning to have a table that shows how ingress annotations are mapped to gateway when i am done with all translation for customers reference.

pkg/ingress2gateway/translate/httproute.go
shuqz: yea originally i did not see how Set is natively supported in GO, but i think we have a set in package

pkg/ingress2gateway/utils/helpers.go
shuqz: yes it will, you can see my example gateway manifest from description. some example, ingress name is: echo. and generated lbc name is `echo-lb-confi-ad8c9c7aee`, gatewaty name is `echo-gateway-de5d75dc9e`, httproute name is `echo-route-793e7248e5`

cmd/lbc-migrate/main.go
shuqz: just trying to be consistent on pattern lol but i can update it

docs/guide/gateway/migrate_from_ingress.md
shuqz: i think i want to keep it here for a complete information

docs/guide/gateway/migrate_from_ingress.md
shuqz: added Prerequisites section

pkg/ingress2gateway/translate/annotation_coverage_test.go
shuqz: done

pkg/ingress2gateway/translate/httproute.go
shuqz: i will return an error as default

pkg/ingress2gateway/translate/annotation_coverage_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"github.com/stretchr/testify/assert"
+	annotations "sigs.k8s.io/aws-load-balancer-controller/pkg/annotations"
+)
+
+// TestAllIngressAnnotationsCovered ensures every Ingress annotation suffix defined in
+// pkg/annotations/constants.go is accounted for in the migration tool.
+//
+// WHY THIS TEST EXISTS:
+// If a new annotation is added to pkg/annotations/constants.go but not categorized here,
```
wweiwei-li, line 28: This is nice

pkg/ingress2gateway/translate/translate.go
wweiwei-li: +1

pkg/ingress2gateway/translate/translate.go
```diff
@@ -0,0 +1,226 @@
...
+func mergeTGAnnotations(ingressAnnotations map[string]string, servicesByKey map[string]corev1.Service, namespace, svcName string) map[string]string {
+	merged := make(map[string]string)
+	for k, v := range ingressAnnotations {
+		merged[k] = v
+	}
+
+	svcKey := fmt.Sprintf("%s/%s", namespace, svcName)
+	if svc, ok := servicesByKey[svcKey]; ok {
+		// Service annotations override for TG-level fields
+		tgSuffixes := []string{
```
wweiwei-li: Do you think if we can just do a full merge.  so we don't need to maintain the hardcode list tgSuffixes ? The mergeTGAnnotations seems like only be used by buildTargetGroupConfig.

pkg/ingress2gateway/translate/lb_config.go
shuqz: removed

pkg/ingress2gateway/translate/icp_overrides.go
shuqz: i will add a check and only add it to HTTPS ones, thanks for catching it

pkg/ingress2gateway/translate/translate.go
```diff
@@ -0,0 +1,226 @@
...
+		migrationTag := fmt.Sprintf("ingress/%s/%s", namespace, ing.Name)
+
+		lbConfig := buildLoadBalancerConfigResource(lbConfigName, namespace, effectiveAnnotations, listenPorts, migrationTag)
+
+		// Apply IngressClassParams overrides directly to the LB config (highest priority)
+		var icp *elbv2api.IngressClassParams
+		if ing.Spec.IngressClassName != nil {
+			icp = ingressClassParamsByClass[*ing.Spec.IngressClassName]
+		}
+		if lbConfig != nil && icp != nil {
```
wweiwei-li: Q: I see lbConfig will be nil if there are no LB-level annotations. Would there be a case where there are no LB annotations but the Ingress does have an IngressClassParams? In that case, do we need to applyIngressClassParams ?

pkg/ingress2gateway/translate/httproute.go
shuqz: thanks for bringing it up! it does not look trivial, i will add it to todo

pkg/ingress2gateway/translate/httproute.go
```diff
@@ -0,0 +1,183 @@
...
+				match.Path = &gwv1.HTTPPathMatch{
+					Type:  &pathType,
+					Value: &pathValue,
+				}
+			}
+			routeRule.Matches = []gwv1.HTTPRouteMatch{match}
+
+			// Build backendRef
+			if path.Backend.Service != nil {
+				port := gwv1.PortNumber(path.Backend.Service.Port.Number)
```
wweiwei-li: ServiceBackendPort could also be name, do we need to handle named port ?

pkg/ingress2gateway/translate/icp_overrides.go
shuqz: each service backend gets its own TargetGroupConfiguration. and for each of them, applyIngressClassParamsToTGProps will be called. so IngressClassParams applies to all target groups under that Ingress. do you have concern on this?

pkg/ingress2gateway/translate/tg_config.go
shuqz: renamed it

pkg/ingress2gateway/translate/translate.go
shuqz: done, also let me switch tgcCreated to use set instead of map

pkg/ingress2gateway/translate/translate.go
shuqz: this key is for svc, which should be unique in format namespace/name.

pkg/ingress2gateway/translate/translate.go
shuqz: emm i am directly passing namespace to functions right now, i guess i can just put namespace in serviceRef struct, it can reduce function attributes.

pkg/ingress2gateway/translate/translate.go
shuqz: good call, it might be too careful to double check in both place, it will increase maintenance burden

pkg/ingress2gateway/translate/translate.go
shuqz: it should, let me handle it

pkg/ingress2gateway/translate/httproute.go
shuqz: i will add a step to resolve name to port number

pkg/ingress2gateway/translate/translate.go
```diff
@@ -121,72 +141,24 @@ func mergeTGAnnotations(ingressAnnotations map[string]string, servicesByKey map[
...
+// serviceRef holds a service reference extracted from an Ingress backend.
 type serviceRef struct {
-	name string
-	port int32
+	namespace string
+	name      string
+	port      int32
 }
 
-// getServiceRefs extracts all unique service references from an Ingress.
```
shuqz, line 159: while making changes to this file, i realize both buildHttp and getServiceRefs will iterate over all ingresses. it is unnecessary, buildHttp can iterate it once and get service ref data

PR #4639, opened 2026-03-23
[Gateway API] Support attaching routes from listeners originating from a ListenerSet

zac-nixon:
### Description

Refactored the route, listener mapper to hopefully clean the code up a bit. This PR adds the ability for users to attach listeners from a ListenerSet to a Gateway. Things left to do:

1/ Add ListenerSet status updates
2/ Automated E2E test add.

See inline comments for more context.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/gateway/routeutils/listener_attachment_helper.go
+++ b/pkg/gateway/routeutils/listener_attachment_helper.go
@@ -37,7 +37,6 @@ func newListenerAttachmentHelper(k8sClient client.Client, logger logr.Logger) li
 // Gateway API rules to determine compatibility between listener and route.
 // Returns: (compatibleHostnames, allowed, failedRouteData, error)
 func (attachmentHelper *listenerAttachmentHelperImpl) listenerAllowsAttachment(ctx context.Context, parentNamespace string, listener gwv1.Listener, route preLoadRouteDescriptor, matchedParentRef gwv1.ParentReference, hostnamesFromHttpRoutes map[int32]sets.Set[gwv1.Hostname], hostnamesFromGrpcRoutes map[int32]sets.Set[gwv1.Hostname]) ([]gwv1.Hostname, *RouteData, error) {
-	// check namespace TODO --- Update for ListenerSet, should be ListenerSet namespace.
 	namespaceOK, err := attachmentHelper.namespaceCheck(ctx, parentNamespace, listener, route)
 	if err != nil {
 		return nil, nil, err
--- a/pkg/gateway/routeutils/loader.go
+++ b/pkg/gateway/routeutils/loader.go
@@ -255,6 +255,12 @@ func (l *loaderImpl) loadChildResources(ctx context.Context, preloadedRoutes map
 
 func generateRouteDataCacheKey(rd RouteData) string {
 	port := ""
+
+	kind := gatewayKind
+	if rd.ParentRef.Kind != nil {
+		kind = string(*rd.ParentRef.Kind)
+	}
+
 	if rd.ParentRef.Port != nil {
 		port = fmt.Sprintf("%d", *rd.ParentRef.Port)
 	}
@@ -266,5 +272,5 @@ func generateRouteDataCacheKey(rd RouteData) string {
 	if rd.ParentRef.Namespace != nil {
 		namespace = string(*rd.ParentRef.Namespace)
 	}
-	return fmt.Sprintf("%s-%s-%s-%s-%s-%s-%s", rd.RouteMetadata.RouteName, rd.RouteMetadata.RouteNamespace, rd.RouteMetadata.RouteKind, rd.ParentRef.Name, namespace, port, sectionName)
+	return fmt.Sprintf("%s-%s-%s-%s-%s-%s-%s-%s", kind, rd.RouteMetadata.RouteName, rd.RouteMetadata.RouteNamespace, rd.RouteMetadata.RouteKind, rd.ParentRef.Name, namespace, port, sectionName)
 }
--- a/pkg/gateway/routeutils/loader_test.go
+++ b/pkg/gateway/routeutils/loader_test.go
@@ -41,7 +41,7 @@ func (m *mockMapper) mapListenersAndRoutes(ctx context.Context, gw gwv1.Gateway,
 		routesByPort:              m.mapToReturn,
 		compatibleHostnamesByPort: make(map[int32]map[string]sets.Set[gwv1.Hostname]),
 		failedRoutes:              m.routeStatusUpdates,
-		matchedParentRefs:         matchedParentRefs,
+		matchedParentRefs:         m.matchedParentRefs,
 		routesPerListener:         m.listenerRouteCount,
 	}, nil
 }
@@ -193,6 +193,7 @@ func Test_LoadRoutesForGateway(t *testing.T) {
 		acceptedKinds            sets.Set[RouteKind]
 		expectedMap              map[int32][]RouteDescriptor
 		expectedPreloadMap       map[int32][]preLoadRouteDescriptor
+		parentRefs               map[string][]gwv1.ParentReference
 		expectedPreMappedRoutes  []preLoadRouteDescriptor
 		mapperRouteStatusUpdates []RouteData
 		expectedReconcileQueue   map[string]bool // generateRouteDataCacheKey -> succeeded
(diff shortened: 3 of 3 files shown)


pkg/gateway/routeutils/attachment_helper.go
```diff
@@ -28,12 +28,12 @@ func doesResourceAttachToGateway(parentRef gwv1.ParentReference, resourceNamespa
 	return nameCheck && nsCheck
 }
 
-func doesResourceAllowNamespace(ctx context.Context, fromNamespaces gwv1.FromNamespaces, labelSelector *metav1.LabelSelector, nsSelector namespaceSelector, resourceNamespace string, gw gwv1.Gateway) (bool, error) {
```
zac-nixon, line 31: We want to re-use the logic between ListenerSet and Gateway.

pkg/gateway/routeutils/listener_attachment_helper.go
```diff
@@ -16,7 +15,7 @@ import (
 // listenerAttachmentHelper is an internal utility interface that can be used to determine if a listener will allow
 // a route to attach to it.
 type listenerAttachmentHelper interface {
-	listenerAllowsAttachment(ctx context.Context, gw gwv1.Gateway, listener gwv1.Listener, route preLoadRouteDescriptor, matchedParentRef *gwv1.ParentReference, hostnamesFromHttpRoutes map[types.NamespacedName][]gwv1.Hostname, hostnamesFromGrpcRoutes map[types.NamespacedName][]gwv1.Hostname) ([]gwv1.Hostname, bool, *RouteData, error)
+	listenerAllowsAttachment(ctx context.Context, parentNamespace string, listener gwv1.Listener, route preLoadRouteDescriptor, matchedParentRef gwv1.ParentReference, hostnamesFromHttpRoutes map[int32]sets.Set[gwv1.Hostname], hostnamesFromGrpcRoutes map[int32]sets.Set[gwv1.Hostname]) ([]gwv1.Hostname, *RouteData, error)
```
zac-nixon, line 18: There was a bug where we only considered conflicting hostnames using listener section name, however we need to use listener port to accurately capture all attached hostnames.

pkg/gateway/routeutils/listener_attachment_helper.go
```diff
@@ -35,26 +34,22 @@ func newListenerAttachmentHelper(k8sClient client.Client, logger logr.Logger) li
...
-func (attachmentHelper *listenerAttachmentHelperImpl) listenerAllowsAttachment(ctx context.Context, gw gwv1.Gateway, listener gwv1.Listener, route preLoadRouteDescriptor, matchedParentRef *gwv1.ParentReference, hostnamesFromHttpRoutes map[types.NamespacedName][]gwv1.Hostname, hostnamesFromGrpcRoutes map[types.NamespacedName][]gwv1.Hostname) ([]gwv1.Hostname, bool, *RouteData, error) {
-	port := matchedParentRef.Port
-	sectionName := matchedParentRef.SectionName
-	// check namespace
-	namespaceOK, err := attachmentHelper.namespaceCheck(ctx, gw, listener, route)
+func (attachmentHelper *listenerAttachmentHelperImpl) listenerAllowsAttachment(ctx context.Context, parentNamespace string, listener gwv1.Listener, route preLoadRouteDescriptor, matchedParentRef gwv1.ParentReference, hostnamesFromHttpRoutes map[int32]sets.Set[gwv1.Hostname], hostnamesFromGrpcRoutes map[int32]sets.Set[gwv1.Hostname]) ([]gwv1.Hostname, *RouteData, error) {
+	// check namespace TODO --- Update for ListenerSet, should be ListenerSet namespace.
+	namespaceOK, err := attachmentHelper.namespaceCheck(ctx, parentNamespace, listener, route)
 	if err != nil {
-		return nil, false, nil, err
```
zac-nixon, line 46: Removed returning false / true for attachment. Users can utilize `RouteData` being nil to signify successful attachment.

pkg/gateway/routeutils/listener_attachment_helper.go
```diff
@@ -35,26 +34,22 @@ func newListenerAttachmentHelper(k8sClient client.Client, logger logr.Logger) li
...
-		return nil, false, &rd, nil
+		return nil, new(GenerateRouteData(false, true, string(gwv1.RouteReasonNotAllowedByListeners), RouteStatusInfoRejectedMessageNamespaceNotMatch, route.GetRouteNamespacedName(), route.GetRouteKind(), route.GetRouteGeneration(), matchedParentRef)), nil
 	}
 
 	// check kind
 	kindOK := attachmentHelper.kindCheck(listener, route)
 	if !kindOK {
-		rd := GenerateRouteData(false, true, string(gwv1.RouteReasonNotAllowedByListeners), RouteStatusInfoRejectedMessageKindNotMatch, route.GetRouteNamespacedName(), route.GetRouteKind(), route.GetRouteGeneration(), gw, port, sectionName)
-		return nil, false, &rd, nil
+		return nil, new(GenerateRouteData(false, true, string(gwv1.RouteReasonNotAllowedByListeners), RouteStatusInfoRejectedMessageKindNotMatch, route.GetRouteNamespacedName(), route.GetRouteKind(), route.GetRouteGeneration(), matchedParentRef)), nil
```
zac-nixon, line 51: This was suggested by my IDE. In go 1.26 we can use `new()` instead of &Value{}.

pkg/gateway/routeutils/listener_attachment_helper.go
```diff
@@ -63,30 +58,28 @@ func (attachmentHelper *listenerAttachmentHelperImpl) listenerAllowsAttachment(c
...
 		if !hostnameOK {
-			rd := GenerateRouteData(false, true, string(gwv1.RouteReasonNoMatchingListenerHostname), RouteStatusInfoRejectedMessageNoMatchingHostname, route.GetRouteNamespacedName(), route.GetRouteKind(), route.GetRouteGeneration(), gw, port, sectionName)
-			return nil, false, &rd, nil
+			return nil, new(GenerateRouteData(false, true, string(gwv1.RouteReasonNoMatchingListenerHostname), RouteStatusInfoRejectedMessageNoMatchingHostname, route.GetRouteNamespacedName(), route.GetRouteKind(), route.GetRouteGeneration(), matchedParentRef)), nil
 		}
 	}
 
 	// check cross serving hostname uniqueness
 	if route.GetRouteKind() == HTTPRouteKind || route.GetRouteKind() == GRPCRouteKind {
-		hostnameUniquenessOK, conflictRoute := attachmentHelper.crossServingHostnameUniquenessCheck(route, hostnamesFromHttpRoutes, hostnamesFromGrpcRoutes)
```
zac-nixon, line 76: The bad thing about fixing the cross serving hostname check (using port instead of section name) means that we lost the route name.

pkg/gateway/routeutils/route_attachment_helper_test.go
zac-nixon, line 1: I was able to consolidate this logic into loader.go, we didn't need all this additional logic.

pkg/gateway/routeutils/loader.go
```diff
@@ -153,24 +152,23 @@ func (l *loaderImpl) LoadRoutesForGateway(ctx context.Context, gw gwv1.Gateway,
...
 				}
 			}
 		}
 	}
 
 	return &LoaderResult{
 		Routes:            loadedRoute,
-		Listeners:         gw.Spec.Listeners,
-		AttachedRoutesMap: attachedRouteMap,
+		AttachedRoutesMap: mapResult.routesPerListener,
```
shuqz, line 165: why do we remove Listeners from loaderResult? is it gonna be added back?

pkg/gateway/routeutils/listener_attachment_helper.go
shuqz: other than the route name, i am wondering if the scope of this check changes? the map was per-listener before, but now it looks like per-port? e.g. for this case:
```
# Gateway with two listeners on the same port, different hostnames
listeners:
- name: api
  port: 443
  hostname: *.example.com
- name: web  
  port: 443
  hostname: web.example.com
---
# HTTPRoute attached to listener "api"
parentRefs:
- name: my-gw
  sectionName: api
hostnames: ["*.example.com"]
---
# GRPCRoute attached to listener "web"
parentRefs:
- name: my-gw
  sectionName: web
hostnames: ["web.example.com"]
```
i think this is a non-conflict case since they attach to different listener, with the code change in this PR, can you verify if it is still non-conflict?

pkg/gateway/routeutils/listener_attachment_helper.go
shuqz: can it be nil? i remember they need some attributes not empty

pkg/gateway/routeutils/route_attachment_helper.go
shuqz, line 1: thanks for the refactor!

pkg/gateway/routeutils/listener_attachment_helper.go
zac-nixon: check out the old code path. there was no case where we returned attachment = true with route data being nil.

pkg/gateway/routeutils/listener_attachment_helper.go
zac-nixon: Shouldn't this be a conflicting case? The traffic logically comes into the load balancer via port 443, not the section names.

shuqz: synced offline

PR #4641, opened 2026-03-24
Change default client-side rate-limiter to correctly match ELBv2 api limits

cezarsa:
### Description

The current default client-side throttling rules for ALBs cause calls to `RegisterTargets|DeregisterTargets` to consume buckets from 2 different Limiters. This was found by reading the code after looking into AWS API request rate metrics when using a very high controller concurrency configuration.

The default ELB throttling rules have a second rule for ELBv2 matching everything (`.*`). However, calls to RegisterTargets do not stop evaluating after matching the first rule, and end up matching that `.*` rule as well. Consuming tokens from both: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/75eb48adadbb836ecd4d4fd5e6401dc7dd704e8c/pkg/aws/throttle/throttler.go#L76-L80

This PR fixes this by using more restrictive Regexps and also adds a new rule specifically for `Modify.*` calls, which have a different rate limit rule according to ELB docs: https://docs.aws.amazon.com/elasticloadbalancing/latest/userguide/elb-api-throttling.html#bucket-capacity-and-refill-rate

### Checklist
- [ ] Manually tested

### DIFF
--- a/docs/deploy/configurations.md
+++ b/docs/deploy/configurations.md
@@ -159,7 +159,7 @@ Once disabled, the controller shall not take any actions on the waf addons of th
 Controller uses the following default throttle config:
 
 ```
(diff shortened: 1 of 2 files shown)


cezarsa:
I thought I was getting crazy but apparently the Describe.* call bucket have increased in the docs from 10:40 to 50:200. I've just pushed a new commit adding that change as well.

PR #4645, opened 2026-03-25
[Gateway API] ListenerSet status updater

zac-nixon:
### Description

Implements a ListenerSet status updater similar to what we have for the route status update. This way when there is many ListenerSets attached to the same Gateway, then the main reconcile loop is still fast. The key difference this status updater has compared to the route status reconciler is that we need to have a complex data structure representing the listener status of the ListenerSet. The solution I came up with was to have a separate map item has the listener statuses, the listener status is looked up during ListenerSet status update and then deleted afterwards.

Summary of changes

- Fix Route Reconciler to allow for ListenerSet parent refs, previously the logic always looked up Gateways.
- Change the reconcile logic to NOT stop on invalid listener, instead the listener is not programmed and there is a status that mentions the listener is invalid.
- Refactored the Listener Status generation logic to work between generating Listener status for Gateways and ListenerSets.
- Introduced a ListenerSet reconciler.

TODO:

- Documentation Updates
- Automated E2E tests

ListenerSet status E2E test: https://gist.github.com/zac-nixon/ea8dbe16d1cb20db89175e9b4cffda0c

### Checklist
- [x] Manually tested

### DIFF
--- a/test/e2e/gateway/alb_listenerset_test.go
+++ b/test/e2e/gateway/alb_listenerset_test.go
@@ -174,6 +174,7 @@ var _ = Describe("test k8s alb gateway with ListenerSet", func() {
 				Expect(observedGW.Status.Addresses).NotTo(BeEmpty())
 				dnsName = observedGW.Status.Addresses[0].Value
 				Expect(dnsName).NotTo(BeEmpty())
+				Expect(*observedGW.Status.AttachedListenerSets).To(Equal(int32(1)))
 			})
 
 			var lbARN string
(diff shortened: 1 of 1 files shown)


controllers/gateway/gateway_controller.go
```diff
@@ -416,38 +414,61 @@ func (r *gatewayReconciler) updateGatewayStatusSuccess(ctx context.Context, lbSt
...
+	}
+
+	for _, rejectedListenerSet := range loaderResults.RejectedListenerSets {
+		r.listenerSetStatusSubmitter.Enqueue(buildRejectedListenerSetStatus(rejectedListenerSet))
+	}
+
 	if needPatch {
 		if err := r.k8sClient.Status().Patch(ctx, gw, client.MergeFrom(gwOld)); err != nil {
-			return errors.Wrapf(err, "failed to update gw status: %v", k8s.NamespacedName(gw))
+			return errors.Wrapf(err, ""+
```
shraddhabang, line 474: Nit: unnecessary ""+""+""

controllers/gateway/listener_status_utils.go
```diff
@@ -6,31 +6,93 @@ import (
...
+	}
+
+	return routeutils.ListenerSetStatusData{
+		ListenerSetMetadata: routeutils.ListenerSetMetadata{
+			ListenerSetName:      listenerSetNamespacedName.Name,
+			ListenerSetNamespace: listenerSetNamespacedName.Namespace,
+			Generation:           results.Generation,
+		},
+		ListenerSetStatusInfo: routeutils.ListenerSetStatusInfo{
+			Accepted:          true,
```
shuqz: looking at `A listener ls-8080 is the only listener on the ListenerSet and it has a protocol conflict with a listener gw-8080 on the Gateway` from https://gateway-api.sigs.k8s.io/geps/gep-1713/#listenerconditions-within-a-listenerset

seems like when there is only one listener and that listener is invalid, then the overall condition is false with 
```
- type: Accepted
  status: False
  reason: ListenersNotValid
- type: Programmed
  status: False
  reason: ListenersNotValid
```

however, for "A listener ls-8080 on the ListenerSet with other valid listeners has a protocol conflict with a listener gw-8080 on the Gateway"
so for a listenerSet with at least one valid listener, overall status is considered as valid 
```
conditions:
- type: Accepted
  status: True
  reason: Accepted
```
how does the logic count all valid or partial valid?

controllers/gateway/listener_status_utils.go
```diff
@@ -6,31 +6,93 @@ import (
...
+			AcceptedReason:    acceptedReason,
+			AcceptedMessage:   acceptedMessage,
+			Programmed:        programmed,
+			ProgrammedReason:  programmedReason,
+			ProgrammedMessage: programmedMessage,
+		},
+	}, buildListenerStatus(results.Generation, results, programmed, generateListenerEntryStatus)
+}
+
+func buildRejectedListenerSetStatus(rejectedListenerSet gwv1.ListenerSet) (routeutils.ListenerSetStatusData, []gwv1.ListenerEntryStatus) {
```
shuqz, line 88: i did not see update on AttachedListenerSets in gateway status throughout the whole PR.

controllers/gateway/listener_status_utils.go
```diff
@@ -113,11 +175,11 @@ func buildAcceptedCondition(gw gwv1.Gateway, reason gwv1.ListenerConditionReason
...
 		LastTransitionTime: metav1.NewTime(time.Now()),
-		ObservedGeneration: gw.GetGeneration(),
+		ObservedGeneration: generation,
 	}
 }
 
-func buildConflictedCondition(gw gwv1.Gateway, reason gwv1.ListenerConditionReason, message string) metav1.Condition {
+func buildConflictedCondition(generation int64, reason gwv1.ListenerConditionReason, message string) metav1.Condition {
 	status := metav1.ConditionTrue
 	if reason != gwv1.ListenerReasonNoConflicts {
```
shuqz, line 195: should be `reason == gwv1.ListenerReasonNoConflicts` cuz we are setting status=false.
so, if noConflict, then conflicted=false

pkg/gateway/routeutils/route_listener_mapper.go
```diff
@@ -139,11 +149,18 @@ func (ltr *listenerToRouteMapperImpl) mapListenersAndRoutes(ctx context.Context,
 	listenerSetListenerSectionNameToListener := make(map[types.NamespacedName]map[gwv1.SectionName]gwv1.Listener)
 
 	for nsn, listenerSet := range listeners.ListenerSetListeners.listenersPerListenerSet {
+		ltr.logger.Info("Got this listener set", "namespace", nsn)
```
shuqz: forget to delete?

controllers/gateway/listener_status_utils.go
zac-nixon: Hmm, so I guess if there are no valid listeners then accepted and programmed need to be false.

shuqz: yes and AttachedListenerSets=0

controllers/gateway/listener_status_utils.go
zac-nixon: It's right here https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4645/changes#diff-2b4f286cea053042f282b5d23adcb512f5a8a41c1ab8bb03f2cdfa517daff3edR458-R460

controllers/gateway/listener_status_utils.go
zac-nixon: fixed

PR #4658, opened 2026-04-01
fix: guard BaseEndpoint assignment for wafregional and shield clients

kahirokunn:
### Description

All service clients in NewDefaultAWSClientsProvider guard the o.BaseEndpoint assignment with if customEndpoint != nil, except wafregional and shield. This is likely an oversight.

Without the guard, when --aws-api-endpoints is not set, nil is unconditionally assigned to o.BaseEndpoint, which overwrites the value inherited from cfg via NewFromConfig (e.g. AWS_ENDPOINT_URL).

This PR adds the same if != nil guard to these two clients, consistent with the other seven.

### Checklist
- [ ] Manually tested


zac-nixon:
/lgtm
/approved

Thanks!

PR #4665, opened 2026-04-03
add calculate attached listenerset unit test

zac-nixon:
cut v3.2.0 release + adds a unit test for calculating # of attached listener sets


shraddhabang (commented):
/lgtm

helm/aws-load-balancer-controller/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: aws-load-balancer-controller
 description: AWS Load Balancer Controller Helm chart for Kubernetes
 version: 3.1.0
-appVersion: v3.1.0
+appVersion: v3.2.0
```
shuqz, line 5: why is version not being updated?

PR #4677, opened 2026-04-07
Add HTTP client timeout to all AWS SDK clients

haouc:
### Description

AWS SDK v2 defaults to no HTTP request timeout, which means API calls can hang indefinitely if AWS is unresponsive. Set a 10-second timeout on all SDK clients by configuring an http.Client in GenerateAWSConfig and in the EC2 IMDS config.

This covers all 9 service clients (EC2, ELBV2, ACM, WAFv2, WAFRegional, Shield, RGT, STS, GlobalAccelerator), the EC2 IMDS client, and any dynamically-created assumed-role ELBV2 clients.

Extracted buildEC2MetadataConfig from NewCloud for testability. Added unit tests to verify both config paths set the timeout.

The in-cluster test confirmed the controller with the 10s HTTP client timeout starts up cleanly, successfully makes AWS API calls, acquires leader election, and runs all three controllers (ingress, service, targetGroupBinding) with zero errors or restarts.

### Checklist
- [X] Manually tested

### DIFF
--- a/pkg/aws/aws_config.go
+++ b/pkg/aws/aws_config.go
@@ -2,6 +2,9 @@ package aws
 
 import (
 	"context"
+	"net/http"
+	"time"
+
 	"github.com/aws/aws-sdk-go-v2/aws"
 	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
 	"github.com/aws/aws-sdk-go-v2/aws/ratelimit"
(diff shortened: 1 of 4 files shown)


pkg/aws/aws_config.go
```diff
@@ -16,6 +19,8 @@ import (
 
 const (
 	userAgent = "elbv2.k8s.aws"
+	// defaultAWSSDKClientTimeout is the timeout for individual HTTP requests made by AWS SDK clients.
+	defaultAWSSDKClientTimeout = 10 * time.Second
```
oliviassss, line 23: 10s is reasonable for individual requests, but since the LBC talks to 9+ AWS services with varying response profiles, I'd prefer if we can have a slightly larger value (like 15s/20s). Also I'm wondering how it handles throttling situation, or because it's per request timeout, won't affect the controller behavior when API is throttling

haouc: Throttling will be errored out quickly and having a timeout shouldn't be impacting existing behaviors on throttling.

haouc: We want to fail fast and 15 sec is a safe compromise and looks fine to me.

PR #4678, opened 2026-04-08
Watch pod events for updating TargetGroupBindings with ReadinessGates

cezarsa:
### Description

This change makes the TargetGroupBinding controller more efficient by not requeueing on a timer when using readiness gates and waiting for pods to become ready.

Previously, when readiness gates were present on pods, the controller would requeue at a fixed interval while it had non-ready endpoints or containers. This is because the controller never knew when the pod containers would be ready, since the readiness gates prevented the EndpointSlice status from being updated when the container status changed.

With this change, we should avoid unnecessary requeues while also making the controller respond more quickly to changes in pods. The change was also hinted as a possible improvement in the original PR that introduced the interval-based requeue: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/1528

### Checklist
- [x] Manually tested


zac-nixon:
I am a little concerned about this performance penalty incurred by this change. Especially in large clusters, this will add a lot of load for processing pods that are not relevant to the load balancers we manage.

cezarsa:
@zac-nixon I don't think that's really gonna be the case. List watchers for Pod resources already exist in the PodInfoRepo implementation (which makes me realize this PR should probably be updated to make them use a controller-runtime Getter instead of a custom ListWatcher), so in a way, the controller is already processing all Pod events, just not storing most of its data.

This PR will essentially just add 2 DeepEqual comparisons and 2 `string.HasPrefix` calls to most events. Which shouldn't have that big of an impact on processing time.

The memory usage increase, though, can be significant, since the controller will be caching full Pods rather than just the PodInfo representation.

I've just tested this on a fairly large cluster with ~13k pods, and this is what it looks like before and after this change:
 
Screenshot 2026-04-14 at 10 41 57

zac-nixon:
Looking at the memory graph, it looks like you're now using ~3x the memory?

cezarsa:
@zac-nixon I've updated the PR adding a new commit to ensure we reuse the PodInfoRepo for watching the Pods, and the complete Pod objects are no longer cached. That improved memory usage a lot, there's only a very small increase on a cluster now with ~25k pods:

Screenshot 2026-05-05 at 19 56 22

zac-nixon (approved):
/ok-to-test
/approved

Thanks for the performance improvement!

zac-nixon (approved):
/lgtm
/approved

PR #4686, opened 2026-04-13
Automate helm ClusterRole RBAC sync from kubebuilder

shraddhabang:
### Issue
We've had repeated issues where developers add new kubebuilder RBAC markers but forget to update the helm chart's ClusterRole, causing helm-installed controllers to have missing permissions. The helm rbac.yaml was maintained by hand, separate from the kubebuilder-generated  role.yaml, and over time they drifted — different resource groupings, missing rules, verb ordering mismatches.

### Description

This PR eliminates that problem by making role.yaml the single source of truth for the controller's ClusterRole rules. The sync now runs automatically as part of make manifests, so there's no extra step to remember. A CI check (`verify-rbac-sync` in `quick-ci`) catches any drift if someone bypasses the Makefile.

### Scope:

This sync only covers the ClusterRole rules (the controller's own permissions). The following are not synced and remain hand-maintained in the helm chart and/or the sync script:

- Leader election Role and RoleBinding (helm-templated with release-specific names/namespaces)
- Conditional secrets permission (clusterSecretsPermissions.allowAllSecrets — helm-only, no kubebuilder equivalent)
- ClusterRoleBinding (helm-templated)
 
If any of these need changes, update `sync-rbac-to-helm.sh` directly.

### Checklist
- [x] Manually tested


hack/verify-rbac-sync.sh
```diff
@@ -0,0 +1,37 @@
...
+set -o nounset
+set -o pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+"${SCRIPT_DIR}/sync-rbac-to-helm.sh"
+
+changed_files=$(git status --porcelain --untracked-files=no -- helm/aws-load-balancer-controller/templates/rbac.yaml || true)
+if [ -n "${changed_files}" ]; then
+   echo "Detected that helm RBAC is out of sync with kubebuilder RBAC; run 'make sync-rbac'"
```
wweiwei-li: The error message is suggesting run make, do we need to also add `sync-rbac` to make file.  Right now it is only part of `make manifests`

wweiwei-li (approved):

helm/aws-load-balancer-controller/templates/rbac.yaml
```diff
@@ -54,83 +54,82 @@ metadata:
...
+  resources: [listenerruleconfigurations/finalizers, loadbalancerconfigurations/finalizers, targetgroupconfigurations/finalizers]
+  verbs: [patch, update]
 - apiGroups: ["gateway.k8s.aws"]
-  resources: [loadbalancerconfigurations/status, targetgroupconfigurations/status, listenerruleconfigurations/status]
-  verbs: [get, patch, watch]
-- apiGroups: ["gateway.networking.k8s.io"]
-  resources: [gatewayclasses, gateways]
-  verbs: [get, list, watch, patch]
+  resources: [listenerruleconfigurations/status, loadbalancerconfigurations/status, targetgroupconfigurations/status]
+  verbs: [get, patch, update]
```
shuqz, line 109: we changed from `watch` to `update`  that's a real permission change instead of order shuffle, and i see we added `patch` permission to resources: [referencegrants]. how are we planning to inform customers. release note?

hack/verify-rbac-sync.sh
```diff
@@ -0,0 +1,37 @@
...
+"${SCRIPT_DIR}/sync-rbac-to-helm.sh"
+
+changed_files=$(git status --porcelain --untracked-files=no -- helm/aws-load-balancer-controller/templates/rbac.yaml || true)
+if [ -n "${changed_files}" ]; then
+   echo "Detected that helm RBAC is out of sync with kubebuilder RBAC; run 'make manifests'"
+   echo "changed files:"
+   printf "%s\n" "${changed_files}"
+   echo "git diff:"
+   git --no-pager diff -- helm/aws-load-balancer-controller/templates/rbac.yaml
+   echo "To fix: run 'make sync-rbac'"
```
shuqz: i do not see sync-rbac in makefile

helm/aws-load-balancer-controller/templates/rbac.yaml
shraddhabang: The [kubebuilder](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/d42a2ae5408514f7ed9ac79898fe88433ec27142/controllers/gateway/gateway_controller.go#L146) had these permissions for referance grants and helm did not have these since they were manually updated. When sync script ran, it picked up kube-builder perms. 

usually customer get the helm perm updates automatically when they upgrade the version. but we will add this in `Actions required` section as well to pick up new values.

hack/verify-rbac-sync.sh
shraddhabang: Good catch updated.

Issue #4651, opened 2026-03-28
Helm Chart optionally install Gateway API's CRDs


tisonkun:
I'm installing the AWS LBC by:

```terraform
module "eks_blueprints_addons" {
  source  = "aws-ia/eks-blueprints-addons/aws"
  version = "~> 1.23"

  cluster_name      = module.eks.cluster_name
  cluster_endpoint  = module.eks.cluster_endpoint
  cluster_version   = module.eks.cluster_version
  oidc_provider_arn = module.eks.oidc_provider_arn

  enable_external_secrets             = true
  enable_metrics_server               = true
  enable_aws_load_balancer_controller = true
  aws_load_balancer_controller = {
    chart_version = var.lbc_chart_version
    set = [
      {
        name  = "vpcId"
        value = module.vpc.vpc_id
      },
      {
        name  = "installCRDs"
        value = "true"
      },
      {
        name  = "controllerConfig.featureGates.ALBGatewayAPI"
        value = "true"
      }
    ]
  }

  depends_on = [module.eks, kubernetes_manifest.gateway_api_crds]
}
```

The "installCRDs" installs only CRDs dedicated to the AWS LBC, not the Gateway API's.

Then the initialized controller would restart forever because the Gateway API CRDs are not available.

Not sure if AWS EKS should provide a way to initiate out-of-the-box Gateway API CRDs, or this Helm Chart should provide a way to install the Gateway API CRDs.

I'm trying to install the CRDs separately by:

```terraform
data "http" "gateway_api_crds" {
  url = "https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.5.0/standard-install.yaml"
}

locals {
  gateway_api_crds = [
    for manifest in provider::kubernetes::manifest_decode_multi(data.http.gateway_api_crds.response_body) :
    { for key, value in manifest : key => value if key != "status" }
  ]
}

resource "kubernetes_manifest" "gateway_api_crds" {
  count      = length(local.gateway_api_crds)
  manifest   = local.gateway_api_crds[count.index]
  depends_on = [module.eks]
}
```

But it fails to plan because the EKS cluster is not initiated in the very first:

```
│ Error: Failed to construct REST client
│ 
│   with module.base.kubernetes_manifest.gateway_api_crds[8],
│   on main.tf line 115, in resource "kubernetes_manifest" "gateway_api_crds":
│  115: resource "kubernetes_manifest" "gateway_api_crds" {
│ 
[...]

PR #4687, opened 2026-04-14
add gateway crds to helm as opt-in

alecrajeev:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4651

### Description

Currently the kubernetes standard gateway CRDs have to be installed before the gateway controller part of `aws-load-balancer-controller` will work. It would be nice to have a way to install these CRDs with this helm chart.

Note: These are opt-in and by default it will not automatically install the CRDs because there may be folks who already have them installed and we do not want them to conflict.

### Checklist
- [x] Manually tested


helm/aws-load-balancer-controller/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: aws-load-balancer-controller
 description: AWS Load Balancer Controller Helm chart for Kubernetes
-version: 3.2.1
```
shraddhabang, line 4: No need to bump these versions yet. We do this as part of our release process.

helm/aws-load-balancer-controller/values.yaml
```diff
@@ -513,3 +513,8 @@ loadBalancerClass:
 
 # creator will disable helm default labels, so you can only add yours
 # creator: "me"
+
+# Whether or not to install the standard kubernetes gateway api crds
+
+gatewayCRDsAPI:
+  enabled: true
```
shraddhabang, line 520: Default should be false since this is opt-in and we don't want to conflict with existing installations.

helm/aws-load-balancer-controller/templates/standard-gatewayapi-crds.yaml
shraddhabang, line 1: Thanks for the contribution! Here's some feedback to make this easier to maintain long-term and more flexible for users.

We want to the ability to opt in to install both standard and experimental CRDs. Instead of embedding Helm directives directly in the 17k-line CRD file, let's keep the raw upstream YAML untouched so we can copy-paste from upstream on updates. Store the source of truth in `config/crd/gateway/upstream/` (for both standard and experimental channels), and extend the `make crds` target to copy them into `helm/aws-load-balancer-controller/files/` — the same way our LBC-specific CRDs already flow from `config/crd/` → `helm/crds/`. The existing `verify-crds.sh` CI check will automatically catch drift without any extra changes.

Then use a single thin template wrapper with independent toggles for each channel (we use standard for L7 and experimental for L4) something like this:

```
# templates/gatewayapi-crds.yaml
{{- if .Values.crds.gatewayAPI.standard.enabled }}
{{ .Files.Get "files/standard-gatewayapi-crds.yaml" }}
{{- end }}
{{- if .Values.crds.gatewayAPI.experimental.enabled }}
{{ .Files.Get "files/experimental-gatewayapi-crds.yaml" }}
{{- end }}
```

```
# values.yaml
crds:
  gatewayAPI:
    standard:
      enabled: false
    experimental:
      enabled: false
```

zac-nixon:
Would love to include this in our next release. Please let me know if you intend to keep working on this :)

helm/aws-load-balancer-controller/templates/standard-gatewayapi-crds.yaml
alecrajeev: Thank you. I was busy with my day job, but just got back to this here: #4906

PR #4490, opened 2025-12-09
Auto discovery default certificate and certificate list

abeowlu:
### Issue
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4090

### Description
AWS ELB API ignores the default certificate when certificate list is provided in TLS negotiation.
The auto discovered default certificate is needed for TLS negotiation. Adding the default certificate discovered to the extre certificate list

brief logic change:
 * add default certificate to extraCert list as AWS ELBs ignores default certificate in certain SNI TLS negotiation scenario
 * [TODO discussion: accept user specified default certificate](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4090) - skipping as `certificate-arn` implements similar functionality, and that logic shortcircuits auto-discovery execution

### Checklist
- [x] Added tests that cover your change (if possible)
- [ ] Added/modified documentation as required (such as the `README.md`, or the `docs` directory)
- [ ] Manually tested
- [x] Made sure the title of the PR is a good description that can go into the release notes

### BONUS POINTS checklist: complete for good vibes and maybe prizes?! :exploding_head:
- [x] Backfilled missing tests for code in same general area :tada:
- [x] Refactored something and made the world a better place :star2:

PR #4708, opened 2026-04-27
Handle default certificate as listener SNI certificate

### LINKED ISSUE #4090: Adding default certificate to the certificate list for SNI client connections
Scenarios in controller certificate auto-discovery;
- where different encryption certificate (ECC, RSA) are auto-discovered
- the ECC certificate is discovered first and in the [LB model is created as the default certificate](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/ab69d951f7f42ce91f5d7880118b8fc7937794f1/pkg/service/model_build_listener.go#L116)
- the RSA encyption certificate is discovered ...and any other certificate, and added to the certificate ilst
- From the ALB documentation, however, the default certificate is never used when a client request specifies the SNI and there is a certificate list
- client attempting to negotiate ECC cipher suite for connection fail

**Motivation**

- the ALB spec would seem to expect the default certificate to also be added to the certificate list
 
- Add the default certificate also to the certificate list. Perhaps appending the certificate twice?

-[ ] Yes, I am willing to contribute a PR to implement this feature

VPC-byte:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it

When an HTTPS listener is configured with multiple ACM certificates, AWS treats the first listener certificate as the default certificate, but it does not always advertise that default certificate through SNI. This can make clients that depend on SNI matching fail unless users duplicate the default certificate in the extra certificate list manually.

This PR includes the listener default certificate in the desired extra/SNI certificate set, while deduplicating by certificate ARN so the existing duplicate-default workaround does not result in duplicate add requests.

#### Which issue(s) this PR fixes

Fixes #4090

#### Special notes for your reviewer

Opened as draft first because I commented on the issue asking whether maintainers prefer automatic reconciliation or an opt-in annotation. The current implementation follows the behavior requested in the issue and keeps duplicate handling idempotent.

#### Testing

```
go test ./pkg/deploy/elbv2 -run Test_buildSDKCertificates -count=1 -v
go test ./pkg/deploy/elbv2 -run "Test_buildSDKCertificates|Test_updateSDKListenerWithExtraCertificatesAddsDefaultCertificateToSNI" -count=1 -v
go test ./pkg/deploy/elbv2 -count=1
```

### DIFF
--- a/pkg/deploy/elbv2/listener_manager.go
+++ b/pkg/deploy/elbv2/listener_manager.go
@@ -430,7 +430,14 @@ func buildSDKCertificates(modelCerts []elbv2model.Certificate) ([]elbv2types.Cer
 	var defaultSDKCerts []elbv2types.Certificate
 	var extraSDKCerts []elbv2types.Certificate
 	defaultSDKCerts = append(defaultSDKCerts, buildSDKCertificate(modelCerts[0]))
-	for _, cert := range modelCerts[1:] {
+
+	extraCertARNs := make(map[string]struct{}, len(modelCerts))
+	for _, cert := range modelCerts {
+		certARN := awssdk.ToString(cert.CertificateARN)
+		if _, exists := extraCertARNs[certARN]; exists {
+			continue
+		}
+		extraCertARNs[certARN] = struct{}{}
 		extraSDKCerts = append(extraSDKCerts, buildSDKCertificate(cert))
 	}
 	return defaultSDKCerts, extraSDKCerts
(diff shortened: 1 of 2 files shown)


shuqz:
Hi, 
Thanks for working on this. Can you first rebase it and then we can put on ok-to-test label?

shuqz:
Hi, 
We notice there is a PR from earlier to address same issue: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4490

We will be prioritize merging #4490

PR #4710, opened 2026-04-29
feat(helm): add opt-in Gateway API CRDs to Helm chart

roulettedares:
Bundles upstream Gateway API v1.5.1 standard and experimental CRDs as opt-in via crds.gatewayAPI.standard.enabled / experimental.enabled (both default false). Raw YAML lives in config/crd/gateway/upstream/ and is copied to helm/files/ by make crds, so verify-crds.sh catches drift automatically. Closes #4687 #4651

### Checklist
- [x] Manually tested


shraddhabang:
Hey @roulettedares ,Thanks for working on this. While testing this, I found out that the current approach of embedding CRD files in the chart and installing via helm install/helm upgrade doesn't work due to the Helm 1MB release secret limit.

```
Error: UPGRADE FAILED: create: failed to create: Secret
"sh.helm.release.v1.aws-load-balancer-controller.v2" is invalid:
data: Too long: must have at most 1048576 bytes

```

I think we need to adopt different approach here. We can keep the CRD files in the same chart but document the install using helm template piped into `kubectl apply --server-side`. This bypasses the release secret entirely since helm template only renders locally — no Helm release is created. This is the same pattern [Envoy Gateway documents](https://github.com/envoyproxy/gateway/tree/main/charts/gateway-crds-helm) for the same reason.

roulettedares:
I didn't hit this issue since I'm installing via argoCD
```
kubectl get gatewayclass  2>&1 && echo "---" && kubectl get pods -n kube-system -l
      app.kubernetes.io/name=aws-load-balancer-controller…)
  ⎿  NAME   CONTROLLER            ACCEPTED   AGE
     alb    gateway.k8s.aws/alb   True       99m
     nlb    gateway.k8s.aws/nlb   True       99m
```
ill update the docs

roulettedares:
@shraddhabang @shuqz @zac-nixon how is this looking?

makarov-roman:
would be very nice to get these, any help needed?

PR #4730, opened 2026-05-07
Set IncludeManagedResources true

wweiwei-li:
### Issue

EC2 recently released a managed resource visibility feature https://aws.amazon.com/about-aws/whats-new/2026/04/ec2-managed-resource-visibility/,  the  DescribeNetworkInterfaces  will not get the ENI for managed respurce by default, need set Set IncludeManagedResources true. 

### Checklist
- [x ] Added tests that cover your change (if possible)
- [ ] Manually tested
- [ x] Made sure the title of the PR is a good description that can go into the release notes


pkg/networking/pod_eni_info_resolver.go
```diff
@@ -262,7 +262,8 @@ func (r *defaultPodENIInfoResolver) resolveViaPodENIAnnotation(ctx context.Conte
 
 	eniIDs := sets.StringKeySet(podKeysByENIID).List()
 	req := &ec2sdk.DescribeNetworkInterfacesInput{
-		NetworkInterfaceIds: eniIDs,
+		NetworkInterfaceIds:     eniIDs,
+		IncludeManagedResources: awssdk.Bool(true),
```
oliviassss, line 266: the change is not necessary for this func as it queries by ID directly, according to EC2 doc
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/amazon-ec2-managed-instances.html#managed-resource-visibility-settings

>The same direct-query-by-ID behavior applies to all affected resource types. You can use describe-volumes, describe-launch-templates, and describe-network-interfaces with specific resource IDs to access hidden managed resources of those types.

PR #4737, opened 2026-05-13
fix checkpoint early-exit

wweiwei-li:
### Issue

fix checkpoint early-exit

### Description
The controller has a performance optimization that skips work when it believes nothing has changed, tracked via a saved checkpoint. Due to a race between two reconcile execution order, one reconcile read a stale pod condition (True) from its local cache — the other reconcile's update to False had not yet propagated. Seeing no change needed, it saved the checkpoint as clean. The scheduled follow-up reconcile that was supposed to update the condition to True then hit the checkpoint early-exit and skipped. 

This bug requires all of following conditions simultaneously:
* Pod readiness gates enabled 
* Same-IP re-registration

This fix is to add a check to the checkpoint early-exit logic, ensuring pods with pending readiness gates are always reconciled

### Checklist
- [x] Manually tested


shraddhabang (approved):

zac-nixon (approved):
/approved
/lgtm

PR #4742, opened 2026-05-15
[feat i2g] multi-ingress TGC with RouteConfigurations

shuqz:
### Description

**Problem:**
- When multiple Ingress resources reference the same backend Service with different TG-level annotations, the migration tool previously only captured the first Ingress's values — silently dropping the rest and emitting duplicate TGC resources with the same name.

**Fix:**
- Accumulate all per-ingress TGC entries across all groups before building TGCs.
- Single entry → `DefaultConfiguration` (simple path, same as before).
- Multiple entries → emit per-ingress `RouteConfigurations` keyed by the generated HTTPRoute name, so the controller's longest-match merge assigns each HTTPRoute its own TG settings.
- Also fixed `applyIngressClassParamsToTGProps` to propagate `icp.Spec.Tags` to TG props (was only handling `TargetType`).

**Key decisions:**
- TLDR: No `reflect.DeepEqual` comparison to detect identical configs — multiple entries always produce RouteConfigurations to avoid subtle equality bugs with pointer-heavy structs.
DETAILS 
```
i was thinking to check if all Ingresses produced identical TargetGroupProps
   using reflect.DeepEqual. If identical, it would collapse them into a single
  DefaultConfiguration instead of emitting RouteConfigurations.

  The problem: TargetGroupProps has fields like Tags *map[string]string, HealthCheckConfig
  *HealthCheckConfiguration, slices, etc. reflect.DeepEqual can report two logically
  equivalent values as different — for example nil vs &map[string]string{}, or nil slice vs
  empty slice. So the "are these identical?" check could give wrong answers, causing the code
   to emit RouteConfigurations when it shouldn't or vice versa.

  so decided it's simpler to skip that comparison entirely: if multiple Ingresses reference
  the same Service, always emit RouteConfigurations regardless of whether the configs match.
  The controller handles this correctly either way — there's no functional downside to having
   RouteConfigurations with identical values.
```
- The controller handles redundant RouteConfigurations with identical values correctly via its longest-match resolution.

### Checklist
- [x] Manually tested
applied same yaml file i used to find the bug and showed in from in-cluster console, it is fixed

### DIFF
--- a/pkg/ingress2gateway/translate/icp_overrides.go
+++ b/pkg/ingress2gateway/translate/icp_overrides.go
@@ -284,6 +284,15 @@ func applyIngressClassParamsToTGProps(props *gatewayv1beta1.TargetGroupProps, ic
 		tt := gatewayv1beta1.TargetType(icp.Spec.TargetType)
 		props.TargetType = &tt
 	}
+	if len(icp.Spec.Tags) > 0 {
+		if props.Tags == nil {
+			tags := make(map[string]string)
+			props.Tags = &tags
+		}
+		for _, t := range icp.Spec.Tags {
+			(*props.Tags)[t.Key] = t.Value
+		}
+	}
 }
 
 // resolveSSLRedirectPort returns the SSL redirect port if configured.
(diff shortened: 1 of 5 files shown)


shraddhabang (approved):

zac-nixon (approved):
/lgtm
/approved

PR #4746, opened 2026-05-20
[feat i2g] bug fix

shuqz:
### Description
  **1. Deterministic group member ordering**
  Sort explicit group members by `group.order` then `namespace/name` before using `members[0]`, matching LBC's runtime `sortGroupMembers` logic.
  **2. `--namespaces` and `--ingress-name` flags**
  - Renamed `--namespace` to `--namespaces` (string slice) for multi-namespace scoping (e.g. `--namespaces team-a,team-b`).
  - Added `--ingress-name` to migrate a single Ingress via direct `Get` call. Requires exactly one `--namespaces` value.
  - see documentation
  **3. Always give explicit `sectionName` in parentRefs**
  ParentRefs now always include explicit `sectionName` per port instead of omitting it when member ports match all Gateway ports. Safer against unintended attachment to future listeners.

### Checklist
- [x] Manually tested

### DIFF
--- a/cmd/lbc-migrate/main.go
+++ b/cmd/lbc-migrate/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"sigs.k8s.io/aws-load-balancer-controller/pkg/ingress2gateway"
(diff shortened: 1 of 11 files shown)


cmd/lbc-migrate/main.go
```diff
@@ -91,20 +94,36 @@ func validateFlags(opts *ingress2gateway.MigrateOptions) error {
...
 		if opts.Kubeconfig != "" {
 			return fmt.Errorf("--kubeconfig can only be used with --from-cluster")
 		}
 	}
 
-	// --namespace and --all-namespaces are mutually exclusive
-	if opts.Namespace != "" && opts.AllNamespaces {
-		return fmt.Errorf("--namespace and --all-namespaces are mutually exclusive")
+	// --namespaces and --all-namespaces are mutually exclusive
+	if len(opts.Namespaces) > 0 && opts.AllNamespaces {
```
shraddhabang, line 112: Lets add unit tests for these validations.

pkg/ingress2gateway/reader/cluster_reader.go
```diff
@@ -94,6 +109,29 @@ func readFromClient(ctx context.Context, k8sClient client.Client, clientSet kube
 	return resources, nil
 }
 
+// listNamespacedResources lists Ingresses and Services for a single namespace
+// (or all namespaces if ns is empty) and appends them to resources.
+func listNamespacedResources(ctx context.Context, k8sClient client.Client, resources *ingress2gateway.InputResources, ns string) error {
```
shraddhabang: Can we update this function to handle all the cases to make the logic simple?
Like it can handle both single ingresses, all namespaces and specific namespaces? The logic is duplicating for single ingresses.

shuqz: updated

cmd/lbc-migrate/main.go
shuqz: added

shraddhabang (approved):

PR #4749, opened 2026-05-20
fix: ignore top-level RegexValues mirrored by AWS in DescribeRules response

### LINKED ISSUE #4744: Gateway API HTTPRoute RegularExpression path causes non-converging reconcile loop (repeated ModifyRule)
**Bug Description**

When using Gateway API `HTTPRoute` with `path.type: RegularExpression`, AWS Load Balancer Controller enters a non-converging reconcile loop and repeatedly calls `ModifyRule` for the same listener rules, even with no config changes.

Root cause: AWS `DescribeRules` mirrors the regex value from `PathPatternConfig.RegexValues` into the top-level `RuleCondition.RegexValues` field. LBC's desired model never sets the top-level field (it stays `nil`), so equality never converges. The `Values` field on `RuleCondition` is already handled via `cmpopts.IgnoreFields` - `RegexValues` was simply never added.

**Steps to Reproduce**

- Step-by-step guide to reproduce the bug:
  1. Enable Gateway API support in AWS LBC (`ALBGatewayAPI=true`).
  2. Apply a `Gateway` + `HTTPRoute` using regex path matching.
  3. Wait for initial reconcile.
  4. Make no further changes.
  5. Observe repeated `modifying listener rule` for same rule ARNs every few seconds.

- Manifests applied while reproducing the issue:
```yaml
apiVersion: gateway.networking.k8s.io/v1beta1
kind: HTTPRoute
metadata:
  name: my-httproute
  namespace: my-namespace
spec:
  parentRefs:
  - name: my-gateway
    sectionName: https
  rules:
  - matches:
    - path:
        type: RegularExpression
        value: ^/api/v1/resource/\w+
    backendRefs:
    - name: my-service
      port: 80
```

- Controller logs/error messages while reproducing the issue:
  - Repeating pattern:
    - `Got request for reconcile`
    - `successfully built model` (contains `pathPatternConfig.regexValues`)
    - `modifying listener rule` (`resourceID: 443:1` and `443:2`, same ARNs repeatedly)
    - `modified listener rule`
    - immediate next `Got request for reconcile`

  - LBC desired model (from logs):
    ```json
    {
      "field": "path-pattern",
      "pathPatternConfig": {
        "regexValues": ["^/api/v1/resource/\\w+"]
      }
    }
    ```

  - AWS `DescribeRules` response:
    ```json
    {
      "Field": "path-pattern",
      "PathPatternConfig": {
        "RegexValues": ["^/api/v1/resource/\\w+"]
      },
      "RegexValues": ["^/api/v1/resource/\\w+"]
    }
    ```

  Every reconcile: desired `RuleCondition.RegexValues = nil` vs actual `RuleCondition.RegexValues = ["^/api/v1/resource/\\w+"]` -> not equal -> `ModifyRule` -> watch event -> immediate reconcile -> repeat.

**Expected Behavior**

After initial reconcile, state converges and no further `ModifyRule` calls occur unless config changes.

**Actual Behavior**

Controller continuously modifies the same listener rules in a loop. The immediate post-`ModifyRule` reconcile is expected (event-driven) - the defect is non-convergence: regex path conditions keep comparing as different, so every reconcile reissues `ModifyRule`.

Frequency: **Always** - 100% reproducible with any `path.type: RegularExpression` rule via Gateway API.

Impact: unnecessary ELBv2 API calls causing **actual throttling in our environment**, observed even with a single service containing 4 listener rules using regex path matching.

**Regression**

Unknown.

**Current Workarounds**

No workarounds available.

**Environment**

- AWS Load Balancer controller version: `v3.1.0`
- Kubernetes version: `v1.32`
- Using EKS (yes/no), if so version?: Yes, `v1.32`
- Using Service or Ingress: Gateway API (HTTPRoute)
- AWS region: `ap-southeast-2`
- How was the aws-load-balancer-controller installed: Helm
  - `helm ls -A | grep -i aws-load-balancer-controller`:
    ```
    aws-loadbalancer-controller  aws-loadbalancer-controller  12  2026-04-08 23:24:44  deployed  aws-load-balancer-controller-3.1.0  v3.1.0
    ```
  - `helm -n aws-loadbalancer-controller get values aws-loadbalancer-controller`:
    ```yaml
    clusterName: <redacted>
    controllerConfig:
      featureGates:
        ALBGatewayAPI: true
    region: ap-southeast-2
    replicaCount: 1
    serviceAccount:
      annotations:
        eks.amazonaws.com/role-arn: arn:aws:iam::<redacted>:role/<redacted>
      create: true
      name: aws-loadbalancer-controller
    syncPeriod: 0h1m0s
    vpcId: vpc-<redacted>
    ```
- Current state of the Controller configuration - relevant args:
  ```
  --sync-period=0h1m0s
  --feature-gates=ALBGatewayAPI=true
  --aws-region=ap-southeast-2
  --cluster-name=<redacted>
  --aws-vpc-id=vpc-<redacted>
  ```
  - `kubectl describe ingressclasses`:
    ```
    Name:         alb
    Controller:   ingress.k8s.aws/alb
    Labels:       app.kubernetes.io/version=v3.1.0
                  helm.sh/chart=aws-load-balancer-controller-3.1.0
    Events:       <none>
    ```
    (Context only; issue reproduces on Gateway API `HTTPRoute`, not Ingress.)
  - `kubectl -n <appnamespace> describe ingress <ingressname>`: N/A - issue reproduces only with Gateway API HTTPRoute regex path matching.
  - `kubectl -n <appnamespace> describe svc <servicename>`: N/A - backend Service unchanged during reproduction.

**Possible Solution**

One-line fix in `pkg/equality/elbv2/compare_option_for_rule_conditions.go`:

```go
// Before
cmpopts.IgnoreFields(elbv2types.RuleCondition{}, "Values"),

// After
cmpopts.IgnoreFields(elbv2types.RuleCondition{}, "Values", "RegexValues"),
```

Failing test case to add to `compare_option_for_rule_conditions_test.go`:

```go
{
    name: "equal - AWS mirrors regexValues to top-level RuleCondition.RegexValues",
    desiredRuleCondition: []types.RuleCondition{
        {
            Field: awssdk.String("path-pattern"),
            PathPatternConfig: &types.PathPatternConditionConfig{
                RegexValues: []string{"^/api/v1/resource/\\w+"},
            },
        },
    },
    actualRuleCondition: []types.RuleCondition{
        {
            Field: awssdk.String("path-pattern"),
            PathPatternConfig: &types.PathPatternConditionConfig{
                RegexValues: []string{"^/api/v1/resource/\\w+"},
            },
            RegexValues: []string{"^/api/v1/resource/\\w+"},
        },
    },
    expected: true, // fails without fix, passes with it
},
```

**Additional Context**

My cluster has both Ingress and Gateway API enabled and active. Reproduction is specific to Gateway API `HTTPRoute` resources with `path.type: RegularExpression`.

pataraktw:
### Issue

[Fixes #4744](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4744)

### Description

When using regex path patterns, AWS `DescribeRules` mirrors `PathPatternConfig.RegexValues`
into the top-level `RuleCondition.RegexValues` field in its response. LBC's desired model
never sets this top-level field, so every reconcile sees a diff and calls `ModifyRule`
unnecessarily, causing a perpetual reconcile loop.

This fix adds `RegexValues` to the ignored top-level `RuleCondition` fields, consistent
with how `Values` is already handled.

Root cause confirmed on v3.1.0 with Gateway API routes using `RegularExpression` match
type — controller called `ModifyRule` on every reconcile cycle despite no actual changes
to the desired state. The same bug exists for Ingress regex path patterns.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/equality/elbv2/compare_option_for_rule_conditions.go
+++ b/pkg/equality/elbv2/compare_option_for_rule_conditions.go
@@ -17,7 +17,7 @@ func CompareOptionForRuleCondition() cmp.Option {
 		cmpopts.IgnoreUnexported(elbv2types.QueryStringConditionConfig{}),
 		cmpopts.IgnoreUnexported(elbv2types.QueryStringKeyValuePair{}),
 		cmpopts.IgnoreUnexported(elbv2types.SourceIpConditionConfig{}),
-		cmpopts.IgnoreFields(elbv2types.RuleCondition{}, "Values"),
+		cmpopts.IgnoreFields(elbv2types.RuleCondition{}, "Values", "RegexValues"),
 	}
 }
 
--- a/pkg/equality/elbv2/compare_option_for_rule_conditions_test.go
+++ b/pkg/equality/elbv2/compare_option_for_rule_conditions_test.go
@@ -79,6 +79,60 @@ func Test_CompareOptionForRuleConditions(t *testing.T) {
 			},
 			expected: true,
 		},
+		{
+			name: "equal - regex path pattern with AWS-mirrored top-level RegexValues",
+			desiredRuleCondition: []types.RuleCondition{
+				{
+					Field: awssdk.String("host-header"),
+					HostHeaderConfig: &types.HostHeaderConditionConfig{
+						Values: []string{"example.com"},
+					},
+				},
+				{
+					Field: awssdk.String("path-pattern"),
+					PathPatternConfig: &types.PathPatternConditionConfig{
+						RegexValues: []string{"^/api/v1/.*"},
+					},
+				},
+			},
+			actualRuleCondition: []types.RuleCondition{
+				{
+					Field: awssdk.String("host-header"),
+					HostHeaderConfig: &types.HostHeaderConditionConfig{
+						Values: []string{"example.com"},
+					},
+				},
+				{
+					Field: awssdk.String("path-pattern"),
+					PathPatternConfig: &types.PathPatternConditionConfig{
+						RegexValues: []string{"^/api/v1/.*"},
+					},
+					RegexValues: []string{"^/api/v1/.*"},
+				},
+			},
+			expected: true,
+		},
+		{
+			name: "not equal - different regex path pattern",
+			desiredRuleCondition: []types.RuleCondition{
+				{
+					Field: awssdk.String("path-pattern"),
+					PathPatternConfig: &types.PathPatternConditionConfig{
+						RegexValues: []string{"^/api/v1/.*"},
+					},
+				},
+			},
+			actualRuleCondition: []types.RuleCondition{
+				{
+					Field: awssdk.String("path-pattern"),
+					PathPatternConfig: &types.PathPatternConditionConfig{
+						RegexValues: []string{"^/api/v2/.*"},
+					},
+					RegexValues: []string{"^/api/v2/.*"},
+				},
+			},
+			expected: false,
+		},
 	}
 
 	for _, tc := range testCase {


zac-nixon (approved):
/lgtm
/approved

Thanks for the investigation and fix!

PR #4757, opened 2026-05-24
Fix feature gates string representation

immanuwell:
### Issue

No related issue found.

### Description

Tiny fix for `FeatureGates.String()`. It was printing the whole `FeatureStatus`, so values looked like `GlobalAcceleratorController={false true}` instead of `GlobalAcceleratorController=false`. That goes sideways when the string is parsed back as `key=bool`, and pflag can hit this path for flag/default rendering.

Repro, before this change:

```
go test ./pkg/config -run TestDefaultFeatureGatesString -count=1
```

It fails with `invalid mapStringBool: GlobalAcceleratorController={false true}`.

Now the output is stable, sorted, and parseable. fwiw this is local CLI/config plumbing, no AWS quota or resource ceiling involved.

### Checklist
- [x] Added/modified documentation as required (not needed)
- [x] Manually tested

Tests:

```
go test ./pkg/config -run TestDefaultFeatureGatesString -count=1
go test ./pkg/config ./pkg/algorithm ./webhooks/... -count=1
go test $(go list ./... | rg -v '/test/e2e|/conformance')
```

`go test ./...` also runs conformance/e2e locally and needs kube config plus cluster-name, so that one fails here for env setup only.

### DIFF
--- a/pkg/config/feature_gates.go
+++ b/pkg/config/feature_gates.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


zac-nixon (approved):
/approved
/lgtm

Thanks for the fix!

zac-nixon:
Not sure why the bot didn't merge this. Sorry

PR #4771, opened 2026-05-28
[feat i2g] add e2e tests for ingress2gateway migration tool

shuqz:
### Description

Adds e2e tests that verify the `lbc-migrate` tool produces Gateway resources whose model matches what the Ingress controller would build for the same Ingress. The verification compares the dry-run-plan annotations from both controllers and asserts there are no unexpected diffs (the existing `pkg/ingress2gateway/console` classifier marks expected migration artifacts — added migration tag, controller-generated names, default health-check drift, etc. — as `Known`).

#### New tests (under `test/e2e/ingress2gateway`, gated by `--enable-migration-tests`)

All three live in a new `Context("dry-run plan comparison")` inside the existing `Describe("Ingress to Gateway Migration Tests")`. They sit alongside the pre-existing full-cutover test, which is now in its own `Context("full cutover workflow")`.

**Test 1 — complex single Ingress (`--from-cluster`)**
- One Ingress with: scheme, target-type=ip, listen-ports HTTP:80, load-balancer-attributes, tags, full health-check overrides, target-group-attributes, a fixed-response action with a `source-ip` condition, plus a per-Service health-check-path override.
- Three paths (longest first to match Gateway-API path-length precedence): `/api/v1/resources` → svc-api, `/api/v1` → svc-static, `/maintenance` → fixed-response action.
- Runs `lbc-migrate --from-cluster --namespaces <ns> --dry-run`, applies the generated YAML, then diffs the two plans.

**Test 2 — cross-namespace IngressGroup with IngressClassParams (`-f` file mode)**
- Two Ingresses in two different namespaces sharing a `group.name`. Member A carries two action annotations (fixed-response + weighted forward to two services) and Member B is a simple member.
- Adds an IngressClassParams that overrides scheme to `internet-facing` (overriding Member A's `internal`) and adds an extra tag — exercises the ICP-priority chain.
- Runs `lbc-migrate -f <temp-file> --dry-run` against a serialized YAML doc containing the IngressClass, ICP, both Ingresses, and three Services.

**Test 3 — multi-group cluster (`--input-dir` + `--split=namespace`)**
- A platform group (2 members in `nsPlatform`) and a standalone Ingress in `nsTeamA`. Member-1's path is longer so Gateway-API path-length precedence aligns with `group.order`.
- Runs `lbc-migrate --input-dir ... --output-dir ... --split=namespace --dry-run`, verifies the per-namespace output layout (`gatewayclass.yaml` at top level + per-namespace subdirs), applies recursively, then compares both plans (platform vs standalone).

#### Other changes

- `pkg/ingress2gateway/translate/httproute.go`: when assembling the route for `spec.defaultBackend`, also call `t.buildLRCForTags` — symmetric with non-default rules at line 154. Without this, user-tag annotations would not be emitted onto the catch-all rule.
- `test/e2e/ingress2gateway/utils.go`: subprocess invocations (`lbc-migrate`, `kubectl apply`) now run through `runShell` which wraps them with a 2-minute `context.WithTimeout`; prevents an unresponsive subprocess from stalling the suite. 

 Dry-run plan annotation cleanup on feature disable

 - When the IngressPlanAnnotation feature gate is disabled (e.g. after migration is complete), the ingress controller now cleans up stale alb.ingress.kubernetes.io/dry-run-plan annotations from all group members on next reconcile. 

### Checklist
- [x] Manually tested
```
Ran 4 of 4 Specs in 730.793 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

Ginkgo ran 1 suite in 12m28.47794825s
Test Suite Passed
```

### DIFF
--- a/pkg/ingress2gateway/console/classify.go
+++ b/pkg/ingress2gateway/console/classify.go
@@ -157,9 +157,9 @@ func isTargetGroupRef(s string) bool {
 	return strings.HasPrefix(s, "#/resources/"+utils.StackResTypeTargetGroup+"/")
 }
 
-// buildUserSpecifiedFields scans Ingress annotations and returns the set of
+// BuildUserSpecifiedFields scans Ingress annotations and returns the set of
 // model field paths that were explicitly configured by the user.
-func buildUserSpecifiedFields(ingressAnnotations map[string]string) UserSpecifiedFields {
+func BuildUserSpecifiedFields(ingressAnnotations map[string]string) UserSpecifiedFields {
 	usf := make(UserSpecifiedFields)
 	for suffix, fieldPath := range annotationToFieldPath {
 		key := annotations.AnnotationPrefixIngress + "/" + suffix
--- a/controllers/ingress/dryrun_test.go
+++ b/controllers/ingress/dryrun_test.go
@@ -14,6 +14,144 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
+func Test_clearDryRunPlanAnnotation(t *testing.T) {
+	tests := []struct {
+		name           string
+		ingAnnotations map[string]string
+	}{
+		{
+			name: "removes annotation when present",
+			ingAnnotations: map[string]string{
+				dryRunPlanAnnotation:               `{"id":"test-stack"}`,
+				"alb.ingress.kubernetes.io/scheme": "internet-facing",
+			},
+		},
+		{
+			name: "no-op when annotation absent",
+			ingAnnotations: map[string]string{
+				"alb.ingress.kubernetes.io/scheme": "internet-facing",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ing := &networking.Ingress{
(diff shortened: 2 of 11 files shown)


pkg/ingress2gateway/console/classify.go
```diff
@@ -157,9 +157,9 @@ func isTargetGroupRef(s string) bool {
 	return strings.HasPrefix(s, "#/resources/"+utils.StackResTypeTargetGroup+"/")
 }
 
-// buildUserSpecifiedFields scans Ingress annotations and returns the set of
+// BuildUserSpecifiedFields scans Ingress annotations and returns the set of
 // model field paths that were explicitly configured by the user.
-func buildUserSpecifiedFields(ingressAnnotations map[string]string) UserSpecifiedFields {
+func BuildUserSpecifiedFields(ingressAnnotations map[string]string) UserSpecifiedFields {
```
zac-nixon, line 162: Looks like this is always used in the console package. Why did you export it?

zac-nixon (approved):

zac-nixon (approved):
/approved
/lgtm

Issue #4775, opened 2026-06-01
enable-endpoint-slices: IPv4 pod addresses registered into IPv6 target group for RequireDualStack services


Sergey-Kizimov:
### Bug Description

When `--enable-endpoint-slices=true` (the default since v3.3.0), the controller attempts to register IPv4 pod addresses into IPv6-only target groups for services with `ipFamilyPolicy: RequireDualStack` and `ipFamilies: [IPv6, IPv4]`. AWS rejects these with:

```
ValidationError: The IP address '10.35.x.x' is not a valid IPv6 address
```

The affected pods' readiness gates (`target-health.elbv2.k8s.aws/*`) are never satisfied and the pods remain stuck in not-ready. Existing pods with stale pre-upgrade registrations continue serving traffic, masking the impact until the next pod reschedule.

### Root Cause

`computeServiceEndpointsData` in `pkg/backend/endpoint_resolver.go` lists **all** EndpointSlices for a service without filtering by address type:

```go
r.k8sClient.List(ctx, epSliceList,
    client.InNamespace(svcKey.Namespace),
    client.MatchingLabels{discovery.LabelServiceName: svcKey.Name})
// returns BOTH addressType=IPv4 and addressType=IPv6 slices
```

For a `RequireDualStack [IPv6, IPv4]` service, Kubernetes creates two EndpointSlices — one per address family. Both are merged into a single flat list and fed into `resolvePodEndpointsWithEndpointsData`. This produces two `PodEndpoint` entries per pod (one with the IPv4 address, one with the IPv6 address).

When `matchPodEndpointWithTargets` runs against a TGB with `ipAddressType: ipv6`, the IPv6 endpoints match existing targets (already registered) while the IPv4 endpoints appear as unmatched and are submitted to `RegisterTargets`. AWS rejects the call since the target group is IPv6-only.

The bug existed since EndpointSlice support was added (PR #2169, Sep 2021) but was dormant because `--enable-endpoint-slices` defaulted to `false`. PR #4353 (merged Sep 2025, shipped in v3.3.0) flipped the default to `true`, exposing the bug.

With `--enable-endpoint-slices=false` (the pre-v3.3.0 default), the legacy `corev1.Endpoints` path is used, which only exposes the **primary** IP family — for `ipFamilies: [IPv6, IPv4]` that is IPv6 — so the bug is never triggered.

### Steps to Reproduce

[...]

PR #4783, opened 2026-06-04
[fix] filter EndpointSlices by IP family for target group registration

shuqz:
### Issue
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4775

### Description

When `--enable-endpoint-slices=true`, the endpoint resolver listed every EndpointSlice for a Service without filtering by address family. For Services with both IPv4 and IPv6 EndpointSlices, this caused the controller to submit IPv4 pod addresses to an IPv6-only target group's `RegisterTargets` call, which AWS rejects with `ValidationError: The IP address '...' is not a valid IPv6 address`. The affected pods' readiness gates never flipped and stayed not-ready.

This change makes `ResolvePodEndpoints` accept the target group's `ipAddressType` and filter EndpointSlices to only those whose `AddressType` matches that family. 

### Checklist
- [x] Manually tested
i was able to reproduce issue from my side, and verified the error is gone after deploying fix. 


pkg/targetgroupbinding/resource_manager.go
```diff
@@ -946,3 +947,11 @@ func (m *defaultResourceManager) getPodAvailabilityZone(ctx context.Context, pod
 
 	return &az, nil
 }
+
+// tgbTargetIPAddressType returns the TGB's IP address type, defaulting to IPv4 when unset.
+func tgbTargetIPAddressType(tgb *elbv2api.TargetGroupBinding) elbv2api.TargetGroupIPAddressType {
```
wweiwei-li: How about using endpointSliceAddressType ? each EndpointSlice has one AddressType

```
func endpointSliceAddressType(tgb *elbv2api.TargetGroupBinding) discovery.AddressType {
	if tgb.Spec.IPAddressType != nil && *tgb.Spec.IPAddressType == elbv2api.TargetGroupIPAddressTypeIPv6 {
		return discovery.AddressTypeIPv6
	}
	return discovery.AddressTypeIPv4
}
```

This way, we could make endpoint_resolve clean, only have kubernetes type

shraddhabang (approved):
/lgtm

PR #4785, opened 2026-06-05
Match Route53 hosted zones by longest suffix in GetHostedZoneID

nbr:
### Description

The [ACM certificate management feature](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/ingress/certificate_management/) validates the certificates it requests using DNS records in Route 53. For each name on a certificate, `GetHostedZoneID` (`pkg/aws/services/route53.go`) finds the hosted zone that should hold the validation record.

The lookup only matched a zone whose name was exactly the certificate name, or the certificate name with its first label dropped. A name two or more labels below the zone apex never matched, and validation failed with:

> no hosted zone found for validation records

With a single `env.example.com` hosted zone, for example, the SAN `*.app.pr-1.env.example.com` gets trimmed once to `app.pr-1.env.example.com`. That still isn't `env.example.com`, so no zone is returned, the validation record is never written, and the certificate stays in `PENDING_VALIDATION`.

This replaces the two checks with a single pass that selects the zone whose labels are the longest suffix of the certificate name. That matches how external-dns resolves a zone: [`ZoneIDName.FindZone`](https://github.com/kubernetes-sigs/external-dns/blob/1f6b9f106a50b29eb573cf1c409cbcb0d6bc4875/provider/zonefinder.go#L50-L75) keeps the matching zone with the longest name (`len(zoneName) > len(suitableZoneName)`), and the AWS provider's [`suitableZones`](https://github.com/kubernetes-sigs/external-dns/blob/1f6b9f106a50b29eb573cf1c409cbcb0d6bc4875/provider/aws/aws.go#L1336) does the same when picking a hosted zone. Apex and one-label-below names resolve to the same zone as before. Names nested deeper now resolve, wildcards match on their non-wildcard labels, and when both a parent and a child zone could match, the more specific child zone is used.

`route53_test.go` covers the apex, nested-wildcard, and parent/child cases.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/aws/services/route53.go
+++ b/pkg/aws/services/route53.go
@@ -59,20 +59,21 @@ func (c *route53Client) GetHostedZoneID(ctx context.Context, domain string) (*st
 
 	recParts := strings.Split(domain, ".")
 
-	// try first if we have an apex record
+	var bestID *string
+	bestLen := -1
 	for _, zone := range zones {
 		zoneParts := strings.Split(strings.TrimRight(*zone.Name, "."), ".")
-		if slices.Equal(recParts, zoneParts) {
-			return zone.Id, nil
+		if len(zoneParts) > len(recParts) {
+			continue
+		}
+		if slices.Equal(recParts[len(recParts)-len(zoneParts):], zoneParts) && len(zoneParts) > bestLen {
+			bestLen = len(zoneParts)
+			bestID = zone.Id
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


the-technat (approved):
Awesome find! That bit defenitely got forgotten.

shraddhabang:
/ok-to-test
Thank you for your contribution. 
/approved

wweiwei-li (approved):

PR #4786, opened 2026-06-05
Fix ACM certificate creation for Ingresses with wildcard hosts

nbr:
### Description

The [`create-acm-cert` annotation](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/ingress/annotations/#create-acm-cert) requests one certificate per ingress, using every host as a SAN and the first host as the Domain Name ([`buildCertificateSpec`](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/3f1e9707ac4ce65504d72b425e82be5245b38e0e/pkg/ingress/model_build_certificates.go#L49)).

- Hosts are a sorted set, so a wildcard like `*.app.example.com` sorts first and becomes the Domain Name.
- The Domain Name is baked into the certificate's resource ID and written as an ACM tag ([`buildCertificateResourceID`](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/3f1e9707ac4ce65504d72b425e82be5245b38e0e/pkg/ingress/model_build_certificates.go#L43), surfaced by [`ResourceTags`](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/3f1e9707ac4ce65504d72b425e82be5245b38e0e/pkg/deploy/tracking/provider.go#L125)).
- ACM rejects `*` in tag values (`[\p{L}\p{Z}\p{N}_.:/=+\-@]*`), so `RequestCertificate` fails for any ingress with a wildcard host:

> ValidationException: Value of the input at 'tags' failed to satisfy constraint: Member must satisfy regular expression pattern: ...

Note this is only the tag value. ACM accepts a wildcard Domain Name on the certificate itself.

#### Fix

- Replace `*` with `wildcard` in the resource ID before it's used as a tag value.
- The Domain Name and SANs sent to ACM are unchanged, so the certificate covers the same names and the resource ID stays unique per ingress.
- Also handles ingresses whose hosts are all wildcards.

#### Tests

- `model_build_certificates_test.go`: an all-wildcard ingress case, plus a test asserting the resource ID contains no `*`.

### Checklist
- [x] Manually tested

### DIFF
--- a/test/e2e/ingress/acm_ingress_test.go
+++ b/test/e2e/ingress/acm_ingress_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 
+	awssdk "github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/acm/types"
 	. "github.com/onsi/ginkgo/v2"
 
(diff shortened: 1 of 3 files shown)


the-technat (approved):
Good catch!

I think your fix is valid. But regarding the case you mentioned with an ingress only containing wildcards, I think it could also be elegant just doing a search/replace action on the tag value. AWS tag values can be up to 255 chars long, and the hash consists of the namespace/name and hostname which all have to be max 64 characters, so replacing * with "wildcard" would be valid too.

nbr:
> Good catch!
> 
> I think your fix is valid. But regarding the case you mentioned with an ingress only containing wildcards, I think it could also be elegant just doing a search/replace action on the tag value. AWS tag values can be up to 255 chars long, and the hash consists of the namespace/name and hostname which all have to be max 64 characters, so replacing * with "wildcard" would be valid too.

@the-technat Thanks for the feedback, I pivoted the PR to address the tag issue directly

the-technat (approved):

shraddhabang:
The fix looks good to me. Thank you for working on this. Can you add an e2e test for this case as well?

nbr:
@shraddhabang e2e tests added in a21b9037df3c80c55aefd912c5fa1b46a1ef2031

test/e2e/ingress/acm_ingress_test.go
```diff
@@ -252,3 +319,11 @@ func ExpectCertTypeToBe(ctx context.Context, certARN string, t types.Certificate
 
 	tf.Logger.Info("Cert of expected type", "arn", certARN, "type", detail.Type)
 }
+
+func ExpectCertDomainNameToBe(ctx context.Context, certARN string, domainName string) {
```
shraddhabang, line 325: Verify ExpectCertDomainNameToBe expectation — since the code still uses hosts[0] (sorted) as DomainName, and * sorts before alphabetical characters, the wildcard host will be the DomainName. The assertion should be changed to expect the wildcard host. Did you run this locally?

nbr: No, I was leaning on [pull-aws-load-balancer-controller-e2e-test](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_aws-load-balancer-controller/4786/pull-aws-load-balancer-controller-e2e-test/2077458099173593088) but now realizing it doesn't actually run my test. The test was written for c1b4e18a6603b1aff7a27c4793e0ddde4aa7da60 before I changed the approach. I'll run the e2e test ASAP. We've been running our fork in prod for about a month now.

shraddhabang:
Thanks for fixing this. 
/approved

zac-nixon:
/lgtm
/approved
🚀

PR #4792, opened 2026-06-15
Fix input order dependency in GW rules precedence

cezarsa:
### Description

Currently, the rule precedence is non-deterministic when some rules include no hostname matches. This is demonstrated by the added test cases.

With no code changes, the test results are as follows:
| Test case  | Result |
| ------------- | ------------- |
| `input order independent result for mixed hostname and no-hostname rules - input order A` | FAIL |
| `input order independent result for mixed hostname and no-hostname rules - input order B` | PASS |
| `input order independent result for mixed hostname and no-hostname rules, with method - input order A` | PASS |
| `input order independent result for mixed hostname and no-hostname rules, with method - input order B` | PASS |

The only thing that changes between test cases with `input order A` and `input order B` is really the order of the input rules.

In practice, this makes the rule ordering non-deterministic, since the input order depends on Go map ordering and the order in which resources are read from the k8s cluster. This bug was found in production while investigating why the listener rules order was flapping.

The fix included here is tentative, and I'm not sure if that's the right approach. I'm open to suggestions on different fixes that make the output deterministic.

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/gateway/routeutils/route_rule_precedence.go
+++ b/pkg/gateway/routeutils/route_rule_precedence.go
@@ -183,11 +183,7 @@ func getHostnameListPrecedenceOrder(hostnameListOne, hostnameListTwo []string) i
 			return precedence
 		}
 	}
-	if len(hostnameListOne) < len(hostnameListTwo) {
-		return -1
-	} else if len(hostnameListOne) > len(hostnameListTwo) {
-		return 1
-	}
+	// can not complete tie breaking at hostname level
 	return 0
 }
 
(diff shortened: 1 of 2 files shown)


cezarsa:
I was wrong in my initial analysis about non-determinism. The result is deterministic based on the route creation timestamp tie-breaker.
The observed non-determinism resulted from differences in the order of HTTPRoute creation on my cluster.

That said, I still think the behavior here is wrong, since the rules have different hostname matches configured and shouldn't rely on the tie-breaker.

shuqz:
Hi, 
Thanks for opening the PR. We followed Gateway API definition to determine precedence order. https://gateway-api.sigs.k8s.io/reference/api-spec/main/spec/#httprouterule
Regarding hostname, it mentioned Exact Hostname Matches has higher priority when wildcard ones. But I do not think there is mention about route with more hostnames should have higher priority than route with less hostnames. If you see it in documentation, do you mind pointing out location?

cezarsa:
No longer needed after https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4809.

PR #4795, opened 2026-06-22
add source nat setting so udp tests pass on ipv6 cluster

zac-nixon:
### Description

Any test that uses a UDP listener fails on an ipv6 cluster, the root cause is that we need to allow for source nating, otherwise the ELB API rejects the configuration. We have to set this value for all subnets in the LB [this is a crazy API setting]. In order to calculate the number of subnets to set the source nat value for, we calculate the unique AZs in the cluster.

### Checklist
- [x] Manually tested


shraddhabang (approved):
/lgtm

shraddhabang (commented):
/lgtm

PR #4802, opened 2026-06-26
Fix deletion guard for Gateway TargetGroupConfigurations still referenced by TCPRoutes

immanuwell:
### Issue

None

### Description

Gateway-targeted `TargetGroupConfiguration`s are a documented path for gateway chaining, but delete reconciliation only guarded Service targets.

So if a `TCPRoute` still points at a `Gateway` backend, deleting the TGC can drop the finalizer and remove the config while it is still referenced. Tiny edge case, but its real.

This patch adds the missing in-use check for `targetReference.kind: Gateway` and covers it with tests.

Repro:
1. Create a `TargetGroupConfiguration` with `targetReference.kind: Gateway`.
2. Create a `TCPRoute` that uses that `Gateway` as a backend.
3. Delete the TGC.
4. Before this patch, the controller can remove the finalizer anyway.
5. With this patch, deletion stays blocked until the `TCPRoute` reference is gone.

### Checklist
- [ ] Manually tested

### DIFF
--- a/controllers/gateway/targetgroup_configuration_controller_test.go
+++ b/controllers/gateway/targetgroup_configuration_controller_test.go
@@ -1,0 +1,118 @@
+package gateway
+
+import (
+	"context"
+	"testing"
+
+	"github.com/go-logr/logr"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	elbv2gw "sigs.k8s.io/aws-load-balancer-controller/apis/gateway/v1beta1"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/k8s"
+	"sigs.k8s.io/aws-load-balancer-controller/pkg/shared_constants"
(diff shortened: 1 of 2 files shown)


immanuwell:
@shraddhabang @shuqz please take a look

controllers/gateway/targetgroup_configuration_controller_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/go-logr/logr"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	elbv2gw "sigs.k8s.io/aws-load-balancer-controller/apis/gateway/v1beta1"
```
dominikhei: Shouldn't this be: sigs.k8s.io/aws-load-balancer-controller/v3/apis/gateway/v1

immanuwell: fixed

immanuwell:
@dominikhei applied your suggestion, please take a look

Issue #4807, opened 2026-07-02
Gateway API: non-deterministic ALB listener rule priority when an HTTPRoute has no hostnames (catch-all flaps above host-scoped routes)


sergeylanzman:
**Bug Description**

With the ALB Gateway API implementation (`--feature-gates=ALBGatewayAPI=true`), the listener rule *priority* assigned to `HTTPRoute` rules is non-deterministic when at least one route has **no `hostnames`** (an all-hosts catch-all). Such a route ties with host-scoped routes during precedence calculation and is then ordered only by route creation timestamp. Because the sort is not stable and its input order depends on Go-map/read order, the resulting ALB rule priorities *flap between reconciles*. When the no-hostname catch-all lands above the host-scoped rules, it shadows them and their traffic is sent to the catch-all backend instead.

**Steps to Reproduce**

- Step-by-step guide to reproduce the bug:
  1. Create an ALB `Gateway` with an `https` listener that has **no** `hostname`.
  2. Apply a no-hostname catch-all HTTPRoute (`PathPrefix: /`) → fallback backend. Create it **first** (so it's the oldest route).
  3. Apply a host-scoped HTTPRoute for `app.example.com` (`PathPrefix: /`) → app backend.
  4. Trigger reconciles repeatedly (e.g., roll the backend pods a few times, or restart the controller) and dump the listener rules after each.
- Manifests applied while reproducing the issue:
```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata: { name: catchall, namespace: default }   # created FIRST -> oldest route
spec:
  parentRefs: [{ kind: Gateway, name: my-alb, sectionName: https }]
  rules:
    - matches: [{ path: { type: PathPrefix, value: / } }]
      backendRefs: [{ name: fallback-svc, port: 80 }]
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata: { name: app-host, namespace: default }
spec:
  parentRefs: [{ kind: Gateway, name: my-alb, sectionName: https }]
  hostnames: [ app.example.com ]
  rules:
    - matches: [{ path: { type: PathPrefix, value: / } }]
      backendRefs: [{ name: app-svc, port: 80 }]
```
- Controller logs/error messages while reproducing the issue (same catch-all rule ARN, priority changing across four reconciles that were triggered only by *unrelated backend endpoint churn* — a pod rollout — with no config change):
```
[...]

PR #4809, opened 2026-07-05
Fix non-transitive rule precedence for routes without hostnames

sergeylanzman:
### Issue

Related to #4807 and #4792.

### Description

`getHostnameListPrecedenceOrder` compares two hostname lists positionally, but only up to `min(len(a), len(b))`:

```go
length := min(len(hostnameListOne), len(hostnameListTwo))
for i := range length {
    precedence := getHostnamePrecedenceOrder(hostnameListOne[i], hostnameListTwo[i])
    if precedence != 0 {
        return precedence
    }
}
return 0
```

When a route has **no hostnames** (a catch-all `/*` HTTPRoute), its list is empty, so `min(0, N) = 0`, the loop never executes, and the function returns `0` ("equal") instead of treating the empty list as the least specific.

**Why this is a problem:** a catch-all becomes hostname-equal to *every* route, so its precedence falls through to path length / creation timestamp. Meanwhile host-specific routes are ordered against each other by hostname specificity (path is never consulted). Mixing these two criteria makes `compareHttpRulePrecedence` **non-transitive** — it contains comparison cycles, e.g.:

```
catch-all  >  console  (equal path -> older creation timestamp wins)
console    >  login    (more specific hostname wins, path ignored)
login      >  catch-all (longer path wins)
```

`sort.Slice` requires the `less` function to be a strict weak ordering; given a non-transitive comparator its output is **undefined** and depends on the input order (which varies between reconciles). In practice this let the catch-all rule intermittently sort **above** host-specific rules, so requests that should match host-scoped rules were served by the catch-all.

**Fix:** after all compared hostnames tie, break by hostname-list length — a shorter (in the extreme, empty) list constrains fewer hostnames and is therefore less specific, so it gets lower precedence. This makes an empty hostname list always lose to any route that has hostnames and restores a strict weak ordering, so the sort result no longer depends on input order.

The same `getHostnameListPrecedenceOrder` is used by both `compareHttpRulePrecedence` and `compareGrpcRulePrecedence`, so the fix covers HTTP and GRPC routes.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/gateway/routeutils/route_rule_precedence.go
+++ b/pkg/gateway/routeutils/route_rule_precedence.go
@@ -183,7 +183,19 @@ func getHostnameListPrecedenceOrder(hostnameListOne, hostnameListTwo []string) i
 			return precedence
 		}
 	}
-	// can not complete tie breaking at hostname level
+	// All compared hostnames tie up to the shorter list's length. A shorter list
+	// (in the extreme, an empty list as used by a catch-all route) constrains
+	// fewer hostnames and is therefore less specific, so it must have lower
+	// precedence. Without this, an empty hostname list returns 0 ("equal") and
+	// precedence falls through to path length / creation timestamp, mixing
+	// incompatible criteria and making the comparator non-transitive.
+	if len(hostnameListOne) != len(hostnameListTwo) {
+		if len(hostnameListOne) > len(hostnameListTwo) {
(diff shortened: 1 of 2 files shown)


zac-nixon:
/lgtm
/approved

thanks for the fix!

shraddhabang (approved):

PR #4794, opened 2026-06-22
[Gateway API]remove overlapping check in httpRoute and GRPCRoute

shuqz:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4793

### Description

- remove rejection for httpRoute and grpcRoute sharing same hostname. 
- this is not breaking change for existing customers, since we are loosing the restriction. what's not allowed before is now allowed. 

### Checklist
- [x] Added tests that cover your change (if possible)
- [ ] Added/modified documentation as required (such as the `README.md`, or the `docs` directory)
- [x] Manually tested
  - before the change, grpc route status update shows 'Accepted=False` for NotAllowedByListeners
  - after the change, grpc route shows `Accepted=true`
  - added e2e test
- [x] Made sure the title of the PR is a good description that can go into the release notes

### BONUS POINTS checklist: complete for good vibes and maybe prizes?! :exploding_head:
- [ ] Backfilled missing tests for code in same general area :tada:
- [ ] Refactored something and made the world a better place :star2:

PR #4818, opened 2026-07-09
fix(gateway): Unified cross-kind route precedence for HTTPRoute/GRPCRoute

shraddhabang:
## Summary

When an HTTPRoute and a GRPCRoute share a hostname on the same HTTPS listener (allowed per GEP-1016), ALB listener rule priority must be determined by route specificity, not route kind.

Previously, `SortAllRulesByPrecedence` sorted HTTP and GRPC rules independently then concatenated all HTTP rules before GRPC rules. This meant a catch-all HTTPRoute (`PathPrefix /`) would always receive ALB Priority 1 while a specific GRPCRoute (`/Service/Method`) received a higher number — allowing the catch-all to intercept gRPC traffic.

## Changes

- **`pkg/gateway/routeutils/route_rule_precedence.go`**: Added a unified `compareRulePrecedenceUnified` comparator that delegates to existing same-kind comparators for intra-kind ordering and uses normalized specificity (hostname → path type → effective path length → headers → additional conditions → timestamp) for cross-kind ordering.
- **`pkg/gateway/routeutils/route_rule_precedence_test.go`**: Added unit tests for cross-kind comparison, plus integration tests with multiple HTTP, multiple GRPC, and mixed route scenarios.
- **`test/e2e/gateway/alb_tests/alb_ip_target_test.go`**: Added e2e test verifying a catch-all HTTPRoute does not intercept specific GRPCRoute traffic.
- **`test/framework/verifier/aws_resource_verifier.go`**: Added `VerifyListenerRulePrecedence` helper for validating ALB rule priority ordering.

## Root Cause

PR #4794 removed the `crossServingHostnameUniquenessCheck` (per GEP-1016 relaxation) to allow HTTPRoute and GRPCRoute on the same listener+hostname. The underlying priority ordering bug pre-existed but was previously unreachable.

## Testing

- All existing unit tests pass (no regression)
- New unit tests cover: cross-kind comparison, multiple HTTP routes, multiple GRPC routes, mixed routes
- E2e test validates gRPC traffic reaches the correct backend even with a catch-all HTTPRoute present (WIP)


pkg/gateway/routeutils/route_rule_precedence.go
```diff
@@ -350,3 +346,130 @@ func getGrpcRoutePathType(method *v1.GRPCMethodMatch) int {
...
+	}
+
+	// 3. Effective path length
+	onePathLen := ruleOne.effectivePathLength()
+	twoPathLen := ruleTwo.effectivePathLength()
+	if onePathLen != twoPathLen {
+		return onePathLen > twoPathLen
+	}
+
+	// 4. Header count
```
shuqz: i think for compareHttpRulePrecedence, we have hasMethod > header count

pkg/gateway/routeutils/route_rule_precedence_test.go
```diff
@@ -151,6 +151,89 @@ func Test_SortAllRulesByPrecedence(t *testing.T) {
...
+	grpcLongSvc := &grpcRouteDescription{
+		route: &gwv1.GRPCRoute{
+			ObjectMeta: v1.ObjectMeta{Name: "grpc-long-svc", Namespace: "ns"},
+			Spec:       gwv1.GRPCRouteSpec{Hostnames: []gwv1.Hostname{"grpc.example.com"}},
+		},
+		rules: []RouteRule{&convertedGRPCRouteRule{rule: &gwv1.GRPCRouteRule{
+			Matches: []gwv1.GRPCRouteMatch{{Method: &gwv1.GRPCMethodMatch{Type: (*gwv1.GRPCMethodMatchType)(awssdk.String("Exact")), Service: awssdk.String("com.company.longpkg.MyService"), Method: awssdk.String("Execute")}}},
+		}}},
+	}
+	// Multi-route test fixtures: mixed HTTP + GRPC
```
shuqz, line 208: can we add more tests for this case, for example, test case includes method-vs-header tie.
i just want to make sure for mixed case, the order is deterministic

pkg/gateway/routeutils/route_rule_precedence_test.go
```diff
@@ -151,6 +151,89 @@ func Test_SortAllRulesByPrecedence(t *testing.T) {
...
+	httpRoot := &httpRouteDescription{
```
shuqz, line 209:
```
  Test_Repro_Cycle — the comparator forms a cycle:
  A<B = true   (HTTP vs HTTP: HasMethod true>false)
  B<C = true   (cross-kind: HeaderCount 3>2)
  C<A = true   (cross-kind: HeaderCount 2>1)
  CYCLE CONFIRMED: A<B<C<A

  Test_Repro_NonDeterministicSort — the same 3 rules in different input orders sort into 3 different priority orders:
  input A,B,C  =>  A > B > C
  input A,C,B  =>  C > A > B
  input B,C,A  =>  B > C > A     ← same rules, different result
  Since ALB priority = sorted index (model_build_listener.go:324-336), those are three different priority assignments
  for identical input.

  The recipe behind the test case

  The cycle needs three rules that tie on the first factors, then split on the diverging one:

  ┌─────┬──────┬───────┬──────────────────────────────────────┬───────────┬─────────┐
  │     │ Kind │ Path  │             Eff. length              │ HasMethod │ Headers │
  ├─────┼──────┼───────┼──────────────────────────────────────┼───────────┼─────────┤
  │ A   │ HTTP │ exact │ 10                                   │ ✅ true   │ 1       │
  ├─────┼──────┼───────┼──────────────────────────────────────┼───────────┼─────────┤
  │ B   │ HTTP │ exact │ 10                                   │ ❌ false  │ 3       │
  ├─────┼──────┼───────┼──────────────────────────────────────┼───────────┼─────────┤
  │ C   │ GRPC │ exact │ 10 (svc 4 + method 4 + 2 separators) │ —         │ 2       │
  └─────┴──────┴───────┴──────────────────────────────────────┴───────────┴─────────┘
```

pkg/gateway/routeutils/route_rule_precedence.go
```diff
@@ -350,3 +346,130 @@ func getGrpcRoutePathType(method *v1.GRPCMethodMatch) int {
...
+// compareCrossKindRulePrecedence compares an HTTPRoute rule against a GRPCRoute
+// rule (or vice versa) using normalized specificity factors so that route kind
+// alone does not determine priority. The comparison order is:
+//  1. Hostname specificity
+//  2. Path type (exact > prefix > regex > none)
+//  3. Effective path length
+//  4. Header count
+//  5. Additional conditions (method, query params — more constraints = more specific)
+//  6. Common tiebreakers (creation timestamp, namespaced name, rule/match index)
+func compareCrossKindRulePrecedence(ruleOne, ruleTwo RulePrecedence) bool {
```
shuqz, line 386: do we wanna add something in documentation regarding how we decide precedence order for mixed case?

pkg/gateway/routeutils/route_rule_precedence_test.go
shraddhabang: Good catch. I will add the `hasMethod` comparison before `headerCount` check to avoid this transitivity.

pkg/gateway/routeutils/route_rule_precedence.go
shraddhabang: Good catch. Will fix this

pkg/gateway/routeutils/route_rule_precedence_test.go
shraddhabang: Yes I can add these

pkg/gateway/routeutils/route_rule_precedence.go
shraddhabang: I think the comments are self-explanatory here. If we add more conditions in future we will need to change those as welll. Instead we can keep these comments as source of truth for everyone on this repo. If someone explicitly asks for these to be included in docs, we can add at that time. But for now, I think these comments are well explained.

Ari4ka (approved):

PR #4819, opened 2026-07-09
docs: add guidance on migrating between services without downtime

### LINKED ISSUE #4810: Target group is deleted and recreated when backend service name changes (no graceful drain)
**Bug Description**
When updating an Ingress backend from one Service to another (both selecting the same pods), the controller deletes the existing target group and creates a new one. This causes a traffic disruption, connections are dropped immediately without respecting `deregistration_delay.timeout_seconds`.

**Steps to Reproduce**
- Step-by-step guide to reproduce the bug:
   1. Create two Services (service-a and service-b) with the same pod selector
   2. Create an Ingress with alb.ingress.kubernetes.io/target-type: ip pointing to service-a
   3. Update the Ingress backend to point to service-b
   4. Observe that the target group is deleted and recreated, dropping active connections
- Manifests applied while reproducing the issue: n/a
- Controller logs/error messages while reproducing the issue: n/a

**Expected Behavior**
When `target-type: ip` is used and the resolved pod IPs are unchanged (both services have the same selector), the controller should update the target group binding in-place rather than performing a delete/recreate cycle. At minimum, the old target group should be drained before removal.

**Actual Behavior**
The controller deletes the existing target group and creates a new one.

**Regression**
Was the functionality working correctly in a previous version ? No

**Current Workarounds**

**Environment**
- AWS Load Balancer controller version: used with eks automode
- Kubernetes version: `v1.34.8-eks-0247562`
- Using EKS (yes/no), if so version?: Yes `v1.34.8-eks-0247562`
- Using Service or Ingress: ingress
- AWS region: ap-southeast-2
- How was the aws-load-balancer-controller installed: eks automode
- Current state of the Controller configuration: eks automode
- Current state of the Ingress/Service configuration: state is fine.
Migrate to a new loadbalancer.

- [x] Yes, I'm willing to submit a PR to fix this issue

**Additional Context**
The controller appears to use the service name as part of the target group identity. Changing the service name triggers a full replacement even when the underlying targets haven't changed. A graceful transition (drain old targets, register new) would prevent downtime during service consolidation.

Thiru-2004-varul:
### Issue
Fixes #4810

### Description
Adds a short section to the blue/green traffic splitting doc explaining how to migrate an Ingress backend from one Service to another without dropping active connections.

Currently, directly changing the backend service name causes the controller to delete the existing target group and create a new one, dropping connections immediately without respecting deregistration_delay.timeout_seconds. This isn't a bug — it's existing behavior — but the recommended alternative (weighted target groups via the actions.* annotation) wasn't documented for this specific use case.

This PR adds a "Migrating between services without downtime" section to docs/guide/use_cases/blue_green/index.md, describing how to use the existing weighted target group feature to shift traffic gradually and avoid dropped connections.

### Checklist
- [x] Manually tested

### DIFF
--- a/docs/guide/use_cases/blue_green/index.md
+++ b/docs/guide/use_cases/blue_green/index.md
@@ -70,6 +70,18 @@ spec:
                 name: blue-green
                 port:
                   name: use-annotation
+
+## Migrating between services without downtime
+
+If you need to switch an Ingress backend from one Service to another (for example, during a service rename or consolidation), directly changing the `service.name` in the Ingress spec causes the controller to delete the existing target group and create a new one. Since this delete happens immediately, any active connections are dropped without respecting `deregistration_delay.timeout_seconds`.
+
+To migrate without dropping connections, use the weighted target group pattern described above instead of changing the backend service directly:
+
+1. Define both the old and new services as weighted target groups in the `actions.*` annotation, starting with 100% weight on the old service and 0% on the new one.
+2. Gradually shift the weight toward the new service over time (for example, 80/20, then 50/50, then 0/100).
+3. Once traffic is fully shifted to the new service, remove the old service from the annotation.
+
+Since both target groups stay registered with the ALB throughout the migration, connections drain naturally and no traffic is dropped.
 ```
 
 


shraddhabang (approved):
/lgtm
Thanks for the changes.

PR #4829, opened 2026-07-10
feat(gateway): Use Gateway API v1 for TCPRoute and UDPRoute

### LINKED ISSUE #4816: [Gateway API] Upgrade TCPRoute and UDPRoute from v1alpha2 to v1 (Gateway API 1.6 GA graduation)

Update TCPRoute and UDPRoute handling to support Gateway API (CRDs) ≥ 1.6 `gateway.networking.k8s.io/v1`.

**Motivation**

TCPRoute and UDPRoute were promoted to **GA (v1) in Gateway API 1.6** ([kubernetes-sigs/gateway-api#4790](https://github.com/kubernetes-sigs/gateway-api/issues/4790)). Users running Gateway API ≥ 1.6 now have `v1` as the served version for these resources. Continuing to use only `v1alpha2` in the LBC means:

- Resources are created under a deprecated API version that may be removed in a future Gateway API release
- Tooling (kubectl, admission webhooks, audit logs) sees `v1alpha2` objects instead of the promoted `v1` API
- Conformance test manifests upstream already moved to `v1` ([kubernetes-sigs/gateway-api#4979](https://github.com/kubernetes-sigs/gateway-api/pull/4979)); the LBC is now behind

PR [#4519](https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4519) explicitly verified `v1alpha2` was the only supported version under Gateway API v1.2.x — that was correct at the time but is now outdated.

Extend the existing auto-detection mechanism introduced in [#4602](https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4602) to also detect which API version is available for TCPRoute and UDPRoute, and use the highest available version:

1. At startup, query the Kubernetes discovery API for the served versions of `tcproutes` and `udproutes` under `gateway.networking.k8s.io`
2. If `v1` is available (Gateway API ≥ 1.6), use `gateway.networking.k8s.io/v1`
3. If only `v1alpha2` is available (Gateway API < 1.6), fall back to `gateway.networking.k8s.io/v1alpha2`
4. Register both versions in the scheme so the controller handles objects created under either version without requiring migration

This is the same pattern used by other Kubernetes controllers when graduating from beta/alpha to stable APIs (e.g., the Ingress `v1beta1` → `v1` transition). It makes the change fully non-breaking: users on Gateway API < 1.6 continue working unchanged, while users on ≥ 1.6 get the stable API.

- **Hard-cut to `v1` with a minimum Gateway API version bump**: simpler implementation but breaks users on Gateway API < 1.6 until they upgrade.
- **Keep `v1alpha2` indefinitely**: avoids any work but creates long-term drift as `v1alpha2` gets deprecated and eventually removed upstream.

- [x] Yes, I am willing to contribute a PR to implement this feature

nicomazzarello:
### Issue

Fixes #4816

### Description

TCPRoute and UDPRoute graduated to GA (`gateway.networking.k8s.io/v1`) in Gateway API 1.6. The controller previously handled these routes only under `v1alpha2`, which on Gateway API 1.6 standard-channel clusters (that serve `v1` only) causes the route watches to fail with `no matches for kind "TCPRoute" in version "gateway.networking.k8s.io/v1alpha2"` and, when the gateway feature flags are defaulted, silently disables `NLBGatewayAPI` because CRD detection only looks for the `v1alpha2` kinds.

Per review feedback, this PR migrates TCPRoute/UDPRoute handling to the `v1` API — one controller release supports one Gateway API release (same approach as the TLSRoute migration in the Gateway API 1.5 upgrade). There is no version fallback: clusters must install Gateway API >= 1.6 CRDs to use TCP/UDP routes.

**How it works**

- Bumps `sigs.k8s.io/gateway-api` (and `/conformance`) to v1.6.0.
- Migrates all TCPRoute/UDPRoute handling (watches, lists, status updates, event handlers) from the `v1alpha2` types to the `v1` types.
- CRD detection now requires the `v1` versions of TCPRoute/UDPRoute and disables `NLBGatewayAPI` (when the feature flags are defaulted) if they are not served.
- Updates e2e resources and docs examples to `v1`; the L4 gateway docs note the Gateway API >= 1.6 requirement.

**Note on the store changes:** client-go v0.36 (pulled in by the gateway-api v1.6.0 bump) added `LastStoreSyncResourceVersion()` and `Bookmark()` to the `cache.Store` interface, so the two custom store wrappers (`pkg/aga/resource_store.go`, `pkg/k8s/secrets_store.go`) gain matching delegation methods.

**Note on e2e:** the TCP/UDP e2e tests require Gateway API >= 1.6 CRDs installed on the test cluster.

### Checklist
- [x] Manually tested


zac-nixon:
Thanks for the PR. I'll let the tests run first, then give the code a look!

zac-nixon:
Most looks good, thanks! I see other Gateway API implementations just going ahead and bumping to the v1 version of TCP / UDP routes without a fallback. I think that this is the correct way to handle it, and it avoids a messy conversion adapter. What do you think about removing the adapter?

nicomazzarello:
Thanks for the review, @zac-nixon! A hard cut to v1 was actually my first instinct too.

The reason I went with the fallback is that TCPRoute/UDPRoute existed in v1alpha2 for some time, so there are still plenty of clusters running Gateway API ≤ 1.5 CRDs where only v1alpha2 is served. With a hard cut, upgrading the controller on those clusters either auto-disables NLBGatewayAPI (when the flags are defaulted, via the CRD detection) or fails the TCP/UDP watches (when the flags are set explicitly) until the CRDs are upgraded to 1.6.

That said, the migration path is easy (the v1alpha2 and v1 schemas are identical), and Gateway API CRDs 1.6 experimental channel still serves v1alpha2. If you're comfortable with that requirement (CRD detection disabling the feature when v1 isn't served), I'm happy to remove the adapter and the rest of the fallback code. It would be a big clean-up and shrink the PR considerably.

zac-nixon:
I think the hard-cut is the clean approach, as it's too burdensome to make the controller work with every version of the Gateway API, we are trying to make one controller release == one gateway api release.

I tried with the Github UI to make the merge clean for you from our recent package upgrade from non-versioned => versioned but it seems I missed a spot. I apologize about that.

nicomazzarello:
No worries, I'll go ahead with the hard-cut and fix the branch. I'll mention you when it's done

nicomazzarello:
Done @zac-nixon. The fallback and the adapter are removed — both route types are now handled through the v1 API only, and CRD detection requires the v1 kinds (disabling NLBGatewayAPI when they're not served).

I'd like your opinion on what I did with the tests — let me know if you'd have gone differently:

- crddetect: detection only queries v1 now, so fixtures with kinds under v1alpha2 are inputs production can never produce. I rewrote the pre-1.6 case as what such a cluster actually looks like to the detector (v1 served without TCPRoute/UDPRoute → NLB disabled), and dropped the alpha2-only case (equivalent to "nothing present").
- pkg/k8s DetectCRDs: the multi-group-version test now queries what production actually queries (gateway.networking.k8s.io/v1 + gateway.k8s.aws/v1beta1). I deliberately kept v1alpha2 in TestDetectCRDs_v1_ignoreNotInInterestSet as the ignored group version — ignoring it is exactly what the controller relies on when a pre-1.6 or experimental-channel cluster serves it.

Tested with the unit suite plus a live EKS cluster on Gateway API 1.6 CRDs: v1 watches clean, feature detection correct, and route status written at v1 end to end.

zac-nixon (approved):

zac-nixon:
Looks perfect! Thanks for making sure to update the documentation too.

Appreciate the hardwork here.

/lgtm /approved

Issue #4836, opened 2026-07-14
ACM certificate management: Problem with Hostname wildcards


mathstlouis:
**Bug Description**

We have been testing the new [Certificate Management](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/ingress/certificate_management/) feature to issue certificates using [AWS PCA](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/ingress/certificate_management/#pca-support). In our staging environments, we use [hostname wildcards](https://kubernetes.io/docs/concepts/services-networking/ingress/#hostname-wildcards) in Ingress objects. There seems to be a bug when setting ACM tags because of the "*" character. Probably because of the "ingress.k8s.aws/resource" tag value that is set from the hostname.

Error logs:
```
operation error ACM: RequestCertificate, https response error StatusCode: 400, RequestID: 01543adb-d3ff-491a-a967-bc27dc7c6116, 
api error ValidationException: 1 validation error detected: Value of the input at 'tags.2.member.value' failed to satisfy constraint:
Member must satisfy regular expression pattern: ([\\p{L}\\p{Z}\\p{N}_.:/=+\\-@]*)
```

**Steps to Reproduce**

- Enable certificate management feature gate with required AWS IAM permissions.
- Create ingress object with annotations "alb.ingress.kubernetes.io/acm-pca-arn" and "alb.ingress.kubernetes.io/create-acm-cert" set. Set one spec.rules[].host with a leading wildcard ex.: "*.foo.com".

**Expected Behavior**
Certificate created with PCA and synced to AWS ACM. 

**Actual Behavior**
Creation in ACM fails with error.

**Environment**
- AWS Load Balancer controller version: 3.4.1
- Kubernetes version: 1.35.6
- Using EKS (yes/no), if so version?: yes, eks.15
- Using Service or Ingress: ingress
- AWS region: ca-central-1

**Contribution Intention (Optional)**

- [ ] Yes, I'm willing to submit a PR to fix this issue
- [x] No, I cannot work on a PR at this time

the-technat:
I'd say this open PR should address this issue: #4786.

shraddhabang:
@the-technat is right. We will be merging the open PR which should fix this issue.

PR #4846, opened 2026-07-18
fix: sanitize wildcard hostnames in certificate resource ID for ACM tags

Dasmat13:
### Description
This pull request resolves a bug where ACM certificate requests fail for Ingresses with wildcard hostnames due to invalid character restrictions in AWS tag values (#4836).

The certificate `resourceID` embeds the `DomainName` and is used as the `ingress.k8s.aws/resource` tracking tag value. However, AWS ACM API validation rejects `*` in tag values (`[\p{L}\p{Z}\p{N}_.:/=+\-@]*`), causing `RequestCertificate` calls to fail with a `400 ValidationException` for wildcard host configurations.

This change sanitizes the `resourceID` by replacing the `*` character with `"wildcard"` in `buildCertificateResourceID()`. The actual `DomainName` and `SubjectAlternativeNames` sent to ACM remain unchanged, ensuring certificate coverage is unaffected while preventing the validation failure.

### Testing
- Added unit tests covering all wildcard/non-wildcard/multi-wildcard cases for `buildCertificateResourceID()`.
- Added an E2E test case in `test/e2e/ingress/acm_ingress_test.go` exercising wildcard-host Ingress certificate validation.
- All unit tests (`go test ./pkg/...`) pass successfully.

### DIFF
--- a/pkg/ingress/model_build_certificates_test.go
+++ b/pkg/ingress/model_build_certificates_test.go
@@ -285,11 +285,18 @@ func Test_buildCertificateResourceID(t *testing.T) {
 			ObjectMeta: metav1.ObjectMeta{Namespace: "awesome-ns", Name: "ing-wild"},
 		},
 	}
-	got := task.buildCertificateResourceID(&acmModel.CertificateSpec{
+	got1 := task.buildCertificateResourceID(&acmModel.CertificateSpec{
 		Type:       acmtypes.CertificateTypeAmazonIssued,
 		DomainName: "*.app.example.com",
 	}, ing)
+	assert.NotContains(t, got1, "*")
+	assert.Contains(t, got1, "wildcard.app.example.com")
 
-	assert.NotContains(t, got, "*")
-	assert.Contains(t, got, "wildcard.app.example.com")
+	// Case 2: Non-wildcard domain
+	got2 := task.buildCertificateResourceID(&acmModel.CertificateSpec{
+		Type:       acmtypes.CertificateTypeAmazonIssued,
+		DomainName: "app.example.com",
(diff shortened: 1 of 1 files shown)


pkg/ingress/model_build_certificates_test.go
```diff
@@ -248,3 +277,36 @@ func Test_buildACMCertificates(t *testing.T) {
...
+
+	// Case 1: Single wildcard
+	got1 := task.buildCertificateResourceID(&acmModel.CertificateSpec{
+		Type:       acmtypes.CertificateTypeAmazonIssued,
+		DomainName: "*.app.example.com",
+	}, ing)
+	assert.NotContains(t, got1, "*")
+	assert.Contains(t, got1, "wildcard.app.example.com")
+
+	// Case 2: Multiple wildcards
```
dominikhei: Since multi-level wildcards aren't supported (at least to my knowledge), would this case ever occur in practice?

Dasmat13: yeah true, multi-level wildcards aren't valid for ACM anyway. I just threw that in as a quick defensive check to make sure the string replace handles multiple asterisks properly. Happy to remove that test case if u prefer keeping it strictly valid formats!

Dasmat13:
@dominikhei Removed the multi-level wildcard test case as suggested and rebased on latest `main`! Thanks for the review.

the-technat:
Isn't this a duplicate of #4786?

Dasmat13:
Looks like #4786 was merged and covers this fix. Closing this as a duplicate. 
Thanks

Issue #4727, opened 2026-05-06
cross account role assume


FernandoMiguel:
**Describe the feature you are requesting**
create ACM validation route53 records in another aws account route53

**Motivation**
EKS cluster with R53 zones in multiple accounts

**Describe the proposed solution you'd like**
Thank you so much for the new feature to be able to handle ACM issuing.
The LBC is only able to create ACM validation records in the same aws account R53 where the LB exists.
But we have R53 zones across multiple accounts.
We would need the LBC to be able to assume a list of roles in other accounts, roles that would have IAM privs to update that own account R53 zones.

**Describe alternatives you've considered**
we currently issue a small list of Certs with terraform and pass those ARNs to the LBC.
but this makes it very hard to create sub.sub.sub domain certs.

the-technat:
Just to confirm: you are talking about Amazon issued certificates right? Not using PCA... (cause there no Route53 records are needed). 

I've also realized that this is missing while adopting this feature this morning in our environment. I think it shouldn't be too difficult to pass in another IAM role to assume for Route53 route creation. But if I'm reading your description carefully it seems like you'd need multiple IAM roles depending on the DNS zone that needs to get updated?

That might become a bit more involved since then we'd need some sort of mapping to pass into the controller which DNS zone needs to get updated with what IAM role...

FernandoMiguel:
@the-technat this is for public certs issued by ACM, not the PCA.
some other ACKs do support multi role, so LBC could borrow that pattern
AWS calls it `CARM`
https://github.com/aws-controllers-k8s/route53-controller/blob/f52bddffa26e4b86423b28933bdf852c83928ce4/helm/values.yaml#L177-L193

Docs: https://github.com/aws-controllers-k8s/community/blob/main/docs/design/proposals/carm/cross-account-resource-management.md

shraddhabang:
Thanks for the feature request! I think this is valid request.  I agree with @the-technat. The approach would be a domain-to-role mapping configuration where you specify which IAM role to assume for Route53 operations based on the domain. The controller would then assume the appropriate cross-account role when creating DNS validation records, similar to how we already handle [cross-account](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v3.3/guide/targetgroupbinding/targetgroupbinding/?h=ass#assumerole-cross-account-targetgroups) target registration on TargetGroupBindings.

[...]

PR #4847, opened 2026-07-19
fix: use public Route 53 zone for Amazon-issued ACM DNS validation in split-horizon DNS

### LINKED ISSUE #4840: create-acm-cert selects a private hosted zone for public ACM DNS validation in split-horizon Route 53; certificate stays in PENDING_VALIDATION
**Bug Description**

Thanks for the certificate-management feature - auto-provisioning ACM certs from an Ingress is really handy. We ran into an edge case with split-horizon Route 53 and wanted to share it with a couple of possible directions.

When create-acm-cert provisions an Amazon-issued (public) ACM certificate, the controller resolves the DNS validation record into the most-specific matching Route 53 hosted zone. In a split-horizon setup (a private zone that is a subdomain of a public zone), that most-specific zone is the private one. Public ACM validation is checked over public DNS, so a record placed in a private-only zone is never resolvable, and the certificate remains in PENDING_VALIDATION - the HTTPS listener never finalizes and the Ingress gets no address.

**Steps to Reproduce**

- Route 53 in one account with a private hosted zone internal.example.com and a public hosted zone example.com.
- Enable the EnableCertificateManagement feature gate; create an Ingress with host app.internal.example.com and annotation alb.ingress.kubernetes.io/create-acm-cert: "true" (no certificate-arn).
- Controller logs show it building the model, requesting the certificate, and creating the validation record — the log lines include only the record name/value, not a zone:
"msg":"created validation record","record":{"Name":"_<hash>.app.internal.example.com.","Type":"CNAME","Value":"_<hash>.xxxxxx.acm-validations.aws."}
- The validation CNAME ends up in the private internal.example.com zone.
- The ACM certificate stays PENDING_VALIDATION and the Ingress never receives an address.

**Expected Behavior**

For a public (Amazon-issued) certificate, the validation record should be created in a hosted zone where public ACM validation can actually succeed - a public zone - or the controller should allow the user to specify which zone to use.

**Actual Behavior**

The validation record is created in the private zone, where public ACM validation can never complete, so the certificate stays PENDING_VALIDATION indefinitely and the affected Ingress never serves HTTPS. It does not stop the controller overall. Reproducible Always in a split-horizon topology.

**Regression**
Was the functionality working correctly in a previous version ? [Yes / No]
If yes, specify the last version where it worked as expected
No - this appears to be the behavior since the certificate-management feature was introduced (PR #4554). Not a regression from a previously working version.

**Current Workarounds**

Pre-issue the certificate (DNS-validated by placing the CNAME in the public zone) and attach it via the certificate-arn annotation, which bypasses the controller's Route 53 logic entirely.

**Environment**
- AWS Load Balancer controller version: v3.3.0
- Kubernetes version: 1.35 (EKS)
- Using EKS (yes/no), if so version?: Yes, 1.35
- Using Service or Ingress: Ingress
- AWS region: eu-central-1
  the upstream chart is rendered to static manifests (Tanka helm template) and applied via ArgoCD (GitOps), so helm ls/helm get values don't apply. Equivalent render command (chart v3.3.0 from https://aws.github.io/eks-charts):
    
    ```bash
    helm template aws-load-balancer-controller eks/aws-load-balancer-controller \
      --version 3.3.0 \
      --namespace alb \
      --set clusterName=<cluster-name> \
      --set region=eu-central-1 \
      --set vpcId=<vpc-id> \
      --set enableCertManager=true \
      --set controllerConfig.featureGates.EnableCertificateManagement=true \
      --set serviceAccount.create=true \
      --set serviceAccount.name=aws-load-balancer-controller \
      --set serviceAccount.annotations."eks\.amazonaws\.com/role-arn"=<controller-iam-role-arn>
      Controller image: public.ecr.aws/eks/aws-load-balancer-controller:v3.3.0.
    ```

    <details>
    
    <summary>kubectl -n <controllernamespace> describe deployment aws-load-balancer-controller</summary>
    
    ```yaml
    Name:                   aws-load-balancer-controller
    Namespace:              alb
    CreationTimestamp:      Wed, 20 May 2026 14:54:03 
    Labels:                 app.kubernetes.io/instance=aws-load-balancer-controller
                            app.kubernetes.io/managed-by=x
                            app.kubernetes.io/name=aws-load-balancer-controller
                            app.kubernetes.io/version=v3.3.0
                            helm.sh/chart=aws-load-balancer-controller-3.3.0
    Annotations:            argocd.argoproj.io/tracking-id: x:apps/Deployment:alb/aws-load-balancer-controller
                            deployment.kubernetes.io/revision: 5
    Selector:               app.kubernetes.io/instance=aws-load-balancer-controller,app.kubernetes.io/name=aws-load-balancer-controller
    Replicas:               2 desired | 2 updated | 2 total | 2 available | 0 unavailable
    StrategyType:           RollingUpdate
    MinReadySeconds:        0
    RollingUpdateStrategy:  25% max unavailable, 25% max surge
    Pod Template:
      Labels:           app.kubernetes.io/instance=aws-load-balancer-controller
                        app.kubernetes.io/name=aws-load-balancer-controller
      Annotations:      automation.x/restartedAt: 2026-07-05T07:07:27Z
                        kubectl.kubernetes.io/restartedAt: 2026-05-24T15:05:12+03:00
                        prometheus.io/port: x
                        prometheus.io/scrape: true
      Service Account:  aws-load-balancer-controller
      Containers:
       aws-load-balancer-controller:
        Image:       public.ecr.aws/eks/aws-load-balancer-controller:v3.3.0
        Ports:       x/TCP (webhook-server), x/TCP (metrics-server)
        Host Ports:  x/TCP (webhook-server), x/TCP (metrics-server)
        Args:
          --cluster-name=x
          --ingress-class=x
          --aws-region=eu-central-1
          --aws-vpc-id=x
          --feature-gates=ALBGatewayAPI=true,EnableCertificateManagement=true
        Liveness:     http-get http://:x/x delay=Xs timeout=Xs period=Xs #success=x #failure=x
        Readiness:    http-get http://:x/readyz delay=Xs timeout=Xs period=Xs #success=x #failure=x
        Environment:  <none>
        Mounts:
          /x/x/x from cert (ro)
      Volumes:
       cert:
        Type:               Secret (a volume populated by a Secret)
        SecretName:         x
        Optional:           false
      Priority Class Name:  x
      Node-Selectors:       <none>
      Tolerations:          <none>
    Conditions:
      Type           Status  Reason
      ----           ------  ------
      Progressing    True    NewReplicaSetAvailable
      Available      True    MinimumReplicasAvailable
    OldReplicaSets:  aws-load-balancer-controller-69d7854d6c (0/0 replicas created), aws-load-balancer-controller-d7f6989b6 (0/0 replicas created), aws-load-balancer-controller-5c6ffd6844 (0/0 replicas created), aws-load-balancer-controller-856d9dc7dc (0/0 replicas created)
    NewReplicaSet:   aws-load-balancer-controller-fcb5cfb5 (2/2 replicas created)
    Events:          <none>
    ```
    
    
    </details>

From reading the code, GetHostedZoneID (pkg/aws/services/route53.go) matches the hosted zone by DNS suffix and doesn't currently consult zone.Config.PrivateZone; listHostedZones returns both public and private zones. PR #4785 ("match by longest suffix", v3.4.1) keeps the most-specific-zone behavior, so it doesn't change this case (apologies if I missed an existing option). Two directions that would close the gap for public certificates:

1. Explicit hosted-zone override — let the user specify the Route 53 hosted zone ID used for ACM validation (e.g. an Ingress annotation alb.ingress.kubernetes.io/acm-validation-hosted-zone-id, or an IngressClassParams field). Deterministic; covers split-horizon, delegated subdomains, and cross-account DNS where suffix inference alone isn't enough.
2. Public-zone-aware automatic selection — for Amazon-issued certificates, restrict candidate zones to public hosted zones and walk up to the nearest public ancestor. This keeps the existing "most specific matching zone" behavior, narrowed to zones that can satisfy public validation; it handles the common split-horizon case with no per-Ingress config.

Either resolves the reported case; both together (auto default + override) would be ideal. It would also help to fail fast with a clear error when no public zone matches and no override is given, rather than create a certificate that can't be validated and leave it stuck.

Happy to help test or contribute a PR if a direction sounds reasonable. 

- [X] Yes, I'm willing to submit a PR to fix this issue

**Additional Context**

Reference: AWS - [Issue an ACM certificate for a domain in a private hosted zone](https://repost.aws/knowledge-center/acm-certificate-domain-private-host-zone) (public-cert validation records must live in a public zone). This affects anyone with split-horizon DNS or overlapping private/public zones.

niv1612:
### Issue

Fixes #4840

### Description

The `create-acm-cert` (`EnableCertificateManagement`) feature writes the ACM DNS validation CNAME into the most-specific matching Route 53 hosted zone. In split-horizon DNS — a private zone that is a subdomain of a public zone — that most-specific zone is the private one. ACM validates Amazon-issued (public) certificates over public DNS, so the record never resolves, the certificate stays in `PENDING_VALIDATION`, and the HTTPS listener never finalizes.

**How it works**

- Adds `GetPublicHostedZoneID` to the `Route53` service, reusing the existing longest-suffix matcher but skipping private zones (`zone.Config.PrivateZone`).
- The ACM create path (`CreateWithValidationRecords`) now selects the validation zone from public zones only (nearest public ancestor). If no public zone matches, it fails fast with an actionable error pointing at the `certificate-arn` annotation, instead of requesting a certificate that hangs in `PENDING_VALIDATION`.
- The delete path attempts cleanup in both the most-specific zone (records written by controller versions predating this fix) and the public zone (records written after it), so validation records are not orphaned across the upgrade.
- The private CA (`acm-pca-arn`) path is unaffected — the synthesizer routes it to `Create`, which never touches validation records.

Unit tests cover `GetPublicHostedZoneID` (split-horizon selects the public zone, private-only fails fast, longest-suffix among public zones) and the split-horizon delete cleaning records in both zones; existing synthesizer create/delete expectations are updated. `docs/guide/ingress/certificate_management.md` documents public-zone selection and the fail-fast fallback to `certificate-arn`.

Manually tested end-to-end against live Route 53 and ACM: split-horizon ingresses wrote the validation CNAME to the public zone and the certificate reached `ISSUED`; an ingress under a private-only domain failed fast before any certificate or record was created; and deleting the ingresses removed the auto-provisioned certificate and its validation record while leaving unrelated records in the shared zone untouched.

### Checklist
- [x] Manually tested

### DIFF
--- a/docs/guide/ingress/certificate_management.md
+++ b/docs/guide/ingress/certificate_management.md
@@ -35,7 +35,7 @@ This reconciliation might fail if there are no certificates present to autodisco
 ### Certificate Validation
 
 Amazon Issued certificates are currently validated using DNS Method and Route53 records. Validation can take [up to 30 minutes](https://docs.aws.amazon.com/acm/latest/userguide/dns-validation.html). 
-E-Mail validation is not supported due to significant higher delays between requesting a certificate and it's issuance. 
+E-Mail validation is not supported due to significant higher delays between requesting a certificate and its issuance. 
 When using a PCA, certificates don't have to be validated.
 
 Because ACM validates Amazon-issued certificates over **public** DNS, the controller writes the validation record into the nearest-ancestor **public** Route53 hosted zone. In split-horizon setups (a private zone that is a subdomain of a public zone), the private zone is skipped so the record lands where ACM can resolve it. If no public hosted zone matches the domain (private-only domain, or the public parent lives in an account the controller can't see), the controller fails fast. In that case, pre-create the certificate yourself and reference it with the [`certificate-arn`](annotations.md#certificate-arn) annotation.
--- a/docs/guide/ingress/certificate_management.md
+++ b/docs/guide/ingress/certificate_management.md
@@ -38,6 +38,8 @@ Amazon Issued certificates are currently validated using DNS Method and Route53 
 E-Mail validation is not supported due to significant higher delays between requesting a certificate and it's issuance. 
 When using a PCA, certificates don't have to be validated.
 
+Because ACM validates Amazon-issued certificates over **public** DNS, the controller writes the validation record into the nearest-ancestor **public** Route53 hosted zone. In split-horizon setups (a private zone that is a subdomain of a public zone), the private zone is skipped so the record lands where ACM can resolve it. If no public hosted zone matches the domain (private-only domain, or the public parent lives in an account the controller can't see), the controller fails fast. In that case, pre-create the certificate yourself and reference it with the [`certificate-arn`](annotations.md#certificate-arn) annotation.
+
 ## Ingress Group Behavior
 
 When using certificate management with [IngressGroups](ingress_class.md#specgroup), each ingress in the group gets its own certificate based on its own hostnames. All certificates are attached to the shared ALB's HTTPS listener.
--- a/pkg/deploy/acm/certificate_manager.go
+++ b/pkg/deploy/acm/certificate_manager.go
@@ -83,7 +83,7 @@ func (c *defaultCertificateManager) CreateWithValidationRecords(ctx context.Cont
 		if _, checked := hostedZoneByDomain[host]; checked {
 			continue
 		}
-		zoneID, err := c.route53Client.GetHostedZoneID(ctx, host)
+		zoneID, err := c.route53Client.GetPublicHostedZoneID(ctx, host)
 		if err != nil {
 			return nil, fmt.Errorf("pre-check failed for domain %q: %w", host, err)
 		}
@@ -226,6 +226,12 @@ func (c *defaultCertificateManager) DeleteWithValidationRecords(ctx context.Cont
 
 	for _, opts := range desc.Certificate.DomainValidationOptions {
 		if opts.ValidationMethod == acmtypes.ValidationMethodDns {
+			if opts.ResourceRecord == nil {
+				// ACM populates ResourceRecord asynchronously; a certificate deleted right
+				// after creation may not have one yet, so there is no record to clean up.
+				c.logger.Info("no resource record on validation option, skipping validation record cleanup", "domain", awssdk.ToString(opts.DomainName))
+				continue
+			}
 			c.logger.Info("deleting validation records for certificate", "certificateARN", arn)
 			id, err := c.route53Client.GetHostedZoneID(ctx, awssdk.ToString(opts.DomainName))
 			if err != nil {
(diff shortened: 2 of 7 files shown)


docs/guide/ingress/certificate_management.md
```diff
@@ -35,9 +35,11 @@ This reconciliation might fail if there are no certificates present to autodisco
 ### Certificate Validation
 
 Amazon Issued certificates are currently validated using DNS Method and Route53 records. Validation can take [up to 30 minutes](https://docs.aws.amazon.com/acm/latest/userguide/dns-validation.html). 
-E-Mail validation is not supported due to significant higher delays between requesting a certificate and it's issuance. 
+E-Mail validation is not supported due to significant higher delays between requesting a certificate and its issuance. 
 When using a PCA, certificates don't have to be validated.
 
+Because ACM validates Amazon-issued certificates over **public** DNS, the controller writes the validation record into the nearest-ancestor **public** Route53 hosted zone. In split-horizon setups (a private zone that is a subdomain of a public zone), the private zone is skipped so the record lands where ACM can resolve it. If no public hosted zone matches the domain (private-only domain, or the public parent lives in an account the controller can't see), the controller fails fast. In that case, pre-create the certificate yourself and reference it with the [`certificate-arn`](annotations.md#certificate-arn) annotation.
```
the-technat, line 41: Just to note: there is another FR for cross-account records: #4727, but that doesn't affect the status quo as you mentioned in the docs.

pkg/deploy/acm/certificate_manager.go
```diff
@@ -226,6 +226,12 @@ func (c *defaultCertificateManager) DeleteWithValidationRecords(ctx context.Cont
 
 	for _, opts := range desc.Certificate.DomainValidationOptions {
 		if opts.ValidationMethod == acmtypes.ValidationMethodDns {
+			if opts.ResourceRecord == nil {
```
the-technat, line 229: Good catch! Noticed this one too in some of my tests, should make it more clear what's going on instead of writing misleading logs.

the-technat (approved):
From the perspective of the original feature author I'd say this looks good and is reasonable. Over to the maintainers for the final verdict 😁.

Issue #4805, opened 2026-06-30
Frontend NLB and external-dns


omallo:
**Describe the feature you are requesting**
When fronting an ALB with a NLB using the `alb.ingress.kubernetes.io/enable-frontend-nlb: "true"` annotation, the AWS Load Balancer Controller creates 2 entries in the ingress status, one for the ALB and one for the NLB. When using the External DNS Controller on AWS, this does not work properly since External DNS (or actually AWS) only creates a DNS record for the first entry. When using a frontend NLB, it would be more natural to create a DNS record only for the NLB since this is what is used from the outside. This is why I would like to request a feature in the AWS Load Balancer Controller, so that only the NLB is added to the ingress status.

**Motivation**
I believe that creating a DNS record for the NLB only is the most typical use case.

**Describe the proposed solution you'd like**
To avoid the feature being a breaking change, I would propose to have a new annotation (flag) which can be set so that only the NLB is added to the ingress status. That way, the External DNS controller would work correctly.

**Describe alternatives you've considered**
The bug had already been reported on the [External DNS Controller project](https://github.com/kubernetes-sigs/external-dns/issues/5661) but it has been deemed specific to the AWS Load Balancer Controller, which is why I'm reporting it here. As described in that bug, a workaround consists in using the `external-dns.alpha.kubernetes.io/target` annotation with the NLB hostname as the target. This works but has the issue that the NLB hostname is not predictable, which means that the annotation needs to be added in a second step after the NLB has been created, which is not very convenient.

dominikhei:
A simple fix would be a new annotation on the ingress manifest, e.g. alb.ingress.kubernetes.io/frontend-nlb-status-only: "true" that causes the controller to only write the NLB hostname to ingress.status.loadBalancer.ingress. 

I'm happy to look into the implementation but would appreciate input from someone more familiar with the codebase on whether this approach makes sense.

PR #4849, opened 2026-07-23
Added frontend-nlb-status-only annotation to fix ExternalDNS record creation when using a frontend NLB

dominikhei:
### Issue

#4805 

The original issue was surfaced in [external-dns#5661](https://github.com/kubernetes-sigs/external-dns/issues/5661), where it was decided that the fix belongs in the AWS Load Balancer Controller rather than ExternalDNS

### Description

When frontend NLB is configured via `alb.ingress.kubernetes.io/enable-frontend-nlb: "true"`, the controller writes both the ALB and NLB hostnames to status.loadBalancer.ingress. When using ExternalDNS on AWS, only a DNS record for the first entry is created, meaning the NLB, which is the actual external-facing endpoint, may not get a DNS record at all.

The existing workaround (external-dns.alpha.kubernetes.io/target) requires a two-step deploy because the NLB hostname isn't known until after provisioning, which makes it impractical for automated workflows.

This PR adds a new boolean annotation:
`alb.ingress.kubernetes.io/frontend-nlb-status-only: "true"`

When set, only the NLB hostname is written to status.loadBalancer.ingress, so ExternalDNS creates a record for the NLB only. The annotation is opt-in to avoid a breaking change. If the NLB DNS is not yet available, the controller falls back to the ALB hostname to avoid leaving status empty.

**Validation:**

Manually validated on EKS: ingress with the annotation shows 1 status entry (NLB only) and a control ingress without it shows 2 entries (ALB + NLB).

### Checklist
- [x] Manually tested

### DIFF
--- a/controllers/ingress/group_controller.go
+++ b/controllers/ingress/group_controller.go
@@ -314,8 +314,9 @@ func (r *groupReconciler) updateIngressStatus(ctx context.Context, lbDNS string,
 
 	ingOld := ing.DeepCopy()
 	if frontendNlbStatusOnly && frontendNlbDNS != "" {
-		// Only write the NLB hostname so that tools like ExternalDNS create DNS
-		// records pointing at the NLB rather than both the ALB and the NLB.
+		// Only write the NLB hostname to status. When using ExternalDNS on AWS,
+		// only the first status entry gets a DNS record, which would be the ALB,
+		// not the NLB that clients should actually reach.
 		if len(ing.Status.LoadBalancer.Ingress) != 1 ||
 			ing.Status.LoadBalancer.Ingress[0].Hostname != frontendNlbDNS {
 			ing.Status.LoadBalancer.Ingress = []networking.IngressLoadBalancerIngress{
--- a/controllers/ingress/group_controller.go
+++ b/controllers/ingress/group_controller.go
@@ -90,7 +90,9 @@ func NewGroupReconciler(cloud services.Cloud, k8sClient client.Client, eventReco
 		stackMarshaller:   stackMarshaller,
 		stackDeployer:     stackDeployer,
 		backendSGProvider: backendSGProvider,
+
 		secretsManager:    secretsManager,
+		annotationParser:  annotationParser,
 
 		groupLoader:           groupLoader,
 		groupFinalizerManager: groupFinalizerManager,
@@ -114,6 +116,7 @@ type groupReconciler struct {
 	stackDeployer     deploy.StackDeployer
 	backendSGProvider networkingpkg.BackendSGProvider
 	secretsManager    k8s.SecretsManager
+	annotationParser  annotations.Parser
 
 	groupLoader           ingress.GroupLoader
 	groupFinalizerManager ingress.FinalizerManager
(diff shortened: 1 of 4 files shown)


docs/guide/ingress/annotations.md
```diff
@@ -1450,3 +1451,36 @@ When this option is set to true, the controller will automatically provision a N
         ```
         alb.ingress.kubernetes.io/frontend-nlb-attributes: dns_record.client_routing_policy=availability_zone_affinity
         ```
+
+- <a name="frontend-nlb-status-only">`alb.ingress.kubernetes.io/frontend-nlb-status-only`</a> controls whether only the frontend NLB hostname is written to `status.loadBalancer.ingress` instead of both the ALB and NLB hostnames.
+
+    By default, when a frontend NLB is enabled, both the ALB and NLB hostnames appear in the ingress status. When using [ExternalDNS](https://github.com/kubernetes-sigs/external-dns) on AWS, only a DNS record for the first entry is created (see [external-dns#5661](https://github.com/kubernetes-sigs/external-dns/issues/5661)), which means the NLB does not get a DNS record. Setting this annotation to `"true"` restricts the status to only the NLB hostname, so ExternalDNS creates a DNS record pointing at the NLB.
+
+    !!!note
+        This annotation has no effect if `alb.ingress.kubernetes.io/enable-frontend-nlb` is not set, or if the NLB has not yet been provisioned. In those cases the ALB hostname is written to status as normal.
```
zac-nixon: I wonder if we just want to block setting any hostname if you have this flag set, until the nlb is provisioned? that way there is no weird host flip / flop when you're provisioning an nlb -> alb stack.

dominikhei: The change would be pretty simple from a code perspective. I don't see any downside apart from the fact that if anything is waiting on ingress status to be populated before considering the ingress ready could be blocked for longer. But since the current behavior would populate it with the ALB hostname, they would be unblocking on the wrong address anyway...

dominikhei: Made the changes

controllers/ingress/group_controller.go
```diff
@@ -306,25 +309,41 @@ func (r *groupReconciler) updateIngressStatus(ctx context.Context, lbDNS string,
...
+		// only the first status entry gets a DNS record. Writing only the NLB hostname
+		// ensures ExternalDNS creates a record pointing at the NLB, not the ALB.
+		if len(ing.Status.LoadBalancer.Ingress) != 1 ||
+			ing.Status.LoadBalancer.Ingress[0].Hostname != frontendNlbDNS {
+			ing.Status.LoadBalancer.Ingress = []networking.IngressLoadBalancerIngress{
+				{Hostname: frontendNlbDNS},
+			}
+		}
+		// In case frontendNlbStatusOnly && frontendNlbDNS == "" nothing gets applied until the NLB is provisioned
+	} else if !frontendNlbStatusOnly {
```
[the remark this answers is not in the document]
dominikhei: This was delibaretly designed that way, when the NLB isn't provisioned yet, nothing is written to status. With the suggested change, the ALB hostname would be written in that window, which means ExternalDNS would create a DNS record pointing at the ALB. Once that record exists you'd have to clean it up, so it's safer to have no record at all until the NLB is ready

dominikhei:
@wweiwei-li I reviwed copilots proposed changes, feel free to have a look again :)

controllers/ingress/group_controller.go
```diff
@@ -306,25 +309,43 @@ func (r *groupReconciler) updateIngressStatus(ctx context.Context, lbDNS string,
 		}
 	}
 
+frontendNlbStatusOnly := false
```
wweiwei-li: can you run `gofmt` ? , the indentation doesn't seem right

wweiwei-li:
Also, can you click the failing EasyCLA check , and complete the sign flow ? It is required check

dominikhei:
> Also, can you click the failing EasyCLA check , and complete the sign flow ? It is required check

the missing CLA is for the copilot commit. I tried signing it with my name but it does not change.

controllers/ingress/group_controller.go
dominikhei: `gofmt -w` didn't change the file, so I assume formatting is good. Visually the nesting also seems to be correct to me.

PR #4853, opened 2026-07-26
fix(gateway): use TCP target group protocol for TLSRoute on passthrough NLB listeners

### LINKED ISSUE #4556: [Gateway API] gateway.k8s.aws/nlb listeners with `protocol: TLS` and `tls.mode: Passthrough` appear to build a TLS listener
**Bug Description**

Gateways of class `gateway.k8s.aws/nlb` whose listeners use `protocol: TLS` and `tls.mode: Passthrough` appear to build a TLS listener.

**Steps to Reproduce**

- Create a Gateway with `gatewayClassName` whose manifest contains `controllerName: gateway.k8s.aws/nlb`
- Add a listener of with `protocol: TLS` and `tls.mode: Passthrough`

**Expected Behavior**

I expected a _TCP_ listener on the AWS NLB side because traffic/streams should proxy to backends without TLS termination (raw streams)

**Actual Behavior**

AWS LB Controller log outputs show a model with a TLS listener attempting to be built which fails due to no discoverable/matching certificates (required for AWS NLB TLS listeners).

**Regression**
Was the functionality working correctly in a previous version ? [No]

**Current Workarounds**

**Environment**
- AWS Load Balancer controller version: v3.0.0
- Kubernetes version: v1.34.1
- Using EKS (yes/no), if so version?: Platform version eks.9
- Using Service or Ingress: `Gateway`
- AWS region: `us-west-2`
  - If helm was used then please show output of `helm -n <controllernamespace> get values <helmreleasename>`
  - If helm was not used, then copy/paste the exact command used to install the controller, including flags and options.
  - `kubectl -n <controllernamespace> describe deployment aws-load-balancer-controller`
  - `kubectl describe ingressclasses`
  - `kubectl -n <appnamespace> describe ingress <ingressname>`
  - `kubectl -n <appnamespace> describe svc <servicename>`

- [x] No, I cannot work on a PR at this time

**Additional Context**

pujitha24:
Motivation:
Issue #4556 reported that a Gateway API NLB listener with `protocol: TLS`
and `tls.mode: Passthrough` incorrectly built a TLS listener. That part
was already fixed in v3.1.0 by downgrading the listener protocol to TCP
for passthrough mode.

However, follow-up reports on the same issue (from 2026-03-04 and
2026-03-10) show the fix was incomplete: when a `TLSRoute` is bound to
such a passthrough listener, the controller still infers the *target
group* protocol as TLS. This produces a real AWS API error when
reconciling the load balancer:

  IncompatibleProtocols: The listener and the following target groups
  have incompatible protocols

because the (now TCP) listener cannot be attached to a TLS target group.
This is a functional failure, not cosmetic: the Gateway never reaches a
working state and reconciliation errors repeatedly.

Approach:
`inferTargetGroupProtocolFromRoute` in
pkg/gateway/model/model_build_target_group.go always returned
`ProtocolTLS` for `TLSRouteKind` on an NLB, regardless of the actual
listener protocol. It now takes the (already passthrough-downgraded)
listener protocol and returns `ProtocolTCP` when the listener protocol
is TCP, keeping `ProtocolTLS` for the `tls.mode: Terminate` case. Users
who explicitly set `targetGroupProps.Protocol` are unaffected, since
that explicit-override path is checked before the inference path.

This only changes NLB behavior for TLSRoute; ALB target group protocol
inference (which cannot take this NLB-only branch) is unaffected.

Validation:
go build ./...
go test -count=1 ./pkg/gateway/model/...
Both pass, including a corrected unit test case in
Test_buildTargetGroupProtocol that had been asserting the buggy
(mismatched) behavior.

Fixes #4556

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/pkg/gateway/model/model_build_target_group.go
+++ b/pkg/gateway/model/model_build_target_group.go
@@ -355,15 +355,15 @@ func (builder *targetGroupBuilderImpl) buildTargetGroupIPAddressType(backendConf
 func (builder *targetGroupBuilderImpl) buildTargetGroupProtocol(targetGroupProps *elbv2gw.TargetGroupProps, route routeutils.RouteDescriptor, listenerProtocol elbv2model.Protocol) (elbv2model.Protocol, error) {
 	// TODO - Not convinced that this is good, maybe auto detect certs == HTTPS / TLS.
 	if builder.loadBalancerType == elbv2model.LoadBalancerTypeApplication {
-		return builder.buildL7TargetGroupProtocol(targetGroupProps, route)
+		return builder.buildL7TargetGroupProtocol(targetGroupProps, route, listenerProtocol)
 	}
 
 	return builder.buildL4TargetGroupProtocol(targetGroupProps, route, listenerProtocol)
 }
 
-func (builder *targetGroupBuilderImpl) buildL7TargetGroupProtocol(targetGroupProps *elbv2gw.TargetGroupProps, route routeutils.RouteDescriptor) (elbv2model.Protocol, error) {
+func (builder *targetGroupBuilderImpl) buildL7TargetGroupProtocol(targetGroupProps *elbv2gw.TargetGroupProps, route routeutils.RouteDescriptor, listenerProtocol elbv2model.Protocol) (elbv2model.Protocol, error) {
 	if targetGroupProps == nil || targetGroupProps.Protocol == nil {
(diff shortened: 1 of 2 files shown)


pujitha24:
/assign @zac-nixon

This is rebased and conflict-free whenever you get a chance to take a look.

PR #4860, opened 2026-07-27
docs: clarify scheme change replacement

kappa8219:
Clarify that changing `alb.ingress.kubernetes.io/scheme` replaces the Application Load Balancer and requires traffic-migration planning.

### DIFF
--- a/docs/guide/ingress/annotations.md
+++ b/docs/guide/ingress/annotations.md
@@ -698,7 +698,7 @@ Access control for LoadBalancer can be controlled with following annotations:
         ```
 
     !!!warning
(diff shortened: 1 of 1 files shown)


kappa8219:
I stumbled this situation while changing aws alb managed gateway from internet-facing to internal. It is important to be aware it is imppossible to do it inplace.

kappa8219:
@zac-nixon @shraddhabang please look

shraddhabang:
Thanks for adding the warning — this is useful information for users. One suggestion: please remove the CloudFormation docs link. This controller doesn't use CloudFormation, so linking to the CFN TemplateReference is confusing.

shraddhabang (approved):
/lgtm

PR #4741, opened 2026-05-15
Fix error when updating health check settings in TGC for Gateway API ALB target groups

shraddhabang:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4740

### Description

Two bugs in `isSDKTargetGroupRequiresReplacementDueToNLBHealthCheck`:
1.  `isL4TargetGroup` switch had empty case bodies that did not fall through (Go semantics), so only TCP_UDP returned true
2.  Guard logic was inverted — skipped replacement for NLB and forced replacement for ALB

### Checklist
- [x] Added tests that cover your change (if possible)
- [ ] Added/modified documentation as required (such as the `README.md`, or the `docs` directory)
- [x] Manually tested
- [x] Made sure the title of the PR is a good description that can go into the release notes

### BONUS POINTS checklist: complete for good vibes and maybe prizes?! :exploding_head:
- [x] Backfilled missing tests for code in same general area :tada:
- [x] Refactored something and made the world a better place :star2:

Issue #4740, opened 2026-05-15
LBC is trying to re-create target groups when `TargetGroupConfiguration` `defaultConfiguration` is updated.


maksym-iv-ef:
**Bug Description**
LBC is trying to re-create target groups when `TargetGroupConfiguration` `defaultConfiguration` is updated when using Gateway API (L7, multiple HTTP and GRPC routes)

**Steps to Reproduce**
1. Provision GTW with multiple HTTP and GRPC routes and use `TargetGroupConfiguration`
    ```yaml
    apiVersion: gateway.k8s.aws/v1beta1
    kind: TargetGroupConfiguration
    metadata:
      annotations:
        argocd.argoproj.io/tracking-id: bar-foo-app-ingest:gateway.k8s.aws/TargetGroupConfiguration:foo-app/foo-app-ingest
      labels:
        app: foo-app-ingest
        app.kubernetes.io/instance: foo-app-ingest
        app.kubernetes.io/managed-by: Helm
        app.kubernetes.io/name: foo-app
        app.kubernetes.io/version: 0.1.0
        cloud: aws
        env: stage
        helm.sh/chart: foo-app-0.1.0
        region: us-east-2
      name: foo-app-ingest
      namespace: foo-app
    spec:
      defaultConfiguration:
        healthCheckConfig:
          healthCheckPath: /readyz
          healthCheckPort: "8080"
          healthCheckProtocol: HTTP
        protocol: HTTPS
        targetGroupAttributes:
        - key: deregistration_delay.timeout_seconds
          value: "30"
        targetType: ip
      targetReference:
        group: ""
        kind: Service
        name: foo-app-ingest
    ```
2. Change `spec.defaultConfiguration.healthCheckConfig.healthCheckProtocol`
3. Observe the LBC logs or GTW events.

**Expected Behavior**
Target Group params are updated

**Actual Behavior**
The error is thrown, apparently, LBC is trying to re-create the target group
```
Failed deploy model due to operation error Elastic Load Balancing v2: CreateTargetGroup, https response error StatusCode: 400, RequestID: 09b8a56d-e39f-4a86-888f-fbd55929e8cd, DuplicateTargetGroupName: A target group with the same name 'k8s-foo-fooin-19ad0c01af' exists, but with different settings
```

**Environment**
- AWS Load Balancer controller version: v3.3.0
- Kubernetes version: 1.35
- Using EKS (yes/no), if so version?: 1.35
[...]

PR #4876, opened 2026-08-11
fix(gateway): avoid DuplicateTargetGroupName for Services with multiple named targetPorts

legal90:
### Issue

It's related to the issue #4740 and the fix #4741, which fixed a *different* cause of the same error (health-check-driven target group replacement in the synthesizer). This PR fixes a separate, build-time cause: target group **name** collisions when one Service is exposed on multiple named `targetPort`s. No dedicated issue filed yet, I can open one if preferred.

### Description

For IP target groups, `GetTargetGroupPort` returns a constant `1` for any named `targetPort`:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/0f415d98b80d0178c5c0630ef0d27cf27cb293a3/pkg/gateway/routeutils/backend_service.go#L86-L87

That port is the only port-derived input to the name hash in `buildTargetGroupName`:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/0f415d98b80d0178c5c0630ef0d27cf27cb293a3/pkg/gateway/model/model_build_target_group.go#L331

So when a single Service is exposed on N named `targetPort`s through one route, all N backends hash to the **same** target group name and `CreateTargetGroup` fails with:

```
DuplicateTargetGroupName: A target group with the same name '...' exists, but with different settings
```
which aborts the entire gateway reconcile. The in-model resource ID already stays unique because it uses `GetIdentifierPort()`, so N distinct target groups are built but collapse to one name.

### Fix 
For non-Instance targets, we should hash the per-backend identifier port instead of the resolved TG port. 
For numeric ports `identifierPort.String()` equals `strconv.Itoa(tgPort)`, so existing names are unchanged. Instance targets keep hashing the NodePort.

### Compatibility / churn
- Numeric IP targets and all Instance targets: name unchanged — no target group churn.
- Only IP targets using named `targetPort`s change. Multi named-port Services are currently broken (no stable name to preserve). A single named-port IP Service would see a one-time target group re-creation on upgrade.

### Checklist
- [x] Added/modified documentation as required (no-op)
- [x] Manually tested

### DIFF
--- a/pkg/gateway/model/model_build_target_group.go
+++ b/pkg/gateway/model/model_build_target_group.go
@@ -283,11 +283,12 @@ func (builder *targetGroupBuilderImpl) buildTargetGroupSpec(gw *gwv1.Gateway, ro
 		return elbv2model.TargetGroupSpec{}, err
 	}
 	tgPort := backendConfig.GetTargetGroupPort(targetType)
+	identifierPort := backendConfig.GetIdentifierPort()
 	targetControlPort, err := builder.buildTargetControlPort(targetGroupProps, tgProtocol, targetType)
 	if err != nil {
 		return elbv2model.TargetGroupSpec{}, err
 	}
-	name := builder.buildTargetGroupName(targetGroupProps, k8s.NamespacedName(gw), route.GetRouteNamespacedName(), route.GetRouteKind(), backendConfig.GetBackendNamespacedName(), tgPort, targetType, tgProtocol, tgProtocolVersion, targetControlPort)
+	name := builder.buildTargetGroupName(targetGroupProps, k8s.NamespacedName(gw), route.GetRouteNamespacedName(), route.GetRouteKind(), backendConfig.GetBackendNamespacedName(), tgPort, identifierPort, targetType, tgProtocol, tgProtocolVersion, targetControlPort)
 
 	if tgPort == 0 {
 		if targetType == elbv2model.TargetTypeIP {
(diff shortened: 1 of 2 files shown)


zac-nixon:
Can you include a unit test that validates this claim:
```
For non-Instance targets, we should hash the per-backend identifier port instead of the resolved TG port.
For numeric ports identifierPort.String() equals strconv.Itoa(tgPort), so existing names are unchanged. Instance targets keep hashing the NodePort.
```

legal90:
@zac-nixon Sure! I updated asserts in `Test_buildTargetGroupName_namedPortUniqueness`, so now they cover all 4 aspects:
- Named ports stay unique (collision fix)
- Non-Instance target names ignore the resolved TG port (new assertion)
- Numeric IP targets keep their historical name (regression guard)
- Instance targets hash the NodePort, independent of the identifier port

zac-nixon:
Fix looks good, and I validated it in my own cluster. I wanted to checkout this statement:

```
Only IP targets using named targetPorts change. Multi named-port Services are currently broken (no stable name to preserve). A single named-port IP Service would see a one-time target group re-creation on upgrade.
```

It turns out we did the right thing in the stack tags, https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/gateway/model/model_build_target_group.go#L630-L636 which means that this change is not disruptive to configurations that already work. This is because the target group name is just cosmetic after creation, and we utilize the resource tags instead, which are correct because of the code I linked.

PR #4878, opened 2026-08-13
feat(gateway custom): allow importers to customize gateway names

zac-nixon:
### Description

Allows importers to change the value of ALB / NLB Gateway controllers, so they can run their own version of the controller.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/gateway/constants/controller_constants.go
+++ b/pkg/gateway/constants/controller_constants.go
@@ -3,10 +3,14 @@ package constants
 import "k8s.io/apimachinery/pkg/util/sets"
 
 /*
-Common constants
+Common constants -- These are assigned as a variable so that importers can change these variables at runtime.
 */
 
 var (
+	// ALBGatewayController gateway controller name for ALB
+	ALBGatewayController = "gateway.k8s.aws/alb"
+	// NLBGatewayController gateway controller name for NLB
+	NLBGatewayController     = "gateway.k8s.aws/nlb"
 	FullGatewayControllerSet = sets.New(ALBGatewayController, NLBGatewayController)
 )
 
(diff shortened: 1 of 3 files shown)


pkg/gateway/constants/controller_constants.go
```diff
@@ -7,6 +7,10 @@ Common constants
 */
 
 var (
+	// ALBGatewayController gateway controller name for ALB
+	ALBGatewayController = "gateway.k8s.aws/alb"
+	// NLBGatewayController gateway controller name for NLB
+	NLBGatewayController     = string("gateway.k8s.aws/nlb")
```
wweiwei-li: nit: no need to have string() wrapper

pkg/gateway/constants/controller_constants.go
wweiwei-li: `ALBRouteResourceGroupVersion` also need to be removed

pkg/gateway/constants/controller_constants.go
```diff
@@ -7,6 +7,10 @@ Common constants
 */
 
 var (
+	// ALBGatewayController gateway controller name for ALB
```
wweiwei-li, line 10: do we need give a comment here to make  the intent explicit ( like Allows importers to change the value of ALB / NLB Gateway controllers, so they can run their own version of the controller)

zac-nixon: done :)

pkg/gateway/constants/controller_constants.go
zac-nixon: done :)

wweiwei-li (approved):

shraddhabang (approved):
/lgtm

PR #4879, opened 2026-08-14
fix: handle DeletedFinalStateUnknown tombstones in TypedInformer's DeleteFunc

ginbear:
### Description
`TypedInformer[T].DeleteFunc` asserts the deleted object directly to `T`:

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/v3.5.0/pkg/k8s/typed_informer.go#L46-L49

When the informer misses the real delete event — for example right after a watch reconnect — and only notices the object is gone on the next relist, client-go delivers a `cache.DeletedFinalStateUnknown` tombstone instead of the real object (this is documented on `cache.ResourceEventHandler` itself). The bare type assertion panics on that tombstone. Because apimachinery's crash handler re-panics after logging ("Observed a panic"), this crashes the whole controller process instead of staying isolated to the informer goroutine.

```
panic: interface conversion: interface {} is cache.DeletedFinalStateUnknown, not *k8s.PodInfo [recovered, repanicked]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x7f6c78c2c635]

goroutine 232 [running]:
k8s.io/apimachinery/pkg/util/runtime.handleCrash({0x7f6c7ba73a58, 0x1db2739f9c20}, {0x7f6c7b508480, 0x1db2739f9b60}, {0x0, 0x0, 0x7f6c75e0397a?})
	/go/pkg/mod/k8s.io/apimachinery@v0.36.2/pkg/util/runtime/runtime.go:116 +0x1a9
panic({0x7f6c7b508480?, 0x1db2739f9b60?})
	/usr/local/go/src/runtime/panic.go:860 +0x13a
sigs.k8s.io/aws-load-balancer-controller/v3/pkg/k8s.(*TypedInformer[...]).Start.func3()
	/workspace/pkg/k8s/typed_informer.go:47 +0x65
k8s.io/client-go/tools/cache.ResourceEventHandlerDetailedFuncs.OnDelete(...)
	/go/pkg/mod/k8s.io/client-go@v0.36.2/tools/cache/controller.go:345
k8s.io/client-go/tools/cache.(*processorListener).run.func1(...)
	/go/pkg/mod/k8s.io/client-go@v0.36.2/tools/cache/shared_informer.go:1365 +0x35d
```

Since only the Pod informer for TargetGroupBinding readiness gates goes
through `TypedInformer`, this is currently the only affected watch.

### What changed

- Unwrap the tombstone before asserting to `T`.
- If the object still doesn't match `T` after unwrapping, log via
  `HandleErrorWithContext` and skip instead of panicking.

### Checklist
- [x] Added tests that cover your change
- [ ] Added/modified documentation as required
- [x] Manually tested


zac-nixon:
Hey, thanks for the debugging here. I thought about this issue a bit, and I don't even think we need the delete events as those don't impact the target group binding, because we are watching endpoint[slice]s too. Would you mind making the delete function empty (e.g. {}).

ginbear:
Thanks a lot for taking the time to look into this and for the clear explanation — that makes much more sense than my original approach! Updated: DeleteFunc is now a no-op, and I removed the tombstone-unwrapping helper (and its test) since it's no longer needed.

zac-nixon:
Thanks!

zac-nixon (approved):

shraddhabang (approved):
/lgtm

PR #4881, opened 2026-08-16
Make pod informer transform idempotent so WatchList streaming sync does not fall back to full LIST

yash97:
### Issue

N/A. Happy to file one if maintainers prefer tracking this as an issue first.

### Description

The pod informer transform (`podInfoConverter`) only accepts `*corev1.Pod` and returns an error for anything else. That breaks streaming initial sync.

Since client-go v0.35 the `WatchListClient` feature is enabled by default, and this repo pins v0.36.2, so the controller attempts a WatchList streaming sync (KEP-3157) for the pod informer on every start. During the stream, the reflector applies the informer transform to objects in its temporary store, then the destination FIFO applies the same transform again on `Replace`. The second application hands the converter its own output, a `*PodInfo`. The converter returns "expect pod object", the streaming sync aborts, and the reflector silently falls back to a conventional LIST.

The fallback is expensive: it decodes every pod in the cluster into memory at once before the transform can slim anything down. On small clusters nobody notices. On a cluster with roughly 380k pods (kwok scale testing), the controller could not finish initial sync inside a 20Gi memory limit and crash looped on OOM, logging `problem wait for podInfo repo sync` on every attempt.

The fix is a two-line early return: when the transform receives a `*PodInfo`, hand it back unchanged. With an idempotent transform the streaming sync completes and peak memory during sync stays close to one decoded pod at a time, instead of the whole cluster.

Testing was done in three layers:

1. The included unit test applies the transform twice and asserts the second application returns the same `*PodInfo` instead of an error.
2. Wire-level verification against a real kube-apiserver (v1.34, envtest, `WatchList=true` feature gate), with a recording `http.RoundTripper` on the pods endpoint. Before the fix the request log shows the streaming attempt followed by the fallback:
   ```
   req[0]: allowWatchBookmarks=true&resourceVersionMatch=NotOlderThan&sendInitialEvents=true&watch=true
   req[1]: limit=500&resourceVersion=0        <- fallback LIST
   ```
   After the fix, req[0] is the only sync request and the informer syncs from the stream.
3. Peak-heap comparison at 20,000 seeded pods (each a few KB, with realistic labels, annotations, env vars), sampling `HeapInuse` every 20ms through initial sync:
   ```
   before: peak 483MiB   (fallback full LIST)
   after:  peak  36MiB   (streaming sync)
   ```

The envtest harness for layers 2 and 3 is not part of this PR to keep it focused, since it needs kubebuilder assets in CI. I can contribute it in a follow-up if there is interest.

On API servers without WatchList support, client-go falls back to a plain LIST exactly as it does today, so behavior there is unchanged.

### Checklist
- [x] Manually tested


zac-nixon (approved):

shraddhabang (approved):
/lgtm

yash97:
Two housekeeping notes:

1. **Rebased out the Copilot Autofix commit** (`d9f27e8`) and re-applied the same change as `cd41d22` -- the autofix commit was authored with an email EasyCLA does not recognize, which turned the CLA check red. The tree is byte-identical; only authorship changed. @zac-nixon thanks for the test hardening -- kept it as-is, just re-attributed so the CLA check passes.

2. **The govulncheck failure is pre-existing on `main`**, not from this PR: the 2026-08-13 Go security release fixed seven stdlib vulns that the pinned Go 1.26.5 toolchain still contains, so every PR currently fails that job. One-line fix in #4883 (`.go-version` -> 1.26.6). Once that merges, a re-run here should go green without further changes.

yash97:
Update: folded the Go 1.26.6 toolchain bump into this PR directly (commit `f9edf17`) instead of the separate #4883, so this PR is self-contained and CI can go green on its own. The bump fixes the seven stdlib vulns govulncheck flags (2026-08-13 Go security release); `.go-version` is the single source for workflows, GOTOOLCHAIN, and the build image.

PR #4882, opened 2026-08-17
Added logic to the target_group_synthesizer to ensure the controller can recover when the maximum unique target groups per ALB is reached

### LINKED ISSUE #4848: Controller cannot recover when maximum unique target groups per ALB is reached
**Bug Description**

The AWS Load Balancer Controller is unable to reconcile an IngressGroup after the Application Load Balancer reaches the maximum number of unique target groups.

This appears to be similar to #4373, but in this case the limiting resource is not the number of listener rules. The controller reaches the AWS quota of 100 unique target groups associated with a single Application Load Balancer.

Once the quota is reached, the controller repeatedly attempts to create a new listener rule referencing an additional target group. The reconciliation fails before obsolete listener rules and target groups are removed.

As a result, adding, updating, or deleting Ingress resources in the affected IngressGroup no longer converges automatically.

***Error Message***
`{"level":"info","ts":"2026-06-23T09:00:58Z","logger":"controllers.ingress","msg":"creating listener rule","stackID":"act-iu-eu-alb","resourceID":"443:3"}

{"level":"error","ts":"2026-06-23T09:00:58Z","msg":"Reconciler error","controller":"ingress","object":{"name":"xxx-iu-eu-alb"},"namespace":"","name":"xxx-iu-eu-alb","reconcileID":"1a47778b-67b3-4480-860a-981c8747b28b","error":"failed to create listener rule: operation error Elastic Load Balancing v2: CreateRule, https response error StatusCode: 400, RequestID: f18fac52-719a-4f11-ba24-a02f940eb045, TooManyUniqueTargetGroupsPerLoadBalancer: You have reached the maximum number of unique target groups that you can associate with a load balancer of type 'application': [100]"}`

**Steps to Reproduce**
- Create an Application Load Balancer managed by the AWS Load Balancer Controller.
- Configure multiple Ingress resources to use the same alb.ingress.kubernetes.io/group.name.
- Configure the Ingress resources so that they create separate target groups.
- Increase the number of Ingress resources and target groups until the ALB has 100 unique target groups associated with it.
- Add another Ingress or modify an existing Ingress so that another target group or listener rule must be created.
- Delete one or more existing Ingress resources that should cause older listener rules and target groups to be removed.
- Observe that reconciliation continues to fail with TooManyUniqueTargetGroupsPerLoadBalancer.

**Expected Behavior**
When the ALB has reached the maximum number of unique target groups, the controller should still be able to process deletions and updates that reduce the number of target groups.

For a model containing both deletions and creations, the controller should preferably:

Delete obsolete listener rules.
Remove obsolete target groups from the load balancer.
Wait until AWS no longer considers those target groups associated with the ALB, if required.
Create the new target groups and listener rules.

Alternatively, the controller should detect that the target-group quota has been reached and prioritize cleanup operations before attempting resource creation.

Deleting an Ingress should always allow the controller to recover from the quota condition without manual changes in AWS.

**Actual Behavior**
The controller attempts to create a listener rule before sufficient obsolete listener rules or target groups have been removed.

AWS rejects the request with:

TooManyUniqueTargetGroupsPerLoadBalancer:
You have reached the maximum number of unique target groups that you can associate with a load balancer of type 'application': [100]

The complete model deployment fails. Consequently, planned deletions are apparently not executed either.

The affected IngressGroup remains stuck in a reconciliation loop.

**Current Workarounds**
We manually deleted approximately 30 listener rules from the affected ALB within AWS Console.

After manually freeing sufficient capacity, the controller was able to continue reconciliation. It then processed the pending deletions and successfully deployed the model:

{"level":"info","ts":"2026-06-23T09:11:49Z","logger":"controllers.ingress","msg":"successfully deployed model","ingressGroup":"xxx-iu-eu-alb"}

No controller restart was required.

**Environment**
- AWS Load Balancer controller version: v2.17.1  
- Kubernetes version: 1.33
- Using EKS (yes/no), if so version?: v1.33.13-eks-8f14419
- Using Service or Ingress: Ingress
- AWS region: eu-central-1

`helm` -n kube-system ls  | grep -i aws-load-balancer-controller
xxx-iu-aws-lb-ctlr            	kube-system	12      	2026-04-14 10:02:36.183285012 +0000 UTC	deployed	aws-load-balancer-controller-1.17.1        	v2.17.1  

Potential approaches could include:

* Execute destructive cleanup operations before create operations when an AWS quota is reached.
* Split reconciliation into cleanup and creation phases.
* Retry model deployment after performing safe deletions.
* Detect TooManyUniqueTargetGroupsPerLoadBalancer and prioritize obsolete listener-rule and target-group deletion.
* Calculate the expected peak number of associated target groups during model deployment.
* Avoid an intermediate state that requires more target groups than both the current and desired state.
* Improve the error event to indicate which Ingress resources or target groups contribute to the quota.

- [X] No, I cannot work on a PR at this time

**Additional Context**
This appears closely related to #4373.

Issue #4373 describes a deadlock when the listener-rule quota is reached. This issue demonstrates similar reconciliation behavior when the quota for unique target groups per ALB is reached.
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4373 

The common behavior seems to be that the controller attempts resource creation before processing deletions required to free capacity.

dominikhei:
### Issue

Fixes #4848

### Description

Previously stale target groups were deleted by `PostSynthesize` as target groups can't be deleted while listener rules still reference them, so deletion is deferred until after the listener rule synthesizer has already cleaned up its stale rules. However if `Synthesize` hits the quota error and returns early, `PostSynthesize` never runs, so the stale TGs never get deleted, so the slot never gets freed, so the next reconcile hits the exact same error again.

This PR introduces logic to fix this by adding an inline delete to `Synthesize`. When the quota error is hit, instead of returning immediately, one stale TG is deleted within `Synthesize` to free a slot and retry the create in the same reconcile cycle. Because the inline delete runs before the listener rule synthesizer has cleaned up stale rules in the current cycle, the delete may fail with `ResourceInUse` on the first attempt. In that case the controller requeues. On the next cycle the listener rules from the previous cycle are already gone, so the stale TG delete succeeds and the create goes through / the current deadlock is resolved. The controller recovers in at most two reconcile cycles instead of being stuck forever.

Compared to the `ListenerRuleSynthesizer` this one for target groups handles quota logic inside `Synthesize` on the struct and not via a separate function. Due to this there needs to be the full environment that `Synthesize` expects
and tests for the new functionality became a bit complex.  

**Note:** The logic for deleting and creating TGs could be made easier to understand and tests could be simplified by rewriting the range loop as an index-based loop with the same three-branch structure as the listener rule synthesizer. Would be interested on some thoughts on that?

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/deploy/elbv2/target_group_synthesizer.go
+++ b/pkg/deploy/elbv2/target_group_synthesizer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	awssdk "github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/smithy-go"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -62,13 +63,39 @@ func (s *targetGroupSynthesizer) Synthesize(ctx context.Context) error {
 	// * unmatched targetGroups might still be use by a listener rule.
 	s.unmatchedSDKTGs = unmatchedSDKTGs
 
+	// sdkTGDeleteIndex tracks how many unmatchedSDKTGs were already deleted inline
+	sdkTGDeleteIndex := 0
 	for _, resTG := range unmatchedResTGs {
 		tgStatus, err := s.tgManager.Create(ctx, resTG)
 		if err != nil {
-			return err
+			if !isTooManyUniqueTargetGroupsErr(err) {
+				return err
+			}
+			// Listener rules are always deleted prior to target groups, so
+			// any old TG that is no longer needed should already be dissociable.
+			// Delete one stale TG to free a slot, then retry the create.
+			// Due to eventual consistency the listener-rule dissociation may not be
+			// fully reflected yet. In that case the delete will error and the
+			// controller will requeue and retry.
+			if sdkTGDeleteIndex >= len(s.unmatchedSDKTGs) {
+				return errors.Wrap(err, "unable to synthesize target groups: TooManyUniqueTargetGroupsPerLoadBalancer and no unused target groups available to delete")
+			}
+			s.logger.V(1).Info("TooManyUniqueTargetGroupsPerLoadBalancer detected, deleting unused target group to free quota",
+				"tgArn", awssdk.ToString(s.unmatchedSDKTGs[sdkTGDeleteIndex].TargetGroup.TargetGroupArn))
+			if delErr := s.tgManager.Delete(ctx, s.unmatchedSDKTGs[sdkTGDeleteIndex]); delErr != nil {
+				return delErr
+			}
+			sdkTGDeleteIndex++
+			// Retry the create now that a slot has been freed.
+			tgStatus, err = s.tgManager.Create(ctx, resTG)
+			if err != nil {
+				return err
+			}
 		}
 		resTG.SetStatus(tgStatus)
 	}
+	// Trim the TGs already deleted so PostSynthesize does not attempt them again.
+	s.unmatchedSDKTGs = s.unmatchedSDKTGs[sdkTGDeleteIndex:]
 	for _, resAndSDKTG := range matchedResAndSDKTGs {
 		tgStatus, err := s.tgManager.Update(ctx, resAndSDKTG.resTG, resAndSDKTG.sdkTG)
 		if err != nil {
@@ -214,6 +241,14 @@ func isL4TargetGroup(protocol elbv2model.Protocol) bool {
 	default:
 		return false
 	}
+}
+
+func isTooManyUniqueTargetGroupsErr(err error) bool {
+	var apiErr smithy.APIError
+	if errors.As(err, &apiErr) {
+		return apiErr.ErrorCode() == "TooManyUniqueTargetGroupsPerLoadBalancer"
+	}
+	return false
 }
 
 func isSDKTargetGroupTargetControlPortDrifted(tgSpec elbv2model.TargetGroupSpec, sdkTG TargetGroupWithTags) bool {
--- a/pkg/deploy/elbv2/target_group_synthesizer_test.go
+++ b/pkg/deploy/elbv2/target_group_synthesizer_test.go
@@ -1,10 +1,13 @@
 package elbv2
 
 import (
+	"context"
 	"testing"
 
 	awssdk "github.com/aws/aws-sdk-go-v2/aws"
 	elbv2types "github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2/types"
+	"github.com/aws/smithy-go"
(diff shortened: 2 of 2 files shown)


zac-nixon:
P1 — The PR handles the error at the wrong AWS operation. The new code catches TooManyUniqueTargetGroupsPerLoadBalancer from tgManager.Create(), which calls CreateTargetGroup. But AWS does not return that error from CreateTargetGroup; its documented quota error is TooManyTargetGroups, the account-level TG quota. TooManyUniqueTargetGroupsPerLoadBalancer is produced when a TG becomes associated with the load balancer, e.g. through CreateRule—which is exactly where issue #4848 reports the failure. So in the reported production scenario, this new branch is never reached. The new test masks this by having mockTargetGroupManager.Create() synthesize an error that the real CreateTargetGroup API doesn't produce.
P1 — The claimed two-reconcile recovery path cannot work when the TG is still referenced. The PR says that if inline target-group deletion gets ResourceInUse, the next reconcile will find that the listener rules from the previous cycle are already gone. But TargetGroupSynthesizer runs before ListenerRuleSynthesizer, and the stack deployer immediately returns when any Synthesize() call errors. Therefore, if the inline TG delete fails with ResourceInUse, listener-rule synthesis never runs in that reconcile. On the next reconcile you reach the same TG first and fail again. There is no mechanism in this patch that makes the stale listener rule disappear between those attempts.

dominikhei:
> P1 — The PR handles the error at the wrong AWS operation. The new code catches TooManyUniqueTargetGroupsPerLoadBalancer from tgManager.Create(), which calls CreateTargetGroup. But AWS does not return that error from CreateTargetGroup; its documented quota error is TooManyTargetGroups, the account-level TG quota. TooManyUniqueTargetGroupsPerLoadBalancer is produced when a TG becomes associated with the load balancer, e.g. through CreateRule—which is exactly where issue #4848 reports the failure. So in the reported production scenario, this new branch is never reached. The new test masks this by having mockTargetGroupManager.Create() synthesize an error that the real CreateTargetGroup API doesn't produce. P1 — The claimed two-reconcile recovery path cannot work when the TG is still referenced. The PR says that if inline target-group deletion gets ResourceInUse, the next reconcile will find that the listener rules from the previous cycle are already gone. But TargetGroupSynthesizer runs before ListenerRuleSynthesizer, and the stack deployer immediately returns when any Synthesize() call errors. Therefore, if the inline TG delete fails with ResourceInUse, listener-rule synthesis never runs in that reconcile. On the next reconcile you reach the same TG first and fail again. There is no mechanism in this patch that makes the stale listener rule disappear between those attempts.

Thank you for the detailed feedback. I will have a look at it once I am back from holidays on Sunday!

Issue #4899, opened 2026-09-08
NLB Service: global `preserve_client_ip.enabled=false` is applied to UDP/TCP_UDP target groups, ELB rejects it and no new listener is deployed


KR-Ravindra:
**Bug Description**

For a `LoadBalancer` Service that mixes TCP and UDP ports and sets the global annotation
`service.beta.kubernetes.io/aws-load-balancer-target-group-attributes: preserve_client_ip.enabled=false`,
the controller copies the attribute onto every target group, including the UDP / TCP_UDP ones. ELB rejects that
attribute for UDP-family target groups, so `ModifyTargetGroupAttributes` fails right after `CreateTargetGroup`,
the target group synthesizer returns the error and the stack deploy aborts before the listener synthesizer runs.
On every reconcile the Service gets:

```
Warning  FailedDeployModel  ...  Failed deploy model due to InvalidConfigurationRequest: preserve_client_ip.enabled cannot be false if the target group protocol is UDP or TCP_UDP
```

Any listener that was not already there (the UDP one, and any TCP port added in the same change) is never created,
while previously created listeners keep working. From the outside the new ports are just dark; the only hint is the
recurring event. The controller already encodes the ELB rule (comment in
`pkg/service/model_build_target_group.go` L724-L728) but only uses it for security-group rule generation, not for
the attribute it sends to ELB.

**Steps to Reproduce**

- Step-by-step:
  1. Apply the Service below with only the TCP port; wait for the NLB and listener to be ready.
  2. Add the UDP port (or apply the full manifest from scratch).
  3. `kubectl describe svc mixed` shows the `FailedDeployModel` event above on every reconcile; the UDP listener
     (and any other listener added in the same change) never appears in the AWS console.

- Manifest:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: mixed
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-type: external
    service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
    service.beta.kubernetes.io/aws-load-balancer-scheme: internal
    service.beta.kubernetes.io/aws-load-balancer-target-group-attributes: preserve_client_ip.enabled=false
spec:
  type: LoadBalancer
  selector: {app: demo}
  ports:
  - {name: tcp-app, port: 7447,  protocol: TCP, targetPort: 7447}
  - {name: udp-app, port: 47998, protocol: UDP, targetPort: 47998}
```

[...]

PR #4900, opened 2026-09-08
Force preserve_client_ip.enabled=true for UDP-family NLB target groups

KR-Ravindra:
### Issue

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4899

## Problem

A `LoadBalancer` Service with mixed TCP and UDP ports and the global annotation
`service.beta.kubernetes.io/aws-load-balancer-target-group-attributes: preserve_client_ip.enabled=false`
never gets its UDP listener (or any other listener added in the same change). ELB refuses that attribute for
UDP-family target groups, `ModifyTargetGroupAttributes` fails right after `CreateTargetGroup`, and the stack
deploy aborts before the listener synthesizer runs. The Service only shows, on every reconcile:

```
Warning  FailedDeployModel  Failed deploy model due to InvalidConfigurationRequest: preserve_client_ip.enabled cannot be false if the target group protocol is UDP or TCP_UDP
```

## Root cause

`pkg/service/model_build_target_group.go`:

- `buildTargetGroup` (L106) calls `buildTargetGroupAttributes` without the target group protocol, and the resulting
  attributes go verbatim into `TargetGroupSpec.TargetGroupAttributes` (L114, L154) for every target group.
- `buildTargetGroupAttributes` (L327-L394) merges global and per-port attributes with no protocol awareness; the only
  validation of `preserve_client_ip.enabled` is `strconv.ParseBool` in `validateAndParseAttributes` (L307-L325).
- The ELB rule that client IP preservation can't be disabled for UDP, QUIC, TCP_QUIC and TCP_UDP target groups is
  already encoded at L724-L730, but only for security group rule generation.

## Fix

- `buildTargetGroupAttributes` now takes `tgProtocol`. For UDP, TCP_UDP, QUIC and TCP_QUIC target groups a
  `preserve_client_ip.enabled=false` attribute is forced to `true` with an Info log naming the service, port and
  protocol. TCP/TLS target groups are untouched, so the global annotation keeps disabling client IP preservation
  where ELB allows it and a mixed-protocol Service deploys.
- Because the forced value equals what ELB reports for those target groups, the attribute reconciler diff is empty
  and no `ModifyTargetGroupAttributes` call is made for that key.
- `t.preserveClientIP` (derived from the same attributes in `buildPreserveClientIPFlag`) is now also correct for
  UDP-family target groups.
- New helper `isClientIPPreservationForced(tgProtocol)`.
- Docs: note under the client IP preservation example in `docs/guide/service/annotations.md`.

Issue #4899 offered two options (clear validation error, or force `true`). This implements the second so that an
existing mixed-protocol Service with the global attribute starts deploying instead of failing differently; switching
to a returned error is a two-line change in the same block if maintainers prefer that.

## How tested

Table-driven cases added to `Test_defaultModelBuilderTask_targetGroupAttrs` in
`pkg/service/model_build_target_group_test.go`: TCP keeps `false`; UDP and TCP_UDP with `false` get `true`; UDP with
`true` unchanged; per-port `false` on the TCP port leaves the UDP target group without the attribute.

Before the fix (new cases run against `main`, ff98381):

```
--- FAIL: Test_defaultModelBuilderTask_targetGroupAttrs (0.00s)
    --- FAIL: Test_defaultModelBuilderTask_targetGroupAttrs/preserve_client_ip.enabled=false_is_forced_to_true_for_UDP_target_groups (0.00s)
        model_build_target_group_test.go:509:
                Error:          Not equal:
                expected: []elbv2.TargetGroupAttribute{{Key:"preserve_client_ip.enabled", Value:"true"}, {Key:"proxy_protocol_v2.enabled", Value:"false"}}
                actual  : []elbv2.TargetGroupAttribute{{Key:"preserve_client_ip.enabled", Value:"false"}, {Key:"proxy_protocol_v2.enabled", Value:"false"}}
    --- FAIL: Test_defaultModelBuilderTask_targetGroupAttrs/preserve_client_ip.enabled=false_is_forced_to_true_for_TCP_UDP_target_groups (0.00s)
        (same diff)
FAIL    sigs.k8s.io/aws-load-balancer-controller/v3/pkg/service 0.026s
```

After the fix:

```
--- PASS: Test_defaultModelBuilderTask_targetGroupAttrs (0.00s)
--- PASS: Test_defaultModelBuilder_buildPreserveClientIPFlag (0.00s)
ok      sigs.k8s.io/aws-load-balancer-controller/v3/pkg/service 0.279s
```

`go test ./pkg/service/...`, `go vet ./...`, `gofmt -l pkg/service/` and `go build ./...` are clean.
Not manually tested against a live NLB.

## Links

- https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/4899
- https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/guide/service/annotations.md (target group attributes / client IP preservation)

### Checklist
- [ ] Manually tested

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.


KR-Ravindra (commented):
Round 1 self-review.

1. Checked `pkg/service/model_build_target_group.go` on main: `buildTargetGroup` calls `buildTargetGroupAttributes` without the protocol (L106), the attributes go into the target group spec unmodified, and the ELB rule that client IP preservation cannot be disabled for UDP/TCP_UDP/QUIC/TCP_QUIC only exists in the security-group code (L724-730). The description matches the code.
2. Ran `go test ./pkg/service/ -run 'Test_defaultModelBuilderTask_targetGroupAttrs|Test_defaultModelBuilder_buildPreserveClientIPFlag'` on this branch: pass. The new helper only changes the attribute for the four UDP-family protocols and only when the parsed value is `false`; TCP/TLS paths are untouched.
3. Forcing `true` rather than returning an error is a design choice the issue left open; the description says so and the error variant is a two-line change in the same block if maintainers prefer it.

Marking ready so a member can `/ok-to-test`.

KR-Ravindra:
Closing: per the discussion on #4899, the per-port target-group-attributes annotation is the intended solution and the controller should not rewrite the attribute.

PR #4909, opened 2026-09-17
Fix TGB networking with empty ports

### LINKED ISSUE #4907: If len(rule.Ports) == 0, the result of computeIngressPermissionsForTGBNetworking is not used, causing missing SGR
**Bug Description**

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/networking/networking_manager.go#L403

This line shadows `permissions`, unlike in the `range.rulePorts` branch, which has the effect that if no port is set, no rule is created, when the intended behaviour is for a rule to be created with ports 0-65535.

**Steps to Reproduce**

Create a TGB like this, replacing the TG and SG values as required.

```
kind: TargetGroupBinding
metadata:
  name: repro
  namespace: tgb-repro
spec:
  targetType: ip
  serviceRef: {name: repro, port: 8080}
  targetGroupARN: <TG_ARN>
  networking:
    ingress:
      - from:
          - securityGroup:
              groupID: <SRC_SG>
        ports: []                                                        # <----- the relevant line
---
apiVersion: v1
kind: Pod
metadata:
  name: app
  namespace: tgb-repro
  labels: {app: repro}
spec:
  containers:
    - name: pause
      image: registry.k8s.io/pause:3.9
---
apiVersion: v1
kind: Service
metadata:
  name: repro
  namespace: tgb-repro
spec:
  selector: {app: repro}
  ports: [{port: 8080, targetPort: 8080}]
```

**Expected Behavior**

An ingress rule on the pod ENI's security group, 0-65535 TCP from the SG given in the TGB.

**Actual Behavior**

Succesful reconcile, no security group rule created.

**Regression**
Was the functionality working correctly in a previous version ? [Yes / No]
If yes, specify the last version where it worked as expected

Unknown

**Current Workarounds**

Something like this would work but is unintuitive:

```
  networking:
    ingress:
      - from:
          - securityGroup:
              groupID: <SG>
        ports:
          - protocol: TCP     # no `port:` → all ports
```

**Environment**
- AWS Load Balancer controller version: 3.1.0
- Kubernetes version: 1.35.6
- Using EKS (yes/no), if so version?: yes
- Using Service or Ingress: yes
- AWS region: n/a
  - If helm was used then please show output of `helm -n <controllernamespace> get values <helmreleasename>`
  - If helm was not used, then copy/paste the exact command used to install the controller, including flags and options.
  - `kubectl -n <controllernamespace> describe deployment aws-load-balancer-controller`
  - `kubectl describe ingressclasses`
  - `kubectl -n <appnamespace> describe ingress <ingressname>`
  - `kubectl -n <appnamespace> describe svc <servicename>`

The fix would probably be something like:

```
                      if len(rule.Ports) == 0 {
                              allTCPPort := elbv2api.NetworkingPort{Protocol: &protocolTCP, Port: nil}
-                             permissions, err := m.computePermissionsForPeerPort(ctx, rulePeer, allTCPPort, pods)
+                             permissionsForPeerPort, err := m.computePermissionsForPeerPort(ctx, rulePeer, allTCPPort, pods)
                              if err != nil {
                                      return nil, err
                              }
-                             permissions = append(permissions, permissions...)
+                             permissions = append(permissions, permissionsForPeerPort...)
                      }
```

- [x] No, I cannot work on a PR at this time

**Additional Context**

This is what caused some confusion in #4242, I think.

dlanov:
### Issue

Fixes #4907

### Description

Fixes TargetGroupBinding networking rules with an empty or unspecified `ports` list.

The API defines an empty `ports` list as allowing all TCP ports, but `computeIngressPermissionsForTGBNetworking` accidentally shadows the accumulated `permissions` variable in this path. The computed permission is therefore discarded and no security group rule is returned.

This change:

* avoids shadowing the outer permissions accumulator;
* appends the computed all-TCP permission to the result;
* adds regression coverage verifying that an empty ports list produces a TCP permission for ports `0-65535`.

The regression test fails on the previous implementation because it returns no permissions for the empty-ports case.

Validation:

* `go test -race ./pkg/networking/... -run Test_defaultNetworkingManager_computeIngressPermissionsForTGBNetworking -count=1 -v`
* `go test ./pkg/networking/...`
* `go vet ./pkg/networking/...`
* `make fast-unit-test`
* `git diff --check`

### Checklist

* [x] Added tests that cover your change (if possible)
* [x] Added/modified documentation as required (such as the `README.md`, or the `docs` directory)
* [ ] Manually tested
* [x] Made sure the title of the PR is a good description that can go into the release notes

* [x] Backfilled missing tests for code in same general area :tada:
* [ ] Refactored something and made the world a better place :star2:


Rocin1738:
Up load API

Rocin1738:
TC ports

Rocin1738:
TCP Ports

Rocin1738:
Ingress Permissions for TGBNetworking

Rocin1738 (commented):
IngresspermissionsforTCPports

Rocin1738 (commented):
@GetPermissionForPeerPort

shraddhabang (approved):
Thanks for working on this

wweiwei-li (approved):

